PR #952, opened 2025-06-11
chore(Sinks): Add SinkCatalog, narrow Descriptors

### LINKED ISSUE #951: [TASK] Add sink catalog, centralize sink binding
### Description

We currently don't have a central sink catalog, instead every frontend has it's own map of names to source descriptors defined somewhere.
Also, every frontend needs to resolve the names of sinks to sink descriptors by themselves.

We should centralize this logic, so that it is part of core nebulastream and can be used by any new frontend, such as the REST Server or the SQL Frontend.

### Acceptance Criteria

There is a SinkCatalog that gets used by the frontends and frontends don't resolve sink names themselves anymore.

Artraxon:
This PR adds a threadsafe catalog for creating, validating and maintaining `SinkDescriptors`.
`SinkDescriptors` are now value-based like `SourceDescriptors`.
Both `Sink- and SourceDescriptors` are slimmed down, properties like `numberOfBuffersInLocalPool`, `location` or `addTimestamp` are moved to the config parameters.
For that, we now also pass other parameter maps into the `createConfigParameterContainerMap`, so that other sources can add standard properties.
Adding the `SinkDescriptors` to the `LogicalSinkOperator` based on the name is now done in a phase in the legacy optimizer to centralize the logic.
The changes are integrated into the systest, where I also pulled out our two-step logic for sink creation into a separate utility class.

## Verifying this change
All systests and unit tests are running, except for the integration tests. They are slated for removal in @alepping upcoming PR anyway.

Sources, Sinks, Optimizer, Systests.

This PR closes #951


nes-sinks/src/SinkCatalog.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <DataTypes/Schema.hpp>
+#include <Sinks/SinkDescriptor.hpp>
+#include <Util/Logger/Logger.hpp>
+
+namespace NES
+{
+std::optional<NES::Sinks::SinkDescriptor> NES::SinkCatalog::addSinkDescriptor(
+    std::string sinkName, const Schema& schema, const std::string_view sinkType, std::unordered_map<std::string, std::string> config)
+{
+    if (std::ranges::all_of(sinkName, [](const char character) { return std::isdigit(character); }))
```
yschroeder97, line 35: Then you can add this check in the future when you introduce that change.

nes-sinks/src/SinkCatalog.cpp
```diff
@@ -0,0 +1,88 @@
...
+    return sinkDescriptorOpt->second;
+}
+
+bool NES::SinkCatalog::removeSinkDescriptor(const std::string& sinkName)
+{
+    const std::unique_lock lock(catalogMutex);
+    return sinks.erase(sinkName) == 1;
+}
+
+bool NES::SinkCatalog::removeSinkDescriptor(const Sinks::SinkDescriptor& sinkDescriptor)
```
yschroeder97, line 71: You pass the sink descriptor itself here and the name to the `contains` method. Keep it consistent with respect to what the actual use is (i.e., do we want to pass the name of the descriptor here).

nes-sinks/src/SinkDescriptor.cpp
```diff
@@ -31,46 +34,61 @@
...
 {
+SinkDescriptor::SinkDescriptor(
+    std::string sinkName,
+    std::shared_ptr<const Schema>&& schema,
+    const std::string_view sinkType,
+    Configurations::DescriptorConfig::Config&& config)
+    : Descriptor(std::move(config)), sinkName(std::move(sinkName)), schema(std::move(schema)), sinkType(sinkType)
+{
+}
+
```
yschroeder97, line 45: Fix some of the newlines (I think you even introduced a clang-format for this, no?)

nes-sources/include/Sources/LogicalSource.hpp
```diff
@@ -32,7 +32,7 @@ class LogicalSource
 {
     friend SourceCatalog;
     friend OperatorSerializationUtil;
-    explicit LogicalSource(std::string logicalSourceName, const std::shared_ptr<Schema>& schema);
+    explicit LogicalSource(std::string logicalSourceName, std::shared_ptr<const Schema>&& schema);
```
yschroeder97, line 35: I don't see a reason why you would pass an rvalue to the shared pointer here.

nes-sources/include/Sources/SourceDescriptor.hpp
```diff
@@ -36,8 +38,12 @@
 
 namespace NES
 {
+class SerializableQueryPlan;
 class SourceCatalog;
 class OperatorSerializationUtil;
+namespace IntegrationTestUtil
```
yschroeder97, line 44: Is this required?

nes-sources/include/Sources/SourceDescriptor.hpp
```diff
@@ -72,34 +74,58 @@ class SourceDescriptor final : public NES::Configurations::Descriptor
...
+    static constexpr int INVALID_NUMBER_OF_BUFFERS_IN_LOCAL_POOL = -1;
+    /// NOLINTNEXTLINE(cert-err58-cpp)
+    static inline const NES::Configurations::DescriptorConfig::ConfigParameter<int64_t> NUMBER_OF_BUFFERS_IN_LOCAL_POOL{
+        "numberOfBuffersInLocalPool",
+        INVALID_NUMBER_OF_BUFFERS_IN_LOCAL_POOL,
+        [](const std::unordered_map<std::string, std::string>& config)
+        { return NES::Configurations::DescriptorConfig::tryGet(NUMBER_OF_BUFFERS_IN_LOCAL_POOL, config); }};
+
+    /// NOLINTNEXTLINE(cert-err58-cpp)
+    static inline const NES::Configurations::DescriptorConfig::ConfigParameter<uint64_t> LOCATION{
```
yschroeder97, line 111: Same as above, remove until required by distributed.

nes-sources/include/Sources/SourceValidationProvider.hpp
```diff
@@ -13,12 +13,14 @@
...
 #include <string>
+#include <string_view>
 #include <unordered_map>
 #include <Configurations/Descriptor.hpp>
 
 namespace NES::Sources::SourceValidationProvider
 {
-NES::Configurations::DescriptorConfig::Config
-provide(const std::string& sourceType, std::unordered_map<std::string, std::string>&& stringConfig);
+std::optional<NES::Configurations::DescriptorConfig::Config>
```
yschroeder97, line 24: Isn't there also `SinkValidation` where this would apply?

nes-sources/private/FileSource.hpp
```diff
@@ -70,7 +70,7 @@ struct ConfigParametersCSV
         { return NES::Configurations::DescriptorConfig::tryGet(FILEPATH, config); }};
 
     static inline std::unordered_map<std::string, NES::Configurations::DescriptorConfig::ConfigParameterContainer> parameterMap
-        = NES::Configurations::DescriptorConfig::createConfigParameterContainerMap(FILEPATH);
+        = NES::Configurations::DescriptorConfig::createConfigParameterContainerMap(SourceDescriptor::parameterMap, FILEPATH);
```
yschroeder97, line 73: What about the `TCPSource` in `nes-plugins` or the sinks? Change it for all implementations or for none

nes-sources/src/LogicalSource.cpp
```diff
@@ -12,22 +12,23 @@
...
-#include <Sources/LogicalSource.hpp>
 #include <fmt/format.h>
 
 namespace NES
 {
 
 
-LogicalSource::LogicalSource(std::string logicalSourceName, const std::shared_ptr<Schema>& schema)
-    : logicalSourceName(std::move(logicalSourceName)), schema(schema)
+LogicalSource::LogicalSource(std::string logicalSourceName, std::shared_ptr<const Schema>&& schema)
```
yschroeder97, line 30: Again, I can not think of any valid reason why you would pass a temporary to a shared pointer here. 
Scenario 1: You construct a temporary with `std::make_shared` when calling this ctor. In this case, it would make more sense to just create the object in here instead of passing from the outside.
Scenario 2: You have an lvalue and use `std::move` to move the shared pointer in this ctor. But shared pointers are for shared ownership, with move you give up the original value, removing the purpose of the shared pointer.

nes-sinks/include/Sinks/SinkCatalog.hpp
```diff
@@ -0,0 +1,44 @@
...
+    std::optional<Sinks::SinkDescriptor> getSinkDescriptor(const std::string& sinkName) const;
+
+    bool removeSinkDescriptor(const std::string& sinkName);
+    bool removeSinkDescriptor(const Sinks::SinkDescriptor& sinkDescriptor);
+
+    bool containsSinkDescriptor(const std::string& sinkName) const;
+    bool containsSinkDescriptor(const Sinks::SinkDescriptor& sinkDescriptor) const;
+
+private:
+    mutable std::recursive_mutex catalogMutex;
```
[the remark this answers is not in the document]
Artraxon: I changed it to folly Synchronized

nes-common/include/ExceptionDefinitions.inc
```diff
@@ -66,6 +66,8 @@ EXCEPTION(UnknownDataType, 2123, "unknown data type")
 EXCEPTION(UnknownSource, 2124, "unknown source")
 EXCEPTION(SourceAlreadyExists, 2125, "source already exists")
 EXCEPTION(QueryCompilerError, 2126, "query compiler / backend throwed an error")
+EXCEPTION(UnknownSink, 2127, "Unknown sink")
```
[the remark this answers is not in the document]
Artraxon: done

nes-configurations/include/Configurations/Descriptor.hpp
```diff
@@ -294,8 +301,33 @@ class DescriptorConfig
...
+                }
+            }
+            else
+            {
+                static_assert(false, "Invalid config parameter type");
+            }
+        };
+        (inserter(parameters), ...);
+        return configParameterMap;
     }
```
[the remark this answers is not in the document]
Artraxon: This allows you to pass other config parameter maps to this factory function.
This enables us to have "well-known" parameters such as location or file_path that lots of sources/sinks might have. Sources and sinks that have them, don't have to reimplement them, and code that depends on these properties (like the systests) can generalize over all of them, instead of having to go through all source/sink specific parameters.

nes-logical-operators/src/Operators/Sinks/SinkLogicalOperator.cpp
```diff
@@ -96,12 +107,13 @@ LogicalOperator SinkLogicalOperator::withChildren(std::vector<LogicalOperator> c
...
 {
-    return {sinkDescriptor->schema};
+    return children | std::ranges::views::transform([](const LogicalOperator& child) { return child.getOutputSchema(); })
+        | std::ranges::to<std::vector>();
 };
 
 Schema SinkLogicalOperator::getOutputSchema() const
 {
-    return sinkDescriptor->schema;
+    return sinkDescriptor.has_value() ? *sinkDescriptor->getSchema() : Schema{};
```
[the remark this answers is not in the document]
Artraxon: We could, but then we'd need to split the class up into a SinkNameLogicalOperator and a SinkDescriptorLogicalOperator like we have with the sources.

My perspective is, that I want to get rid of the SourceNameLogicalOperator in the near future with the new query binder anyway, so I would splitting it up into two classes to just remove one of them in the near future seams a bit unnecessary.

nes-nebuli/include/NebuLI.hpp
```diff
@@ -37,11 +39,15 @@ namespace NES::CLI
...
 {
     std::shared_ptr<const SourceCatalog> sourceCatalog;
+    std::shared_ptr<const SinkCatalog> sinkCatalog;
 
 public:
     [[nodiscard]] LogicalPlan optimize(const LogicalPlan& plan) const;
     LegacyOptimizer() = default;
-    explicit LegacyOptimizer(const std::shared_ptr<const SourceCatalog>& sourceCatalog) : sourceCatalog(sourceCatalog) { }
+    explicit LegacyOptimizer(std::shared_ptr<const SourceCatalog> sourceCatalog, std::shared_ptr<const SinkCatalog> sinkCatalog)
+        : sourceCatalog(std::move(sourceCatalog)), sinkCatalog(std::move(sinkCatalog))
```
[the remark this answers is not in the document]
Artraxon: In theory you are right at the moment, but I don't think this would be good moving forward.

We will have an ever growing amount of rewrite rules and optimization stages, where everyone requires access to different catalogs/data.

By using inversion-of-control, we can separate the dependencies of the optimization stages from the invocation of the stages.

This allows us to apply all optimization stages in one place, and for example also add them as plugins.

And, then of course the optimization stages have shared ownership of the catalogs for the time that their instances exist.

nes-nebuli/private/LegacyOptimizer/SinkBindingRule.hpp
```diff
@@ -0,0 +1,32 @@
...
+#include <Plans/LogicalPlan.hpp>
+#include <Sinks/SinkCatalog.hpp>
+
+namespace NES::LegacyOptimizer
+{
+class SinkBindingRule
+{
+public:
+    explicit SinkBindingRule(std::shared_ptr<const SinkCatalog> sinkCatalog) : sinkCatalog(std::move(sinkCatalog)) { }
+    void apply(LogicalPlan& queryPlan) const;
```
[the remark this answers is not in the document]
Artraxon: See my comment above, we should keep the interface consistent and use IOC.

nes-nebuli/private/YAML/YAMLBinder.hpp
```diff
@@ -43,6 +47,7 @@ struct SchemaField
 struct Sink
 {
     std::string name;
+    std::vector<SchemaField> schema;
```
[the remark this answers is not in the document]
Artraxon: Because sinks can be shared between queries, and you could otherwise infer different schemas for the same sink in different queries, which for a lot of sink types would just be logically undefined or produce the wrong output. 

I created an issue for adding inline-sink declarations, so that one can specify a sink without the schema, which is then guaranteed to be local to one query https://github.com/nebulastream/nebulastream/issues/1001 .

But if you create a data systems object that can be referenced in multiple queries, you need to specify a schema so the correct usage can be verified.

nes-nebuli/private/YAML/YAMLBinder.hpp
```diff
@@ -81,14 +86,21 @@ struct BoundQueryConfig
 
 class YAMLBinder
```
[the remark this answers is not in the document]
Artraxon: Hmm, there was one comment referencing future sink catalogs, I removed now the whole struct because it was not used.

nes-nebuli/src/LegacyOptimizer/SinkBindingRule.cpp
```diff
@@ -0,0 +1,46 @@
...
+    queryPlan.rootOperators
+        = queryPlan.rootOperators
+        | std::ranges::views::transform(
+              [this](const LogicalOperator& rootOperator) -> LogicalOperator
+              {
+                  auto sinkOperator = rootOperator.tryGet<SinkLogicalOperator>();
+                  INVARIANT(sinkOperator.has_value(), "Root operator must be sink");
+                  /// Check to centralize the sink binding logic
+                  /// TODO #897 move sink binding to new query binder
+                  PRECONDITION(
```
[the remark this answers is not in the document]
Artraxon: I guess, I didn't pay that much attention. fixed.

nes-nebuli/src/LegacyOptimizer/SinkBindingRule.cpp
```diff
@@ -0,0 +1,46 @@
...
+                      not sinkOperator->sinkDescriptor.has_value(), "Sink Descriptors should only be set during the sink binding phase");
+                  const auto sinkDescriptor = sinkCatalog->getSinkDescriptor(sinkOperator->sinkName);
+                  if (not sinkDescriptor.has_value())
+                  {
+                      throw UnknownSink("{}", sinkOperator->sinkName);
+                  }
+                  sinkOperator->sinkDescriptor = sinkDescriptor;
```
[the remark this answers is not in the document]
Artraxon: As mentioned above, this would be better in the current state, but we want to get rid of the SinkNameLogicalOperator in the near future anyway.

nes-nebuli/src/YAML/YAMLBinder.cpp
```diff
@@ -201,41 +187,31 @@ std::vector<SourceDescriptor> YAMLBinder::bindRegisterPhysicalSources(const std:
     return boundSources;
 }
 
-BoundQueryConfig YAMLBinder::parseAndBind(std::istream& inputStream)
+std::vector<Sinks::SinkDescriptor> YAMLBinder::bindRegisterSinks(const std::vector<Sink>& unboundSinks)
 {
-    BoundQueryConfig config{};
-    auto& [plan, sinks, logicalSources, physicalSources] = config;
-    try
+    std::vector<Sinks::SinkDescriptor> boundSinks{};
```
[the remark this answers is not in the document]
Artraxon: Again, it's not my component, this should be decided by for example @ls-1801 .

nes-sinks/include/Sinks/SinkDescriptor.hpp
```diff
@@ -14,36 +14,91 @@
...
+#include <Identifiers/NESStrongType.hpp>
 #include <Util/Logger/Formatter.hpp>
 #include <SerializableOperator.pb.h>
 
+namespace NES
+{
+class SerializableQueryPlan;
+class OperatorSerializationUtil;
+class SinkCatalog;
+namespace IntegrationTestUtil
```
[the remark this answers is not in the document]
Artraxon: I removed the unused ones, for SinkCatalog and Serialization easier for the friend declarations

nes-sinks/src/SinkDescriptor.cpp
Artraxon: I got lazy and didn't rebase and merge. Done

nes-sources/include/Sources/LogicalSource.hpp
Artraxon: changed it to pass by value and creating the shared ptr in the constructor

nes-sources/include/Sources/SourceDescriptor.hpp
Artraxon: no, removed

nes-sources/include/Sources/SourceDescriptor.hpp
Artraxon: removed

PR #961, opened 2025-06-17
fix(parser): fixes parsing of nested unions

### LINKED ISSUE #912: [BUG] SQLParser rejects nested unions in a subquery
### Version / Commit

main

### What happened?

Currently, the SQL Parser does not accept nested unions in a subquery.

```sql
SELECT timestamp
FROM (
    SELECT * FROM stream UNION SELECT * FROM stream2 UNION SELECT * FROM stream2
)
INTO sink
```

The query should combine the data from all sources and project only the timestamp from each tuple

### How to reproduce

_No response_

### Operating System

linux

### Deployment details

_No response_

### Anything else

_No response_

### Code of Conduct

bgrb1:
The bug mentioned in issue 912 occurs whenever there is a nested union structure, and the top union itself is part of a subquery. 
The cause is in the `AntlrSQLQueryPlanCreator`. The `exitSetOperation` function will always read both its children from the top level queryPlans stack. However, whenever a union is a part of a subquery, it will not push its produced plan to the this global stack, but to the helper of the current subquery context. Which means that in a nested union structure, a parent union will not see the produced plan of a child union. 

To fix this, a set operation will treat both its children like subqueries, and create its own helper, from which it will also read its children. The union case in exitPrimaryQuery is also removed, so that the function will always push its produced plan to the next helper if there is one. 

Also, some usage of move semantics is added on the way. 

## Verifying this change
The unit tests, including the parser unit tests succeed + all systests. Also, the bug testcase found in nes-systests/bug/UnionInSubquery.test succeeds now, indicating that the bug was fixed successfully.  

Parser, systests

This PR closes #912


ls-1801:
merged in #909

PR #963, opened 2025-06-18
Fix Join Varsized DataTypes

alepping:
Prior, we did not support the following query:
```sql
SELECT CONCAT(text1, text2) AS concatedText FROM streamWithText INTO sinkStreamWithConcatedText
```

The problem was that when we joined `text1` and `text2`, which are both varsized fields, we would return UNDEFINED as the result type.
This PR changes the `join` function to return VARSIZED, if both data types are varsized.

## Verifying this change
Adds test for the specific query above (`nes-systests/function/varsized/Concat.test`)

### DIFF
--- a/nes-systests/function/varsized/Concat.test
+++ b/nes-systests/function/varsized/Concat.test
@@ -1,0 +1,21 @@
+# name: function/varsized/Concat.test
+# description: Simple concat of varsized data types test
+# groups: [Function, VarSized]
+
+Source streamWithText VARSIZED text1 VARSIZED text2 VARSIZED text3 INLINE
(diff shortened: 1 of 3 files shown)


nes-systests/function/varsized/Concat.test
```diff
@@ -0,0 +1,21 @@
+# name: selection/Selection_NotEquals.test
+# description: Simple selection tests
+# groups: [Selection]
```
adrianmichalke, line 3: This is a copy paste error I guess?

adrianmichalke (approved):

PR #975, opened 2025-06-24
Simple Demo/Tutorial

keyseven123:
This PR adds the markdown for a self-hosted demo/tutorial that shows how to deploy queries


ls-1801 (changes requested):
There is so much copy paste in these scripts.

1. I think we can reduce a lot of the bash code by just improving the docker compose.
2. I would like to see some basic test in our ci that checks if stuff works.
3. I would prefer the system stat source to be a custom source instead.

PR #978, opened 2025-06-25
Added engine to pipeline setup call

keyseven123:
This PR passes the nautilus engine during the query compilation phase to the pipeline setup call. This allows any operator to query compile additional code that is not being called from a typical pipeline invocation. One example is the cleanup function for an aggregation or hash-join slice.

## Verifying this change
This change is tested by running our CI tests.

- ExecutionEngine
- QueryCompiler


adrianmichalke (changes requested):
re-request once rebased

PR #984, opened 2025-07-01
[TEST] Codecov

ls-1801:


keyseven123 (changes requested):
Please also remove `cmake/CodeCoverage.cmake`

Artraxon (approved):

PR #987, opened 2025-07-02
test(Systest): Use serialization logic in systest

ls-1801:
This PR changes the localquerysubmitter to actually use the serialization logic, which catches bad serialization logic in our systests.

### DIFF
--- a/nes-systests/systest/src/QuerySubmitter.cpp
+++ b/nes-systests/systest/src/QuerySubmitter.cpp
@@ -24,8 +24,10 @@
 #include <Listeners/QueryLog.hpp>
 #include <Runtime/Execution/QueryStatus.hpp>
 #include <Runtime/QueryTerminationType.hpp>
+#include <Serialization/QueryPlanSerializationUtil.hpp>
 #include <grpcpp/create_channel.h>
 #include <grpcpp/security/credentials.h>
+
 #include <ErrorHandling.hpp>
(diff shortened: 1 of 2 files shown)


keyseven123 (approved):

alepping (approved):
I'm fine with it. Better to 'unnecessarily' serialize/deserialize and test that functionality, then to add an additional mode that adds additional complexity imo.
@ls-1801 
whats the plan for remote execution? that should rely on serialization in any case. Is there a good reason why remote execution is not the default?

adrianmichalke:
@ls-1801 my main concern was runtime. But the delta seems to be minimal. 

To fully check for correct serialization, we would need to check the equality `queryplan = serialized(deserialized(queryplan))`. If we do not want a separate systest mode your code should include it IMO.

adrianmichalke:
In the systest tool you get the check for free as you are already doing the (de-)serialization.

Just write instead:
   ```
    const auto equivalentPlan = QueryPlanSerializationUtil::deserializeQueryPlan(QueryPlanSerializationUtil::serializeQueryPlan(plan));
    if (equivalentPlan == plan)
    {
       return worker.registerQuery();
    }
    else
    {
       return CannotSerialize("error during serialization of plan: {}", plan);
    }
```

ls-1801:
I mean we can integrate that check, and just always assert that.

PR #988, opened 2025-07-02
feature(adding install script)

### LINKED ISSUE #985: [TASK] Provide Easy Way of Building NebulaStream from Source
### Description

As part of making NebulaStream more developer friendly, we should mention all of our linux dependencies and provide a small bash script that installs them. 

### Acceptance Criteria

- [ ] Bash script or another way that installs all deps, runs cmake, builds the project
- [ ] Add a link to the script/docker files in https://github.com/nebulastream/nebulastream/blob/main/docs/technical/development.md#non-container-development-environment

tim-nick:
Adding install script

This change is about adding a script which separates out the installation of the dependencies from docker into a bash script so users can also install them locally.

This PR closes #985

### DIFF
--- a/docker/dependency/Base.dockerfile
+++ b/docker/dependency/Base.dockerfile
@@ -1,72 +1,23 @@
-# The Base Dockerfile installs and configures all relevant tooling to build the dependencies and NebulaStream.
-# Currently our compiler toolchain is based on llvm-18 using libc++.
-# Additionally we install a recent CMake version and the mold linker.
 FROM ubuntu:24.04
 
+# 1) Copy & run your install script
+COPY scripts/install_sys_dependencies.sh /usr/local/bin/install_script.sh
+RUN chmod +x /usr/local/bin/install_script.sh \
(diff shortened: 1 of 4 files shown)


docker/dependency/Base.dockerfile
```diff
@@ -1,72 +1,23 @@
...
-    && make install -j $(nproc)\
-    && cd .. \
-    && rm -rf cmake-${CMAKE_VERSION}.tar.gz cmake-${CMAKE_VERSION} \
-    && cmake --version
+# 2) Install Ninja in-container
+RUN apt update -y \
+ && apt install -y ninja-build
 
-# default cmake generator is ninja
+# 3) Default CMake generator and llvm toolchain version since it's only set in non-docker builds at default
```
keyseven123, line 21:
```suggestion
# 3) Default CMake generator and llvm toolchain version since it's only set in non-docker builds per default
```

keyseven123 (approved):

keyseven123:
Yes, I will close it for now.

PR #1000, opened 2025-07-07
fix(engine): Addresses flakey test

ls-1801:
This commit fixes a flakey test, caused by a race between
the engine shutdown and the query status listener reporting
the running event. Although unlikely, it is possible for the
running query status report to be skipped during query shutdown
even if the query has already produced multiple tuplebuffer.

The fix waits for the running event to be reported.

## Verifying this change
- Fixes a issue in a test

- N/A


nes-query-engine/tests/QueryEngineTest.cpp
```diff
@@ -88,6 +88,10 @@ TEST_F(QueryEngineTest, singleQueryWithShutdown)
         ctrl->injectData(std::vector(DEFAULT_BUFFER_SIZE, std::byte(0)), NUMBER_OF_TUPLES_PER_BUFFER);
         ctrl->injectData(std::vector(DEFAULT_BUFFER_SIZE, std::byte(0)), NUMBER_OF_TUPLES_PER_BUFFER);
         test.sinkControls[pipeline]->waitForNumberOfReceivedBuffersOrMore(4);
+
+        /// The Tests asserts that a query reaches the running state, to prevent flakey tests. Even if the query already produced 4 buffers
```
adrianmichalke, line 92: nit:
```suggestion
        /// The Tests asserts that a query reaches the running state, to prevent flakey tests. Even if the query already produced 4 buffers
```
```suggestion
        /// The tests asserts that a query reaches the running state, to prevent flakey tests. Even if the query already produced 4 buffers
```

adrianmichalke (approved):

keyseven123 (approved):

PR #1005, opened 2025-07-08
feat(CI): Test gh arm server

ls-1801:


.github/workflows/build_and_test.yml
```diff
@@ -250,6 +233,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.ref }}
+      - name: ccache
+        uses: hendrikmuhs/ccache-action@v1.2.18
+        with:
+          key: ${{ matrix.arch }}-${{ matrix.build_type }}-${{matrix.sanitizer}}-${{matrix.stdlib}}
```
adrianmichalke: What's the rational of choosing in this case the standalone ccache action vs. the `prepare-github` action that includes it as well?

.github/steps/prepare-github/action.yml
```diff
@@ -0,0 +1,33 @@
+name: prepare github ci server instance
```
keyseven123: I know that this is a nitpick but I would like to have a consistent usage of description as well as capitalization for the name and description. If you also agree than I can tackle this in a separate PR. Just let me know

.github/steps/run-in-container/action.yml
```diff
@@ -0,0 +1,23 @@
...
+    required: true
+
+runs:
+  using: 'composite'
+  steps:
+    - name: run in container
+      shell: bash
+      run: |
+        docker run \
+             -v $(pwd):$(pwd) -e CCACHE_DIR=$(pwd)/.ccache --workdir $(pwd) -u $(id -u):$(id -g) \
```
keyseven123: Why is it given that the ccache dir is located in the current working dir?

.github/workflows/build_and_test.yml
```diff
@@ -14,28 +14,22 @@ on:
...
-      options: --user root
-    timeout-minutes: 40
-    runs-on: [ self-hosted, linux, Build, "${{matrix.arch}}" ]
+    name: "${{matrix.arch}}-${{matrix.stdlib}}-${{matrix.build_type}}-${{ matrix.sanitizer }} Test"
+    runs-on: ${{ matrix.runner }}
     strategy:
       fail-fast: false
       matrix:
-        arch: [ x64, arm64 ]
+        arch: [ arm64, x64 ]
```
keyseven123: Is there a reason other than lexical sorting?

keyseven123 (changes requested):

nes-systests/systest/CMakeLists.txt
```diff
@@ -107,7 +107,7 @@ endif ()
 
 if (NOT CODE_COVERAGE)
     ## We run all join and aggregation tests with different no. worker threads and different join strategies
-    set(workerThreads 1 2 4 8)
+    set(workerThreads 1 2 4)
```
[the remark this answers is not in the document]
ls-1801: the problem is that there are only 4 cpus on the github servers. We test larger numbers of threads with the large scale tests anyways

.github/steps/prepare-github/action.yml
ls-1801: lets do it in a different pr

keyseven123 (approved):

PR #1006, opened 2025-07-09
Hotfix systest small data

adrianmichalke:
9bc26cf68e2de225ea3ab1da471acdb1f2be8805 seems to have broken the small data usage in systest. This PR is a hotfix for this issue.

**chore(CMake): use absolute paths for systest-lib target_compile_definition**

**fix(CMake): fixes that small data files could not be found. We symlink now to the files.**


adrianmichalke:
covered by #1007

PR #1007, opened 2025-07-09
fix(Systest): Use CMake ExternalData dir

ls-1801:

Addresses the issue of missing systest input files. Previously, systest assumed that testdata always
reside in cmake-build-debug, which was the case until recently, where small test input files were moved back
into the main repository.

When building CMake without the `ENABLE_LARGE_TEST` flag, there are no external references (no .md5 files) that
would need to be resolved, thus ExternalData would not bother to move/link them into the build directory.
This is not an issue for our ctest, which uses the systest tool. As in CMake, we declared them with `ExternalData_Add_Test`, which resolves the path correctly and uses the build directory.

When building our systest tool, we supply a fixed compiler define that points to the CMake build directory, which
wouldn't work if the external data does not move the files there.

The fix correctly resolves the ExternalData `Data{...}` definition and uses the resolved path as a compiler define instead
of hardcoding it.

## Verifying this change
This PR adds a smoke test to the clean-build CI job, ensuring this issue does not recur. On the same note, this also adds a check that both nebuli and the single-node-worker can run after a clean build.

-CI
-Systest


.github/workflows/build_and_test.yml
```diff
@@ -50,6 +50,9 @@ jobs:
         run: |
           source .github/.env/test-${{matrix.sanitizer}}.env
           cmake --build build -j ${{steps.number-of-jobs.outputs.number_of_jobs}} -- -k 0
+          
+          # Verify basic functionality of systest tool
+          build/nes-systests/systest/systest --exclude-groups large -n 100 -- --worker.queryOptimizer.executionMode=INTERPRETER --worker.numberOfBuffersInGlobalBufferManager=20000 --worker.queryEngine.numberOfWorkerThreads=${{steps.number-of-jobs.outputs.number_of_jobs}}
```
adrianmichalke, line 55: Cool, as it is a clean build we would detect errors like the wrong located small data? 

Have you tested if it fails without the two fix commits?

keyseven123 (changes requested):

nes-systests/systest/CMakeLists.txt
```diff
@@ -15,11 +15,27 @@ add_subdirectory(checksum)
...
 )
+if (ENABLE_LARGE_TESTS)
+    set(TEST_DATA_PATH DATA{../testdata/,RECURSE:,REGEX:.*})
+else ()
+    # By default this only includes small files located in testdata/small
+    set(TEST_DATA_PATH DATA{../testdata/,RECURSE:,REGEX:small.*})
+endif ()
 set(ExternalData_TIMEOUT_ABSOLUTE 1200)
+ExternalData_Expand_Arguments(
+        test-data
```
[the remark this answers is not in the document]
ls-1801: It's a cmake target, not a constant.

.github/workflows/build_and_test.yml
ls-1801: yes it fails without the commit.

adrianmichalke (approved):

PR #1009, opened 2025-07-12
refactor(LogicalOperators): enhance logical operator representations

yschroeder97:
This pull request enhances the logical operator representations. 
The changes will be used extensively by the upcoming distributed PR.

**LogicalPlan**:
- Change visibility of `rootOperators` to private
- Change `replaceOperator` and `replaceSubtree` to take the id of the operator to replace for more flexibility
- Introduce `getOperatorById` to retrieve an operator
- Support immutable access patterns by adding `withRootOperators` and `withAdditionalRoots` that create a new plan
- Add convenience free function `addRootOperators` to add additional roots without affecting existing ones

**LogicalOperator**:
- Support dynamic traits by introducing a `TraitSet` to each logical operator, where optimizer passes can add traits to at the time of query planning
- Add  `addTraits` free function that returns new logical operators that adds traits to the given operator

**Trait**:
- Change `TraitSet` to use an actual `std::unordered_set` such that we avoid adding duplicate traits
- Add `DefaultTrait` that can be extended by marker traits (traits without custom data). This avoids having to implement the 4 methods (`hash`, `operator==`, `getType`, and `serialize`) for them.

## Verifying this change
This change is tested by
- Additional unit tests in `LogicalPlanTest`

- LogicalOperators
- QueryOptimizer (trait extended)
- Nebuli (root operator visibility)
- Systests (root operator visibility)

This PR closes #922.

### DIFF
--- a/nes-logical-operators/include/Operators/IngestionTimeWatermarkAssignerLogicalOperator.hpp
+++ b/nes-logical-operators/include/Operators/IngestionTimeWatermarkAssignerLogicalOperator.hpp
@@ -17,6 +17,7 @@
 #include <string>
 #include <string_view>
 #include <vector>
+
 #include <DataTypes/Schema.hpp>
 #include <Identifiers/Identifiers.hpp>
 #include <Operators/LogicalOperator.hpp>
@@ -35,6 +36,7 @@ public:
     [[nodiscard]] bool operator==(const LogicalOperatorConcept& rhs) const override;
     [[nodiscard]] SerializableOperator serialize() const override;
 
+    [[nodiscard]] LogicalOperator withTraitSet(TraitSet traitSet) const override;
     [[nodiscard]] TraitSet getTraitSet() const override;
 
     [[nodiscard]] LogicalOperator withChildren(std::vector<LogicalOperator> children) const override;
@@ -58,6 +60,7 @@ protected:
     static constexpr std::string_view NAME = "IngestionTimeWatermarkAssigner";
 
     std::vector<LogicalOperator> children;
+    TraitSet traitSet;
     Schema inputSchema;
     Schema outputSchema;
     std::vector<OriginId> inputOriginIds;
--- a/nes-logical-operators/include/Plans/LogicalPlan.hpp
+++ b/nes-logical-operators/include/Plans/LogicalPlan.hpp
@@ -38,44 +38,51 @@ public:
     LogicalPlan() = default;
     explicit LogicalPlan(LogicalOperator rootOperator);
     explicit LogicalPlan(QueryId queryId, std::vector<LogicalOperator> rootOperators);
+    explicit LogicalPlan(QueryId queryId, std::vector<LogicalOperator> rootOperators, std::string originalSql);
+
+    LogicalPlan(const LogicalPlan& other) = default;
+    LogicalPlan& operator=(const LogicalPlan& other);
+    LogicalPlan(LogicalPlan&& other) noexcept;
+    LogicalPlan& operator=(LogicalPlan&& other) noexcept;
 
     [[nodiscard]] bool operator==(const LogicalPlan& otherPlan) const;
     friend std::ostream& operator<<(std::ostream& os, const LogicalPlan& plan);
 
-    std::vector<LogicalOperator> rootOperators;
-
     [[nodiscard]] QueryId getQueryId() const;
-    [[nodiscard]] const std::string& getOriginalSql() const;
+    [[nodiscard]] std::string getOriginalSql() const;
+    [[nodiscard]] std::vector<LogicalOperator> getRootOperators() const;
+
+    [[nodiscard]] LogicalPlan withRootOperators(const std::vector<LogicalOperator>& operators) const;
 
     void setOriginalSql(const std::string& sql);
     void setQueryId(QueryId id);
 
 private:
-    /// Holds the original SQL string
-    std::string originalSql;
     QueryId queryId = INVALID_QUERY_ID;
+    std::vector<LogicalOperator> rootOperators;
+    std::string originalSql; /// Holds the original SQL string
 };
 
 /// Get all parent operators of the target operator
 [[nodiscard]] std::vector<LogicalOperator> getParents(const LogicalPlan& plan, const LogicalOperator& target);
 
 /// Replace `target` with `replacement`, keeping target's children
-[[nodiscard]] std::optional<LogicalPlan>
-replaceOperator(const LogicalPlan& plan, const LogicalOperator& target, LogicalOperator replacement);
+[[nodiscard]] std::optional<LogicalPlan> replaceOperator(const LogicalPlan& plan, OperatorId target, LogicalOperator replacement);
 
 /// Replace `target` with `replacement`, keeping the children that are already inside `replacement`
-[[nodiscard]] std::optional<LogicalPlan>
-replaceSubtree(const LogicalPlan& plan, const LogicalOperator& target, const LogicalOperator& replacement);
+[[nodiscard]] std::optional<LogicalPlan> replaceSubtree(const LogicalPlan& plan, OperatorId target, const LogicalOperator& replacement);
 
 /// Adds a new operator to the plan and promotes it as new root by reparenting existing root operators and replacing the current roots
 [[nodiscard]] LogicalPlan promoteOperatorToRoot(const LogicalPlan& plan, const LogicalOperator& newRoot);
+
+[[nodiscard]] LogicalPlan addRootOperators(const LogicalPlan& plan, const std::vector<LogicalOperator>& rootsToAdd);
 
 template <class T>
 [[nodiscard]] std::vector<T> getOperatorByType(const LogicalPlan& plan)
 {
     std::vector<T> operators;
     std::ranges::for_each(
-        plan.rootOperators,
+        plan.getRootOperators(),
         [&operators](const auto& rootOperator)
         {
             auto typedOps = BFSRange(rootOperator)
@@ -86,12 +93,14 @@ template <class T>
     return operators;
 }
 
+[[nodiscard]] std::optional<LogicalOperator> getOperatorById(const LogicalPlan& plan, OperatorId operatorId);
+
 template <typename... TraitTypes>
 [[nodiscard]] std::vector<LogicalOperator> getOperatorsByTraits(const LogicalPlan& plan)
 {
     std::vector<LogicalOperator> matchingOperators;
     std::ranges::for_each(
-        plan.rootOperators,
+        plan.getRootOperators(),
         [&matchingOperators](const auto& rootOperator)
         {
             auto ops = BFSRange(rootOperator);
--- a/nes-logical-operators/src/Operators/LogicalOperator.cpp
+++ b/nes-logical-operators/src/Operators/LogicalOperator.cpp
@@ -12,14 +12,15 @@
     limitations under the License.
 */
 
-#include <memory>
+#include <Operators/LogicalOperator.hpp>
+
 #include <string>
 #include <string_view>
 #include <utility>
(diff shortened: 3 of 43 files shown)


nes-logical-operators/include/Operators/IngestionTimeWatermarkAssignerLogicalOperator.hpp
```diff
@@ -35,6 +35,8 @@ class IngestionTimeWatermarkAssignerLogicalOperator final : public LogicalOperat
     [[nodiscard]] bool operator==(const LogicalOperatorConcept& rhs) const override;
     [[nodiscard]] SerializableOperator serialize() const override;
 
+    [[nodiscard]] LogicalOperator withTraitSet(TraitSet traitSet) const override;
+    [[nodiscard]] LogicalOperator withAdditionalTraits(TraitSet traitSet) const override;
```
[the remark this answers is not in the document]
yschroeder97: Done.

nes-logical-operators/src/Plans/LogicalPlanBuilder.cpp
```diff
@@ -197,9 +197,8 @@ LogicalPlanBuilder::checkAndAddWatermarkAssigner(LogicalPlan queryPlan, const st
 }
 
 LogicalPlan LogicalPlanBuilder::addBinaryOperatorAndUpdateSource(
-    const LogicalOperator& operatorNode, LogicalPlan leftLogicalPlan, LogicalPlan rightLogicalPlan)
+    const LogicalOperator& operatorNode, const LogicalPlan& leftLogicalPlan, const LogicalPlan& rightLogicalPlan)
 {
-    leftLogicalPlan.rootOperators.push_back(rightLogicalPlan.rootOperators[0]);
-    return promoteOperatorToRoot(leftLogicalPlan, operatorNode);
+    return promoteOperatorToRoot(leftLogicalPlan.withAdditionalRoots(rightLogicalPlan.getRootOperators()), operatorNode);
```
Artraxon, line 202: concatenation can be done here with a bit of ranges codes

```suggestion
auto newRootOperators = std::ranges::views::join(std::array{leftLogicalPlan.getRootOperators(), rightLogicalPlan.getRootOperators()});
+    return promoteOperatorToRoot(leftLogicalPlan.withRootOperators(newRootOperators | std::ranges::to<std::vector>()), operatorNode);
```

nes-logical-operators/src/Operators/LogicalOperator.cpp
```diff
@@ -113,3 +119,15 @@ LogicalOperator LogicalOperator::withInferredSchema(std::vector<Schema> inputSch
     return self->withInferredSchema(std::move(inputSchemas));
 }
 }
+
+namespace NES
+{
+
+LogicalOperator addTraits(const LogicalOperator& op, const TraitSet& traitSet)
+{
+    auto result = op.getTraitSet();
```
Artraxon, line 128: Imo if it is only used in testing it can also just be a test utility, because it swallows for example that traits are just overwritten, which depending on the context maybe should be an error. Or at least indicate in the function name that it is overwriting existing traits of the same type.

nes-logical-operators/src/Serialization/OperatorSerializationUtil.cpp
```diff
@@ -128,7 +129,7 @@ SourceDescriptor OperatorSerializationUtil::deserializeSourceDescriptor(const Se
     /// TODO #815 the serializer would also a catalog to register/create source descriptors/logical sources
     const auto physicalSourceId = PhysicalSourceId{sourceDescriptor.physicalsourceid()};
     const auto& sourceType = sourceDescriptor.sourcetype();
-    const auto workerId = WorkerId{sourceDescriptor.workerid()};
+    const auto& workerId = WorkerId{sourceDescriptor.workerid()};
```
Artraxon, line 132: Isn't this a dangling reference? This should stay assigned by value. Besides the fact, even if it is not dangling, you don't need to use a reference for an 8 byte type in general, because 
1. reference usage is harder to optimize, a by value assignment that is only used to be passed to another function would be inlined. In some cases, the c++ standard even guarantees that, but only for value semantics, not for references.
2. References are internally also usually implemented with pointers, so you also have 8 bytes, but need to dereference once.

nes-query-optimizer/include/Traits/Trait.hpp
```diff
@@ -70,6 +76,15 @@ struct Trait
     Trait(const Trait& other);
     Trait(Trait&&) noexcept;
 
+    /// Compares this trait with another trait for equality.
+    /// @param other The trait to compare with.
+    /// @return bool True if the traits are equal, false otherwise.
```
adrianmichalke, line 81: nit: docu redundant IMO

nes-logical-operators/src/Operators/LogicalOperator.cpp
```diff
@@ -113,3 +119,15 @@ LogicalOperator LogicalOperator::withInferredSchema(std::vector<Schema> inputSch
     return self->withInferredSchema(std::move(inputSchemas));
 }
 }
+
+namespace NES
+{
```
adrianmichalke, line 124:
```suggestion
```

adrianmichalke (changes requested):

nes-logical-operators/include/Plans/LogicalPlan.hpp
```diff
@@ -38,34 +38,40 @@ class LogicalPlan
...
 
 /// Get all parent operators of the target operator
 [[nodiscard]] std::vector<LogicalOperator> getParents(const LogicalPlan& plan, const LogicalOperator& target);
 
 /// Replace `target` with `replacement`, keeping target's children
-[[nodiscard]] std::optional<LogicalPlan>
-replaceOperator(const LogicalPlan& plan, const LogicalOperator& target, LogicalOperator replacement);
+[[nodiscard]] std::optional<LogicalPlan> replaceOperator(const LogicalPlan& plan, OperatorId target, LogicalOperator replacement);
 
 /// Replace `target` with `replacement`, keeping the children that are already inside `replacement`
```
Artraxon, line 73: I am not sure if we only want to do it by operator id, because the equivalence relation is call-side context specific. I think there is value in having something that guarantees to compare by operator ID, but it should be an overload.

I think it would be better to keep the generic interface where we don't guarantee by what we compare, and just forward it to the replace Operator by ID for now.
We might add other such functions in the future, or parametrize the replaceOperator function by passing a functor for comparison.

nes-query-optimizer/include/Traits/Trait.hpp
```diff
@@ -136,14 +151,34 @@ struct Trait
...
     };
 
     std::unique_ptr<Concept> self;
 };
 
-using TraitSet = std::vector<Trait>;
+struct TraitHash
+{
+    size_t operator()(const Trait& trait) const noexcept { return trait.hash(); }
+};
```
Artraxon, line 166: why not just implement std::hash so that it is the default and you don't need to specify it every time?

Artraxon (changes requested):
Overall I think a good iteration over our IR that improves safety a lot, I didn't like all the raw member field access from the beginning.

We just have to be really careful where we put what kind of utility functions and whether they are in line with the semantics that people would expect/need.

nes-logical-operators/include/Plans/LogicalPlan.hpp
```diff
@@ -38,34 +38,40 @@ class LogicalPlan
...
 
-    std::vector<LogicalOperator> rootOperators;
-
     [[nodiscard]] QueryId getQueryId() const;
-    [[nodiscard]] const std::string& getOriginalSql() const;
+    [[nodiscard]] std::string getOriginalSql() const;
+    [[nodiscard]] std::vector<LogicalOperator> getRootOperators() const;
+
+    [[nodiscard]] LogicalPlan withRootOperators(const std::vector<LogicalOperator>& operators) const;
+    [[nodiscard]] LogicalPlan withAdditionalRoots(const std::vector<LogicalOperator>& operators) const;
```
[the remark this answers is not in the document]
yschroeder97: I didn't put it here because I was simply "guessing" that someone will need, but because it IS used in distributed, and I have repeatedly been in the situation there where I want to add new roots to the plan, keeping the old ones.

nes-logical-operators/include/Plans/LogicalPlan.hpp
yschroeder97: I don't really understand what you mean by "equivalence relation is call-side context specific". 

The reasons I did this change were
a) it only compares for the id currently anyway, so why not allow for more freedom in the current implementation? 
b) in distributed, I have code that uses exactly this freedom. I store operator id's in a map that point to some other information. Then, at some later point, I want to find the operators in the plan, but at this point, they have changed their other contents. 

We could either keep it this way until we explicitly want something else (with my change, its not like something that was previously possible will not be possible anymore, because you can just extract the id of the operator), or if you insist on the change, I can make an overload. But I do NEED the functionality of just passing in the operator by id.

nes-logical-operators/src/Operators/LogicalOperator.cpp
yschroeder97: Its not only used in testing. Everything in this PR has its roots in the changes I saw as necessary and implemented in distributed. The only reason some of these functions are not used or only in testing is that otherwise I would need to pull distributed into this PR. But the whole point of this PR is to make the upcoming distributed PR smaller.

The idea with using a proper unordered set was that traits will only be overwritten if they are identical via their hash, not if they are of the same type but contain different data, for example. 

I can add a comment though, explaining this behaviour. Would this be enough for now?

nes-query-optimizer/include/Traits/Trait.hpp
yschroeder97: True, removed.

adrianmichalke (approved):

nes-logical-operators/src/Operators/LogicalOperator.cpp
Artraxon: hmm, I am ok with the function itself, but it should be clear from the signature that it doesn't overwrite. Maybe call it tryAddTraits or something similar

nes-query-optimizer/include/Traits/Trait.hpp
```diff
@@ -45,6 +48,28 @@ struct TraitConcept
...
+
+    [[nodiscard]] const std::type_info& getType() const override { return typeid(Derived); }
+
+    [[nodiscard]] SerializableTrait serialize() const override
+    {
+        SerializableTrait trait;
+        trait.set_trait_type(getType().name());
+        return trait;
+    }
 };
```
Artraxon, line 73: I think having these as a mixin availale is a good idea, but please make these methods final

nes-query-optimizer/include/Traits/Trait.hpp
```diff
@@ -136,14 +165,36 @@ struct Trait
             return false;
         }
 
+        [[nodiscard]] size_t hash() const override { return data.hash(); }
```
Artraxon, line 168: you could also just call std::hash in the type erased wrapper, but since there is an equals here as well... I guess its fine (resolve me)

Artraxon (changes requested):

nes-logical-operators/src/Operators/LogicalOperator.cpp
yschroeder97: Called it `addAdditionalTraits`, which should clarify that it adds them to the already xisting ones.

nes-query-optimizer/include/Traits/Trait.hpp
yschroeder97: Done.

Artraxon (approved):

PR #1011, opened 2025-07-12
fix(logger): remove NES_FATAL_ERROR

### LINKED ISSUE #1010: [BUG] NES_FATAL_ERROR logging fn is not fatal
### Version / Commit

0

### What happened?

In places like https://github.com/nebulastream/nebulastream/blob/81de5eb1f1c35191d96db3ad462439663b152155/nes-input-formatters/src/CSVInputFormatter.cpp#L391, `NES_FATAL_ERROR` is used for error handling. I was able to hit the case during fuzzing, but nebulastream did neither crash nor exit but just logged the error and chugged along happily.

I expect `NES_FATAL_ERROR` terminate nebulastream in some way.

Image

### How to reproduce

```cpp
NES::Logger::setupLogging("client.log", NES::LogLevel::LOG_ERROR);
NES_FATAL_ERROR("'fatal' error");
std::unreachable();
```

### Operating System

0

### Deployment details

_No response_

### Anything else

_No response_

### Code of Conduct

fwc:
Currently, `NES_FATAL_ERROR` logs on `spdlog::level::critical`, but it does not cause nebulastream to terminate in any way.

~~With this change, `NES_FATAL_ERROR` actually becomes, well, _fatal_.~~ Removes `NES_FATAL_ERROR` to reduce confusion [as suggested by @adrianmichalke ](https://github.com/nebulastream/nebulastream/pull/1011#issuecomment-3069985944)

Fixes #1010


keyseven123 (approved):

adrianmichalke:
Hm, after looking at the current occurrences of NES_FATAL_ERROR, most occurrences should not lead to a complete termination without option to catch and handle the error. How about just swapping them against exceptions?

fwc:
> Consequently, we could remove the logging option NES_FATAL_ERROR to avoid confusion

I think this is a good Idea :) I will adapt the PR accordingly.

If one wants to log & kill, one could still do e.g.

```
INVARIANT(false, "this should not be reachable");
```

keyseven123 (approved):

adrianmichalke (approved):
Very Cool!

PR #1013, opened 2025-07-14
fix use-after-free when tuplebuf debugging enabled

fwc:
Reapplies 86149e978e2a8e9454ee17a6192fd682e9ea0024 as it as silently reverted in faa1d27a878d26707c8be872c52f2a5f492f15bc.

Fixes failure of [`tupleBuffer_debugging` tests](https://github.com/nebulastream/nebulastream/actions/runs/16257008404/job/45894977179)


adrianmichalke (approved):
Thx, good catch

ls-1801 (approved):

PR #1015, opened 2025-07-14
feat(Systest): `--log-path` option to set the logging path

adrianmichalke:


keyseven123 (commented):
I am a little bit confused about this PR about what I should use in the future to set the correct logging path if I use the docker. I can either set it via the `HOST_LOGGING_PATH` env or via `--logPath`.
IMO, we can get rid of the `HOST_LOGGING_PATH` env. This would also clean up the method `SystestExecutor::setupLogging()`.

Artraxon (approved):

PR #1019, opened 2025-07-15
feat(QueryCompiler): enable dump compilation of results via --worker.dumpCompilationResult

adrianmichalke:


nes-single-node-worker/src/SingleNodeWorker.cpp
```diff
@@ -70,6 +72,7 @@ std::expected<QueryId, Exception> SingleNodeWorker::registerQuery(LogicalPlan pl
         auto queryPlan = optimizer->optimize(plan);
         listener->onEvent(SubmitQuerySystemEvent{queryPlan.getQueryId(), explain(plan, ExplainVerbosity::Debug)});
         auto request = std::make_unique<QueryCompilation::QueryCompilationRequest>(queryPlan);
+        request->dumpCompilationResult = static_cast<DumpMode>(configuration.workerConfiguration.dumpCompilationResult.getValue());
```
Artraxon, line 75: you can remove the static cast

nes-query-compiler/src/Phases/LowerToCompiledQueryPlanPhase.cpp
```diff
@@ -137,7 +141,8 @@ void processSink(const Predecessor& predecessor, const std::shared_ptr<Pipeline>
     it->predecessor.emplace_back(predecessor);
 }
 
-std::unique_ptr<ExecutablePipelineStage> getStage(const std::shared_ptr<Pipeline>& pipeline, ExecutionMode executionMode)
+std::unique_ptr<ExecutablePipelineStage>
+getStage(const std::shared_ptr<Pipeline>& pipeline, ExecutionMode executionMode, DumpMode dumpCompilationResult)
 {
```
Artraxon, line 146: long term we'll probably also want to move this class to a PIMPL implementation so that we can just set these options once as member fields and then can access them from every function. But for now its fine I think (resolved)

nes-query-compiler/private/Phases/LowerToCompiledQueryPlanPhase.hpp
```diff
@@ -15,10 +15,12 @@
...
 #include <memory>
+#include <Configuration/WorkerConfiguration.hpp>
+#include <Util/DumpMode.hpp>
 #include <CompiledQueryPlan.hpp>
 #include <PipelinedQueryPlan.hpp>
 
 namespace NES::QueryCompilation::LowerToCompiledQueryPlanPhase
 {
-std::unique_ptr<CompiledQueryPlan> apply(const std::shared_ptr<PipelinedQueryPlan>& pipelineQueryPlan);
+std::unique_ptr<CompiledQueryPlan> apply(const std::shared_ptr<PipelinedQueryPlan>& pipelineQueryPlan, DumpMode dumpCompilationResults);
```
Artraxon, line 25: I'd turn this subnamespace into a class and pass the parameter via IOC, I send you a patch

nes-runtime/include/Configuration/WorkerConfiguration.hpp
```diff
@@ -49,6 +54,9 @@ class WorkerConfiguration final : public BaseConfiguration
            "Number buffers in source local buffer pool. May be overwritten by a source-specific configuration (see SourceDescriptor).",
            {std::make_shared<NumberValidation>()}};
 
+    EnumOption<DumpMode> dumpCompilationResult = {
```
Artraxon, line 57: Do we have some way of automatically turning a enum option into all its values? If not, just leave it how it is

Artraxon (changes requested):

nes-runtime/include/Configuration/WorkerConfiguration.hpp
keyseven123: Maybe `magic_enum` has something?

keyseven123, line 57: I am sorry to be pedantic here. I would rename it to `dumpIRofQueryCompilation` as it dumps more than a result. It dumps all intermediate steps.

nes-query-compiler/src/Phases/LowerToCompiledQueryPlanPhase.cpp
```diff
@@ -154,9 +159,29 @@ std::unique_ptr<ExecutablePipelineStage> getStage(const std::shared_ptr<Pipeline
             INVARIANT(false, "Invalid backend");
         }
     }
-    options.setOption("toConsole", true);
-    options.setOption("toFile", true);
-
+    switch (dumpCompilationResult)
```
keyseven123: Maybe add a link to the [Nautilus Docs](https://github.com/nebulastream/nautilus/blob/main/docs/options.md)

keyseven123 (changes requested):

nes-runtime/include/Configuration/WorkerConfiguration.hpp
adrianmichalke: addressed

keyseven123 (approved):

PR #1025, opened 2025-07-18
feat(AVX): Added AVX support

keyseven123:


adrianmichalke (changes requested):
Can we add a `option(...)` in our main CMake for this? I guess we should always try to enable it in benchmark mode

keyseven123:
> Can we add a `option(...)` in our main CMake for this? I guess we should always try to enable it in benchmark mode

AVX is already enabled in the benchmark mode. In general the AVX flags are activated as soon as the option `NES_BUILD_NATIVE` is set. I also do not think that somebody wants to activate AVX while not building for the native processor.

ls-1801 (approved):

PR #1028, opened 2025-07-21
Enable concat in projection

alepping:
Prior, we could not use `CONCAT` in projections.
The PR adds new tests in `nes-systests/function/varsized/Concat.test` and introduces fixes to enable these tests.


nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -845,11 +845,6 @@ void AntlrSQLQueryPlanCreator::exitFunctionCall(AntlrSQLParser::FunctionCallCont
             /// Check if the function is a constructor for a datatype
             if (funcName == "CONCAT")
             {
-                if (helpers.top().functionBuilder.size() != 2)
-                {
-                    throw InvalidQuerySyntax(
-                        "Concat requires two arguments, but got {} at {}", helpers.top().functionBuilder.size(), context->getText());
-                }
```
adrianmichalke, line 852: Whats wrong with that check?

PR #1033, opened 2025-07-25
fixed access of MemoryProvider and MemoryLayout constructors

### LINKED ISSUE #980: [BUG] memoryProvider subclasses not checking schemas memory layout
### Version / Commit

6a0254040bd4f242f7e3a09f7e6554e971c1c81f

### What happened?

When creating a Row- or ColumnTupleBufferProvider directly using std::make_shared(), the Schema can include a memoryLayout that differs from the subclass instance. 
In this case the Schemas memoryLayout gets ignored resulting in unexpected behavior.

I would expect an error when trying to initialize ColumnTupleBudfferProvider with a schema with memoryLayout being row and vice versa.

### How to reproduce

create a Schema using ROW_LAYOUT.
create a ColumnTupleBufferProvider using the previously created schema.

### Operating System

Ubuntu 24.04.2 LTS

### Deployment details

_No response_

### Anything else

_No response_

### Code of Conduct

HerrTumorius:
This pull request changes how we create _TupleBufferMemoryProvider_ and _MemoryLayouts_.
Before they were several ways to do that, allowing to create e.g. _ColumnTupleBufferProvider_ and _ColumnLayout_ with a _Schema_ that has _MemoryLayoutType_ `ROW_LAYOUT`.
Now the _MemoryLayoutType_ of the _Schema_ is determining which instance is created.
Usage: 
```
MemoryLayout::create(bufferSize, schema) 
TupleBufferMemoryProvider::create(bufferSize, schema)
```

returns a _MemoryLayout_ which can be converted into the _MemoryLayoutType_ used in the _Schema_,
e.g. `NES::Util::as<ColumnLayout>(createdMemoryLayout)`

Similar for _TupleBufferMemoryProvider_

## Verifying this change
This changes are tested by existing Column, Row and Dynamic MemoryLayoutTests now using the create and as functions.

- ExecutionEngine
- QueryCompiler

## Documentation

This PR closes #980


nes-nautilus/src/Nautilus/Interface/MemoryProvider/TupleBufferMemoryProvider.cpp
```diff
@@ -114,16 +115,23 @@ TupleBufferMemoryProvider::~TupleBufferMemoryProvider() = default;
...
+            return std::make_shared<RowTupleBufferMemoryProvider>(RowTupleBufferMemoryProvider{rowMemoryLayout});
+        }
+        case Schema::MemoryLayoutType::COLUMNAR_LAYOUT: {
+            auto columnMemoryLayout
+                = NES::Util::as<Memory::MemoryLayouts::ColumnLayout>(Memory::MemoryLayouts::MemoryLayout::create(bufferSize, schema));
+            return std::make_shared<ColumnTupleBufferMemoryProvider>(ColumnTupleBufferMemoryProvider{columnMemoryLayout});
+        }
+            /// should never get accessed since creating the memoryLayout already checks schemas memoryLayoutType.
+        default:
+            throw NotImplemented("Currently only row and column layout are supported");
```
ls-1801, line 134: I would not have these default cases. 

If you remove the default case, then anyone adding a new MemoryLayoutType is required to add an implementation. If you have a default case, this would compile, but throw at runtime.

Remove the default case and add a std::unreachable() after the switch to make it compile.

ls-1801 (changes requested):
please remove the default cases from the two switch statements.

PR #1058, opened 2025-08-07
feat(Utils): Add strong type wrapper around string for URIs

### LINKED ISSUE #1008: [TASK] Strong Type for URI
### Description

We validate URIs in the config validation, but then pass out a string.
We should have a strong type for URIs, so that you can't pass something non-valid by accident.

### Acceptance Criteria

There is a strong type for URIs and all places where URIs are currently passed as string (for example SystestRunner) use it.

CatchACode:
Add a strong type wrapper for URI. This uses Boost URL to validate that the input string is a valid URI. As such `boost-url` was added as a dependency

This closes #1008


adrianmichalke (approved):

nes-common/include/Util/URI.hpp
```diff
@@ -0,0 +1,89 @@
...
+#include <yaml-cpp/node/convert.h>
+#include <yaml-cpp/node/node.h>
+
+namespace NES
+{
+
+class URI
+{
+public:
+    URI() noexcept = default;
```
keyseven123, line 33: Please write some documentation. Why do we need the class? Why can we not use boost::url directly?

nes-common/include/Util/URI.hpp
```diff
@@ -0,0 +1,89 @@
...
+        if (!result)
+        {
+            throw std::invalid_argument(result.error().message());
+        }
+        uri = boost::urls::url(*result);
+    }
+
+    static std::optional<URI> tryParse(const std::string_view input) noexcept
+    {
+        if (!boost::urls::parse_uri(input))
```
keyseven123, line 47: So I assume that parsing is the sole reason for using `boost::urls`?

vcpkg/vcpkg.json
```diff
@@ -41,6 +41,7 @@
     "nautilus",
     "replxx",
     "spdlog",
-    "yaml-cpp"
+    "yaml-cpp",
+    "boost-url"
```
keyseven123, line 45: Just the general advice on us being careful of integrating boost. I just remember the pain and hassle it was to get rid of it.

keyseven123 (changes requested):

vcpkg/vcpkg.json
CatchACode: We could use another library for URL parsing, like [ada-url](https://github.com/ada-url/ada) to avoid bringing in other boost. Or we could just use a [regexp](https://mathiasbynens.be/demo/url-regex) that fits our use cases

nes-single-node-worker/src/SingleNodeWorkerStarter.cpp
```diff
@@ -38,11 +38,11 @@ int main(const int argc, const char* argv[])
         NES::GRPCServer workerService{NES::SingleNodeWorker(*configuration)};
 
         grpc::ServerBuilder builder;
-        builder.AddListeningPort(configuration->grpcAddressUri, grpc::InsecureServerCredentials());
+        builder.AddListeningPort(configuration->grpcAddressUri.toString(), grpc::InsecureServerCredentials());
         builder.RegisterService(&workerService);
 
         const auto server = builder.BuildAndStart();
-        NES_INFO("Server listening on {}", static_cast<const std::string&>(configuration->grpcAddressUri));
+        NES_INFO("Server listening on {}", configuration->grpcAddressUri.toString());
```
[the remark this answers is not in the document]
CatchACode: I have expanded the `FMT_OSTREAM` macro onto `ScalarOption<T>` you can now use the option type directly in the logging macro

keyseven123 (approved):

PR #1072, opened 2025-08-14
1046 23h nightly runs

### LINKED ISSUE #1046: [TASK] 24h Nightly Runs
### Description

Building on top of systest's generator source, we would like to regularly stress test NebulaStream for 24h to find e.g. potential memory issues.

As a starting point, each nightly run should start one systest query with endless data generation. The CI workflow should report (during the following nightly run?) if the NES succeeded. One should be able to configure the run queries in the workflow file.   

### Acceptance Criteria

- [ ] nightly job starts 24h systest workflow

CatchACode:
This PR adds a 23h long Systest and a workflow to the Nightly CI that runs this Systest. For this End the a new void sink was added as plugin, as the number of tuples outputted is indeterminate and quickly grows to multiple GB. As such neither the existing checksum sink or file sink were ideal. The length was chosen as 23h so that there is a small buffer where building and other depended upon CI jobs can finish.

This PR closes #1046

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -73,3 +73,12 @@ jobs:
     with:
       dev_image_tag: ${{ needs.get-dev-images.outputs.image-tag }}
       ref: ${{ github.event.pull_request.head.ref }}
+
+  test-nebuli-and-single-node-worker:
+    needs: [ get-dev-images ]
+    name: "Run a query using Nebuli and the single node worker"
+    uses: ./.github/workflows/execute_query.yml
+    with:
+      ref: ${{ github.event.pull_request.head.sha }}
+      dev_image_tag: ${{ needs.get-dev-images.outputs.image-tag }}
+      runtime_seconds: '$((1*60))'


ls-1801 (changes requested):
I'd prefer to have a simpler approach:

1. Instead of using systest I would want to use the single-node-worker + client approach, after all that's what we are testing.
2. Instead of a void sink we can use a file sink that writes to /dev/null
3. I think we can keep the entire logic in a github workflow:

```bash
nes-single-node-worker --... &
QUERY_ID=$(nes-nebuli register -s localhost:8080 -x nightly-query.yml)
sleep 23h
nes-nebuli stop -s localhost:8080 $QUERY_ID
pkill -SIGTERM nes-single-node-worker
```

CatchACode:
I have reworked the PR to use nebuli and the single node worker. Everything is contained within a single new workflow. I tried using a normal file sink writing to `/dev/null` but this does not work as the file sink attempts to delete the output file if it exists.

ls-1801 (changes requested):
1. Cleanup the VoidSink
2. Make GH Action a composible workflow which we can invoke from the PR workflow or the Nightly. The duration should be configurable.
3. Maybe it would make sense to run the single-node-worker in the /usr/bin/time command that way we can also monitor memory usage.

PR #1076, opened 2025-08-18
fix(QueryEngine): Correcter behavior for start/stopping pipelines fast

Artraxon:
Fixes a bug that currently occurs regularly in the repl-test when starting and stopping queries after another really fast.

Previously, if the engine was requested to shut down, Start and Stop Pipeline tasks returned false. This prevented the onComplete handler of a task from running, which in turn prevented the query from releasing and completing.

QueryEngine


ls-1801:
I hopefully fixed the issue in #1092

PR #1079, opened 2025-08-19
extract duplicate TimestampField class

fwc:


ls-1801 (approved):

keyseven123 (approved):
If you want, you could also extract `getTimestampLeftAndRight()`. From purely glancing over it, the code seems almost identical. But this is also fine and can be done in a later PR.

PR #1085, opened 2025-08-21
Vcpkg triplet refactor, small cmake & docker improvements

fwc:


fwc:
> What is the reason behind having a custom triplet for bits?

I'm not sure if I understand the question. The refactor extracts code snippets from the triplets and puts these pieces (one might say "these bits") into the `bits` folder. So the files in the `bits` folder are not triplets but rather pieces (or "bits") of the triplets.

This way, if eg dependency `foo` needs the '-Wno-error-bar` flag set with `tsan`, we do not have to change both `tsan` triplet files but only the `bits/sanitizers/tsan.cmake` file.

keyseven123 (approved):

PR #1090, opened 2025-08-25
Fixed Systest benchmarking gaslighting

keyseven123:
If we used the benchmark flag `-b` and also have a non parsing or compiling query, we would simply skip it and not throwing an error. The current version uses the general runQueries() method with a custom SuccessCallback

## Verifying this change
This change is tested by running our CI tests and the newly added benchmark systest.

- Systests


nes-systests/systest/src/SystestRunner.cpp
```diff
@@ -353,15 +320,15 @@ std::vector<RunningQuery> runQueriesAtLocalWorker(
 {
     auto embeddedQueryManager = std::make_unique<EmbeddedWorkerQueryManager>(configuration);
     QuerySubmitter submitter(std::move(embeddedQueryManager));
-    return runQueries(queries, numConcurrentQueries, submitter);
+    return runQueries(queries, numConcurrentQueries, submitter, [](const auto&) { return ""; });
```
ls-1801, line 323: I prefer a `discardPerformanceMessage` function instead, which is just a noop.

ls-1801 (approved):

PR #1092, opened 2025-08-25
Fix start/stop race condition in query engine 

ls-1801:
This PR addresses a race condition which could occur when a query start races with a query stop. 
The issue was caused by the pipelineSetupCallback which is used once all pipelines have been started, and is intended to start all sources, finding an already destroyed running query plan object. 
The running query plan object has been destroyed by a concurrent query stop.

The fix disables the callback (i.e. it is never called) before destroying the running query plan.
This means either the running query plan is destroyed before the callback is invoked, which will skip the callback, or the callback is invoked which will block the destructor of the running query plan until the callback completes.
The running query plan then has to stop the sources.

Additionally, this PR refactors the Task Callbacks which are invoked after a Task has been handled.
Previously, the success callback was invoked when a task was handled successfully and a failure callback was invoked when the task threw an exception.
This means that tasks that were not handled successfully, but did not throw (i.e. when the task is skipped during termination) would not call any callback.
This PR introduces a third callback `onComplete` which is called regardless, and moves most of the existing callbacks into the `onComplete` category.

## Verifying this change
- Covered by systests and the new repl-test

- QueryEngine

## Documentation

### DIFF
--- a/nes-query-engine/QueryEngine.cpp
+++ b/nes-query-engine/QueryEngine.cpp
@@ -819,7 +819,7 @@ void QueryCatalog::start(
             const auto timestamp = std::chrono::system_clock::now();
             if (const auto locked = state.lock())
             {
-                locked->transition(
+                const auto didTransition = locked->transition(
                     [](Starting&& starting) /// NOLINT(cppcoreguidelines-rvalue-reference-param-not-moved)
                     {
                         RunningQueryPlan::dispose(std::move(starting.plan));
@@ -835,8 +835,11 @@ void QueryCatalog::start(
                         StoppingQueryPlan::dispose(std::move(stopping.plan));
                         return Terminated{Terminated::Stopped};
                     });
-                listener->logQueryStatusChange(queryId, QueryState::Stopped, timestamp);
-                statistic->onEvent(QueryStop(ThreadPool::WorkerThread::id, queryId));
+                if (didTransition)
+                {
+                    listener->logQueryStatusChange(queryId, QueryState::Stopped, timestamp);
+                    statistic->onEvent(QueryStop(ThreadPool::WorkerThread::id, queryId));
+                }
             }
         }
 
@@ -874,22 +877,19 @@ void QueryCatalog::stopQuery(QueryId id)
 {
     const std::unique_ptr<RunningQueryPlan> toBeDeleted;
     {
-        CallbackRef ref;
         const std::scoped_lock lock(mutex);
         if (auto it = queryStates.find(id); it != queryStates.end())
         {
             auto& state = *it->second;
             state.transition(
-                [&ref](Starting&& starting) /// NOLINT(cppcoreguidelines-rvalue-reference-param-not-moved)
+                [](Starting&& starting) /// NOLINT(cppcoreguidelines-rvalue-reference-param-not-moved)
                 {
-                    auto [stoppingQueryPlan, callback] = RunningQueryPlan::stop(std::move(starting.plan));
-                    ref = callback;
+                    auto stoppingQueryPlan = RunningQueryPlan::stop(std::move(starting.plan));
                     return Stopping{std::move(stoppingQueryPlan)};
                 },
-                [&ref](Running&& running) /// NOLINT(cppcoreguidelines-rvalue-reference-param-not-moved)
+                [](Running&& running) /// NOLINT(cppcoreguidelines-rvalue-reference-param-not-moved)
                 {
-                    auto [stoppingQueryPlan, callback] = RunningQueryPlan::stop(std::move(running.plan));
-                    ref = callback;
+                    auto stoppingQueryPlan = RunningQueryPlan::stop(std::move(running.plan));
                     return Stopping{std::move(stoppingQueryPlan)};
                 });
         }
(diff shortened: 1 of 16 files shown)


nes-query-engine/QueryEngine.cpp
```diff
@@ -55,6 +55,43 @@
 namespace NES
 {
 
+namespace
+{
+
+/// This function is unsafe because it requires the lifetime of the RunningQueryPlanNode exceed the lifetime of the callback
+auto injectQueryFailureUnsafe(RunningQueryPlanNode* node, BaseTask::onFailure failure)
```
Artraxon, line 62: why is it not possible to do this safely? Why not use a shared pointer in the caller and pass that?

Artraxon (changes requested):
thank you for debugging this and also the other iterative improvements

nes-query-engine/QueryEngine.cpp
```diff
@@ -355,19 +370,20 @@ class ThreadPool : public WorkEmitter, public QueryLifetimeController
...
+        /// Boolean return value indicates if the onSuccess should be called
+        bool operator()(WorkTask& task) const;
+        bool operator()(StopQueryTask& stopQuery) const;
+        bool operator()(StartQueryTask& startQuery) const;
+        bool operator()(StartPipelineTask& startPipeline) const;
+        bool operator()(PendingPipelineStopTask& pendingPipelineStop) const;
+        bool operator()(StopPipelineTask& stopPipelineTask) const;
+        bool operator()(StopSourceTask& stopSource) const;
+        bool operator()(FailSourceTask& failSource) const;
+
```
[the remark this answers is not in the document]
ls-1801: I may have to take their TaskCallback e.g. if I have to repeat the task.

nes-query-engine/QueryEngine.cpp
ls-1801: I mean that's the reason why it's called unsafe. There is a safe version that uses a weak_ptr. But in the location where the unsafe version is used, there is only a unique_ptr.

But actually there is no need to use a pointer but you can use a reference, granted you could still have lifetime issues.

nes-query-engine/Task.hpp
```diff
@@ -32,65 +36,129 @@ struct RunningQueryPlanNode;
...
+    onFailure onFailureCallback;
+};
+
+class BaseTask
+{
+public:
+    using onComplete = TaskCallback::onComplete;
+    using onSuccess = TaskCallback::onSuccess;
+    using onFailure = TaskCallback::onFailure;
+
```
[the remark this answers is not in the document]
ls-1801: no i removed them

nes-query-engine/Task.hpp
```diff
@@ -32,65 +36,129 @@ struct RunningQueryPlanNode;
...
+/// The failure callback is invoked if the task fails with an exception.
+/// Callbacks are move only, thus every callback is invoked at most once.
+class TaskCallback
 {
 public:
-    using onComplete = std::function<void()>;
-    using onFailure = std::function<void(Exception)>;
+    using onComplete = absl::AnyInvocable<void()>;
+    using onSuccess = absl::AnyInvocable<void()>;
+    using onFailure = absl::AnyInvocable<void(Exception)>;
```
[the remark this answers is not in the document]
Artraxon: I personally don't think thats necessary

Artraxon (approved):

nes-query-engine/QueryEngine.cpp
```diff
@@ -626,15 +629,15 @@ bool ThreadPool::WorkerThread::operator()(StartQueryTask& startQuery) const
     return false;
 }
 
-bool ThreadPool::WorkerThread::operator()(const StopSourceTask& stopSource) const
+bool ThreadPool::WorkerThread::operator()(StopSourceTask& stopSource) const
```
alepping, line 632: You need to pass by reference, because thats the 'signature' of the variant, thus you 'reconstruct' a StopSourceTask below in `addTaskOrDoNextTask`?

nes-query-engine/Task.cpp
```diff
@@ -14,22 +14,165 @@
...
+}
+
+void completeTask(Task& task)
+{
+    std::visit([](auto& specificTask) { return specificTask.complete(); }, task);
+}
+
+void failTask(Task& task, Exception exception)
+{
+    std::visit([&](auto& specificTask) { return specificTask.fail(std::move(exception)); }, task);
```
alepping, line 175:
```suggestion
std::visit([exception = std::move(exception)](auto& specificTask) { return specificTask.fail(std::move(exception)); }, task);
```

nes-query-engine/Task.hpp
```diff
@@ -32,67 +36,97 @@ struct RunningQueryPlanNode;
...
```
alepping, line 50: As mentioned in an earlier comment, since onSuccess and onComplete have the same signature, they can be accidentally misused.

nes-query-engine/Task.cpp
alepping: (resolve me): My main gripe was `[&]`, but thanks for pointing out the unnecessary move

alepping (changes requested):
It seems like you did not change the signature of the `TaskCallback`, but you intended to?
Also, this question is still unanswered:
Did you manage to reproduce the race condition (locally) with the changes in the TestSource.cpp?

alepping (approved):
Misinterpreted your changes. Thanks for the fix!

PR #1095, opened 2025-08-26
Fixes incorrectly thrown "Schema Mismatch" errors for sinks with duplicate fields in compareSchemas

### LINKED ISSUE #1002: [BUG] Systest Tool does not accept duplicate fields
### Version / Commit

main

### What happened?

```
Source stream UINT64 id INLINE
1
2
3
4
5
6
7

Sink stream_sink UINT64 stream$id UINT64 stream$id

SELECT id, id FROM stream INTO stream_sink
----
1,1
2,2
3,3
4,4
5,5
6,6
7,7
```

Produces error:

```
Error: 

Schema Mismatch
---------------
+ 'Field(name: stream$id, DataType: DataType(type: UINT64))' is unexpected field in actual result schema.

All Results match
```

Systest should verify that both columns are equal or reject duplicate fields, with a error message.

### How to reproduce

_No response_

### Operating System

linux

### Deployment details

_No response_

### Anything else

_No response_

### Code of Conduct

LeoRaasch:
This pull request fixes a bug that would incorrectly throw a Schema Mismatch error for systests using sinks with duplicate fields. This was caused by the `compareSchemas` method  in  `SystestResultCheck.cpp` only using the first of the duplicate fields of the actual schema to match the duplicate fields of the expected schema. Thus, the other duplicates remained unmatched, causing the error later on.

The same function used to create a map between the indexes of actual and expected fields, so that the query results could even be compared if the order of fields in the actual schema is wrong. However, if we are dealing with duplicated fields, the correct field index map is ambiguous. Rather than trying every possible mapping of indexes, we decided to remove the map and only compare the results if the actual schema is equal to the expected one.

In order to make up for the potential loss of error information, we created a SchemaDiff class that compares two schemas and gives detailed information on why the schemas do not match.

- Removed `ExpectedToActualFieldMap` struct. Every function that used this map now assumes that the fields of the result are in the right order
- Added `SchemaDiff` struct to compare schemas and give detailed output if the schemas mismatch

## Verifying this change
This change is tested by

- Newly added `nes-systests/bug/DuplicateFieldsSink.test` systest, which matches the test causing the bug in the corresponding issue
- `SchemaDiffTest` suite in `SchemaDiffTest.cpp` to test the SchemaDiff struct

- nes-systests
- nes-data-types (Schema only)

## Documentation
- Left in-code documentation explaining the changes

This PR closes #1002

### DIFF
--- a/nes-data-types/include/DataTypes/Schema.hpp
+++ b/nes-data-types/include/DataTypes/Schema.hpp
@@ -23,6 +23,7 @@
 #include <string_view>
 #include <unordered_map>
 #include <utility>
+#include <variant>
 #include <vector>
 
 #include <DataTypes/DataType.hpp>
@@ -145,8 +146,13 @@ struct SchemaDiff
         size_t actualSize{};
     };
 
-    std::optional<SizeMismatch> sizeMismatch;
-    std::vector<FieldMismatch> fieldMismatches;
+    struct NoMismatch
+    {
+    };
+
+    /// Variant to hold all possible diff outcomes: No mismatches, a size mismatch, or field mismatches.
+    /// Will be default constructed as NoMismatch.
+    std::variant<NoMismatch, SizeMismatch, std::vector<FieldMismatch>> result;
 
     static SchemaDiff of(const Schema& expectedSchema, const Schema& actualSchema);
 
--- a/nes-data-types/src/DataTypes/Schema.cpp
+++ b/nes-data-types/src/DataTypes/Schema.cpp
@@ -21,6 +21,7 @@
 #include <string>
 #include <string_view>
 #include <utility>
+#include <variant>
 #include <vector>
 #include <DataTypes/DataType.hpp>
 #include <Util/Logger/Logger.hpp>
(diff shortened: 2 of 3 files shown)


ls-1801:
@LeoRaasch I have an implementation for a SchemaDiff helper here: [8c8446727560f9a9c1db72f1a48f43d612771070](https://github.com/nebulastream/nebulastream/commit/8c8446727560f9a9c1db72f1a48f43d612771070)

keyseven123 (changes requested):
I agree with @ls-1801 . Please re-request once you include his feedback.

nes-data-types/tests/UnitTests/API/SchemaTest.cpp
```diff
@@ -359,4 +359,396 @@ TEST_F(SchemaTest, withoutSourceQualifierTest)
     }
 }
 
+TEST_F(SchemaTest, diffSchemasIsDifferentTest)
```
keyseven123, line 362: I would move this to a new SchemaDiffTest file and then have each block a separate TEST

keyseven123 (changes requested):
Just one small thing, otherwise good job!

nes-data-types/include/DataTypes/Schema.hpp
```diff
@@ -128,6 +128,27 @@ class Schema
...
 
+/// Checks if two schemas are equal and keeps track of every mismatch and missing / unexpected fields
+struct SchemaDiff
+{
+    struct SchemaMismatch
+    {
+        size_t index{};
+        Schema::Field expectedField;
+        Schema::Field actualField;
+    };
```
ls-1801, line 139: Given that the schema diff is order-sensitive, we can make the check a lot easier by simply asserting that the size matches and that each field matches. 

It is probably overkill to try to guess what order could be the correct order and thus determine which fields are missing or additional

nes-data-types/tests/UnitTests/API/SchemaTest.cpp
```diff
@@ -359,4 +359,396 @@ TEST_F(SchemaTest, withoutSourceQualifierTest)
...
+        Schema actualSchema;
+        actualSchema.addField("field1", DataType::Type::INT64);
+        /// field2 is missing
+        actualSchema.addField("field3", DataType::Type::BOOLEAN);
+
+        auto diff = SchemaDiff::of(expectedSchema, actualSchema);
+
+        EXPECT_EQ(diff.missingFields.size(), 1);
+        EXPECT_EQ(diff.missingFields[0].name, "field3");
+        EXPECT_EQ(diff.missingFields[0].dataType.type, DataType::Type::BOOLEAN);
```
ls-1801, line 492: This is problematic; this should just report that there is a size mismatch. Reporting that Field3 is missing is just misleading.

ls-1801 (changes requested):
I would dial down the SchemaDiff class to first assert that the size matches and than report individual field missmatches.

nes-data-types/include/DataTypes/Schema.hpp
```diff
@@ -128,6 +128,32 @@ class Schema
...
+
+    struct SizeMismatch
+    {
+        size_t expectedSize{};
+        size_t actualSize{};
+    };
+
+    std::optional<SizeMismatch> sizeMismatch;
+    std::vector<FieldMismatch> fieldMismatches;
```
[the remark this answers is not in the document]
LeoRaasch: That is a good idea. I implemented this in the newest commit.

PR #1109, opened 2025-09-05
fix(benchmarks): missing snake case parameter

Artraxon:
We missed one large scale benchmark when changing parameters to snake case, this fixes it.


ls-1801 (approved):

keyseven123 (approved):
Maybe we can add them to the PR run or at least the nightly run?

PR #1127, opened 2025-09-16
feat(grpc): enable health service

ls-1801:
This commit enables the grpc health service, which is provided by grpc out of the box. The health service exposes a grpc endpoint that can be queried to detect if the worker is running. 
This commit installs the `grpc-health-probe` into the single-node-worker and nebuli docker images. The probe can be used to specify a running condition in docker-compose or kubernetes.

### DIFF
--- a/docker/nebuli/Nebuli.dockerfile
+++ b/docker/nebuli/Nebuli.dockerfile
@@ -13,13 +13,17 @@ RUN cd /home/ubuntu/src \
 # the binary is linked against libc++, thus we install it
 FROM ubuntu:24.04 AS app
 ENV LLVM_TOOLCHAIN_VERSION=19
-RUN apt update -y && apt install curl gpg -y
+RUN apt update -y && apt install wget curl gpg -y
 RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor -o /etc/apt/keyrings/llvm-snapshot.gpg \
     && chmod a+r /etc/apt/keyrings/llvm-snapshot.gpg \
     && echo "deb [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/llvm-snapshot.gpg] http://apt.llvm.org/"$(. /etc/os-release && echo "$VERSION_CODENAME")"/ llvm-toolchain-"$(. /etc/os-release && echo "$VERSION_CODENAME")"-${LLVM_TOOLCHAIN_VERSION} main" > /etc/apt/sources.list.d/llvm-snapshot.list \
     && echo "deb-src [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/llvm-snapshot.gpg] http://apt.llvm.org/"$(. /etc/os-release && echo "$VERSION_CODENAME")"/ llvm-toolchain-"$(. /etc/os-release && echo "$VERSION_CODENAME")"-${LLVM_TOOLCHAIN_VERSION} main" >> /etc/apt/sources.list.d/llvm-snapshot.list \
     && apt update -y \
     && apt install -y libc++1-${LLVM_TOOLCHAIN_VERSION} libc++abi1-${LLVM_TOOLCHAIN_VERSION}
+
+RUN GRPC_HEALTH_PROBE_VERSION=v0.4.40 && \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
+    chmod +x /bin/grpc_health_probe
 
 COPY --from=build /tmp/bin /usr/bin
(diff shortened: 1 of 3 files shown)


docker/single-node-worker/SingleNodeWorker.dockerfile
```diff
@@ -13,13 +13,17 @@ RUN cd /home/ubuntu/src \
...
 RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor -o /etc/apt/keyrings/llvm-snapshot.gpg \
     && chmod a+r /etc/apt/keyrings/llvm-snapshot.gpg \
     && echo "deb [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/llvm-snapshot.gpg] http://apt.llvm.org/"$(. /etc/os-release && echo "$VERSION_CODENAME")"/ llvm-toolchain-"$(. /etc/os-release && echo "$VERSION_CODENAME")"-${LLVM_TOOLCHAIN_VERSION} main" > /etc/apt/sources.list.d/llvm-snapshot.list \
     && echo "deb-src [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/llvm-snapshot.gpg] http://apt.llvm.org/"$(. /etc/os-release && echo "$VERSION_CODENAME")"/ llvm-toolchain-"$(. /etc/os-release && echo "$VERSION_CODENAME")"-${LLVM_TOOLCHAIN_VERSION} main" >> /etc/apt/sources.list.d/llvm-snapshot.list \
     && apt update -y \
     && apt install -y libc++1-${LLVM_TOOLCHAIN_VERSION} libc++abi1-${LLVM_TOOLCHAIN_VERSION}
 
+RUN GRPC_HEALTH_PROBE_VERSION=v0.4.40 && \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
+    chmod +x /bin/grpc_health_probe
```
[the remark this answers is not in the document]
kykrueger: Interesting problem, it seems like there are other binaries for ARM available. We could use our envvar for system architecture to get the right one depending on which image we are building.

docker/nebuli/Nebuli.dockerfile
```diff
@@ -13,13 +13,17 @@ RUN cd /home/ubuntu/src \
 # the binary is linked against libc++, thus we install it
 FROM ubuntu:24.04 AS app
 ENV LLVM_TOOLCHAIN_VERSION=19
-RUN apt update -y && apt install curl gpg -y
+RUN apt update -y && apt install wget curl gpg -y
```
keyseven123, line 16: Do we need these? They should be part of the `Base.dockerfile`

docker/nebuli/Nebuli.dockerfile
```diff
@@ -13,13 +13,17 @@ RUN cd /home/ubuntu/src \
...
 RUN curl -fsSL https://apt.llvm.org/llvm-snapshot.gpg.key | gpg --dearmor -o /etc/apt/keyrings/llvm-snapshot.gpg \
     && chmod a+r /etc/apt/keyrings/llvm-snapshot.gpg \
     && echo "deb [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/llvm-snapshot.gpg] http://apt.llvm.org/"$(. /etc/os-release && echo "$VERSION_CODENAME")"/ llvm-toolchain-"$(. /etc/os-release && echo "$VERSION_CODENAME")"-${LLVM_TOOLCHAIN_VERSION} main" > /etc/apt/sources.list.d/llvm-snapshot.list \
     && echo "deb-src [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/llvm-snapshot.gpg] http://apt.llvm.org/"$(. /etc/os-release && echo "$VERSION_CODENAME")"/ llvm-toolchain-"$(. /etc/os-release && echo "$VERSION_CODENAME")"-${LLVM_TOOLCHAIN_VERSION} main" >> /etc/apt/sources.list.d/llvm-snapshot.list \
     && apt update -y \
     && apt install -y libc++1-${LLVM_TOOLCHAIN_VERSION} libc++abi1-${LLVM_TOOLCHAIN_VERSION}
 
+RUN GRPC_HEALTH_PROBE_VERSION=v0.4.40 && \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
```
alepping, line 25: How does it work for arm?

alepping (changes requested):
Just wondering whether `grpc_health_probe-linux-amd64` is hardcoded for amd64 and how we support ARM then?

docker/nebuli/Nebuli.dockerfile
ls-1801: good catch

PR #1129, opened 2025-09-16
TCP Source Refactor

ls-1801:
This PR refactors the TCP Source. 

In hindsight this is not required for the distributed PR, and I don't mind if we decide against modifications to the TCP source.

---

> [!NOTE]
> Refactors TCPSource with RAII socket/connection abstractions, timeout-aware connect/receive and reconnection, simplifies config, and adds a new `CannotProduceSourceBuffer` exception.
> 
> - **Sources • TCP**:
>   - **Refactor `TCPSource`**:
>     - Introduces `AddressInfo`, `OwnedDescriptor`, `Socket`, and `Connection` for RAII and error-checked socket ops.
>     - Uses non-blocking `connect` with `select`-based timeout; per-receive timeouts; automatic reconnect on `recv` errors.
>     - Switches to `std::expected`/`std::variant`; stores connection as `std::optional<Connection>`.
>   - **API/Logic updates**:
>     - `open` resolves host, creates socket, and connects via new abstractions.
>     - `fillTupleBuffer` reads via `Connection::receive`; maps failures to `CannotProduceSourceBuffer`; returns EoS on zero bytes.
>     - `close` clears connection; `toString` simplified.
>   - **Config cleanup**:
>     - Removes `tuple_delimiter`, `socket_buffer_size`, and `bytes_sed_for_socket_buffer_size_transfer`.
>     - `socket_port` now `uint16_t`; parameter map adjusted.
> - **Exceptions**:
>   - Adds `CannotProduceSourceBuffer` (code `4007`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2841216d52ebc445f5392edce6ee12411b080b69. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/nes-plugins/Sources/TCPSource/TCPSocket.cpp
+++ b/nes-plugins/Sources/TCPSource/TCPSocket.cpp
@@ -1,0 +1,308 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <TCPSocket.hpp>
+
+#include <array>
+#include <cerrno>
+#include <cstdint>
+#include <functional>
+#include <memory>
+#include <string>
+#include <netdb.h>
+#include <poll.h>
+#include <unistd.h>
+#include <Util/Files.hpp>
+#include <fmt/format.h>
+#include <sys/eventfd.h>
+#include <sys/socket.h>
+#include <ErrorHandling.hpp>
+
+namespace NES
+{
+
+namespace
+{
+
+struct ResolvedAddress
+{
+    std::unique_ptr<addrinfo, decltype(&freeaddrinfo)> info{nullptr, &freeaddrinfo};
+};
+
+using ResolveResult = std::variant<ResolvedAddress, TCPSocket::ConnectError>;
+
+ResolveResult resolve(const std::string& host, uint16_t port)
+{
+    addrinfo hints{};
+    hints.ai_family = AF_INET;
+    hints.ai_socktype = SOCK_STREAM;
+
+    addrinfo* rawResult = nullptr;
+    const std::string portStr = std::to_string(port);
+    const auto errorCode = getaddrinfo(host.c_str(), portStr.c_str(), &hints, &rawResult);
+    if (errorCode != 0)
+    {
+        return TCPSocket::ConnectError{fmt::format("Failed getaddrinfo with error: {}", gai_strerror(errorCode))};
+    }
+    return ResolvedAddress{.info = {rawResult, &freeaddrinfo}};
+}
+
+} /// anonymous namespace
+
+/// --- StopEventFD ---
+
+class StopEventFD
+{
+public:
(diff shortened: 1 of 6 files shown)


nes-plugins/Sources/TCPSource/TCPSource.cpp
```diff
@@ -48,225 +50,343 @@
...
+{
+    if (this == &other)
     {
-        ::close(sockfd); /// close socket to clean up state
-        throw wrapExternalException("Could not establich connection!");
+        return *this;
     }
+    descriptor = other.descriptor;
+    other.descriptor = -1;
+    return *this;
```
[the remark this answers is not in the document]
ls-1801: good catch

nes-plugins/Sources/TCPSource/TCPSource.hpp
```diff
@@ -39,6 +38,68 @@
...
+    static std::expected<AddressInfo, std::string> from(const std::string& host, uint16_t port);
+
+    auto operator*() const { return *result; }
+
+    auto operator->() const { return result.operator->(); }
+
+    std::unique_ptr<addrinfo, decltype(&freeaddrinfo)> result;
+};
+
+struct OwnedDescriptor
```
alepping: nit: I'd leave a comment on why the `OwnedDescriptor` is beneficial.
Why is a strong type not sufficient.

tekdogan (changes requested):
We used to raise an error when TCP server closes the connection. Now, it just stops the source without hinting the connection error (highlighted part is after the connection lost).

image

nes-plugins/Sources/TCPSource/TCPSocket.cpp
```diff
@@ -0,0 +1,308 @@
...
+TCPSocket::RecvResult TCPSocket::recvExact(std::span<uint8_t> buffer, std::chrono::milliseconds timeout, const std::stop_token& stopToken)
+{
+    /// Lazily create StopEventFD on first call with a stoppable token
+    if (!stopEvent && stopToken.stop_possible())
+    {
+        stopEvent = std::make_unique<StopEventFD>(stopToken);
+    }
+
+    /// Negative = block forever, 0 = non-blocking, positive = that many ms
+    const int timeoutMs = timeout.count() >= 0 ? static_cast<int>(timeout.count()) : -1;
```
tekdogan: When we set `flush_interval_ms` to zero:

- What explained in the README is: "When set to 0 (default), there is no timeout and the source will wait indefinitely for a connection." So the expected behavior is to wait.

- What is happening in the code: it polls 0 from your logic here, and returns `Timeout{}` below when no data available (it skips the if clause). This timeout returns null ptr in the `TCPSource::fillTupleBuffer()`, which can lead to CPU spin of the outer loop. Maybe this is by design (spin as long as no data arrives), but it could degrade the performance.

nes-plugins/Sources/TCPSource/TCPSource.cpp
```diff
@@ -14,275 +14,177 @@
...
+                            [&](std::span<uint8_t> data) -> std::optional<FillTupleBufferResult>
+                            {
+                                ++generatedBuffers;
+                                return FillTupleBufferResult::withBytes(data.size());
+                            },
+                            [](TCPSocket::Timeout&) -> std::optional<FillTupleBufferResult> { return std::nullopt; },
+                            [](Stopped&) -> std::optional<FillTupleBufferResult> { return FillTupleBufferResult::eos(); },
+                            [&](TCPSocket::EoS&) -> std::optional<FillTupleBufferResult>
+                            {
+                                if (autoReconnect)
```
tekdogan: `autoReconnect` is only used during the receiving state, but not during the connecting state above.

PR #1139, opened 2025-09-20
feat: Refactor TestTupleBuffer to use std::span instead of raw pointers

### LINKED ISSUE #1136: [TASK] Replace pointer with span in TestTupleBuffer
### Description

We should refactor the usage of pointers in our `TestTupleBuffer` with `std::span`.

### Acceptance Criteria

- [ ] No more pointers in `TestTupleBuffer`
- [ ] All unit and system-level tests pass

CellophaneFlowers:
Refactor: Replace raw pointers with std::span in TestTupleBuffer
Purpose of the Change and Brief Change Log
This pull request refactors the TestTupleBuffer to replace the usage of raw uint8_t* pointers with std::span<const uint8_t>, enhancing memory safety and clarity. The change list is as follows:
Modified DynamicField to use std::span<const uint8_t> for representing field data instead of a raw pointer.
Updated DynamicTuple to construct DynamicField instances with std::span.
Adjusted all internal accesses within DynamicField to utilize the std::span interface (e.g., data.data()).
Verifying this change
This change is tested by:
Ensuring all existing unit and system-level tests for TestTupleBuffer continue to pass.
Verifying that tuple reads and writes through DynamicTuple and DynamicField function correctly with the std::span implementation.
What components does this pull request potentially affect?
MemoryLayout (indirectly, as it provides offsets)
Runtime (specifically TupleBuffer, as it provides the underlying buffer for TestTupleBuffer)
Testing utilities and integration tests that rely on TestTupleBuffer.
Documentation
The change is reflected in the necessary documentation, e.g., design document, in-code documentation.
All necessary methods (no getter, setter, or constructor) have a documentation.
Issue Closed by this pull request:
This PR closes #1136
image


nes-memory/TestTupleBuffer.cpp
```diff
@@ -50,7 +51,7 @@ DynamicField DynamicTuple::operator[](const std::size_t fieldIndex) const
     auto* bufferBasePointer = buffer.getBuffer<uint8_t>();
     const auto offset = memoryLayout->getFieldOffset(tupleIndex, fieldIndex);
     auto* basePointer = bufferBasePointer + offset;
-    return DynamicField{basePointer, memoryLayout->getPhysicalType(fieldIndex)};
+    return DynamicField{std::span<const uint8_t>(basePointer, memoryLayout->getPhysicalType(fieldIndex).getSizeInBytes()), memoryLayout->getPhysicalType(fieldIndex)};
```
keyseven123, line 54: The conversion of the span, could be done here.

keyseven123 (changes requested):
Thanks for creating the PR. Couple of general high-level points.
- Please remove the file `.TestTupleBuffer.cpp.swp`.
- Please run the format target

If you want to go one step further, you could have a sstd::span of the type in the DynamicField. Then we do not need to cast the read and write operations to the specific type but rather would do the cast once in the constructor.

PR #1142, opened 2025-09-23
Fix(systest) Update systest to create sinks and sources via NES-SQL

### LINKED ISSUE #1113: [TASK] Update Systest to create sources and sinks with NES SQL syntaxt
### Description

Currently, the systests use a bespoke syntax to create sources and sinks. 

In this task, this behavior shall be changed to allow the creation of sources and sinks with the same SQL syntax as the one used in the REPL parser. It should still be possible to define the input data from the defined sources within the test files.  

### Acceptance Criteria

* The systest component can create logical and physical sources based on SQL queries within the test files that have the same `CREATE` syntax as used by the REPL parser.
* The systest component still accepts source input data from within the test file.
* All systests are updated to follow the new logic and continue to successfully test the system.

greenBene:
Currently, systest use a special logic to create sources and sinks. The goal of this PR is to unify the creation of sinks and sources in systest by using the same NES-SQL syntax as used by the repl parser.  

## Verifying this change
All existing systest shouls continue to work

- systest

This PR closes #1113


ls-1801 (changes requested):
Nice! some minor suggestions/questions.

nes-sql-parser/src/StatementBinder.cpp
```diff
@@ -64,17 +64,15 @@ namespace NES
...
 
 public:
     using Literal = std::variant<std::string, int64_t, uint64_t, double, bool>;
 
     Impl(
         const std::shared_ptr<const SourceCatalog>& sourceCatalog,
-        const std::shared_ptr<const SinkCatalog>& sinkCatalog,
         const std::function<LogicalPlan(AntlrSQLParser::QueryContext*)>& queryBinder)
-        : sourceCatalog(sourceCatalog), sinkCatalog(sinkCatalog), queryBinder(queryBinder)
+        : sourceCatalog(sourceCatalog), queryBinder(queryBinder)
```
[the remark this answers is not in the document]
greenBene: The idea is, that the StatementBinder creates Statements structs that should be logically correct, but not yet requires that the mentioned sources or sinks already exist. This should be done by the StatementHandler.

This allows us to manipulate the structs before they are actually created. This behavior is required in systest so the same query parser can be used here to parse `CREATE`  statements, as the systest system creates unique sinks for each `SELECT` query. 

Because of this change, the `sinkCataloge`  is no longer used in the StatementBinder and thus has been removed from the constructor of `StatementBinder::Impl`.

nes-sql-parser/src/StatementBinder.cpp
```diff
@@ -326,19 +324,29 @@ class StatementBinder::Impl
...
+                return bindConfigOptions(physicalSourceDefAST->options->namedConfigExpression());
+            }
+            return std::unordered_map<std::string, std::unordered_map<std::string, Literal>>{};
+        }();
 
-        auto configOptions = bindConfigOptions(physicalSourceDefAST->options->namedConfigExpression());
-        const auto parserConfigIter = configOptions.find("PARSER");
-        if (parserConfigIter == configOptions.end() || not parserConfigIter->second.contains("TYPE"))
+
+        auto configMap = [&]()
```
Artraxon, line 337: it would be good to keep a name here that indicated that this is the parser configs raw values, because this looks like its either everything or the source config

alepping:
Very cool, really looking forward to these changes and in general the unification of how we create sources/sinks.

From playing around with the new systest syntax:
1. forgetting the `;` at the end of a SQL statement causes a crash: `Loading test file:///home/rudi/dima/nebulastream-public/nes-systests/operator/projection/Projection.test failed: invalid query syntax; invalid query syntax; Antlr exception during parsing:  in CREATE LOGICAL SOURCE stream(id UINT64, value UINT64, timestamp UINT64)
CREATE PHYSICAL SOURCE FOR stream TYPE File;`, which is ok imo. It would be nicer if the parser could tell you that a `;` might be missing from the end of the statement.
2. An empty line between CREATE PHYSICAL SOURCE and ATTACH causes a crash: `Loading test file:///home/rudi/dima/nebulastream-public/nes-systests/operator/projection/Projection.test failed: invalid config parameter; Failed validation of config parameter: file_path, in: File`, this is a problem imo. The message is not helpful and since the SQL statement ends in an `;` it is not intuitive that no newline is allowed to follow after that. I'd expect that you fix the behavior and a newline is allowed.
3. A newline after ATTACH INLINE also causes a crash: `failed: unexpected token in sql logic test file; Should never run into the INVALID token during systest file parsing, but got line: 1,1,1000.`. Thats ok, but can be improved.
4. (relates to 1.) No `;` expected after a query statement, but `;` expected after CREATE statement (behavior should be uniform)
5. I noticed weird spacing in systest files (CREATE SINK), e.g., `JoinMultipleStream.test`: `CREATE SINK sinkStreamStream2Stream4Stream4_1(streamstream2stream4stream4_1.start UINT64, streamstream2stream4stream4_1.end	 UINT64, streamstream2stream4.start	 UINT64, streamstream2stream4.end	 UINT64, streamstream2.start	 UINT64, streamstream2.end	 UINT64, stream.id	 UINT64, stream.value	 UINT64, stream.timestamp	 UINT64, stream2.id2	 UINT64, stream2.value2	 UINT64, stream2.timestamp	 UINT64, stream4.id4	 UINT64, stream4.value4	 UINT64, stream4.timestamp	 UINT64, stream4_1.id4_1	 UINT64, stream4_1.value4_1	 UINT64, stream4_1.timestamp UINT64) TYPE File;`

Artraxon (changes requested):
Overall a really good change, finally moving to one unified interface.

I think my only complaint is that by default you should take values and move them. That way they also work with rvalues, which otherwise would require an additional copy.

It's not a big concern here since its not performance critical, but it's good to get in the habit of doing it that way.
Also, with pass-by-value and move you clearly indicate that you give up ownership over the values, and clang-tidy will warn you if you try to access a value after you have moved it. 
Basically, it gives you a very light, dumber version of rusts ownership semantics (which is still better then nothing).

nes-logical-operators/include/Operators/Windows/Aggregations/CountAggregationLogicalFunction.hpp
```diff
@@ -38,7 +38,7 @@ class CountAggregationLogicalFunction : public WindowAggregationLogicalFunction
     [[nodiscard]] std::string_view getName() const noexcept override;
 
 private:
-    static constexpr std::string_view NAME = "Count";
+    static constexpr std::string_view NAME = "COUNT";
```
alepping, line 41: Why exactly is it now necessary to set the name `in-code` in capital letters.
This sudden need requires massive amounts of small changes throughput the codebase that are painful when rebasing.
For this particular function, I only see a use in:
`serialize()`
Why not apply a `Util::toUpperCase` in `FunctionSerializationUtil::deserialize`?

nes-logical-operators/src/Operators/Windows/JoinLogicalOperator.cpp
```diff
@@ -107,8 +107,8 @@ JoinLogicalOperator JoinLogicalOperator::withInferredSchema(std::vector<Schema>
         return sourceNameLeft.value() + sourceNameRight.value() + Schema::ATTRIBUTE_NAME_SEPARATOR;
     }(leftInputSchema, rightInputSchema);
 
-    copy.windowMetaData.windowStartFieldName = newQualifierForSystemField + "start";
-    copy.windowMetaData.windowEndFieldName = newQualifierForSystemField + "end";
+    copy.windowMetaData.windowStartFieldName = newQualifierForSystemField + "START";
```
alepping, line 110: if possible, also apply a toUpper at a central place

nes-plugins/Sources/GeneratorSource/SinusGeneratorRate.cpp
```diff
@@ -34,7 +36,18 @@ std::optional<std::tuple<double, double>> SinusGeneratorRate::parseAndValidateCo
...
+
+    std::vector<std::string_view> params;
+
+    for (const auto& line : configString | std::views::split(','))
+    {
+        for (const auto& subLine : line | std::views::split('\n'))
+        {
+            params.emplace_back(Util::trimWhiteSpaces(std::string_view(subLine)));
+        }
+    }
```
alepping, line 48: Can't you use your new string utils function here?

alepping (changes requested):
Thanks for the changes!
See comments for suggestions.

nes-sql-parser/src/StatementBinder.cpp
greenBene: Agree. Renamed to `parserConfigLiteralMap`.

nes-common/src/Util/Strings.cpp
```diff
@@ -264,4 +264,29 @@ void removeDoubleSpaces(std::string& input)
     input.erase(newEnd, input.end());
 }
 
+std::vector<std::string_view> splitOnMultipleDelimiters(const std::string_view input, const std::string_view delimiters)
```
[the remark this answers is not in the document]
greenBene: Good point and thank you for the patch! :) I applied it.

nes-plugins/Sources/GeneratorSource/SinusGeneratorRate.cpp
greenBene: True! I updated the code accordingly.

nes-logical-operators/include/Operators/Windows/Aggregations/CountAggregationLogicalFunction.hpp
greenBene: Move the upper case logic to `AntlrSQLQueryPlanCreator::exitNamedExpression` and reverted the changed names of the aggregation functions

nes-logical-operators/src/Operators/Windows/JoinLogicalOperator.cpp
greenBene: I don't think there is a more central place for the window start and end suffixes and thus would leave the code as is.

nes-logical-operators/include/Operators/Windows/Aggregations/CountAggregationLogicalFunction.hpp
greenBene: Further, I created the systest `ColumnNames.test` to verify that the aggregation function names are applied as expected.

Artraxon (approved):

PR #1161, opened 2025-10-01
feat(PhysicalOperator): adds return state to open() function

alepping:
The open function can now return an OpenFunctionReturnState that allows the compiledFunction to not necessarily call the close function (or other stuff in the future).
This becomes important, e.g., if an operator wants to repeatedly call open without closing.
For example, the compiled raw scan (see PR #1153) wants to repeat a task, if its SequenceShredder detects a buffer as out of range of its ring structure. Currently, there is no simple way to avoid the `close()` call, which calls emit, which then emits the TupleBuffer and marks the SequenceNumber as seen and its chunk as 'seenLastChunk'. Later, when the task is repeated, the buffer with the same SequenceNumber is emitted again, causing a crash.

With this change, an operator can return from an open() call, stating that it is not finished yet.
The return value is an enum, which we can easily adapt/customize later on.

### DIFF
--- a/nes-physical-operators/include/Aggregation/AggregationProbePhysicalOperator.hpp
+++ b/nes-physical-operators/include/Aggregation/AggregationProbePhysicalOperator.hpp
@@ -34,7 +34,7 @@ public:
         std::vector<std::shared_ptr<AggregationPhysicalFunction>> aggregationPhysicalFunctions,
         OperatorHandlerId operatorHandlerId,
         WindowMetaData windowMetaData);
-    void open(ExecutionContext& executionCtx, RecordBuffer& recordBuffer) const override;
+    OpenReturnState open(ExecutionContext& executionCtx, RecordBuffer& recordBuffer) const override;
 
 private:
     std::vector<std::shared_ptr<AggregationPhysicalFunction>> aggregationPhysicalFunctions;
--- a/nes-physical-operators/include/EmitPhysicalOperator.hpp
+++ b/nes-physical-operators/include/EmitPhysicalOperator.hpp
@@ -41,7 +41,7 @@ public:
 
     void terminate(ExecutionContext&) const override { /*noop*/ }
 
-    void open(ExecutionContext& ctx, RecordBuffer& recordBuffer) const override;
+    OpenReturnState open(ExecutionContext& ctx, RecordBuffer& recordBuffer) const override;
     void execute(ExecutionContext& ctx, Record& record) const override;
     void close(ExecutionContext& ctx, RecordBuffer& recordBuffer) const override;
     void emitRecordBuffer(
--- a/nes-physical-operators/include/Join/HashJoin/HJProbePhysicalOperator.hpp
+++ b/nes-physical-operators/include/Join/HashJoin/HJProbePhysicalOperator.hpp
@@ -44,7 +44,7 @@ public:
 
     /// As the second phase gets triggered by the first phase, we receive a tuple buffer containing all information for performing the probe.
     /// Thus, we start a new pipeline and therefore, we create new Records from the built-up state.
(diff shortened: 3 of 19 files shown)


ls-1801 (changes requested):
1. I don't like that this change effectively touches every operator, although it is really only used in a single operator. I would strive for a solution that requires logic in just one operator (see below).

2. I am not sure if the enum variable should be a nautilus::val. I don't really see any test that triggers the repeat behavior and all the things that I have tried, looped indefinitely (see below).


Approach for local handling:
```c++
struct RawScanOperatorState final : OperatorState
{
    bool shouldClose;
};

void RawScanPhysicalOperator::open(ExecutionContext& executionCtx, RecordBuffer& recordBuffer) const
{
    PRECONDITION(child.has_value(), "Cannot call open without a child value.");
    executionCtx.setLocalOperatorState(this->id, std::make_unique<RawScanOperatorState>());
    static_cast<RawScanOperatorState&>(*executionCtx.getLocalState(this->id)).shouldClose
        = this->rawInputFormatScan->scan(executionCtx, recordBuffer, child.value());
}

std::optional<PhysicalOperator> RawScanPhysicalOperator::getChild() const
{
    return child;
}

void RawScanPhysicalOperator::setChild(PhysicalOperator child)
{
    this->child = child;
}

void RawScanPhysicalOperator::close(ExecutionContext& executionCtx, RecordBuffer& recordBuffer) const
{
    const auto& state = static_cast<const RawScanOperatorState&>(*executionCtx.getLocalState(this->id));
    if (state.shouldClose)
    {
        PhysicalOperatorConcept::close(executionCtx, recordBuffer);
    }
}
```

Looping test case:
```c++
/// Threads may process buffers out of order. This test simulates a scenario where the second thread process the second buffer first.
TEST_F(SpecificSequenceTest, testPipelineResizing)
{
    std::vector<InputFormatterTestUtil::ThreadInputBuffers> inputBuffers;
    inputBuffers.push_back(InputFormatterTestUtil::ThreadInputBuffers{.sequenceNumber = SequenceNumber(512), .rawBytes = "789\n"});
    inputBuffers.push_back({.sequenceNumber = SequenceNumber(1), .rawBytes = "123456789,123456"});
    for (size_t i = 1; i < 511; i++)
    {
        inputBuffers.push_back({.sequenceNumber = SequenceNumber(i + 1), .rawBytes = ""});
    }

    using namespace InputFormatterTestUtil;
    using enum TestDataTypes;
    using TestTuple = std::tuple<int32_t, int32_t>;
    runTest<TestTuple>(TestConfig<TestTuple>{
        .numRequiredBuffers = 512+1, /// 512 buffers for raw data, 1 buffer for results
        .sizeOfRawBuffers = 16,
        .sizeOfFormattedBuffers = 20,
        .parserConfig = {.parserType = "CSV", .tupleDelimiter = "\n", .fieldDelimiter = ","},
        .testSchema = {INT32, INT32},
        .expectedResults = {WorkerThreadResults<TestTuple>{{{TestTuple(123456789, 123456789)}}}},
        .rawBytesPerThread = std::move(inputBuffers)});
}
```

keyseven123 (commented):
I am still unsure whether we should use this or if the operator should handle this for now, as suggested by @ls-1801. 
This PR enables us to extend the `OpenReturnState` to include some error codes that can be triggered in Nautilus calls, such as insufficient memory or other exceptions directly in nautilus code. On the other hand, this is currently only used by one operator, so I understand your point, @ls-1801. 
The main question is: Will we need this in the next two months for another feature? Could the network sink benefit from it or a possible udf/inference operator? Do we want to add more fine-grained error return return codes than a general throw in the execution?
If we are not certain of a feature, then I would not include the enum for now. I know that I agreed to it. Thinking about it again, I think that this is too invasive for now.

PR #1163, opened 2025-10-02
fix(QueryCompiler): Disable mlir multithreading for nautilus

keyseven123:
This is a fix that disables mlir multithreading in the query compiler. We disable multithreading in MLIR by default to not interfere with NebulaStream's thread model

## Verifying this change
This change is tested by running our test suite.


ls-1801 (approved):

greenBene (approved):
Thanks for also fixing the typo! :)

PR #1168, opened 2025-10-02
Nix support for sanitizer builds

adrianmichalke:
Adds support for building with sanitizers via `-DUSE_SANITIZER=...` as with the vcpkg build

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -10,9 +10,13 @@ on:
 
 jobs:
   nix-build:
-    name: "Nix Build"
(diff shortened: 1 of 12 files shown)


keyseven123 (dismissed):
I tested it locally, and it tells me that 
```
Manually-specified variables were not used by the project:

    USE_SANITIZER
```

ls-1801 (changes requested):
Quite frankly, I don't understand how and if this is working as intended. Please give us an walkthrough how this change modifies the dependencies to be build with sanitizers.

PR #1169, opened 2025-10-06
fix(Descriptor): correctly handle missing config parameters

alepping:
Prior, the validateAndFormat would crash if a user did not specify a default argument, if the `configParameter.validate(config);` tried to access the map at the position of the configParameter, which is usually the case.

Now, we check whether the parameter exists in the config first and we check if it has a default parameter. We throw a meaningful error message, if it does not exist and has no default parameter.

### DIFF
--- a/nes-configurations/include/Configurations/Descriptor.hpp
+++ b/nes-configurations/include/Configurations/Descriptor.hpp
@@ -180,20 +180,20 @@ public:
         /// Next, try to validate all config parameters.
         for (const auto& [key, configParameter] : SpecificConfiguration::parameterMap)
         {
-            for (const auto& [keyString, configParameterString] : config)
+            /// If the user did not specify a parameter that is optional, use the default value (if available).
+            if (not config.contains(key))
             {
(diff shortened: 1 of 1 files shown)


nes-configurations/include/Configurations/Descriptor.hpp
```diff
@@ -180,22 +180,22 @@ class DescriptorConfig
         /// Next, try to validate all config parameters.
         for (const auto& [key, configParameter] : SpecificConfiguration::parameterMap)
         {
-            for (const auto& [keyString, configParameterString] : config)
+            /// If the user did not specify a parameter that is optional, use the default value (if available).
+            if (not config.contains(key))
             {
-                NES_DEBUG("key: {}, value: {}", keyString, configParameterString);
```
greenBene, line 185: Just want to double check that the removal of the `NES_DEBUG` statement was a conscious choice. :) If so, all good.

greenBene (approved):
All good :)

PR #1172, opened 2025-10-06
Nightly Fixes

ls-1801:
Apparently the introduction of UnpooledChunks re-introduced a use-after-free when using the TupleBuffer Debug option causing a chrash in the Nightly CI.
This PR also renames the cmake option from `CMAKE_NES_DEBUG_TUPLE_BUFFER_LEAKS` to `NES_DEBUG_TUPLE_BUFFER_LEAKS`

This PR also addresses two clang tidy errors that were reported during the nightly job.

## Verifying this change
This change is tested by
Systest. The issue caused a crash 100% of the time.

- Memory, debugging with NES_DEBUG_TUPLE_BUFFER_LEAKS

## Documentation

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -34,7 +34,7 @@ jobs:
         stdlib: [ 'libstdcxx', 'libcxx' ]
         # To reduce the load on our CI, we do not run all our tests with Benchmark.
         # The difference between Benchmark and Release is anyway that we disable any asserts (invariants or preconditions)
-        build_type: [ 'Debug', 'RelWithDebInfo', 'Release']
+        build_type: [ 'Debug', 'RelWithDebInfo', 'Release' ]
         sanitizer: [ 'address', 'undefined', 'thread', 'none' ]
         exclude:
           # Due to limited capacity, we disable sanitizer and libstdc++ tests on arm machines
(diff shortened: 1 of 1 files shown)


nes-memory/TupleBufferImpl.cpp
```diff
@@ -187,15 +187,15 @@ bool BufferControlBlock::release()
...
-        const auto recycler = std::move(owningBufferRecycler);
-        numberOfTuples = 0;
-        recycleCallback(owner, recycler.get());
 #ifdef NES_DEBUG_TUPLE_BUFFER_LEAKS
         {
             std::unique_lock lock(owningThreadsMutex);
             owningThreads.clear();
         }
 #endif
+        const auto recycler = std::move(owningBufferRecycler);
```
keyseven123, line 196: I was the culprit for this change. How long do the tests with `NES_DEBUG_TUPLE_BUFFER_LEAKS` run for one single c++ lib and architecture? I would like not to have to do this repeatedly.

keyseven123 (changes requested):

.github/workflows/build_and_test.yml
```diff
@@ -222,3 +222,63 @@ jobs:
...
+        # Upload all log and stats files if any of the tests has failed.
+        if: ${{ failure() && !github.event.act }}
+        with:
+          name: logs-${{ matrix.arch }}-${{ matrix.stdlib }}-${{ matrix.sanitizer }}-${{ matrix.build_type }}-${{ matrix.additional_build_flags.name }}
+          path: |
+            build/**/*.log
+            build/**/*.stats
+      - name: Upload test results to Codecov
+        if: ${{ !cancelled() && !github.event.act }}
+        uses: codecov/test-results-action@v1
```
keyseven123, line 274: Are we sure, we want to run the code cov for this?

.github/workflows/build_and_test.yml
```diff
@@ -222,3 +222,63 @@ jobs:
...
+      options: --user root
+    timeout-minutes: 40
+    runs-on: [ self-hosted, linux, Build, "${{matrix.arch}}" ]
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [ x64 ]
+        stdlib: [ 'libcxx' ]
+        build_type: [ 'Debug' ]
+        sanitizer: [ 'none' ]
```
keyseven123, line 245: If we have only these, then we can reduce the number of lines here, IMO.  This job is quite large for simply adding one test run with an additional cmake flag. Can we not reduce or combine it with the others somehow?

PR #1182, opened 2025-10-09
Distributed

ls-1801:
TODO

## Verifying this change
TODO

## Documentation
TODO

---
This is **part 4 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #1182 👈 
- <kbd>&nbsp;3&nbsp;</kbd> #1212 
- <kbd>&nbsp;2&nbsp;</kbd> #1210 
- <kbd>&nbsp;1&nbsp;</kbd> #1536

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -100,6 +100,8 @@ jobs:
             ctest --test-dir build -j -L Systest --output-on-failure --output-junit build/junit-systest.xml --timeout 4800 ${ADDITIONAL_CTEST_ARGS}
       - uses: ./.github/steps/run-in-container
         name: Run Docker Tests
+        # TODO #1540: remove this if once tests are reorganized, currently tsan is just too slow
+        if: ${{ !(matrix.build_type == 'Debug' && matrix.sanitizer == 'thread') }}
         with:
           image_name: nebulastream/nes-development:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |


adrianmichalke:
@ls-1801, I have committed the nix support changes at: 3e9c6ef4de0d9e32de80e31ace17b844cf3adeee.
We might want to containerize the started targets via `virtualisation.oci-containers` in the future. For now, `--port-offset` helps if there are port conflicts.

nes-nebuli/include/DistributedQuery.hpp
```diff
@@ -0,0 +1,174 @@
...
+};
+
+class DistributedQuery
+{
+    std::unordered_map<GrpcAddr, std::vector<LocalQueryId>> localQueries;
+
+public:
+    [[nodiscard]] auto iterate() const
+    {
+        return localQueries
```
greenBene: Why is the implementation in the .hpp and not the .cpp?

nes-nebuli/src/LegacyOptimizer/BottomUpPlacement.cpp
```diff
@@ -0,0 +1,314 @@
...
+    /// Each row is a constraint lowerBound <= x1*ar1 + x2*ar2 + ... + <= upperBound
+    /// Where lowerBound and upperBound for columns restrict the values of individual variables, i.e., placement of a operator on a node.
+
+    /// Example with 2 operators, 3 nodes = 6 variables = 6 columns:
+    ///   p[0][0] p[0][1] p[0][2] p[1][0] p[1][1] p[1][2]
+    ///      ↓       ↓       ↓       ↓       ↓       ↓
+    ///    col 0   col 1   col 2   col 3   col 4   col 5
+    ///
+    /// To make it easier to work with operatorPlacementMatrix implements a mapping from the placement matrix to the columns in the
+    /// constraint model. If an operator o is placed on a node n than placementMatrix[o][n] = true
```
greenBene: Typo: than -> then

nes-nebuli/src/LegacyOptimizer/BottomUpPlacement.cpp
```diff
@@ -0,0 +1,314 @@
...
+    for (const LogicalOperator& op : BFSRange(logicalPlan.getRootOperators().front()))
+    {
+        for (const Topology::NodeId& node : topology | std::views::keys)
+        {
+            auto index = static_cast<int>(reverseIndex.size());
+            operatorPlacementMatrix[{op.getId(), node}] = index;
+            /// By default we allow placemnet on every node. we allow values from (0,1) and limit the solution to integers which gives us
```
greenBene: Typo: placemnet -> placement

nes-systests/systest/remote-test/systest_remote_test.bats
alepping: Collected observations after playing around:
- removing the first two `setup_file()` does not seem to make a difference on my end (VCPKG setup)
- `assert_json_equal` and `assert_json_contains` do not seem to be used (returning 1 in their respective first lines does not do anything)
- only if a query fails, the execution log contains which tests were loaded/executed and passed (I'd prefer to see it even if all queries passed, but I'd be fine with both versions)
- the cmake setup of the test is only hidden behind `if (ENABLE_DOCKER_TESTS)`, but should also be hidden behind `ENABLE_BATS_TESTS`
- we skip all queries with a ` configuration override` in the `runQueriesAtRemoteWorker` function, which only becomes visible if a query fails (or if one looks into the code) and could easily lead to confusion

docs/development/systests.md
```diff
@@ -193,3 +193,184 @@ The endless mode runs tests in an infinite loop i.e. for regression testing (`--
...
+For tests that require specific worker placement, use the `SOURCE.HOST` and `SINK.HOST` options:
+
+**Named Sources and Sinks:**
+
+```sql
+CREATE LOGICAL SOURCE nameStream(firstName VARSIZED, lastName VARSIZED);
+
+CREATE PHYSICAL SOURCE FOR nameStream
+TYPE File
+SET("sink-node:9090" AS `SOURCE`.`HOST`);
```
alepping: (resolve me): very cool!

nes-nebuli/private/LegacyOptimizer/OperatorPlacement.hpp
```diff
@@ -0,0 +1,34 @@
...
+#include <Plans/LogicalPlan.hpp>
+#include <NetworkTopology.hpp>
+
+namespace NES
+{
+
+/// Obtain the placement for op, return empty optional if placement trait is not present in op
+Topology::NodeId getPlacementFor(const LogicalOperator& op);
+
+class OperatorPlacer
```
alepping: since this is an interface, a class description would be helpful

nes-nebuli/src/DistributedQuery.cpp
```diff
@@ -0,0 +1,240 @@
...
+    return NES::DistributedQueryId(generateHorseName());
+}
+
+NES::DistributedQueryState NES::DistributedQueryStatus::getGlobalQueryState() const
+{
+    auto any = [&](std::initializer_list<QueryState> state)
+    {
+        return std::ranges::any_of(
+            localStatusSnapshots | std::views::values,
+            [&](const std::unordered_map<LocalQueryId, std::expected<LocalQueryStatus, Exception>>& localMap)
```
alepping:
```suggestion
            [this](const std::unordered_map<LocalQueryId, std::expected<LocalQueryStatus, Exception>>& localMap)
```

nes-nebuli/src/LegacyOptimizer/LegacyOptimizer.cpp
```diff
@@ -55,6 +63,20 @@ LogicalPlan LegacyOptimizer::optimize(const LogicalPlan& plan) const
 
     originIdInferencePhase.apply(newPlan);
     typeInference.apply(newPlan);
-    return newPlan;
+
+
+    BottomUpOperatorPlacer(workerCatalog).place(newPlan);
```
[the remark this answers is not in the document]
alepping: Also, why hardcode the BottomUpPlacer? What does the Abstract 'OperatorPlacement' exist for then?

Since we only have one operator placement, I'd get rid of the abstract OperatorPlacement.
It is trivial to add an abstraction on top later on.
Currently, the abstraction is just unnecessary and confusing imo.

nes-nebuli/src/LegacyOptimizer/QueryDecomposition.cpp
```diff
@@ -0,0 +1,195 @@
...
+        throw PlacementFailure("No path from {} to {} found", startNode, endNode);
+    }
+    const auto [path] = paths.front();
+    INVARIANT(path.size() >= 2, "Path from {} to {} must contain at least 2 nodes", startNode, endNode);
+
+    LogicalOperator currentOp = op;
+    for (size_t i = 0; i < path.size() - 1; ++i)
+    {
+        const auto& upstreamNode = path.at(i);
+        const auto& downstreamNode = path.at(i + 1);
```
alepping:
```suggestion
    for (const auto [i, hostAddr] : path | NES::views::enumerate)
    {
        const auto& upstreamNode = hostAddr;
        const auto& downstreamNode = path.at(i + 1);
```

alepping: (resolve me): `std::views::slide(2)` :`-(

nes-nebuli/src/LegacyOptimizer/QueryDecomposition.cpp
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    return currentOp;
+}
+
+LogicalOperator decomposePlanRecursive(DecompositionContext& context, const LogicalOperator& op);
+
+LogicalOperator assignOperator(DecompositionContext& context, const LogicalOperator& op, const LogicalOperator& child)
+{
+    auto assignedChild = decomposePlanRecursive(context, child);
```
alepping:
```suggestion
    const auto assignedChild = decomposePlanRecursive(context, child);
```

nes-nebuli/src/LegacyOptimizer/QueryDecomposition.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+    const auto opPlacement = getPlacementFor(op);
+    const auto childPlacement = getPlacementFor(child);
```
alepping:
```suggestion
    const auto opNode = getPlacementFor(op);
    const auto childNode = getPlacementFor(child);
```

nes-nebuli/src/LegacyOptimizer/QueryDecomposition.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+QueryDecomposer::QueryDecomposer(
+    SharedPtr<const WorkerCatalog> workerCatalog, SharedPtr<const SourceCatalog> sourceCatalog, SharedPtr<const SinkCatalog> sinkCatalog)
+    : workerCatalog(std::move(workerCatalog)), sourceCatalog(std::move(sourceCatalog)), sinkCatalog(std::move(sinkCatalog))
+{
+}
+
+DecomposedLogicalPlan<HostAddr> QueryDecomposer::decompose(const LogicalPlan& placedPlan)
+{
+    PRECONDITION(placedPlan.getRootOperators().size() == 1, "BUG: query decomposition requires a single root operator");
```
alepping: do we call decompose on a query plan that runs on a single (worker) node?

nes-nebuli/tests/testdata/basic_single_node.yaml
alepping: review stop 16.12.25 (70/140)

nes-nebuli/tests/testdata/bridge_join.yaml
alepping: just a suggestion, but I'd add ascii art on top to show, with one glance, what the query/test does

alepping: Instead of generating it via claude, you could also just copy paste the plan from our log, that should be more precise

nes-nebuli/tests/testdata/disconnected.yaml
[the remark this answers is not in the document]
alepping: I played around. So generally, creating a topology without that contains nodes that are not connected to any other node does not lead to a crash, is this intended?

nes-nebuli/tests/testdata/not_enough_capacities.yaml
alepping: Connects to above comments, but it would be nice to know why the capacity is not enough here (comment)
Also, when logging the exception I got:
`C++ exception with description "could not place all operators; Placement is not possible`, which is not very descriptive. It would be nice to know that it was due to 'not enough capacities' and ideally how many 'capacities' were required and how many available

nes-plugins/Sinks/NetworkSink/NetworkSink.cpp
```diff
@@ -0,0 +1,229 @@
...
+}
+
+bool BackpressureHandler::empty() const
+{
+    return stateLock.rlock()->buffered.empty();
+}
+
+NetworkSink::NetworkSink(BackpressureController backpressureController, const SinkDescriptor& sinkDescriptor)
+    : Sink(std::move(backpressureController))
+    , tupleSize(sinkDescriptor.getSchema()->getSizeOfSchemaInBytes())
```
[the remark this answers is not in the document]
alepping: generally though, we should not use the schema anymore to get the size of the tuple (in bytes)

alepping (changes requested):

LeoRaasch (changes requested):

Artraxon:
Hey, I'll assign myself as a reviewer as well for the IR/Optimizer parts

nes-nebuli/include/DistributedQuery.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+private:
+    DistributedQueryId queryId{DistributedQueryId::INVALID};
+    DecomposedLogicalPlan<GrpcAddr> plan;
+    LogicalPlan globalPlan;
+};
+}
+
+FMT_OSTREAM(NES::DistributedQueryState);
+FMT_OSTREAM(NES::DistributedQuery);
```
Artraxon: what about DistributedQueryStatus?

nes-nebuli/include/NetworkTopology.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+    /// Upstream --> from the DAG sink to the DAG source
+    /// Downstream --> from the DAG source to the DAG sink
+    enum Direction : uint8_t
+    {
+        Upstream,
+        Downstream
+    };
+
+    [[nodiscard]] static Direction reverse(Direction direction) { return direction == Upstream ? Downstream : Upstream; }
```
Artraxon: can be static constexpr. A bit odd to not attach it to direction, but i don't care that much

nes-nebuli/src/LegacyOptimizer/OperatorPlacement.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+Topology::NodeId getPlacementFor(const LogicalOperator& op)
+{
+    auto placement = getTrait<PlacementTrait>(op.getTraitSet())
+                         .transform([](const PlacementTrait& placementTrait) { return Topology::NodeId(placementTrait.onNode); });
+    if (!placement.has_value())
+    {
+        throw PlacementFailure("Operator does not have a placement trait");
+    }
+    return placement.value();
```
Artraxon: I think most of this can be replaced by traitset.get<PlacementTrait>, which asserts the existence of the trait as an invariant. 

Wouldn't this also be more correct, is there any way how the trait cannot be set except for a programming mistake in the earlier parts of the optimizer?

nes-nebuli/private/LegacyOptimizer/OperatorPlacement.hpp
Artraxon: This interface never gets used anywhere and only has one implementation, I would favor removing it and introducing an abstraction layer for the logic once we have multiple strategies.

nes-nebuli/src/LegacyOptimizer/BottomUpPlacement.cpp
```diff
@@ -0,0 +1,314 @@
...
+    {
+        return 0;
+    }
+    return 1;
+}
+
+LogicalOperator addPlacementTrait(const LogicalOperator& op, const std::unordered_map<OperatorId, Topology::NodeId>& placement)
+{
+    auto oldTraitSet = op.getTraitSet();
+    USED_IN_DEBUG auto addedTrait = oldTraitSet.tryInsert(PlacementTrait(placement.at(op.getId()).getRawValue()));
```
Artraxon: I have never seen the USED_IN_DEBUG macro before, how does it behave? Does it not compile out the insertion into the traitSet?

nes-nebuli/src/LegacyOptimizer/LegacyOptimizer.cpp
Artraxon: Imo BottomUpOperatorPlacer can just follow the the other stages and implement an apply method

nes-nebuli/src/LegacyOptimizer/LegacyOptimizer.cpp
```diff
@@ -55,6 +63,20 @@ LogicalPlan LegacyOptimizer::optimize(const LogicalPlan& plan) const
...
+    DecomposedLogicalPlan<HostAddr> decomposed = QueryDecomposer(workerCatalog, sourceCatalog, sinkCatalog).decompose(newPlan);
```
Artraxon: It might even make sense to move the decomposition out of the optimizer since it doesn't make any decision. A bit similar to how the lowering doesn't make any decisions anymore and just use the existing informations.


It's not super important though

nes-nebuli/src/QueryManager/QueryManager.cpp
```diff
@@ -29,182 +33,344 @@
...
+    std::vector<Exception> exceptions;
 
-        /// The query is expected to be moved into the started state pretty quickly after lowering, it is very unlikely to even observe
-        /// the status not changing immediatly, so a rapid polling interval is appropriate.
-        constexpr auto statusPollInterval = std::chrono::milliseconds(10);
-        constexpr size_t statusRetries = 14;
-        NES_DEBUG("Starting query {} was successful. Waiting for state to change", queryId);
-        for (size_t i = 0; i < statusRetries; ++i)
+    std::chrono::system_clock::time_point queryStartTimestamp = std::chrono::system_clock::now();
+    for (const auto& [grpcAddr, localQueryId] : query.iterate())
```
Artraxon: wouldn't steady_clock be the better choice? its running in one thread and in case of an embedded engine the method should run quite fast, which could lead to problems with system time, since it is not monotonic

nes-sql-parser/AntlrSQL.g4
```diff
@@ -72,6 +72,7 @@ optionsClause: (SET '(' options=namedConfigExpressionSeq ')');
 
 createSinkDefinition: SINK sinkName=identifier schemaDefinition TYPE type=identifier optionsClause?;
 
+createWorkerDefinition: WORKER hostaddr=STRING AT grpcaddr=STRING optionsClause?;
```
Artraxon: Aparrently at was already a keyword in our grammar, but I would prefer it if we could do without because it is non standard an not needed to be syntactically unambiguous. 
I would maybe be ok with it if the thing following the AT is a catalog object (like the FOR in create physical source), but it is just a mandatory string option.

nes-common/include/ExceptionDefinitions.inc
```diff
@@ -73,6 +73,12 @@ EXCEPTION(QueryCompilerError, 2126, "query compiler / backend throwed an error")
 EXCEPTION(UnknownSinkName, 2127, "unknown sink name")
 EXCEPTION(SinkAlreadyExists, 2128, "sink already exists")
 
+/// 23XX Distributed
```
keyseven123: Nit-pick but please move 23XX after 22XX

nes-nebuli/src/DistributedQuery.cpp
```diff
@@ -0,0 +1,250 @@
...
+{
+constexpr std::array HORSE_BREEDS = std::to_array<std::string_view>(
+    {"arabian",    "thoroughbred", "quarter",  "morgan",       "appaloosa", "paint",     "mustang",   "andalusian", "friesian",
+     "clydesdale", "percheron",    "belgian",  "shire",        "shetland",  "welsh",     "connemara", "hanoverian", "warmblood",
+     "lipizzaner", "palomino",     "buckskin", "standardbred", "tennessee", "icelandic", "haflinger"});
+
+constexpr std::array ATTRIBUTES = std::to_array<std::string_view>(
+    {"swift",    "brave",    "noble",   "wild",        "gallant",   "proud",   "mighty",  "gentle",   "fierce",   "graceful",
+     "spirited", "majestic", "elegant", "strong",      "agile",     "loyal",   "bold",    "free",     "dashing",  "swift",
+     "valiant",  "fearless", "regal",   "magnificent", "brilliant", "radiant", "blazing", "charging", "prancing", "soaring"});
```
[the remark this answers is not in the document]
keyseven123: I also think it will create some raised eyebrows, if someone has a query `arabian_swift_12` that failed

keyseven123 (changes requested):

docs/development/systests.md
```diff
@@ -193,3 +193,184 @@ The endless mode runs tests in an infinite loop i.e. for regression testing (`--
...
+**Example: Two-Node Topology**
+
+```yaml
+workers:
+  - host: "sink-node:9090"
+    grpc: "sink-node:8080"
+    capacity: 10000
+  - host: "source-node:9090"
+    grpc: "source-node:8080"
+    capacity: 4
```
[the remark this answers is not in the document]
ls-1801: This is just the documentation; you can pick any number you want. There is not one size fits all, because we have so many different queries.

nes-nebuli/apps/repl/tests/expect-tests/distributed/many-nodes.exp
```diff
@@ -0,0 +1,134 @@
...
+setup "8-node.yaml"
+
+# Spawn nebuli inside the docker container
+spawn docker compose exec -t nebuli nebuli -d -f JSON
+
+wait_for_prompt
+
+send_sql {
+    CREATE WORKER "sink-node:9090" AT "sink-node:8080";
+    CREATE WORKER "source-node-1:9090" AT "source-node-1:8080"
```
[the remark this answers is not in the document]
ls-1801: Yes, this is intentional. The worker name uses the host address (port 9090) while the AT clause specifies the gRPC address (port 8080) — these are two different address types in WorkerConfig.

nes-nebuli/src/LegacyOptimizer/BottomUpPlacement.cpp
ls-1801: Done!

nes-nebuli/src/DistributedQuery.cpp
ls-1801: Done!

nes-nebuli/src/DistributedQuery.cpp
```diff
@@ -0,0 +1,240 @@
...
+            localStatusSnapshots | std::views::values,
+            [&](const std::unordered_map<LocalQueryId, std::expected<LocalQueryStatus, Exception>>& localMap)
+            {
+                return std::ranges::any_of(
+                    localMap | std::views::values,
+                    [&](const std::expected<LocalQueryStatus, Exception>& local)
+                    { return local.has_value() && std::ranges::contains(state, local.value().state); });
+            });
+    };
+    auto all = [&](std::initializer_list<QueryState> state)
```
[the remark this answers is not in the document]
ls-1801: Done!

nes-nebuli/src/LegacyOptimizer/QueryDecomposition.cpp
```diff
@@ -0,0 +1,195 @@
...
+LogicalOperator createNetworkChannel(
+    DecompositionContext& context, const LogicalOperator& op, const Topology::NodeId& startNode, const Topology::NodeId& endNode)
+{
+    /// Ask the topology for a path of nodes that connects upstream and downstream, currently we use any of them
+    const auto paths = context.workerCatalog->getTopology().findPaths(startNode, endNode, Topology::Direction::Downstream);
+    if (paths.empty())
+    {
+        throw PlacementFailure("No path from {} to {} found", startNode, endNode);
+    }
+    const auto [path] = paths.front();
```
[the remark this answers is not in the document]
ls-1801: Done!

nes-nebuli/src/LegacyOptimizer/QueryDecomposition.cpp
ls-1801: Done!

nes-nebuli/src/LegacyOptimizer/QueryDecomposition.cpp
ls-1801: Done!

nes-nebuli/src/LegacyOptimizer/LegacyOptimizer.cpp
ls-1801: Done, renamed place() to apply() to follow the phase convention.

nes-systests/systest/remote-test/systest_remote_test.bats
ls-1801: This is outdated — the duplicate setup_file() functions have been consolidated into one. The other observations are addressed: ENABLE_BATS_TESTS guard is added, and config override skipping is now documented.

docs/development/systests.md
```diff
@@ -193,3 +193,184 @@ The endless mode runs tests in an infinite loop i.e. for regression testing (`--
...
+CREATE PHYSICAL SOURCE FOR nameStream
+TYPE File
+SET("sink-node:9090" AS `SOURCE`.`HOST`);
+ATTACH INLINE
+Alice,Smith
+Bob,Jones
+
+CREATE SINK firstNameLastNameSink(firstNameLastName VARSIZED)
+TYPE File
+SET("sink-node:9090" AS `SINK`.`HOST`);
```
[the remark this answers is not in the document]
ls-1801: Good point — cycle validation has been added to the WorkerCatalog (via Topology::addNode). It now detects cycles using DFS and rejects topologies with cyclic dependencies.

docs/development/systests.md
```diff
@@ -193,3 +193,184 @@ The endless mode runs tests in an infinite loop i.e. for regression testing (`--
...
+2,200,2000
+3,300,3000
+```
+
+> [!NOTE]
+> Explicit worker assignment overrides the automatic placement strategy. Use this for tests that specifically validate distributed query execution or data routing.
+
+### Remote Tests
+
+By default, systest runs all workers embedded within a single process, using in-memory communication channels for multi-worker topologies. This provides fast test execution and simplified debugging.
```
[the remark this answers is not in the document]
ls-1801: Done\! Clarified to "Compared to remote tests with real network overhead, this provides faster test execution and simplified debugging."

nes-nebuli/src/LegacyOptimizer/QueryDecomposition.cpp
ls-1801: The suggestion would cause an out-of-bounds access on the last iteration (when i == path.size()-1, accessing i+1 would be invalid). The current indexed loop correctly stops at size()-1 to ensure i+1 is always valid.

nes-nebuli/include/DistributedQuery.hpp
ls-1801: Removed — the operator<< was declared but never implemented, so this was dead code.

nes-nebuli/include/DistributedQuery.hpp
```diff
@@ -0,0 +1,174 @@
...
+#include <ErrorHandling.hpp>
+#include <WorkerConfig.hpp>
+#include <WorkerStatus.hpp>
+
+namespace NES
+{
+using DistributedQueryId = NESStrongStringType<struct DistributedQueryId_, "invalid">;
+DistributedQueryId getNextDistributedQueryId();
+
+enum class DistributedQueryState : uint8_t
```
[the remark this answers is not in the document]
ls-1801: DistributedQueryState is not stored or mutated directly — it is computed on-the-fly by getGlobalQueryState() from the individual LocalQueryStatus snapshots. There is no state machine; it's a derived view of the cluster state at a point in time. Added a doc comment to the enum clarifying this.

nes-nebuli/src/DistributedQuery.cpp
```diff
@@ -0,0 +1,240 @@
...
+    {
+        return DistributedQueryState::Registered;
+    }
+
+    if (all({QueryState::Stopped, QueryState::Running, QueryState::Started, QueryState::Registered}))
+    {
+        return DistributedQueryState::Running;
+    }
+
+    INVARIANT(false, "unreachable. i think");
```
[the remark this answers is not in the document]
ls-1801: Done!

nes-nebuli/include/NetworkTopology.hpp
ls-1801: Done!

nes-plugins/Sinks/NetworkSink/NetworkSink.cpp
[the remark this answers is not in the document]
ls-1801: Addressed in 9980c79: added a comment clarifying that `getSizeOfSchemaInBytes()` IS the physical tuple layout size for row-based storage. `TupleBuffer` has no per-row-size API (`getCapacity()` returns buffer-level capacity, not tuple size), so `getSizeOfSchemaInBytes()` is the correct call here.

nes-nebuli/src/QueryManager/QueryManager.cpp
ls-1801: Fixed in 0ede885: duration measurement switched from `system_clock` to `steady_clock` in `QueryManager::start()`.

nes-nebuli/src/QueryManager/QueryManager.cpp
```diff
@@ -29,182 +33,344 @@
...
+
+    if (not exceptions.empty())
+    {
+        return std::unexpected{exceptions};
+    }
+
+    /// Poll all queries until there status has changed to something other than Registered.
+    /// We do this so the function can guarantee that the next call to status will guarantee that the query is not in the Registered
+    /// State anymore.
+    auto waitForStatusChange = query.iterate() | std::ranges::to<std::vector>() | std::ranges::to<std::unordered_set>();
```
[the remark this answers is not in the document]
ls-1801: Fixed in 270fe33: removed the intermediate `ranges::to<vector>()` pipe. The ref-tuples from `iterate()` are now copied to pair values before `ranges::to<vector>()` because `erase_if` requires move-assignable elements -- but the redundant double conversion is eliminated.

nes-nebuli/src/QueryManager/QueryManager.cpp
```diff
@@ -29,182 +33,344 @@
...
+
+    for (const auto& [grpcAddr, localQueryId] : query.iterate())
+    {
+        try
+        {
+            INVARIANT(backends.contains(grpcAddr), "Local query references node ({}) that is not part of the cluster", grpcAddr);
+            const auto result = backends.at(grpcAddr).status(localQueryId);
+            localStatusResults[grpcAddr].emplace(localQueryId, result);
+        }
+        catch (std::exception& e)
```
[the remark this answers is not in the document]
ls-1801: Addressed in 270fe33: added an explanatory comment on this catch block. These functions call `stop()` on backends that may throw if a backend is already stopped or unregistered. The catch provides a defensive last-resort log without rethrowing.

nes-frontend/src/Statements/StatementHandler.cpp
```diff
@@ -147,14 +173,33 @@ std::expected<DropPhysicalSourceStatementResult, Exception> SourceStatementHandl
...
+            return host;
+        }
+        return std::visit(
+            Overloaded{
+                [](const DefaultHost& defaultHost) -> std::string { return defaultHost.hostName; },
+                [](const RequireHostConfig&) -> std::string
+                { throw InvalidStatement("Could not handle sink statement. `SINK`.`HOST` was not set"); }},
+            hostPolicy);
+    }();
+    sinkConfig.erase("host");
```
keyseven123: Why do we need to erase it here again?

PR #1189, opened 2025-10-16
docs: rework README.md

adrianmichalke:
- Updates nebulastream description
- adds quickstart guide
- Adds pointers to important docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,44 +21,134 @@
 
 ----
 
-NebulaStream is our attempt to develop a general-purpose, end-to-end data management system for the IoT.
-It provides an out-of-the-box experience with rich data processing functionalities and a high ease-of-use.
+NebulaStream is an end-to-end data-management system for cloud-edge-sensor deployments.
+The platform combines ease of use, extensibility, and efficiency to let teams focus on business logic while the engine optimizes the data path.
 
-NebulaStream is a joint research project between the DIMA group at TU Berlin and BIFOLD.
+**Core Pillars**
+- **Ease of Use** - Rich, out-of-the-box functionality for a variety of stream processing queries.
+- **Extensibility** - Uniform plugin registries for custom connectors, formats, operators, and optimizations.
+- **Efficiency** - Hardware-tailored code, adaptive execution, and interleaved I/O to handle large workloads.
 
-Learn more about Nebula Stream at https://www.nebula.stream or take a look at our [documentation](docs).
+## Vision
 
-# Clang-Format, Clang-Tidy, License & Pragma Once Check, and Ensure Correct Comments
-We use `clang-format` and `clang-tidy` to ensure code quality and consistency.
-To run the checks, you can use the target `format`. 
+NebulaStream is a joint research project at BIFOLD with contributors from the DIMA Group at TU Berlin and the DFKI IAM Group.
+It aims to execute thousands of queries over millions of heterogeneous sources in massively distributed environments.
+We advance this vision through five core technologies:
+- **Heterogeneous Hardware Support** - Targets diverse CPU architectures and accelerators such as GPUs and TPUs.
+- **Code Generation** - Compiles every query to efficient native code for low latency and energy usage.
+- **In-Network Processing** - Pushes operators toward data sources to reduce network traffic.
+- **Adaptive Resource Management** - Reacts to topology or workload changes without interrupting queries.
 
+The system architecture spans sensor to cloud:
+1. **Sources & Sinks** - Built-in connectors (e.g., JDBC, MQTT, TCP) and formats (e.g., CSV, JSON) with plugin hooks for custom components; extend them via the plugin framework in the [Extensibility guide](docs/guide/extensibility.md).
+2. **I/O Handling** - Thread-shared source processing and asynchronous callbacks minimize waiting time.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -21,44 +21,150 @@
...
+EOF
+
+$BUILD_DIR/nes-nebuli/nes-nebuli register \
+  -s localhost:8080 \
+  -i demo.sql \
+  -x \
+  -w
+```
+
+The `-x` flag starts the query immediately after registration and `-w` waits for completion. The generated CSV appears at `demo-output.csv`. Inspect or retire the query with additional `nebuli` subcommands such as `show`, `stop`, or `unregister` as needed.
```
ls-1801: This definitely does not work. Have you tested these? There is `ATTACH INLINE? in the SQL frontend. There is no -x flag for the SQL frontend either.

ls-1801 (changes requested):

Artraxon (changes requested):

README.md
```diff
@@ -21,44 +21,150 @@
...
-We use `clang-format` and `clang-tidy` to ensure code quality and consistency.
-To run the checks, you can use the target `format`. 
+NebulaStream is a joint research project at BIFOLD with contributors from the DIMA Group at TU Berlin and the DFKI IAM Group.
+It aims to execute thousands of queries over millions of heterogeneous sources in massively distributed environments.
+We advance this vision through five core technologies:
+- Heterogeneous Hardware Support - Targets diverse CPU architectures and accelerators such as GPUs and TPUs.
+- Code Generation - Compiles every query to efficient native code for low latency and energy usage.
+- In-Network Processing - Pushes operators toward data sources to reduce network traffic.
+- On-Demand Gathering - Applies processing as early as possible along the data path.
+- Adaptive Resource Management - Reacts to topology or workload changes without interrupting queries.
```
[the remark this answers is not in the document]
yschroeder97: We could also create a table and have crosses and checks what is already achieved, what is in progress, and what works.

README.md
```diff
@@ -21,44 +21,150 @@
...
 
+The system architecture spans sensor to cloud:
+1. Sources & Sinks - Built-in connectors (e.g., JDBC, MQTT, TCP) and formats (e.g., CSV, JSON) with plugin hooks for custom components; extend them via the plugin framework in the [Extensibility guide](docs/guide/extensibility.md).
+2. I/O Handling - Thread-shared source processing and asynchronous callbacks minimize waiting time.
```
yschroeder97: We are currently still running the thread-per-source model, and do not use callbacks. 
I am not sure what @alepping thinks about whether it is worth starting another effort and integrating the async I/O branch into main after the benchmarks.

README.md
```diff
@@ -21,44 +21,150 @@
...
+
+```shell
+BUILD_DIR=cmake-build-debug
+cmake --build $BUILD_DIR -j --target systest
+$BUILD_DIR/systest/systest -t nes-systests/function/arithmetical/FunctionAdd.test:1
+```
+
+Append worker configuration overrides after `--` (for example `-- --worker.default_query_execution.execution_mode=INTERPRETER`). See the [systest guide](docs/development/systests.md) for authoring custom `.test` files and the complete CLI reference.
+
+#### Start nebuli with a worker
```
yschroeder97: You have never actually introduced what "nebuli' is.

yschroeder97 (changes requested):
Thanks for revisiting the README :)

I agree with @Artraxon that we need to somehow differentiate between what users can actually expect and what is future work. 
It is okay to have both, but it will only lead to people being annoyed when they expect to run NES on their GPU and find out it doesn't work.

Otherwise, you could also add a small summary on the most important targets, so that users/devs know what systest, nebuli, and the worker are and for what they should be used.

README.md
```diff
@@ -21,44 +21,150 @@
...
 ----
 
-NebulaStream is our attempt to develop a general-purpose, end-to-end data management system for the IoT.
-It provides an out-of-the-box experience with rich data processing functionalities and a high ease-of-use.
+NebulaStream is a post-cloud, end-to-end data-management system for cloud-edge-sensor deployments across the IoT.
+The platform combines Ease of Use, Extensibility, and Efficiency to let teams focus on business logic while the engine optimizes the data path.
 
-NebulaStream is a joint research project between the DIMA group at TU Berlin and BIFOLD.
+**Core Pillars**
+- Ease of Use - Rich, out-of-the-box functionality a variety of stream processing queries.
```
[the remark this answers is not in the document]
adrianmichalke: addressed

README.md
adrianmichalke: changed to: `Start the client with a worker`

README.md
```diff
@@ -21,44 +21,150 @@
...
+### Execute Queries & Tests
+
+#### Run systest for a query
+Build the `systest` executable in your chosen build directory and execute a specific query. The example below targets the first query in the arithmetic function suite and stores artifacts in `cmake-build-debug/systest-run`:
+
+```shell
+BUILD_DIR=cmake-build-debug
+cmake --build $BUILD_DIR -j --target systest
+$BUILD_DIR/systest/systest -t nes-systests/function/arithmetical/FunctionAdd.test:1
+```
```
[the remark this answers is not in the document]
adrianmichalke: Added a comment regarding the docker container env

README.md
yschroeder97: Yes, I think so!

PR #1191, opened 2025-10-16
feat: add nautilus function inlining 

bgrb1:
This PR is linked to [this nautilus PR](https://github.com/nebulastream/nautilus/pull/96), which adds the capability to (recursively) inline proxy functions called through `invoke`, which e.g. allows to condense the record buffer methods with a call depth of 3 into flat code with direct access to the underlying data structure. This happens on the LLVM Intermediate Representation level at query query compilation time, by using information stored during the compilation of NES (see below).

Here is an example of the change to the first basic block of an example query when applying inlining (in LLVM IR):
[before.txt](https://github.com/user-attachments/files/22250656/before.txt)
[after.txt](https://github.com/user-attachments/files/22250657/after.txt)
Some of the original function calls wouldve also had further subcalls. 

For the exact inner workings, please refer to the nautilus PR linked above. In short, certain functions (and their respective cmake targets) can be annotated with `NAUT_INLINE`. The LLVM IR of annotated functions is intercepted during compilation by an LLVM pass, stored as a string in bitcode form in the compiled code of NES, and then inserted into a registry data structure at startup. Nautilus detects that the address of an invoked function is present in the registry, retrieves the LLVM IR code, and inlines it. Besides the LLVM IR of the target function, the LLVM pass also collects the symbol dependencies of the function, which allows to 1) link the inlined code against NES at runtime (so that they can use non-inlinable functions or global variables), and 2) recursively inline functions that call other inlinable functions, even from other source files. 

While this approach slightly increases the size of the NES binaries, as well as the memory footprint, I would argue that probably only a very tiny fraction of the code is going to be marked for inlining, and that the LLVM bitcode of a function isnt so much larger than its compiled version, which is also stored in memory already. An alternative approach would always be to switch to file based storage eventually. 

There still may exist some functions that cant be inlined so far , e.g. if they use some unhandled LLVM Constants. However, normally these errors should be caught and only result in a console warning. Also the feature is entirely opt-in, so problematic functions/cmake-targets can be untagged at any time to fix a broken inline function. The function name should always be printed to the error log when there is a problem, to indicate which function needs to be untagged.
Furthermore, inlining is still disabled for ARM and undefined behavior sanitizers due to some open problems. 

To evaluate the potential performance benefits, I ran the benchmarks locally, with the join mode forced to NLJ (this implementation had more inline potential). While many queries had little performance improvement, there was a big speedup in some join heavy queries. For example, the largest query from the Nexmark_multiple_GB benchmark had a speedup of over 200% from about 2900s to 900s. The peak speedup was 4.3 from 66s to 15s.
For the future it would maybe be interesting to optimize which calls to actually inline. While trying to optimize performance, I found that some of the inline functions are pretty big (up to 100 LLVM instructions) and bloat the compiled query code, yet still have a strong positive performance impact when inlining. I'd assume there is trade-off where the performance would degrade again at some point, due to instruction cache pressure from replicating the code of big functions multiple times (depending on how well LLVM does at code deduplication here). 

## Verifying this change
This change is tested by:
systests

- Dependencies (upgrades nautilus dependency)
- Query compiler
- Physical operators 
- Memory

---

> [!NOTE]
> Enables Nautilus IR-level function inlining by adding an LLVM pass plugin, annotating hot proxy/functions with NAUT_INLINE, wiring inlining into key CMake targets, and updating the nautilus dependency.
> 
> - **Build/Compiler**:
>   - Add `llvm-passes` with LLVM plugin registering `NautilusInlineRegistrationPass` and `nautilus_inline()` CMake helper; apply to `nes-memory`, `nes-nautilus`, `nes-physical-operators`, `nes-query-engine`, `nes-runtime`.
>   - Enable inlining at runtime compilation (`options.setOption("engine.Inline", true)`).
>   - Bump `nautilus` dependency (Nix and vcpkg refs/hashes).
> - **Core/Runtime/Operators**:
>   - Annotate hot methods and proxies with `NAUT_INLINE`; add proxy helpers and include `nautilus/inline.hpp` across memory, runtime, emit, NLJ build/probe/slice, paged vector, and tuple buffer code.
>   - Introduce `getPagedVectorRefProxy`, slice/window handler proxies, and allocation/emit proxies; minor refactors to use `invoke` with these.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 549d1b768aadf5649b099e956c70c55bed58f8fb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


keyseven123 (changes requested):
Quite cool changes. 
Rerequest me once the PR is ready

keyseven123 (changes requested):

ls-1801 (approved):
I am happy with the changes in nebulastream. I am a little hesitant to move the changes regarding the compiler pass to nautilus, because we haven't really tested it on different platforms and I would assume it could be tricky to make that available via cmake.

alepping (approved):

keyseven123 (approved):
> update: llvm pass will now be built with nautilus directly and is then imported into NES via cmake. For nix, the pass is disabled because of differences in the build environment.



Can you add a comment in the package.nix? Otherwise, LGTM

PR #1192, opened 2025-10-16
Nix support for sanitizer and libcxx builds 

adrianmichalke:
Adds support for building with sanitizers via `-DUSE_SANITIZER=<thread/undefined/address>` and `-DUSE_LIBCXX_IF_AVAILABLE=true`


keyseven123 (changes requested):
Not sure what happened, but none of the Docker images are getting built. Re-request once they are, and I can test it locally.

adrianmichalke:
@keyseven123 it builds now. A Nautilus patch was malformed for vcpkg

PR #1197, opened 2025-10-21
Nix cache

adrianmichalke:
Adds support for the nix magic cache to the ci for the used dependencies. It has a rate limit that we might hit quite often, in those cases we fall back on building locally.
Also uses now a dedicated nix install action

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -20,21 +20,19 @@ jobs:
           ref: ${{ inputs.head_sha }}
           fetch-depth: 0
           submodules: recursive
-      - name: Install Nix
-        uses: cachix/install-nix-action@v31
-        with:
-          extra_nix_config: |
-            experimental-features = nix-command flakes
-      - name: Build with Nix
+
+      - uses: DeterminateSystems/nix-installer-action@main
+
+      - name: Magic Nix Cache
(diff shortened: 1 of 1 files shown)


ls-1801 (changes requested):
1. I am not sure if i can see the cache in action, i.e., i don't know if the cache is actually working. Can you confirm that it does?
2. There are some interesting logs produced by the action, are these okay?

I am not sure if that log is important.
```
 Important: The Determinate Nix Installer will stop distributing upstream Nix no sooner than January 1, 2026.
Timeline
* November 10: we are changing the installer to default to Determinate Nix.
  You can add the `--prefer-upstream-nix` flag now to keep upstream Nix as the default.
* January 1: we are removing support for installing upstream Nix.
  The `--prefer-upstream-nix` flag will not have an effect any longer.
The DeterminateSystems/nix-installer-action GitHub Action is also affected.
```

adrianmichalke:
@ls-1801, feel free to test it. From what I noticed, the change cuts the configuration time from ~5 min to ~1 min. Sanitizer builds should benefit even more.

The system.txt log file you noticed is uploaded by every job with the same content, its not nix job specific. On failures, we upload additional logs. The main log ´44_Run Nix Build_Nix Build.txt´ is always attached and should cover most issues.

ls-1801 (approved):

PR #1200, opened 2025-10-26
fix race condition in Emit

ls-1801:
The existing implementation incremented the chunk number separatly from the calculation of the last chunk flag.
This could lead to situations where a chunk is emitted with chunk number 1 and the last flag and concurrently the next buffer could be emitted with chunk number 2 and without the last flag.
This PR changes to logic to calculate both atomically.

## Verifying this change
The problem only occurred on the distributed branch, however it could technically happen on current main. The EmitPhysicalOperatorTest using undos thread fuzzing could reporoduce the issue.

- ExecutionEngine

### DIFF
--- a/nes-physical-operators/tests/EmitPhysicalOperatorTest.cpp
+++ b/nes-physical-operators/tests/EmitPhysicalOperatorTest.cpp
@@ -118,7 +118,7 @@ public:
         auto schema = Schema{}.addField("A_FIELD", DataType::Type::UINT32);
         auto layout = std::make_shared<RowLayout>(512, schema);
         EmitPhysicalOperator emit{OperatorHandlerId(0), std::make_shared<Interface::BufferRef::RowTupleBufferRef>(layout)};
-        handlers.emplace(OperatorHandlerId(0), std::make_shared<EmitOperatorHandler>());
+        handlers.insert_or_assign(OperatorHandlerId(0), std::make_shared<EmitOperatorHandler>());
         return emit;
     }
 


nes-physical-operators/src/EmitOperatorHandler.cpp
```diff
@@ -23,19 +23,30 @@
...
-    return newChunkNumber;
-}
-
-void EmitOperatorHandler::removeSequenceState(const SequenceNumberForOriginId seqNumberOriginId)
-{
-    seqNumberOriginIdToOutputChunkNumber.wlock()->erase(seqNumberOriginId);
-    seqNumberOriginIdToChunkStateInput.wlock()->erase(seqNumberOriginId);
+    SequenceNumberForOriginId seqNumberOriginId(buffer.getSequenceNumber(), buffer.getOriginId());
+    auto chunkNumberCounter = seqNumberOriginIdToOutputChunkNumber.wlock();
+    const auto newChunkNumber = ChunkNumber((*chunkNumberCounter)[seqNumberOriginId]++ + ChunkNumber::INITIAL);
```
keyseven123: Please extract the `++` into its own line. I can guarantee you that this will lead to problems down the line.

keyseven123 (changes requested):

nes-physical-operators/src/EmitOperatorHandler.cpp
ls-1801: I refactored the method to remove the two hashmaps, I hope its a little bit more clear what happens in the function

PR #1207, opened 2025-10-26
refactor(Optimizer): prepare optimizer for distributed

ls-1801:
This PR removes the `LegacyOptimizer` and introduces the `GlobalOptimizer` keeping the modular phase-based architecture.

The Optimizer receives its state via a QueryPlanningContext.

This PR introduces strong types for different stages of the query plan.

All usage of the legacy optimizer has been updated.

The GlobalOptmizer will be extended in the distributed branch.

## Verifying this change
- All tests pass


keyseven123 (dismissed):
I still see a filename `LegacyOptimizer`. Is this PR up-to-date? Please rerequest if so/change the pr description.

ls-1801:
See @adrianmichalke comment, I reverted the rename. 

I updated the pr Titel.

nes-nebuli/include/LegacyOptimizer/QueryPlanning.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+    explicit QueryPlanner(QueryPlanningContext context) : context(std::move(context)) { }
+
+public:
+    QueryPlanner() = delete;
+
+    static QueryPlanner with(QueryPlanningContext context) { return QueryPlanner{std::move(context)}; }
+
+    PlanStage::OptimizedLogicalPlan plan(PlanStage::BoundLogicalPlan&& boundPlan) &&;
+};
```
adrianmichalke: What's the purpose of the `QueryPlanner`? For now, it does only apply `optimize()`. I would be against adding it if it has no other reason

adrianmichalke (changes requested):
I have two more remarks. Otherwise, I think it's ready.

PR #1209, opened 2025-10-26
Move Statements and add Explain/WorkerStatus

ls-1801:
This PR changes the use of logical source within the statements to use just a name, wrapped in a strong type.
This PR moves statements into a dedicated folder in nebuli. This allows them to be shared by different frontends in the next pr.

This PR adds support for `chrono` and optional in the JSONOutputFormatter. This allows empty optionals to be omitted from the generated JSON.

A new statement handler `TopologyStatementHandler` was added, currently it allows querying the WorkerStatus from the worker. In the distributed branch the handler can add or remove worker.

The query statement handler can now handle an ExplainStatement.

## Verifying this change

Changes are currently untested but will be used in the next pr and the distributed branch

---

> [!NOTE]
> No file changes detected; the diff is empty.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit b8322acae4fc96d24cd79516165cb01b27529f85. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


nes-nebuli/include/Statements/StatementHandler.hpp
```diff
@@ -168,12 +184,73 @@ class QueryStatementHandler final : public StatementHandler<QueryStatementHandle
...
+}
+
+template <typename... HandlerT>
+void handleStatements(const std::vector<Statement>& statements, HandlerT&... handler)
+{
+    for (const auto& statement : statements)
+    {
+        tryCall(statement, handler...);
+    }
+}
```
Artraxon: I am not quite sure I I quite understand where and how this set of three methods is going to be used.

I am a bit afraid that this interface is a bit deceiving, as people writing a frontend will instead of calling handlers and dealing with the potential error in the correct way will just throw it in here and then start building controlflow around catching exceptions.

Can you point me to some code in the distributed branch that uses this? If you want to use this as a utility for a fail-fast mode maybe it makes more sense to move that to that frontend, because I could not find any usage in the current frontends.

nes-nebuli/test/test_large.sql
Artraxon: why delete this tests?

nes-systests/systest/src/QuerySubmitter.cpp
```diff
@@ -25,8 +25,8 @@
 #include <QueryManager/QueryManager.hpp>
 #include <Runtime/Execution/QueryStatus.hpp>
 #include <Serialization/QueryPlanSerializationUtil.hpp>
+#include <Util/PlanRenderer.hpp>
```
greenBene: Is this header required by this file if no code changed?

greenBene (approved):
Thank you for this change! :) I especially like that `CreatePhysicalSourceStatement` is now decoupled from logical sources and just contains a name of the logical source it refers to!

nes-nebuli/test/input/test_large.sql
ls-1801: The test has become increasingly flaky, and I can't be bothered to fix it, because the next PR introduces them again

nes-nebuli/include/Statements/StatementOutputAssembler.hpp
```diff
@@ -71,8 +74,21 @@ constexpr std::array<std::string_view, 4> sinkDescriptorOutputColumns{"sink_name
 using QueryIdOutputRowType = std::tuple<QueryId>;
```
alepping: Why use this alias?
The full type is shorter and gives the reader more information.
```
QueryIdOutputRowType
std::tuple<QueryId>
```

alepping: Imo, too many aliases end up being very confusing. It is nice to be able to just 'read the code' as much as possible. I see the benefit for the other aliases below though.

PR #1210, opened 2025-10-26
Replace QueryId with LocalQueryId

ls-1801:

In preparation for the distributed branch this PR promotes the QueryId to a class which currently only contains a LocalQueryId, the upcoming distributed PR will add a distributed id. In combination both are used to uniquely identify a query on a set of multiple workers.

The LocalQueryId will be a UUID to prevent accidental collisions across workers, and worker restarts (i.e. queryId 1 will not be the same after the worker is restarted).
The changes touches a lot of files so it is separated into a dedicated PR.

## Verifying this change
Existing Tests Pass

---
This is **part 2 of 4 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #1182 
- <kbd>&nbsp;3&nbsp;</kbd> #1212 
- <kbd>&nbsp;2&nbsp;</kbd> #1210 👈 
- <kbd>&nbsp;1&nbsp;</kbd> #1536


keyseven123 (commented):
So we will not have a global query id? Or how do we track what local query ids belong to what queries?

Artraxon (changes requested):
While I understand that the probability of local UUID collisions is extremely low, I don't really understand why this cannot just be autoincremented.

I mean, if we then introduce global query ids, collisions between local ids of different workers shouldn't matter, right?

nes-logical-operators/src/Plans/LogicalPlan.cpp
```diff
@@ -71,48 +71,57 @@ LogicalPlan& LogicalPlan::operator=(const LogicalPlan& other)
...
+LogicalPlan::LogicalPlan(LocalQueryId localQueryId, std::vector<LogicalOperator> rootOperators, std::string originalSql)
+    : localQueryId(localQueryId), rootOperators(std::move(rootOperators)), originalSql(std::move(originalSql))
 {
 }
 
 LogicalPlan promoteOperatorToRoot(const LogicalPlan& plan, const LogicalOperator& newRoot)
 {
     auto root = newRoot.withChildren(plan.getRootOperators());
-    return LogicalPlan(plan.getQueryId(), {std::move(root)}, plan.getOriginalSql());
+    return LogicalPlan({std::move(root)}, plan.getOriginalSql());
```
Artraxon: should the local query ID not be preserved?

Artraxon (changes requested):
As I mentioned already, I am not super happy with just using UUIDs instead of managing the global-local mapping properly, but I am fine with this as a temporary hack to not increase the scope of distributed further.

I hope that the future work on the coordinator will replace this.

keyseven123 (approved):

PR #1213, opened 2025-10-28
chore(Tidy): ignore `bugprone-use-after-move` with folly `tryWriteUntil`

ls-1801:
Silents a clang-tidy false positive.
No move happens if the write does not succeed. If a move happens, we return.

### DIFF
--- a/nes-systests/sources/Generator.test
+++ b/nes-systests/sources/Generator.test
@@ -123,3 +123,19 @@ SELECT * FROM generatorInline INTO generator_sink_inline;
 7
 8
 9
+
+CREATE LOGICAL SOURCE generator_with_timeout(id FLOAT64);
+CREATE PHYSICAL SOURCE FOR generator_with_timeout TYPE Generator SET(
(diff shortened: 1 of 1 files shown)


nes-plugins/Sources/GeneratorSource/GeneratorSource.cpp
```diff
@@ -76,7 +76,14 @@ void GeneratorSource::open()
...
 {
     auto totalElapsedTime = std::chrono::system_clock::now() - generatorStartTime;
-    NES_TRACE("Generated {} buffers in {}. Closing GeneratorSource.", generatedBuffers, totalElapsedTime);
+    if (generatedBuffers == 0)
+    {
+        NES_WARNING("Generated {} buffers in {}. Closing GeneratorSource.", generatedBuffers, totalElapsedTime);
+    }
+    else
+    {
+        NES_INFO("Generated {} buffers in {}. Closing GeneratorSource.", generatedBuffers, totalElapsedTime);
```
Artraxon: While it is kind of ok now, when we have auto instantiating sources in the future we will need to downgrade these again because not outputting tuples from a generator source might be expected query behavior in some cases.

PR #1216, opened 2025-11-03
Run 23h test with address sanitizer

adrianmichalke:


ls-1801 (changes requested):
not a big fan of this, the workflow call could be way more generic

PR #1220, opened 2025-11-03
CI: Run error code uniqueness check

adrianmichalke:
Ensures that our error codes are and remain unique. I prefer this CI check script to a `static_assert` because it does not cause any additional compilation time.

### DIFF
--- a/scripts/format.sh
+++ b/scripts/format.sh
@@ -172,6 +172,14 @@ python3 scripts/check_preamble.py || FAIL=1
 
 DISTANCE_MERGE_BASE=$DISTANCE_MERGE_BASE python3 scripts/check_todos.py || FAIL=1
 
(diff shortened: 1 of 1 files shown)


ls-1801 (changes requested):
What's wrong with just adding:

```bash
# error: no duplicate exception codes
DUPLICATES=$(grep '^EXCEPTION(' nes-common/include/ExceptionDefinitions.inc | sed -E 's/.*,[[:space:]]*([0-9]+).*/\1/' | sort | uniq -d)
if [ -n "$DUPLICATES" ]; then
    log_error "Duplicate Exception Codes found:"
    echo "$DUPLICATES"
fi

```

To the format script?

keyseven123 (approved):

PR #1223, opened 2025-11-06
fix(test): increase delay in DelayedTaskSubmitterTest to prevent flakiness

ls-1801:
The `DelayedTaskSubmitterTest` is flakey because it requires the internal thread to check the condition variable. 
Previous it used a delay of 10ms which is a little bit tight and caused the test to be flakey. 
The PR increases the timeout and documents why it is required in the first place.


alepping (approved):
@ls-1801 
I assume you tried to think about a solution that does not depend on a timeout, since that is the real problem.
Btw. did you ever try to run all tests with `stress -c ..` on and check if some of the tests fail because of timeouts?

PR #1230, opened 2025-11-14
fix(logical-operators): Use qualified detail namespace

Artraxon:
We are using the `NES::detail` namespace unqualified in LogicalOperator.hpp (so just `detail`) which for some reason worked, but @gaturchenko noticed that it doesn't when including the header in a plugin because it is ambiguous with NES::Nautilus::detail.

This PR makes all `NES::detail` usages in LogicalOperator.hpp qualified.


keyseven123 (approved):
As for any other PR reviewer, I plan on flattening the namespaces in `nes-nautilus`

PR #1234, opened 2025-11-17
Bump nautilus version

alepping:
Bumps nautilus version to current main.
Adds small fix for ambiguity in val.hpp.
The new version of Nautilus allows us to pass function attributes to `nautilus::invoke` calls, which can help the LLVM optimizer generating better code.

## Tests
Passes all tests


keyseven123 (approved):
LGTM. Can you give an example what attributes are supported?

PR #1242, opened 2025-11-21
fix(LogicalOperators): Cast to checked in builders and cast correctly to erased

Artraxon:
Currently using a couple of methods of the LogicalOperator wrapper doesn't compile because the return type is missing an explicit getAs to Checked.

Also getAs doesn't work as expected when casting to the ErasedLogicalOperator because it doesn't fulfill the concept, I added an explicit template specialization for it.


keyseven123 (approved):
Can we not have a test to check this? 
If too complex for now, this is fine.

PR #1245, opened 2025-11-25
Hotfix nightly stress test

keyseven123:
This PR fixes our nightly CI by running on our self-hosted to avoid running into a hard time limit of 6h
Additionally, we ported the queries to SQL and added more queries.

### DIFF
--- a/.github/workflows/execute_query.yml
+++ b/.github/workflows/execute_query.yml
@@ -92,7 +92,7 @@ jobs:
             echo "Waiting and monitoring worker..."
             ELAPSED_SECONDS=0
             MAX_SECONDS=${{ inputs.runtime_seconds }}
-            echo "Stress test runtime set to $ELAPSED_SECONDS seconds"
+            echo "Stress test runtime set to $MAX_SECONDS seconds"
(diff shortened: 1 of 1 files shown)


ls-1801 (changes requested):
Re-Request once final

.github/workflows/execute_query.yml
```diff
@@ -65,47 +70,47 @@ jobs:
...
             # wait and monitor
-            SECONDS=0
-            MAX_SECONDS=${{ inputs.runtime_seconds}}
-            echo "Stress test runtime set to $MAX_SECONDS seconds"
-            
-            while [ $SECONDS -lt $MAX_SECONDS ]; do
+            echo "Waiting and monitoring worker..."
+            ELAPSED_SECONDS=0
+            MAX_SECONDS=${{ inputs.runtime_seconds }}
+            echo "Stress test runtime set to $ELAPSED_SECONDS seconds"
```
ls-1801: doesn't that always echo 0 seconds?

ls-1801 (changes requested):

PR #1251, opened 2025-11-27
Bump cpptrace to 1.x.x

### LINKED ISSUE #1193: [TASK] Update ExceptionHandling Logic to match new cpptrace version
We recently updated the vcpkg baseline which introduced a breaking change with cpptrace. 
With version 1.0.0 cpptrace no longer uses macros for attaching trace to non traced exceptions, but instead uses c++ lambdas.

```c++
CPPTRACE_TRY {
    foo();
} CPPTRACE_CATCH(const std::logic_error& e) {
    handle_logic_error(e);
} CPPTRACE_CATCH_ALT(const std::exception& e) {
    handle_exception(e);
} CPPTRACE_CATCH_ALT(...) {
    handle_unknown_exception();
}
```

Newer Version:
```c++
cpptrace::try_catch(
    [&] { // try block
        foo();
    },
    [&] (const std::runtime_error& e) {
        handle_logic_error(e);
    },
    [&] (const std::exception& e) {
        handle_exception(e);
    },
    [&] () { // `catch(...)`
        handle_unknown_exception();
    }
);
```

While the author claims that the switch is straightforward, this requires us to rewrite some of our exception handlers, because returning from inside the lambda does not return from the function, thus this breaks the original functionality.

bgrb1:
This PR bumps the version of cpptrace to 1.0.4. With version 1.x.x there is a breaking change which removes the CPPTRACE_CATCH_ALT makro, and replaces it with a lambda-based construct. This required rewriting some exception handlers, especially because it changes the behavior of return statements (return from the lambda vs return from the parent function). This is mainly relevant for our GrpcService, where there is now a utility function to add-in the default Grpc exception handling via cpptrace lambdas. The PR also includes a small change to the stacktrace formatting, as the new lambda construct bloats stacktraces with cpptrace internals. So, for readability, cpptrace internals are filtered out

## Verifying this change
This change is tested by
- manual testing (exception injection via throw)
- some recorded stacktraces to show that the tracing works as expected: 
[stacktraces.txt](https://github.com/user-attachments/files/23790155/stacktraces.txt)
- successful compilation 

- error handling 
- grpc service 
- query engine

This PR closes #1193

### DIFF
--- a/.nix/cpptrace/package.nix
+++ b/.nix/cpptrace/package.nix
@@ -56,14 +56,18 @@ let
     in
     (if useLibcxx then libcxxStdenv else clangStdenv).mkDerivation rec {
       pname = "cpptrace";
-      version = "0.8.3";
+      version = "1.0.4";
 
-      src = fetchFromGitHub {
-        owner = "jeremy-rifkin";
-        repo = "cpptrace";
-        rev = "v${version}";
-        hash = "sha512-T+fmn1DvgAhUBjanRJBcXc3USAJe4Qs2v5UpiLj+HErLtRKoOCr9V/Pa5Nfpfla9v5H/q/2REKpBJ3r4exSSoQ==";
-      };
+  src = fetchFromGitHub {
+    owner = "jeremy-rifkin";
+    repo = "cpptrace";
+    rev = "v${version}";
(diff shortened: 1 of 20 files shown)


alepping (approved):

adrianmichalke (approved):

ls-1801 (changes requested):
there is a format check that would catch `catch(...)` and ask you to replace it with the old cpptrace macros. Could you also adapt the format check in format.sh:162

ls-1801 (approved):

PR #1266, opened 2025-12-04
feat(systest): Sequential systest query execution

gaturchenko:
This pull request adds the option to run queries in the same order as they are listed in the test file. This is achieved by passing the `--sequential` flag to the `systest` binary. Same as with the `-b` or `-n 1` options, one query will be run at a time. Lastly, running with this option ignores grouping of the queries based on config overrides, so `runQueriesAtRemoteWorker` will be called for every query.

- Systest


adrianmichalke:
Hi @gaturchenko, the clang-format check is failing. 
See https://github.com/nebulastream/nebulastream/actions/runs/19942081886/job/57182200458?pr=1266 
Please apply our format script onto your commit

keyseven123 (commented):
Do we need this anymore, as we have #1289 ?

PR #1274, opened 2025-12-11
Fix clang tidy warnings

yannisxiros:
This pull request fixes some warnings appearing in nightly

### DIFF
--- a/nes-common/include/ErrorHandling.hpp
+++ b/nes-common/include/ErrorHandling.hpp
@@ -162,7 +162,7 @@ Exception wrapExternalException(std::string contextMsg);
 [[nodiscard]] ErrorCode getCurrentErrorCode();
 
 [[nodiscard]] std::optional<ErrorCode> errorCodeExists(uint64_t code) noexcept;
-[[nodiscard]] std::optional<ErrorCode> errorTypeExists(std::string_view name) noexcept;
+[[nodiscard]] std::optional<ErrorCode> errorTypeExists(std::string_view codeOrTypeStr) noexcept;
 }
 
 namespace fmt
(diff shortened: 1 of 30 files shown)


adrianmichalke:
@yannisxiros cool, looks legit! Does it allow us to remove categories out of ´WarningsAsErrors´ in ´.clang-tidy´? If so, this change should be included here as well

nes-common/include/ErrorHandling.hpp
```diff
@@ -38,7 +38,7 @@ namespace NES
 /// We use a trick to keep the shorthand version of error code (e.g., `return ErrorCode::UnknownException;`) via the following using.
 namespace ErrorCodeDetail
 {
-enum ErrorCode
+enum ErrorCode : std::uint16_t
```
keyseven123: is u8 not enough?

nes-nautilus/tests/UnitTests/VarValTest.cpp
```diff
@@ -85,7 +85,7 @@ TEST_F(VarValTest, SimpleMove)
     auto testVarValMove = []<typename T>(const T value)
     {
         const VarVal varVal = nautilus::val<T>(value);
-        const VarVal varValCopy = varVal;
+        const VarVal varValCopy = varVal; /// NOLINT
```
keyseven123: What is the problem with this line?

yannisxiros: clang warning states that it would be more performant to just take a const ref instead of copying, but this test should use the copy constructor

nes-common/include/ErrorHandling.hpp
ls-1801: We have error codes in the 10 thousands

PR #1301, opened 2025-12-19
Add dumpGraph option for the worker

nomorehumor:
New PR, old one unfortunately got closed after I made a mistake and force-pushed it into the feature branch (GitHub has closed the PR and doesn't allow me to reopen it)

Previous PR: #1246

Adds support for `--worker.dumpGraph [true|false]` boolean flag
The flag controls, whether the Nautilus compilation graph will be dumped 
Basically it sets the Nautilus option `dump.graph` to whatever value is passed (see docs [here](https://github.com/nebulastream/nautilus/blob/main/docs/options.md))

## Verifying this change
Run systest with the mentioned flag, you should see the dumped graph in the dump directory

- Worker Configuration
- QueryCompiler

## Documentation
- The change is reflected in the Nautilus documentation, there is not seems to be a centralised documentation for all worker options in nebulastream repo

### DIFF
--- a/nes-runtime/include/Util/DumpMode.hpp
+++ b/nes-runtime/include/Util/DumpMode.hpp
@@ -30,9 +30,9 @@ public:
 
     DumpMode(Options option, bool dumpGraph) : option(option), dumpGraph(dumpGraph) { }
 
-    Options getDumpOption() const { return option; }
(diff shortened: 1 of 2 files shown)


keyseven123 (changes requested):
Please use absolute include paths. Otherwise, LGTM

PR #1307, opened 2025-12-23
feat(inference): Adds basic ML inference

gaturchenko:
**(for example:)** 
This pull request adds support for ML inference via the dedicated operator.

## Verifying this change
This change is verified by the dedicated `InferenceDifferentDataTypes.test` test

This PR affects different parts of the system:
* Extends the Anltr grammar
* Adds new logical and physical operators
* Extends existing `.proto` definitions
* Adds an optimizer rule and a lowering pass
* Adds new dependencies to the VCPKG manifest

## Documentation
The basic usage example is documented in the test file. A more detailed markdown documentation is a work-in-progress

### DIFF
--- a/nes-logical-operators/include/Operators/Windows/Aggregations/WindowAggregationLogicalFunction.hpp
+++ b/nes-logical-operators/include/Operators/Windows/Aggregations/WindowAggregationLogicalFunction.hpp
@@ -52,9 +52,6 @@ public:
 
     [[nodiscard]] virtual std::string_view getName() const noexcept = 0;
 
-    ///@return If this is set, the aggregations lift function will be called with buffers in sequence number order.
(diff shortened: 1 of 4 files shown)


Artraxon:
I assigned myself for the optimizer/IR parts and will give feedback in the next days

docker/dependency/Base.dockerfile
```diff
@@ -26,7 +26,9 @@ RUN apt update -y && apt install \
     openssl \
     ccache \
     ninja-build \
+    bison \
     pkg-config \
+    pipx \
```
Artraxon: For what do you need bison?

I guess it also a philosophical question whether pipx belongs to base or dependency, but I am fine with keeping it here

PR #1308, opened 2025-12-25
Setsum plugin

Mancer1:
Setsum plugin 

pretty much same as checksum plugin, except the output is in 8 columns. so some stylistic improvements may be needed.


Mancer1:
to the wrong repo

PR #1312, opened 2026-01-08
column layout support for systests

### LINKED ISSUE #759: [TASK] systests not working with column layout
### Description

Currently we support variable data layouts namely row and columnar layout. 
But we don't test the columnar layout in the systests as they seem to only be working using only the row layout.

One idea is to use a new data layout conversion operator, since the sources currently only save data into row layout.

### Acceptance Criteria

- [ ] add another version of systests, that validate the current systests are also working, if all operators are using the columnar data layout.
- [ ] all tests pass

EricBenschneider:

This pull request adds support for columnar layout. 
This change introduces a `LogicalMemorySwapOperator` that swaps the memory layout between row and columnar layout.

Using the new phase `DecideMemoryLayout` of the `QueryOptimizer` allows to insert the memory swaps in between operators. This behavior is adjustable by setting the `rowLayoutRatio` of the `QueryExecutionConfiguration`.
A ratio of `1` refers to the whole query being executed using row layout and a ratio of `0` to using column layout.
Since the current source and sink implementation expects data to be in row layout, when using column layout, all operator are swapped to column between source and sink. 
Using a ratio of `0.5` leads to roughly half of the operators being in row with an added swap pipeline for every change in layout. 

The logical memory swap gets lowered to a scan and emit pipeline that both use a different `memoryLayoutType` in their `TupleBufferRef`.
 

## Verifying this change
This change is tested by extending all systests to run using the row layout ratios `0` and `1`.

- ExecutionEngine
- QueryCompiler
- QueryEngine

## Documentation

This PR closes #759

### DIFF
--- a/nes-query-optimizer/src/Phases/DecideMemoryLayout.cpp
+++ b/nes-query-optimizer/src/Phases/DecideMemoryLayout.cpp
@@ -50,55 +50,83 @@ LogicalOperator DecideMemoryLayout::apply(const LogicalOperator& logicalOperator
     }
     else
     {
-        if (conf.layoutStrategy == MemoryLayoutStrategy::SWAP_ALL_COL)
-        {
-            tryInsert(traitSet, MemoryLayoutTypeTrait{MemoryLayoutType::COLUMNAR_LAYOUT});
-        }
-        else
-        {
-            tryInsert(traitSet, MemoryLayoutTypeTrait{MemoryLayoutType::ROW_LAYOUT});
+        /// row_layout_ratio likelihood of the layout being row
+        std::random_device rd;
+        std::mt19937 gen(rd());
+
+        if (!traitSet.contains<MemoryLayoutTypeTrait>())
+        { /// check if not already set
+            if (std::uniform_real_distribution<> dis(0.0, 1.0); dis(gen) <= conf.rowLayoutRatio.getValue())
+            {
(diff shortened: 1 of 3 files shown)


nes-query-optimizer/src/Phases/DecideMemoryLayout.cpp
```diff
@@ -16,27 +16,120 @@
...
+    }
+    else if (logicalOperator.getName() == "Sink")
+    { /// add swap before sink
+        /// sink currently only in row layout
+        tryInsert(traitSet, MemoryLayoutTypeTrait{MemoryLayoutType::ROW_LAYOUT});
+    }
+    else
+    {
+        /// row_layout_ratio likelihood of the layout being row
+        std::random_device rd;
```
alepping: the decision should not be based purely on randomness

nes-query-optimizer/src/Phases/DecideMemoryLayout.cpp
```diff
@@ -16,27 +16,120 @@
...
+            else
+            {
+                tryInsert(traitSet, MemoryLayoutTypeTrait{MemoryLayoutType::COLUMNAR_LAYOUT});
+            }
+        }
+    }
+
+    auto childs = logicalOperator.getChildren();
+
+    if (conf.rowLayoutRatio.getValue() > 0 && conf.rowLayoutRatio.getValue() < 1.0)
```
keyseven123: When is this if false? Do we need it?

PR #1314, opened 2026-01-08
Fixes race condition in query start/stop

ls-1801:
The raw pointer used during the query setup completed callback could be dangling if the query was terminated during setup.
The problem is that the Callback could potentially execute concurrently to the destructor of the RunningQueryPlan.
This is no longer possible with the new Callback implementation as the destructor will block if the callback is executing.

----

As it turns out removing once race condition just revealed a second race condition and a potential deadlock.
The code in the engine was changed to never destruct/stop query plans while holding the Query State lock. The code will move the query plan out of the AtomicState and stop or destroy after the Query state was changed.

Lastly the RunningSource now has a lock to prevent concurrent access to the underlying source thread. A source could be started and immediately request end of stream which would cause a different thread to write to the thread object, while the first thread is still starting the thread.

### DIFF
--- a/nes-common/tests/Util/include/BaseUnitTest.hpp
+++ b/nes-common/tests/Util/include/BaseUnitTest.hpp
@@ -93,7 +93,13 @@ private:
     std::unique_ptr<std::thread> waitThread;
     std::shared_ptr<std::promise<bool>> testCompletion;
     std::atomic<bool> testCompletionSet{false};
-    static constexpr uint64_t WAIT_TIME_SETUP = 5;
+
+#if defined(__has_feature) && __has_feature(thread_sanitizer)
+    /// We double the timeout for tests when running with the thread sanitizer, as it significantly slows down execution for some tests
+    static constexpr std::chrono::minutes WAIT_TIME_SETUP{10};
+#else
+    static constexpr std::chrono::minutes WAIT_TIME_SETUP{5};
+#endif
 };
 
 /**
--- a/nes-common/tests/Util/src/BaseUnitTest.cpp
+++ b/nes-common/tests/Util/src/BaseUnitTest.cpp
@@ -81,7 +81,7 @@ void TestWaitingHelper::startWaitingThread(std::string testName)
         [this, testName = std::move(testName)]() mutable
         {
             auto future = testCompletion->get_future();
-            switch (future.wait_for(std::chrono::minutes(WAIT_TIME_SETUP)))
+            switch (future.wait_for(WAIT_TIME_SETUP))
             {
                 case std::future_status::ready: {
                     CPPTRACE_TRY


nes-query-engine/RunningQueryPlan.cpp
```diff
@@ -270,52 +180,53 @@ std::pair<std::unique_ptr<RunningQueryPlan>, CallbackRef> RunningQueryPlan::star
...
-        pipelineSetupCallbackRef,
-        [runningPlan = runningPlan.get(),
-         queryId,
-         listener = std::move(listener),
-         &controller,
-         &emitter,
-         sources = std::move(sources)]() mutable
+    /// The Setup Callback tracks the task lifetimes of al pipeline setup tasks. Either a task will fail and terminate the query, which will cancel the setup callback.
+    /// or the setup task completes and destroys the callback reference, until the last setup task which will call the setup callback.
+    /// The Setup Callback it self is used to start the sources once all pipelines have been set up.
```
alepping: Very nice comment generally (together with what I remember of your DD).
Given that other ppl may not be familiar with your DD, I'd slightly change the comment:
```md
    /// The QueryEngine uses the setup callback to start the sources once all pipelines have been set up, effectively starting the query.
    /// The setup callback tracks the lifetimes of all pipeline setup tasks. Either a task will fail and terminate the query, which will cancel the setup callback.
    /// Or the setup task completes and destroys the callback reference.
    /// When the last setup task is destroyed, it sets the guard counter of the callback reference to 0, triggering the execution of setup callback.
```

nes-query-engine/RunningQueryPlan.cpp
```diff
@@ -270,52 +180,53 @@ std::pair<std::unique_ptr<RunningQueryPlan>, CallbackRef> RunningQueryPlan::star
...
+    internal.allPipelinesStarted.setCallback(
+        [
+            /// This raw pointer is guaranteed to be alive. As the running callback holds the Callbacks owner object. The callback
+            /// guarantees, that its owner is either alive during the callback execution or that the callback will never execute
```
alepping: You might want to say, why we need to access the raw pointer in the first place.

nes-query-engine/RunningQueryPlan.hpp
```diff
@@ -185,10 +123,12 @@ struct RunningQueryPlan final
         WorkEmitter&,
         std::shared_ptr<QueryLifetimeListener>);
 
-    /// Stopping a RunningQueryPlan will:
-    /// 1. Initialize pipeline termination asynchronously.
-    /// 2. Block until all sources are terminated.
-    static std::unique_ptr<StoppingQueryPlan> stop(std::unique_ptr<RunningQueryPlan> runningQueryPlan);
+    /// Stopping a RunningQueryPlan will return:
+    /// The stopped query plan which keeps the pipelines alive until the soft stop was propagated.
+    /// The callback to initialize the soft stop, by destroying sources. This can only happen outside the AtomicState transition, otherwise
```
yschroeder97: What exactly can only happen outside the AtomicState transition?
I assume destroying the sources, but its unclear what "this" means.

nes-query-engine/RunningQueryPlan.hpp
```diff
@@ -185,10 +123,12 @@ struct RunningQueryPlan final
...
+    /// this might deadlock
```
yschroeder97: Again what is "this"? What causes the deadlock?

nes-query-engine/RunningQueryPlan.cpp
```diff
@@ -362,25 +260,31 @@ std::pair<std::unique_ptr<RunningQueryPlan>, CallbackRef> RunningQueryPlan::star
...
-    internal.sources.clear();
+    return {
+        std::make_unique<StoppingQueryPlan>(
+            std::move(internal.qep), std::move(internal.listeners), std::move(internal.allPipelinesExpired)),
+        [callbackOwner = std::move(internal.allPipelinesStarted),
+         pipelines = std::move(internal.pipelines),
+         sources = std::move(internal.sources)]() mutable
+        {
+            /// Destroying the sources needs to ensure that there is no concurrency on the sources vector.
+            /// The setup callback may run concurrently so it is essntial that the callback owner is destroyed before
```
yschroeder97:
```suggestion
            /// The setup callback may run concurrently so it is essential that the callback owner is destroyed before
```

nes-query-engine/RunningQueryPlan.hpp
ls-1801: see above

yschroeder97 (approved):

PR #1328, opened 2026-01-15
fix(CSVFormatter): handles delimiters in double quotes correctly

adrianmichalke:
now supports:

```
# name: formatter/CSVWithComma.test
# description: successful handling should ignore delimiters in double quotes
# groups: [Formatter]

CREATE LOGICAL SOURCE stream(text VARSIZED, text2 VARSIZED);
CREATE PHYSICAL SOURCE FOR stream TYPE FILE;
ATTACH FILE small/comma.csv

CREATE SINK file_sink(stream.text VARSIZED, stream.text2 VARSIZED) TYPE File;

SELECT * FROM stream INTO file_sink;
----
"a","1,2,3"
",","4,5,6"
b,"7,,8"
c,""
d,","
"e","18,,19"
"f","20,21,22,23,24"
g,"27,,28,29"

SELECT * FROM stream WHERE text2 == VARSIZED(",") INTO file_sink;
----
d,","

SELECT * FROM stream WHERE text2 != VARSIZED(",") INTO file_sink;
----
"a","1,2,3"
",","4,5,6"
b,"7,,8"
c,""
"e","18,,19"
"f","20,21,22,23,24"
g,"27,,28,29"
```


alepping (changes requested):
The reason a `","` value is not supported, is not because of bad handling of quotes, but because `,` is an essential structural symbol for C(omma)SV data.
Similarly, a `'\n'` value is also not supported.
The current logic can handle escaped values that are non-structural symbols (characters other than `,` and `\n`)
The correct solution would be to generally support escaping of structural characters.
That includes allowing for escaped newlines, which is tricky because the escape sequences may be split across buffers.
In that case, the indexer incorrectly interprets an escaped newline as a structural symbol.

I see two possible possible ways forward:
1. I wonder why you need an escaped `,` in the first place. Can't you use another character, for whatever the `,` represents? If it is not necessary, I would strongly prefer not to push forward with these changes.
2. I have a prototype that can correctly  index CSV with escaped commas and newlines. I have not created a PR though, escaped newlines introduce ambiguity, where a CSVInputFormatIndexer may not certainly know if the chunk started in an escaped sequence or not.

nes-systests/formatter/CSVWithComma.test
```diff
@@ -0,0 +1,34 @@
...
+CREATE PHYSICAL SOURCE FOR stream TYPE FILE;
+ATTACH FILE small/comma.csv
+
+CREATE SINK file_sink(stream.text VARSIZED, stream.text2 VARSIZED) TYPE File;
+
+SELECT * FROM stream INTO file_sink;
+----
+"a","1,2,3"
+",","4,5,6"
+b,"7,,8"
```
alepping: @keyseven123 and @adrianmichalke 
do we really need CSV that contains both unquoted and quoted strings?
From my point of view, that is a separate issue from supporting commas in quoted strings.
If so, than we might as well support (RFC-4180), which also allows newlines in quoted strings.

Image

keyseven123 (changes requested):
Re-request once the changes discussed are implemented

keyseven123 (approved):

PR #1333, opened 2026-01-19
fix(PipeliningPhase): Corrected access of Sink Format via wrong method

### LINKED ISSUE #1332: [BUG] Pipeliningphase accesses sink format via wrong method
### Version / Commit

f34fa6d

### What happened?

In `buildPipeLineRecursively()` in `PipeliningPhase.cpp`, line 211, the sink (output) format is accessed via `getSinkType()`, which returns the concrete instance of Sink (File, Checksum, Print, Void) and not the format (Native, CSV, JSON).

Should use `getFormatType()` instead.

### How to reproduce

Currently does not cause any issues since it is used to check for "Native" Sink format, which will probably only be relevant for distributed NES, since it would lead to the addition of an unnecessary scan + inputformatting operator in this case.

### Operating System

Ubuntu 24.04.3 LTS

### Deployment details

Bug is in code, not deployment specific

### Anything else

_No response_

### Code of Conduct

LeoRaasch:
This pull request fixes a small bug in which the the sink format is accessed via `getSinkType()` instead of `getFormatType()`.
Location: PipeliningPhase.cpp, Line 211

Even though getFormatType() can be called for any Sink Descriptor, Void and Checksum Sink do not include this parameter in their config (Even though the parameter itself is defined in `SinkDescriptor.hpp`). This is why must catch an out_of_range error here.

### Small changelog
- Calls `getFormatType()` instead 
- `getFormatType()` now is able to catch the out of range error that accurs for sinks that do not use an output format. Because of this, the return type has been changed to an optional string view

## Verifying this change
This change is tested by
- Nothing currently, it's relevant for Source - Sink Pipelines with nothing in between, where input and output format are "Native". This seems to mainly occur in a distributed setting on intermediate nodes.

- QueryCompiler
- Sink Descriptor

## Documentation
- Not needed

This PR closes #1332

### DIFF
--- a/nes-query-compiler/src/Phases/PipeliningPhase.cpp
+++ b/nes-query-compiler/src/Phases/PipeliningPhase.cpp
@@ -17,6 +17,8 @@
 #include <cstdint>
 #include <memory>
 #include <optional>
+#include <stdexcept>
+#include <string>
 #include <unordered_map>
 #include <utility>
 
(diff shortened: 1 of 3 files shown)


nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -208,7 +210,19 @@ void buildPipelineRecursively(
...
+
+            /// Currently, not all sinks configs actually include a INPUT_FORMAT parameter, so getFormatType might throw an error.
+            /// Once we refactor output-formatting, we should change it so that every sink has a format type parameter in its config.
+            try
+            {
+                sinkFormat = toUpperCase(sink->getDescriptor().getFormatType());
+            }
+            catch (std::out_of_range& error)
+            {
+                sinkFormat = "";
```
keyseven123: I would like to see a warning here. Something like: "No format for sink ...". 
This would help people who implement a new sink but have forgotten to implement a new format to see this warning in the logs.

PR #1352, opened 2026-02-04
Systest event support

adrianmichalke:
Add InlineEventSources that allows: 
```
- `<CRASH>`: terminate the local worker process used by systest.
- `<RESTART>`: start a fresh worker
- `<DELAY <milliseconds>>`
```

Example:
```sql
CREATE LOGICAL SOURCE stream(i64 INT64, ts UINT64);
CREATE PHYSICAL SOURCE FOR stream TYPE TCP;
ATTACH INLINE
10,0
<CRASH>
12,61
<RESTART>
12,60
13,170

CREATE SINK sink(stream.start UINT64, stream.end UINT64, stream.cnt UINT64) TYPE File;

SELECT start, end, COUNT(i64) as cnt FROM stream WINDOW TUMBLING(ts, size 100 ms) INTO sink;
----
0,100,2
100,200,1
```

useful for testing e.g. checkpointing e2e


ls-1801 (commented):
I don't want to block, this but I am not going to accept it. 

You have code ownership over systest, so it's your decision if you want to have this feature.

My 2 cents are that this adds 1000 lines of code for something that could be handled by 20 lines of bash. It adds so much complexity to have a single system test that can use a nice api. I think that the api is also flawed in its current state as you cannot really generate any reasonable volume of data.

keyseven123 (changes requested):
I really like the idea of giving us the option of using events in SLTs. Having this directly in the SLT enables us to have everything in the same file and it adds determinism. 

 I think though that the current implementation adds a lot of files/lines for this. Especially adding a completely separate runner increases the maintenance a lot. Maybe they can be combined into a single implementation.

alepping (changes requested):
@adrianmichalke 
I really like these changes in general.
This:
```sql
10,0
<CRASH>
<RESTART>
12,60
```
is awesome.

However, this should work together with the changes introduced in the distributed branch imo.
First, because the distributed branch allows us to setup multiple workers in a specific topology, which I think works well with these changes here, making them even more interesting/powerful.
Second, because the distributed branch introduces docker-based testing (`.bats` / `.expect`) and I would really like to avoid fragmentation of our testing logic.
I think it makes a lot of sense to use docker-based testing, also for testing against external systems (which could also be interesting for you btw.).

Tell me if I misunderstand and this should be handled separately from the distributed branch, but otherwise I'd argue that this should build on top of the distributed branch and use its logic.

PR #1354, opened 2026-02-05
Add mqtt source with test container systests

### LINKED ISSUE #602: [SUB-EPIC] Add testing infrastructure for I/O-related components
### Description

Part of [EPIC on source redesign](https://github.com/nebulastream/nebulastream-public/issues/595).

As NES is intended to be able to interact with numerous systems via sources and sinks, we need to have tests for these interactions.
For this, we need machinery to be able to incorporate external systems in our tests.
This may include ways to spin up these external systems, provide mechanism to interact with them in a unified way, and shut them down again.
As part of the refactoring effort of building an asynchronous execution environment for sources, the `MQTTSource` represents the first truly external system which needs to be tested properly. 

### Acceptance Criteria

- Examples that showcase the use of external systems

alepping:
Adds support for testing sources that rely on external software, e.g., a mosquitto broker & producer, or an MSSQL Server database.
Introduces `.bats` testing (taken from distributed branch) to generate a docker-compose file.
The docker-compose file can be configured to include docker containers that run mosquitto brokers or external database systems.
Currently the data producers in the docker containers must exist before we start the query.
This requires configuring the MQTT broker to retain messages, otherwise the MQTT producer would send the contents of the test file to the broker, which then drops the messages, before we run the query.

Depends on #1353

This PR closes #602


ls-1801 (changes requested):
I would very much want to wait until the distributed pr is merged, which implements most of the logic for the docker tests, and also handles docker in docker

PR #1360, opened 2026-02-09
feat(serialization) Reflection based serialization

greenBene:
This PR changes the serialization approach of Operators from protobuf to a reflect-cpp based approach 
* to simplify the writing of individual serializations of operators and,
* to enable the flexibility to serialize any kind of operator specific data without the need to extend the the protobuf data model 

## Verifying this change
This change is tested by
* all preexisting tests must continue to run successfully 

* logical operators
* logical functions
* serialization & deserialization

### DIFF
--- a/.github/workflows/get_dev_images.yml
+++ b/.github/workflows/get_dev_images.yml
@@ -145,7 +145,7 @@ jobs:
     if: needs.detect-dependency-changes.outputs.build-dependency == 'true'
     runs-on: [ self-hosted, linux, Build, "${{ matrix.arch }}", dep-builder ]
     concurrency:
-      group: build-dev-image-lock-${{ needs.detect-dependency-changes.outputs.tag }}
+      group: build-dev-image-lock-${{ needs.detect-dependency-changes.outputs.tag }}-${{matrix.arch}}-${{matrix.stdlib}}-${{matrix.sanitizer}}
       # To reduce the amount of work, i.e., building the same deps multiple times, we only allow one job to continue.
       # We do not cancel the in progress one, as it might have run for quite some time and is close to finishing.
       cancel-in-progress: false


.clang-tidy
```diff
@@ -222,4 +222,4 @@ CheckOptions:
   # Workaround for header bug around std::expected
   # It always tries to insert the libc++ implementation headers instead of the stdlib headers.
   # Also, it sometimes even tries to remove the stdlib headers as not needed, then inserts the libc++ headers.
-  misc-include-cleaner.IgnoreHeaders: 'expected;__.*expected\.h;__.*unexpected\.h'
+  misc-include-cleaner.IgnoreHeaders: 'expected;__.*expected\.h;__.*unexpected\.h;LogicalFunctionReflection\.hpp;TraitReflection\.hpp'
```
ls-1801: What's the problem with including these?

nes-common/include/Util/Reflection.hpp
```diff
@@ -0,0 +1,161 @@
...
+#include <ErrorHandling.hpp>
+
+namespace NES
+{
+
+struct Reflected
+{
+    Reflected() : value(std::nullopt) { }
+
+    ///NOLINTNEXTLINE
```
ls-1801: Please don't blindly supress all clang tidy warnings

///NOLINTNEXTLINE(clang-tidy-rule-name) preferably an explanation here, unless obvious

ls-1801 (changes requested):
Happy to see that this removes some inconsistencies in our serialization. I would ask you to provide a document that explains what needs to be implemented when you create a new operator/function/trait to implement proper serialization, it does not need to go into details how it works just what needs to be implemented to have serialization support.

nes-query-optimizer/include/Traits/Trait.hpp
```diff
@@ -139,6 +137,7 @@ struct TypedTrait
     /// @tparam T The type of the trait. Must satisfy TraitConcept concept.
     /// @param op The trait to wrap.
     template <typename T>
+    requires(!std::same_as<T, rfl::Error>)
```
Artraxon: same as with logical functions. I am surprised we didn't need it for the logical operators?

Artraxon (changes requested):
Good Job!

nes-common/include/Util/Reflection.hpp
```diff
@@ -0,0 +1,161 @@
...
+    Reflected operator()(const Reflected& field) const { return field; }
+};
+
+template <>
+struct Unreflector<Reflected>
+{
+    Reflected operator()(const Reflected& field) const { return field; }
+};
+
+} /// namespace NES
```
adrianmichalke: nit: rm

nes-logical-operators/src/Operators/Sources/SourceNameLogicalOperator.cpp
```diff
@@ -128,9 +128,17 @@ std::string SourceNameLogicalOperator::getLogicalSourceName() const
...
 {
     PRECONDITION(false, "no serialize for SourceNameLogicalOperator defined. Serialization happens with SourceDescriptorLogicalOperator");
+    throw CannotSerialize("no serialize for SourceNameLogicalOperator defined. Serialization happens with SourceDescriptorLogicalOperator");
+}
+
+SourceNameLogicalOperator Unreflector<SourceNameLogicalOperator>::operator()(const Reflected&) const
+{
+    PRECONDITION(false, "no serialize for SourceNameLogicalOperator defined. Serialization happens with SourceDescriptorLogicalOperator");
+    throw CannotDeserialize(
+        "no serialize for SourceNameLogicalOperator defined. Serialization happens with SourceDescriptorLogicalOperator");
```
adrianmichalke: Im not sure if I like this pattern. Usually we just have the Precondition followed by a std::unreachable

nes-configurations/src/Configurations/ConfigSerialization.cpp
```diff
@@ -0,0 +1,85 @@
...
+Reflected Reflector<ProjectionList>::operator()(const ProjectionList&) const
+{
+    throw CannotSerialize("ProjectionList is not expected to be reflected");
+}
+
+ProjectionList Unreflector<ProjectionList>::operator()(const Reflected&) const
+{
+    throw CannotDeserialize("ProjectionList is not expected to be unreflected");
+}
+
```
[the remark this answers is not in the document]
greenBene: Thanks for pushing this! I originally thought that the SQL parser still relies on the above mentioned data types, but I validated this assumption and noticed, that no operator/function relies on them anymore for creation. 

While the types were still referenced in some Registry functions, they were only used during serialization which was replaced by reflection in this PR. All Operators are currently directly created (without using the registry) in the SQL parsing/binding, and while some functions use the registry for creation, they don't use the above mentioned config types. 

Thus, I removed the now unused code.

nes-logical-operators/include/Functions/LogicalFunction.hpp
```diff
@@ -133,6 +135,7 @@ struct TypedLogicalFunction
     /// @tparam T The type of the function. Must satisfy LogicalFunctionConcept concept.
     /// @param op The function to wrap.
     template <typename T>
+    requires(!std::same_as<T, rfl::Error>)
```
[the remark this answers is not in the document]
greenBene: Okay, I found a way that achieves the same thing without requiring the rfl header: I require that the given type has a getChildren function that returns a logical Function vector. Is this similar to what you meant by psedo concept?

```cpp
    requires requires(const T& function)
    {
        { function.getChildren() } -> std::convertible_to<std::vector<LogicalFunction>>;
    }
```

nes-query-optimizer/include/Traits/Trait.hpp
greenBene: Same as above:

```cpp
    requires requires(const T& trait)
    {
        { trait.getName() } -> std::convertible_to<std::string_view>;
    }
```

nes-logical-operators/include/Operators/Sources/InlineSourceLogicalOperator.hpp
```diff
@@ -41,7 +41,6 @@ class InlineSourceLogicalOperator
         std::unordered_map<std::string, std::string> parserConfig);
 
     [[nodiscard]] bool operator==(const InlineSourceLogicalOperator& rhs) const;
-    static void serialize(SerializableOperator&);
```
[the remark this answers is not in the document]
greenBene: The InlineSourceLogicalOperators are not directly serialized/reflected, but are replaced by SourceDescriptorLogicalOperators before. If this logic changes we can add the reflection logic quite easily, but I would not add the code if it is not used :)

nes-logical-operators/src/Serialization/WindowTypeReflection.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        if (const auto tumblingWindow = std::dynamic_pointer_cast<Windowing::TumblingWindow>(windowType))
+        {
+            return std::make_pair("TumblingWindow", reflect(tumblingWindow));
+        }
+        if (const auto slidingWindow = std::dynamic_pointer_cast<Windowing::SlidingWindow>(windowType))
+        {
+            return std::make_pair("SlidingWindow", reflect(slidingWindow));
+        }
+        throw CannotDeserialize("Invalid Window Type");
```
ls-1801: Error message should distinguish serialize vs deserialize and include the type when available.
```suggestion
        throw CannotSerialize("Cannot serialize unknown window type");
```

nes-logical-operators/src/Serialization/WindowTypeReflection.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        if (type == "TumblingWindow")
+        {
+            return std::make_shared<Windowing::TumblingWindow>(unreflect<Windowing::TumblingWindow>(config));
+        }
+        if (type == "SlidingWindow")
+        {
+            return std::make_shared<Windowing::SlidingWindow>(unreflect<Windowing::SlidingWindow>(config));
+        }
+        throw CannotDeserialize("Invalid Window Type");
```
ls-1801: Same here — include the type string for debugging.
```suggestion
        throw CannotDeserialize("Cannot deserialize unknown window type {}", type);
```

nes-logical-operators/include/Serialization/WindowTypeReflection.hpp
```diff
@@ -14,15 +14,25 @@
...
+
+#include <memory>
+#include <string>
+#include <Util/Reflection.hpp>
 
 namespace NES
 {
-class TraitSetSerializationUtil
+
+Reflected reflectWindowType(std::shared_ptr<Windowing::WindowType> windowType);
```
ls-1801: No need for `shared_ptr` here — just take a const ref. The function only reads the window type.
```suggestion
Reflected reflectWindowType(const Windowing::WindowType& windowType);
```

nes-configurations/src/Configurations/Descriptor.cpp
```diff
@@ -198,4 +173,96 @@ std::string Descriptor::toStringConfig() const
...
+
+    return reflect(reflectedConfig);
+}
+
+DescriptorConfig::Config Descriptor::unreflectConfig(const Reflected& rfl)
+{
+    auto reflectedConfig = unreflect<ReflectedDescriptorConfig>(rfl);
+    DescriptorConfig::Config config;
+
+    for (auto [key, pair] : reflectedConfig.config)
```
ls-1801: Iterates by value, copying every map entry. Should be by const ref.
```suggestion
    for (const auto& [key, pair] : reflectedConfig.config)
```

nes-configurations/src/Configurations/Descriptor.cpp
```diff
@@ -198,4 +173,96 @@ std::string Descriptor::toStringConfig() const
...
+        {
+            config[key] = unreflect<std::string>(value);
+        }
+        else if (type == "EnumWrapper")
+        {
+            config[key] = unreflect<EnumWrapper>(value);
+        }
+        else
+        {
+            throw CannotDeserialize();
```
ls-1801: Empty error message — include the unknown type string for debugging.
```suggestion
            throw CannotDeserialize("Unknown descriptor config type {}", type);
```

nes-data-types/include/DataTypes/Schema.hpp
```diff
@@ -118,6 +119,45 @@ class Schema
 /// Returns a copy of the input schema without any source qualifier on the schema fields
 Schema withoutSourceQualifier(const Schema& input);
 
+namespace detail
+{
+struct ReflectedField
```
ls-1801: `ReflectedField` only serializes `DataType::Type` (the enum), not the full `DataType`. This will break when `DataType` gains additional fields (e.g., a nullable flag). The serialization should reflect/unreflect the entire `DataType` instead of just the enum to be future-proof.

nes-logical-operators/include/Operators/Windows/JoinLogicalOperator.hpp
```diff
@@ -119,7 +85,31 @@ class JoinLogicalOperator final : public OriginIdAssigner
...
+{
+    JoinLogicalOperator operator()(const Reflected& reflected) const;
 };
 
 static_assert(LogicalOperatorConcept<JoinLogicalOperator>);
 }
+
+namespace NES::detail
+{
+struct ReflectedJoinLogicalOperator
```
ls-1801: The old serialization included `windowMetaData` (`windowStartFieldName`, `windowEndFieldName`) but the new `ReflectedJoinLogicalOperator` doesn't. Is this data intentionally dropped, or should it be included to avoid losing it on a round-trip serialize/deserialize?

nes-query-optimizer/src/Traits/MemoryLayoutTypeTrait.cpp
```diff
@@ -82,16 +71,17 @@ std::string_view MemoryLayoutTypeTrait::getName() const
...
+}
+
+Reflected Reflector<MemoryLayoutTypeTrait>::operator()(const MemoryLayoutTypeTrait& trait) const
+{
+    return reflect(detail::ReflectedMemoryLayoutTypeTrait{trait.memoryLayout});
+}
+
+MemoryLayoutTypeTrait Unreflector<MemoryLayoutTypeTrait>::operator()(const Reflected& reflected) const
+{
+    auto [memoryLayout] = unreflect<MemoryLayoutTypeTrait>(reflected);
```
ls-1801: Bug: `unreflect<MemoryLayoutTypeTrait>` should be `unreflect<detail::ReflectedMemoryLayoutTypeTrait>`. The structured binding expects the reflected detail struct, not the trait itself. Compare with `ImplementationTypeTrait.cpp` and `OutputOriginIdsTrait.cpp` which correctly use their detail types.
```suggestion
    auto [memoryLayout] = unreflect<detail::ReflectedMemoryLayoutTypeTrait>(reflected);
```

nes-logical-operators/src/Operators/Sinks/SinkLogicalOperator.cpp
```diff
@@ -191,22 +191,18 @@ SinkLogicalOperator SinkLogicalOperator::withSinkDescriptor(SinkDescriptor sinkD
...
+{
+    auto [descriptor, name] = unreflect<detail::ReflectedSinkLogicalOperator>(reflected);
+    if (descriptor.has_value())
     {
-        serializableOperator.add_children_ids(child.getId().getRawValue());
+        return SinkLogicalOperator{descriptor.value()};
     }
-
-    serializableOperator.mutable_sink()->CopyFrom(proto);
+    return SinkLogicalOperator{name};
```
ls-1801: When a `sinkDescriptor` is present, the `sinkName` is ignored during deserialization. Is this intentional? The name is serialized (line 196) but silently dropped on the round-trip if there's a descriptor.

nes-logical-operators/include/Functions/ArithmeticalFunctions/FloorLogicalFunction.hpp
```diff
@@ -50,11 +50,33 @@ class FloorLogicalFunction final
...
 };
 
 static_assert(LogicalFunctionConcept<FloorLogicalFunction>);
 
 
 }
 
+namespace NES::detail
+{
+struct ReflectedFlootLogicalFunction
```
ls-1801: Typo: `ReflectedFlootLogicalFunction` should be `ReflectedFloorLogicalFunction`. Appears in 3 places (header line 76, cpp line 96, cpp line 101).
```suggestion
struct ReflectedFloorLogicalFunction
```

nes-logical-operators/src/Operators/Windows/WindowedAggregationLogicalOperator.cpp
```diff
@@ -257,163 +257,52 @@ const WindowMetaData& WindowedAggregationLogicalOperator::getWindowMetaData() co
...
+        return unreflect<WindowedAggregationLogicalOperator>(arguments.reflected);
     }
 
-    auto logicalOperator = WindowedAggregationLogicalOperator(keys, windowAggregations, windowType);
-    if (arguments.inputSchemas.empty())
-    {
-        throw CannotDeserialize("Cannot construct WindowedAggregation");
-    }
-    return logicalOperator.withInferredSchema(arguments.inputSchemas);
+    PRECONDITION(false, "Operator is only build directly via parser or via reflection, not using the registry");
```
adrianmichalke: I dont fully understand the precondtion message. Dont we just not provide the correct arguments we reach this code path?

Also, it this error because of user input (exception) or because we have a bug (INVARIANT, PRECONDITION)?

nes-logical-operators/src/Functions/CastToTypeLogicalFunction.cpp
```diff
@@ -96,7 +89,20 @@ LogicalFunctionGeneratedRegistrar::RegisterCastToTypeLogicalFunction(LogicalFunc
...
+Reflected Reflector<CastToTypeLogicalFunction>::operator()(const CastToTypeLogicalFunction&) const
+{
+    PRECONDITION(false, "CastToTypeLogicalFunction not expected to be reflected");
+    throw CannotSerialize("CastToTypeLogicalFunction not expected to be reflected");
+}
+
+CastToTypeLogicalFunction Unreflector<CastToTypeLogicalFunction>::operator()(const Reflected&) const
+{
+    PRECONDITION(false, "CastToTypeLogicalFunction not expected to be unreflected");
+    throw CannotDeserialize("CastToTypeLogicalFunction not expected to be unreflected");
```
adrianmichalke:
```suggestion
CastToTypeLogicalFunction Unreflector<CastToTypeLogicalFunction>::operator()(const Reflected&) const
{
    PRECONDITION(false, "CastToTypeLogicalFunction not expected to be unreflected");
    std::unreachable;
```

nes-common/include/Util/Reflection.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+template <typename T, typename ReflType>
+requires requires(T data, ReflType rfl) {
+    { Reflector<T>{}(data) } -> std::same_as<ReflType>;
+    { rfl::to_generic(rfl) } -> std::same_as<rfl::Generic>;
+}
+Reflected reflect(const T& data)
+{
+    return Reflected{rfl::to_generic(Reflector<T>{}(data))};
+}
```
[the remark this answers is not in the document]
greenBene: Nice catch, thank you!

nes-logical-operators/include/Serialization/LogicalFunctionReflection.hpp
```diff
@@ -0,0 +1,82 @@
...
+    TypedLogicalFunction<> operator()(const Reflected& rfl) const
+    {
+        auto [name, data] = unreflect<detail::ReflectedLogicalFunction>(rfl);
+
+        LogicalFunctionRegistryArguments argument;
+        argument.reflected = data;
+        auto logicalFunction = LogicalFunctionRegistry::instance().create(name, argument);
+        if (!logicalFunction.has_value())
+        {
+            throw CannotDeserialize("Failed to unreflect logical function");
```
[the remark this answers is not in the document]
greenBene: Added

nes-logical-operators/src/Operators/Sinks/SinkLogicalOperator.cpp
greenBene: I choose this implementation based on the available constructors: If the `SinkLogicalOperator` is created using a `SinkDescriptor`, then the following constructor (`SinkLogicalOperator.cpp:44ff`) is used: 

```cpp
SinkLogicalOperator::SinkLogicalOperator(SinkDescriptor sinkDescriptor)
    : sinkName(sinkDescriptor.getSinkName()), sinkDescriptor(std::move(sinkDescriptor))
{
}
```

Since the `SinkDescriptor` also encodes the name of the sink, the sinkName will be set. 
If on the other hand a `SinkLogicalOperator` is created using the name, then the Descriptor will be set in the LegacyOptimizer within the function `SinkBindingRule::apply(LogicalPlan&)` based on the given name. 

The name of the sink encoded in the `SinkDescriptor` and the name of the sink encoded in the `SinkLogicalOperator` must thus always be the same. 

To ensure that this behavior during the deserialization, I added a validation step that throws a `CannotDeserialize` if this happens

nes-logical-operators/include/Functions/ArithmeticalFunctions/FloorLogicalFunction.hpp
greenBene: Good catch! Thanks!

nes-logical-operators/src/Functions/CastToTypeLogicalFunction.cpp
```diff
@@ -96,7 +89,20 @@ LogicalFunctionGeneratedRegistrar::RegisterCastToTypeLogicalFunction(LogicalFunc
...
-    return CastToTypeLogicalFunction(arguments.dataType, arguments.children[0]);
+    PRECONDITION(false, "Operator is only build directly via parser or via reflection, not using the registry");
+    std::unreachable();
+}
+
+Reflected Reflector<CastToTypeLogicalFunction>::operator()(const CastToTypeLogicalFunction&) const
+{
+    PRECONDITION(false, "CastToTypeLogicalFunction not expected to be reflected");
+    throw CannotSerialize("CastToTypeLogicalFunction not expected to be reflected");
+}
```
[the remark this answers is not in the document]
greenBene: Good catch, I missed this :) Thank you!

nes-logical-operators/src/Functions/CastToTypeLogicalFunction.cpp
greenBene: Thanks!

PR #1367, opened 2026-02-12
feat(ChainedHashMap): Added support to var sized in ChainedHashMapTest

keyseven123:
This PR adds support for `VARSIZED` in `ChainedHashmapTest` so that we test our chained hash map also with variable-sized data.

## Verifying this change

### DIFF
--- a/nes-nautilus/tests/TestUtils/include/NautilusTestUtils.hpp
+++ b/nes-nautilus/tests/TestUtils/include/NautilusTestUtils.hpp
@@ -23,11 +23,13 @@
 #include <string_view>
 #include <utility>
 #include <vector>
+
 #include <DataTypes/DataType.hpp>
 #include <DataTypes/Schema.hpp>
 #include <Nautilus/Interface/BufferRef/LowerSchemaProvider.hpp>
 #include <Nautilus/Interface/BufferRef/TupleBufferRef.hpp>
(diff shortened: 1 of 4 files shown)


nes-nautilus/tests/TestUtils/include/NautilusTestUtils.hpp
```diff
@@ -61,10 +64,37 @@ struct RecordWithFields
...
     Record record;
     std::vector<Record::RecordFieldIdentifier> fields;
 };
 
+struct RecordWithFieldsHash
+{
+    std::size_t operator()(const RecordWithFields& record) const
+    {
+        uint64_t hashValue = 0;
+        const std::unique_ptr<HashFunction> hashFunction = std::make_unique<MurMur3HashFunction>();
```
ls-1801: why is this on the heap?

ls-1801 (approved):

nes-nautilus/tests/TestUtils/include/NautilusTestUtils.hpp
keyseven123: Because `MurMur3HashFunction` does not provide the `calculate()` method we need. 
Not sure if the following is better:
```suggestion
        MurMur3HashFunction murmur3Hash;
        HashFunction& hashFunction = murmur3Hash;
```

Issue #1378, opened 2026-02-14
[TASK] Improve robustness of PhysicalPlanBuilder DAG flip algorithm


zeuchste:
## Description

The `PhysicalPlanBuilder::flip()` method in `nes-query-optimizer/src/PhysicalPlanBuilder.cpp` (lines ~61-127) reverses the direction of a physical plan DAG from sink-to-source to source-to-sink. The algorithm has several robustness gaps:

### Issues

**1. No cycle detection**
- The algorithm assumes input is a DAG but never validates this
- If the input contains a cycle (due to a bug in a rewrite rule), the recursive traversal could infinite-loop or silently produce a corrupt plan

**2. Incomplete post-flip verification**
- Only checks that new roots are `SourcePhysicalOperator` (lines ~120-125)
- Does not verify that edge count is preserved, that all nodes are reachable from new roots, or that the graph remains a DAG after flipping

**3. Raw pointer in visited set**
- `visited` set stores raw pointers extracted from `shared_ptr` objects
- If ownership changes during traversal (e.g., a node goes out of scope), the set contains dangling pointers

### Suggested Improvements
- Add a cycle detection check before flipping (e.g., DFS with coloring)
- Add post-flip assertions: edge count preserved, all nodes reachable, new leaves are sinks
- Use `shared_ptr` in the visited set instead of raw pointers
- Add unit tests for various DAG shapes (linear chain, diamond, multi-source fan-in)

PR #1382, opened 2026-02-16
Fix(#1378) QueryOptimizer: harden PhysicalPlanBuilder::flip()

zeuchste:
## Summary
- Add DFS coloring (White/Gray/Black) cycle detection to `PhysicalPlanBuilder::flip()` to prevent infinite loops on buggy input
- Replace raw pointer keys (`PhysicalOperatorWrapper*`) with `shared_ptr`-based keys using custom hasher/equality to eliminate dangling pointer risk
- Add post-flip `INVARIANT` assertions: edge count preservation, full node reachability from new roots, and sink-leaf verification
- Add 4 unit tests: LinearChainFlip, DiamondShapeFlip, MultiOperatorChainFlip, EdgeCountPreserved

## Test plan
- [x] All 4 new `PhysicalPlanBuilderFlipTest` tests pass in Docker
- [x] All 10 existing optimizer tests (`DecideJoinTypesTest`, `DecideMemoryLayoutTest`) pass — no regressions
- [x] CI full matrix (format check, sanitizers, clang-tidy)

### DIFF
--- a/nes-query-optimizer/src/PhysicalPlanBuilder.cpp
+++ b/nes-query-optimizer/src/PhysicalPlanBuilder.cpp
@@ -151,15 +151,10 @@ PhysicalPlanBuilder::Roots PhysicalPlanBuilder::flip(const Roots& rootOperators)
 
     auto collectNodes = [&visitStateMap, &allNodes](const PhysicalOpPtr& node, auto&& self) -> void
     {
-        if (!node)
-        {
-            return;
-        }
         auto [it, inserted] = visitStateMap.try_emplace(node, VisitState::Unvisited);
-        if (it->second == VisitState::InProgress)
-        {
-            INVARIANT(false, "Cycle detected in physical plan DAG during flip — node visited twice on the same DFS path");
-        }
+        PRECONDITION(
+            it->second != VisitState::InProgress,
+            "Cycle detected in physical plan DAG during flip — node visited twice on the same DFS path");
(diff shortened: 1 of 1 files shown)


adrianmichalke (approved):

nes-query-optimizer/src/PhysicalPlanBuilder.cpp
```diff
@@ -58,71 +61,163 @@ PhysicalPlan PhysicalPlanBuilder::finalize() &&
...
+    size_t operator()(const PhysicalOpPtr& ptr) const { return std::hash<PhysicalOperatorWrapper*>{}(ptr.get()); }
+};
+
+/// Custom equality for shared_ptr keys — compares underlying raw pointer identity.
+struct SharedPtrEqual
+{
+    bool operator()(const PhysicalOpPtr& lhs, const PhysicalOpPtr& rhs) const { return lhs.get() == rhs.get(); }
+};
+
+#ifndef NO_ASSERT
```
Artraxon: This doesn't need to behind an ifdef, it just gets compiled out anyway. 
Would also make it possible to use it as an utility function e.g. in tests.
And ifdefs make code really hard to read in general, so we avoid them (or hide them) whenever possible.

Artraxon (changes requested):
Sorry, just a minor thing, otherwise looks good to me

ls-1801:
@Artraxon you can also just squash and merge

PR #1383, opened 2026-02-17
Adds VTune Hooks via ITT to provide Analysis Insights

ls-1801:
Adds Tasks which allows us to track specific regions of the execution (mostly explaining why execution would stall).
Tasks which where added:

WorkerThread executing work task
WorkerThread waiting for a new task

Sources waiting for backpressue
Sources waiting for inflight buffers to be processed
Sources waiting for the admission queue to accept buffer

BufferManager blocking for Buffers

The overhead is signifcantly less than using the GoogleEventTrace. There is also no scenario where events could be dropped.

## Verifying this change
Running a performance analysis with VTune will show task segments in the Timeline. (Ill post a picture soon)

- ExecutionEngine
- Sources
- Memory

## Documentation


nes-query-engine/QueryEngine.cpp
```diff
@@ -60,6 +61,12 @@ namespace NES
 
 namespace
 {
+__itt_domain* taskExecutionDomain = __itt_domain_create("engine.task");
+__itt_string_handle* executeTask = __itt_string_handle_create("Execute Task");
+__itt_string_handle* pipeline_id = __itt_string_handle_create("Pipeline Id");
+__itt_string_handle* local_query_id_low = __itt_string_handle_create("Local-QueryId (low)");
+__itt_string_handle* local_query_id_high = __itt_string_handle_create("Local-QueryId (high)");
+__itt_string_handle* global_query_id = __itt_string_handle_create("Global Query Id");
```
zeuchste: I think they are not used somewhere

nes-query-engine/TaskQueue.hpp
```diff
@@ -70,17 +73,33 @@ class TaskQueue
...
+
+        /// blocking write
+        __itt_task_begin(taskQueueDomain, __itt_null, __itt_null, blockingWrite);
         while (!stoken.stop_requested())
         {
             /// The order of operation upholds the invariant
             /// NOLINTNEXTLINE(bugprone-use-after-move) no move happens if the write does not succeed. If a move happens, we return.
             if (admission.tryWriteUntil(std::chrono::steady_clock::now() + StopTokenCheckInterval, std::forward<T>(task)))
             {
+                __itt_task_end(taskQueueDomain);
```
zeuchste: should this not be after the release?

zeuchste (commented):
LGTM only the names should be more consistent and we should maybe add more comments what is measured here

nes-common/include/Thread.hpp
```diff
@@ -114,17 +114,28 @@ class Thread
...
+        if (WorkerNodeId.getRawValue() != WorkerId::INVALID)
+        {
+            setThreadName(ThreadName, WorkerNodeId.getRawValue());
+        }
     }
 
     ///Sets the currents thread's name.
     ///threadName has to be non-empty and will be truncated to PTHREAD_NAME_LENGTH character
-    static void setThreadName(const std::string_view threadName)
+    static void setThreadName(const std::string_view threadName, const std::string_view worker = "")
```
keyseven123: If the worker is not set, maybe local?

keyseven123 (changes requested):
Maybe it would be good to also add a markdown on how to use them in vtune.

adrianmichalke (changes requested):
Cool change! I agree with @keyseven123 that a usage guide would be cool. I would like to try it myself.

There seems to be a nix package as well: `intel-oneapi-vtune`. Not sure if we would like to load it always though

PR #1386, opened 2026-02-17
Handle malformed string values

### LINKED ISSUE #500: [TASK] Source Parser: Handle malformed string values
### Description

Currently, we throw an error if we receive a string value, e.g., "3_14" that we cannot format (currently, we only support decimal -points-, so only "3.14" is allowed).
Malformed string values are the norm. Thus, we should not throw if we cannot parse a single value.
However, since we do not support NULL/Undefined values right now, there is no clean way to continue without throwing.
We could log an error and discard the value. However, that might go unnoticed by the user and lead the unexpected and difficult to find/debug errors. In the case of floats, use [NaN-Floats](https://en.wikipedia.org/wiki/NaN). However, we cannot handle NaN-floats in our other operators, e.g., casting a NaN-float to an int.

The only clean solution seems to be to produce a NULL/Undefined value.

### Acceptance Criteria

_No response_

keyseven123:
This PR adds support for malformed strings. If the value cannot be parsed and the field is nullable, we return null. Otherwise, we keep throwing an exception.

## Verifying this change
This change is tested by `MalformedStrings.test`.

- ExecutionEngine

This PR closes #500 .


nes-common/include/Util/Strings.hpp
```diff
@@ -82,11 +82,14 @@ T from_chars_with_exception(std::string_view input)
...
-    if (ec == std::errc())
+    auto base = 10;
+    if (input.size() > 2 and input[0] == '0' and (input[1] == 'x' or input[1] == 'X'))
     {
-        return value;
+        /// If the string starts with 0x, we need to increment the input and set the base to 16
+        base = 16;
+        input = input.substr(2);
     }
+    auto [ptr, ec] = std::from_chars<T>(input.data(), input.data() + input.size(), value, base);
```
greenBene: Please use more speaking names. I think `ptr` is okay, but `ec` is not as commonly used.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONFIF.cpp
```diff
@@ -90,21 +94,30 @@ VarVal SIMDJSONFIF::parseStringIntoNautilusRecord(
             const auto& fieldName = metaData->getFieldNameInJsonAt(fieldIndex);
 
             /// Get the value from the document and convert it to a span of bytes
-            const std::string_view value = accessSIMDJsonFieldOrThrow(currentDoc, fieldName);
-            auto valueBytes = std::as_bytes(std::span(value));
+            const std::optional<std::string_view> value = accessSIMDJsonFieldOrThrow(currentDoc, fieldName, isNullable);
+            if (not value.has_value())
+            {
+                /// If we can not access/parse the field, we return the same as for a null value
```
greenBene: Comment is slightly misleading as we just return the same as for a null value if we cannot access/parse the field _and it is nullable_. While it is kinda implied, I'd prefer to mention it directly.

nes-common/include/Util/Strings.hpp
keyseven123: Done

ls-1801 (changes requested):

nes-common/src/Util/Strings.cpp
```diff
@@ -36,7 +37,15 @@ std::optional<float> from_chars<float>(const std::string_view input)
...
+        /// Clearing errno before calling stod()
+        std::size_t pos = 0;
+        errno = 0;
+        const auto value = std::stof(str, &pos);
+        if (pos != str.size() or errno == ERANGE)
+        {
+            return {};
+        }
+        return value;
     }
```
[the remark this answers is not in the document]
keyseven123: Yes, only `std::strtof` sets `errno`. Thought `std::stof` did the same.

nes-systests/systest/src/SystestResultCheck.cpp
```diff
@@ -217,8 +217,34 @@ FMT_OSTREAM(::ResultErrorString);
...
+        return false;
+    }
+    const auto boolInt = std::stoi(str);
+    return static_cast<bool>(boolInt);
+}
+
 bool compareStringAsTypeWithError(const NES::DataType::Type type, const ExpectedResultField& left, const ActualResultField& right)
 {
+    /// If both string values are NULL, they are equal
+    if (NES::toLowerCase(left.getRawValue()) == "null" and NES::toLowerCase(right.getRawValue()) == "null")
```
ls-1801: `toLowerCase` is called up to 4 times each on `left.getRawValue()` and `right.getRawValue()` across lines 238-243. Cache these in local variables to avoid redundant string allocations.

nes-systests/formatter/MalformedStrings.test
```diff
@@ -0,0 +1,176 @@
...
+7,3.14,false,"valid"
+8,NULL,true,"malformed_double_dot"
+9,NULL,true,"malformed_multiple_dots"
+10,3.14,false,"valid"
+11,NULL,true,"double_overflow"
+12,NULL,true,"double_underflow"
+13,NULL,true,"mixed_alphanumeric"
+14,NULL,true,"space_between_digits"
+15,1.234,false,"dot_separator"
+16,0x1F,false,"hexadecimal"
```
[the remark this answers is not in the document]
keyseven123: Done

nes-systests/systest/src/SystestResultCheck.cpp
```diff
@@ -217,8 +217,34 @@ FMT_OSTREAM(::ResultErrorString);
...
+{
+    if (NES::toLowerCase(str) == "true")
+    {
+        return true;
+    }
+    if (NES::toLowerCase(str) == "false")
+    {
+        return false;
+    }
+    const auto boolInt = std::stoi(str);
```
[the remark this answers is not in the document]
keyseven123: Done

PR #1389, opened 2026-02-19
Track and report total compilation time in systest

adrianmichalke:
Systest now reports after a run:
`Total execution time: 2355 ms (2 seconds) [Compilation(sum): 3079 ms (30.1%), Runtime(sum): 7161 ms]`

Compilation(sum) and Runtime(sum) are normalized runtimes as single-threaded execution

### DIFF
--- a/grpc/SingleNodeWorkerRPCService.proto
+++ b/grpc/SingleNodeWorkerRPCService.proto
@@ -47,9 +47,10 @@ message StopQueryRequest {
 enum QueryState {
     Registered = 0;
     Started = 1;
-    Running = 2;
-    Stopped = 3;
-    Failed = 4;
+    Compiling = 2;
+    Running = 3;
+    Stopped = 4;
+    Failed = 5;
 }
 
 message Error {
@@ -65,9 +66,10 @@ message QueryStatusRequest {
 
 message QueryMetrics {
    optional uint64 startUnixTimeInMs = 1;
-   optional uint64 runningUnixTimeInMs = 2;
-   optional uint64 stopUnixTimeInMs = 3;
-   optional Error error = 4;
+   optional uint64 compilingUnixTimeInMs = 2;
+   optional uint64 runningUnixTimeInMs = 3;
+   optional uint64 stopUnixTimeInMs = 4;
+   optional Error error = 5;
 }
 
 message QueryStatusReply {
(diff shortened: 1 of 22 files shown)


ls-1801 (changes requested):
Please consider just using the setup time (which is almost exposed). I had an implementation here: https://github.com/nebulastream/nebulastream/commit/38b8b2ee7ee0cd2178179d477f3296c4c174831f#diff-fb858e4f7df7f4f73126b33f6e08b6493e95e5a81fb1605f3f2e48e94bf0da95

I dont know about giving the Systest tool access to a static in the CompiledExecutablePipelineStage, that just breaks all kind of encapsulation, and also wouldnt work for a remote test. 

In general I would want to avoid any kind of tracing/measurement scattered every where in the code. It just makes it so much harder to comprehend.

If you really want to measure the exact compilation time, than please find another way than having a global which is accessed from systest. You should use some event which can be exposed via the query log.

keyseven123 (changes requested):
I think that it is generally better to use the query engine listener for this. For example, we could add a new state `COMPILING` and then get the start and end time via the query engine listener. As I am not a fan of the global/adding  new interface calls to the executable stage

Artraxon (changes requested):
I agree with what the others say, this just raises so many red flags. How would you even know which query the statistics belong to?

ls-1801 (changes requested):
In its current iteration, the compilation time would be virtually equal to the setup time. 
You just emit both events right after each other.

I would suggest just using the setup time to display in systest (currently, no operator is really doing anything in the setup besides compiling). 

I would rely on some performance tooling (like vtune) to trace the actual compilation time very precisely, you can have a look at the (arguably stale) vtune pr

keyseven123 (approved):
Just one nit-pick.

ls-1801 (changes requested):
I think we had a discussion about this.

My suggestion was to create a QueryMetrics (QueryStats, name is up for debate) object, which is handed to every (source)/pipeline on destruction, so every source or pipeline has the chance to save some stats. The query metric object will then be available via the query log api.

Re-Request once relevant again

Artraxon (changes requested):
Rerequest once relevant again

PR #1400, opened 2026-02-23
Config file for disabling systest files and groups by default

adrianmichalke:
Adds a config file for disabling systest files and groups by default.

By default we disable now the large scale tests. Current config:
```
# Permanent exclusions for systest startup.
# Examples:
# disabled_test_files:
#   - <relative/path/to/test>.test
# exclude_groups:
#   - <group name>
#
# - large scale tests are always skipped below
exclude_groups:
  - large
```


keyseven123 (approved):

alepping (changes requested):
I'm unsure about the current state of these changes.
If it is just about disabling large tests by default, there would be a way simpler solution that requires far less code.

In general, I like the idea of a global config file.
However, the interactions with i) command line args ii) in-systest-file config are non clear to me and in general non-trivial. Which has priority? What if one excludes a group and another includes a group, etc.

PR #1404, opened 2026-02-25
Makes parse function optimizable for LLVM

alepping:
Adds nomodref, willReturn, and noUnwind attribute to parse function which allows LLVM to optimize out unused parse functions and to move it around (essentially treats it as pure function).

Prior to this change, LLVM treated the parse call as conservative as possible, meaning that even if the return value of the parse call was not used at all in the code (dead code elimination), it would keep the parse calls, because of possible side-effects.

### DIFF
--- a/nes-input-formatters/include/InputFormatters/RawValueParser.hpp
+++ b/nes-input-formatters/include/InputFormatters/RawValueParser.hpp
@@ -42,6 +42,7 @@ template <typename T>
 nautilus::val<T> parseIntoNautilusRecord(const nautilus::val<int8_t*>& fieldAddress, const nautilus::val<uint64_t>& fieldSize)
 {
     return nautilus::invoke(
+        nautilus::FunctionAttributes{nautilus::ModRefInfo::NoModRef, true, true},
         +[](const char* fieldAddress, const uint64_t fieldSize)
         {
             const auto fieldView = std::string_view(fieldAddress, fieldSize);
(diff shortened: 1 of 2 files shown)


ls-1801 (changes requested):
Waiting until the CI Passes, i am curious what needs to be addressed

Artraxon (changes requested):
please rerequest when CI is green, but I don't have much to contribute here in terms of review.

I will also take your word, but do you happen to have some benchmark numbers?

nes-input-formatters/include/InputFormatters/RawValueParser.hpp
keyseven123: It is just for my understanding but should this not be `Ref`? We are reading from the memory address field address or what counts as a ref/read?

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONFIF.hpp
```diff
@@ -158,6 +159,7 @@ class SIMDJSONFIF final : public FieldIndexFunction<SIMDJSONFIF>
         nautilus::val<const SIMDJSONMetaData*> metaData) const
     {
         return nautilus::invoke(
+            nautilus::FunctionAttributes{.modRefInfo = nautilus::ModRefInfo::Ref, .willReturn = true, .noUnwind = true},
```
ls-1801: `ModRefInfo::Ref` may be incorrect here. simdjson's on-demand API is a forward-only streaming parser — accessing `currentDoc[fieldName]` advances internal parser state, which is a mutation. Marking this as read-only could allow LLVM to reorder or eliminate calls, corrupting parser state. Please verify that field access through the on-demand API truly has no side effects, or use `ModRef` instead.

ls-1801 (changes requested):

Artraxon (changes requested):
please rerequest when lukas points where addressed

PR #1412, opened 2026-02-26
feat(Serialization): ReflectionContext in Unreflector

Artraxon:
Currently, an Unreflector needs to be able to unreflect just given the reflected data.
This introduces a problem, since our IR does not exist in a vacuum, but is only valid in a given runtime context.
Currently, we are simply ignoring the problem by creating an invalid IR and validating/binding it to the context later on.

Since a goal of the ongoing schema inference PR is to permit only valid IR as much as possible, this creates a problem I previously solved in my branch by including the context in the registry factory functions.

I could just opt not use our reflection mechanism, but that doesn't seem like a good solution, especially since we are going to run into this problem more often in the future.

This PR solves this issue by allowing an Unreflector to receive some configuration object from a `ReflectionContext`.
The only change in the existing code is that all unreflectors receive a `ReflectionContext` cref as a second argument and must make any further unreflect call on that context object.

See the `ReflectionTest` in the commons unit test folder for an example.

The biggest challenge with this PR is that calling any cppreflect unreflection loses context, since we cannot pass the context to the cppreflect unreflectors and then back to ours. 
To remedy this, I had to replace the cppreflect reflection and unreflection with our own calls, which only use cppreflect for the actual reflection, i.e., finding the fields, but then recursively call our own reflectors/unreflectors with the context.

There are two positive side effects of this PR:
1. I was able to get rid of the `std::optional<LogicalFunction>` hack. I still don't understand why we needed it in the first place, but our own, simpler callchain doesn't do anything unexpected and arguably follows simpler logic than reflectcpp.
2. We don't have to include the full reflection header with reflectcpp everywhere; in a lot of places, a forward header is enough now. Note that we still need the full reflection header for logical functions and operators since the concept definition depends on it. 
 

## Verifying this change
This change is tested by a new unit test and the system test suite.

nes-common


nes-common/include/Util/Reflection/PairReflection.hpp
```diff
@@ -0,0 +1,62 @@
...
+                using ValueType = std::decay_t<decltype(value)>;
+                if constexpr (std::is_same_v<ValueType, rfl::Generic::Array>)
+                {
+                    if (value.size() != 2)
+                    {
+                        throw CannotDeserialize("Expected array of size 2 for std::pair");
+                    }
+                    return std::make_pair(context.unreflect<T1>(Reflected{value[0]}), context.unreflect<T2>(Reflected{value[1]}));
+                }
+                throw CannotDeserialize("Expected array for std::pair");
```
adrianmichalke: Can we print here something more meaningful from `data` to have more context?

adrianmichalke (changes requested):
LGTM, but it would be good if we could add more context to the exception messages. Otherwise debugging malfromatted reflected data will get unnecessary hard.

nes-common/include/Util/Reflection/UnorderedMapReflection.hpp
```diff
@@ -0,0 +1,108 @@
...
+                        K key;
+                        if constexpr (std::is_same_v<K, std::string>)
+                        {
+                            key = keyStr;
+                        }
+                        else if constexpr (std::is_arithmetic_v<K>)
+                        {
+                            if constexpr (std::is_integral_v<K>)
+                            {
+                                key = static_cast<K>(std::stoull(keyStr));
```
yschroeder97: Use `std::stoll` + `static_cast` for signed integer key types, keep `std::stoull` for
  unsigned.

nes-common/include/Util/Reflection/ReflectionCore.hpp
```diff
@@ -0,0 +1,471 @@
...
+        }
+    }
+
+    template <typename T>
+    T unreflect_null() const
+    {
+        if constexpr (requires {
+                          T{};
+                          std::is_default_constructible_v<T>;
+                      })
```
yschroeder97:
```suggestion
        if constexpr (std::is_default_constructible_v<T>)
```

yschroeder97 (changes requested):
Nice and appreciated changes. 

You could try to be more lean in the PR description, giving short examples of the before-and-after of the changes. 
It was a bit tedious to read it.

nes-common/include/Util/Reflection/ReflectionCore.hpp
```diff
@@ -0,0 +1,471 @@
...
+                    return unreflect_primitive<T>(value);
+                }
+                /// null - handle optional types
+                else if constexpr (std::is_same_v<ValueType, std::nullopt_t>)
+                {
+                    return unreflect_null<T>();
+                }
+                else
+                {
+                    throw CannotDeserialize("Unexpected variant type for {}, received", NAMEOF_TYPE(T), NAMEOF_TYPE(ValueType));
```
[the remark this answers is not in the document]
Artraxon: done

nes-common/include/Util/Reflection/ReflectionCore.hpp
Artraxon: default-constructible = T() not T{}, and might do automatic narrowing.

But I was able to remove std::is_default_constructible since T{} implies it.

nes-common/include/Util/Reflection/UnorderedMapReflection.hpp
Artraxon: good catch, done

nes-common/include/Util/Reflection/ReflectionCore.hpp
```diff
@@ -0,0 +1,471 @@
...
+public:
+    explicit ReflectionContext() = default;
+
+    template <typename... Args>
+    explicit ReflectionContext(Args... args) : context{(std::pair{std::type_index{typeid(Args)}, std::any{std::move(args)}}, ...)}
+    {
+    }
+
+    template <typename T>
+    [[nodiscard]] T unreflect(const Reflected& data) const;
```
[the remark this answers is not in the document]
Artraxon: i am surprised i was able to, i though i needed a forward declaration, done

nes-common/tests/UnitTests/Util/ReflectionTest.cpp
```diff
@@ -0,0 +1,189 @@
...
+    std::unordered_map<std::type_index, std::any> contextMap;
+    contextMap[std::type_index(typeid(InnerData))] = config;
+    const ReflectionContext context{contextMap};
+
+    auto deserialized = context.unreflect<std::optional<InnerData>>(reflected);
+
+    EXPECT_FALSE(deserialized.has_value());
+}
+
+///NOLINTEND(bugprone-unchecked-optional-access
```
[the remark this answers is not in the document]
Artraxon: good catch, done

nes-common/tests/UnitTests/Util/ReflectionTest.cpp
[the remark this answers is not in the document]
Artraxon: done

nes-common/tests/UnitTests/Util/ReflectionTest.cpp
```diff
@@ -0,0 +1,189 @@
...
+TEST(ReflectionTest, OptionalEmpty)
+{
+    const std::optional<InnerData> opt = std::nullopt;
+
+    auto reflected = reflect(opt);
+
+    const InnerDataDeserializationConfig config{.includeOptionalField = true};
+    std::unordered_map<std::type_index, std::any> contextMap;
+    contextMap[std::type_index(typeid(InnerData))] = config;
+    const ReflectionContext context{contextMap};
```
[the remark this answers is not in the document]
Artraxon: done

nes-common/tests/UnitTests/Util/ReflectionTest.cpp
```diff
@@ -0,0 +1,254 @@
...
+    EXPECT_EQ(deserialized->optionalField.value(), 42);
+}
+
+TEST(ReflectionTest, OptionalEmpty)
+{
+    const std::optional<InnerData> opt = std::nullopt;
+
+    auto reflected = reflect(opt);
+
+    const InnerDataDeserializationConfig config{.includeOptionalField = true};
```
yschroeder97:
```suggestion
```

yschroeder97 (approved):
Please do address the one remark on an unused varaible.

nes-common/include/Util/Reflection.hpp
```diff
@@ -13,151 +13,74 @@
...
 
-template <>
-struct Unreflector<Reflected>
-{
-    Reflected operator()(const Reflected& field) const { return field; }
-};
+    /// Reflect each field and add to the object
+    [&]<size_t... Is>(std::index_sequence<Is...>)
+    {
+        ((obj[reflect_field_at_index<T, Is>(data).first] = *reflect_field_at_index<T, Is>(data).second), ...);
```
adrianmichalke, line 79:
```suggestion
auto reflected = reflect_field_at_index<T, Is>(data);
        ((obj[reflected.first] = *reflected.second), ...);
```

nes-common/include/Util/Reflection/ReflectionCore.hpp
```diff
@@ -0,0 +1,476 @@
...
+        return Reflected{rfl::Generic{static_cast<int64_t>(data)}};
+    }
+    else if constexpr (std::is_floating_point_v<T>)
+    {
+        return Reflected{rfl::Generic{static_cast<double>(data)}};
+    }
+    else if constexpr (std::is_enum_v<T>)
+    {
+        /// Serialize enums as their underlying integer value
+        return Reflected{rfl::Generic{static_cast<int64_t>(static_cast<std::underlying_type_t<T>>(data))}};
```
adrianmichalke, line 368: Will deserilaized back to enum? Or will this keep a int64?

nes-common/include/Util/Reflection/ReflectionCore.hpp
```diff
@@ -0,0 +1,476 @@
...
+Reflected reflect(const T& data)
+{
+    if constexpr (std::is_same_v<T, bool> || std::is_same_v<T, std::string>)
+    {
+        return Reflected{rfl::Generic{data}};
+    }
+    else if constexpr (std::is_integral_v<T>)
+    {
+        return Reflected{rfl::Generic{static_cast<int64_t>(data)}};
+    }
```
adrianmichalke, line 360: will this work for unsigned ints?

adrianmichalke (changes requested):
Some remaining issues in the comments. Plus, please address:
- Update the serialization.md
- Update the serialization in InlineSourceLogicalOperator
I havnt checked the follow up PR(s). If any of the comment is already addressed there just say so.

nes-common/include/Util/Reflection/ReflectionCore.hpp
Artraxon: yes, because c++ guarantees that for any uint64_t u static_cast<uint64_t>(static_cast<int64_t>(u)) == u . reflectcpp only has int64_t in its backing variant rf::Generic , but I added a test to make sure that none of the code in between breaks it by accident.

nes-common/include/Util/Reflection/ReflectionCore.hpp
Artraxon: It will serialize back to enum, see the reflection test EnumSerialization

nes-common/include/Util/Reflection.hpp
Artraxon: i tried it, but it doesn't work, because the whole expression is inside a pack expansion, and so needs to be together. But with c++26 we could replace this with constexpr for each.

adrianmichalke (changes requested):
LGTM, but have you checked my comments on:

> Update the serialization.md
Update the serialization in InlineSourceLogicalOperator

Artraxon:
@adrianmichalke sorry, forgot about those, addressed them now

adrianmichalke (approved):

PR #1414, opened 2026-02-28
chore(Format): Added to log_warn message if file ends on Test

keyseven123:
This PR adds a small change to the `log_warn` message in our `format.sh` script. If the file ends with `(test|Test|tests|Tests)`, it adds `(most probably fine, as it seems to be a unit test file)`


keyseven123:
> I would keep the wording concise: "fine for tests" just to keep the output of the format script more concise instead emitting a book.
> 
> But it's not important

Being concise does not clutter up the terminal that much.

PR #1415, opened 2026-03-01
feat(serialization): Add registry dispatch for serialization

Artraxon:
Currently, we are using the auto-declared registry function to achieve the plugin-name based runtime dispatch of the correct factory function for logical functions and traits.
This is fine the creating logical functions in the binder, but for unreflection, this is cumbersome as there are already unreflectors defined and there is just the mapping missing.
This leads to a lot of error prone boiler plate and possibly invalid state in the logical function registry functions.

This becomes even more problematic with the unreflection context, as we'd either loose the unreflection context when stepping into the registry functions or would have to pass (a reference to) the context inside the registry arguments.

## New Solution
This PR solves these issues, by allowing one to create unreflection registries with the cmake macro create_unreflection_registry.
It takes three Arguments: Name of the registry, cmake target that should include the registry, and two format strings, where that determine the return type of the registry per added type and the second one specifies the .hpp that needs to be included to resolve the type.

For example, for LogicalFunctions the call is
 ```cmake
create_unreflection_registry(LogicalFunction nes-logical-operators
        "\${PLUGIN_NAME}LogicalFunction"
        "\${PLUGIN_NAME}LogicalFunction.hpp")
```
and for LogicalOperators its
```cmake
create_unreflection_registry(LogicalOperator nes-logical-operators
        "TypedLogicalOperator<\${PLUGIN_NAME}LogicalOperator>"
        "\${PLUGIN_NAME}LogicalOperator.hpp")
```

To register a plugin, use the add_unreflection_plugin macro, which takes the registry name, the plugin name, and optionally KEY <PLUGIN_KEY>, which overwrites the lookup name used by the registrar, which otherwise is the plugin name.

IMPORTANT: The unreflection registry DOES NOT add you source files to any target, unlike the existing registries.
Also, you need to declare the cmake library before creating the unreflection registry, see the nes-logical-operators CMakeLists.txt for an example.

 
In contrast to the existing registries, this registry does not generate any registrar code, instead it only generates some glue code that registers the plugins at runtime, during static initialization, in the registry.
This glue code is bundled in a glue library for the component of the registry and linked with WHOLE_ARCHIVE, so that static registration code does not get optimized out. 
It does NOT require your plugin to be part of the component of the registry, it just needs to be able to resolve the includes.

If you require more custom plugin loading, you are free to just call the registration functions yourself in whatever way you see fit, just ensure that you handle plugin name collisions gracefully.

To use it, just call unreflect on the registry, like 
```c++
template <>
struct Unreflector<TypedLogicalFunction<>>
{
    TypedLogicalFunction<> operator()(const Reflected& rfl, const ReflectionContext& context) const
    {
        auto [name, data] = context.unreflect<detail::ReflectedLogicalFunction>(rfl);

        auto logicalFunction = LogicalFunctionRegistry::instance().unreflect(name, data, context);
        if (!logicalFunction.has_value())
        {
            throw CannotDeserialize("Failed to unreflect logical function of type {}", name);
        }
        return logicalFunction.value();
    }
};
```

## Verifying this change
All systests run.

Registry system, Logical Functions and Traits.

## Old Solution
This PR solves these issues, by allowing one to set an option for a registry (in cmake) that lets cmake generate a table for unreflectors automatically. In contrast to the normal registry factory functions, it not only generates the declaration of the functions, but also the implementations in separate .cpp files that automatically call the defined unreflectors.
Like with the existing registry functions, this approach does not require full recompilation if the header of a logical function has changed.

To use it, use `enable_unreflection_for_registries(PluginType)` before creating the registry.
Also, the registry must inherit from `BaseRegistryWithUnreflection` .

I also added an option `disable_factory_functions_for_registries` that you can use to disable the generation of declarations of the normal factory functions.


keyseven123 (commented):
I do not think that I am the right person to review this PR. I would suggest @alepping and/or @greenBene

greenBene (changes requested):
Thank you for this PR :) I think the core idea is really helpful! 

Three major points: 
1) I think you are mixing several changes in this PR that are not related to the core of the registry dispatch for serialization. Maybe sth went wrong when you cut the files from the Context PR? 
2) A lot of the logic (introduced by the files that I assume belong to the Context PR) feels quite verbose. I would question if they are necessary and if they maybe introduce a heavy maintenance burden because of their complexity. 
3) Related to this PR itself: I am not to happy about the new cmake registry interface. E.g. For traits we now need to call three distinct functions:

```cmake
enable_unreflection_for_registries(Trait)
disable_factory_functions_for_registries(Trait)

# Create all registries: RewriteRule uses normal factory functions, Trait uses unreflection-only
create_registries_for_component(RewriteRule Trait)
```

I would prefer a single function call per registry, that allows configuration. Then it would be clear on registry creation what is expected of the created registry. Sth. like the following:

```cmake
create_registries_for_component(Trait UNREFLECTION)
create_registries_for_component(RewriteRule FACTORY)
create_registries_for_component(LogicalFunction UNREFLECTION FACTORY)

```

alepping (changes requested):
@Artraxon looks really good overall

My two main gripes are the complexity of the PluginRegistrationUtil.cmake and Registry.hpp (commented both explicitly).
Creating a second registry header and simplifying the cmake logic would make this an awesome PR/contribution imo

Artraxon:
I have to update the PR description, but i completely reworked the registry after talking to @alepping

alepping (changes requested):
Personally, I'm happy with the registry split. While there is some overlap, cleanly separating the already complex registries just adds less of a cognitive burden. Note however, that there are some differences concerning requirement of capitalization and other stuff where the registries differ. I do think that streamlining them as much as possible, without merging them, is the way to go.

One real pain point that I have with this PR is the boilerplate code we dump on people who implement operators. Also, non-trivial, templated boilerplate.

I tried to create a simplified version and I'd like to hear your opinion, its on this branch:
`serialization-registry-alepping-suggestion-v1`

Note that the first commit introduces a poc of what I imagine and the second commit applies it to all operators.

nes-logical-operators/src/Functions/ArithmeticalFunctions/CMakeLists.txt
```diff
@@ -22,3 +22,16 @@ add_plugin(Pow LogicalFunction nes-logical-operators PowLogicalFunction.cpp)
...
+add_unreflection_plugin(LogicalFunction Ceil)
+add_unreflection_plugin(LogicalFunction Div)
+add_unreflection_plugin(LogicalFunction Exp)
+add_unreflection_plugin(LogicalFunction Floor)
+add_unreflection_plugin(LogicalFunction Modulo KEY Mod)
+add_unreflection_plugin(LogicalFunction Mul)
+add_unreflection_plugin(LogicalFunction Pow)
+add_unreflection_plugin(LogicalFunction Round)
+add_unreflection_plugin(LogicalFunction Sqrt)
+add_unreflection_plugin(LogicalFunction Sub)
```
greenBene, line 37: I am not too happy to have two plugin calls when registering a plugin, esp. if it is redundant for all LogicalFunction plugins. Can't we define the need for an unreflection plugin in the registry definition?

greenBene (changes requested):

nes-logical-operators/src/Functions/ArithmeticalFunctions/CMakeLists.txt
Artraxon: so, this was the result of extensive talks with Aljoscha, but no, you can't, and its also not the pattern that we apply in other parts of the code base.
We have one register or add_plugin call per plugin and registry, because it would otherwise make the registry code way too complicated. Aljoscha at some tried at some point to combine them, but it was too complicated for little benefit.

For example, for the sources you might have up to four register calls, like for example for the TCPSource in nes-plugins, (one for Source, SourceValidation, InlineData and FileData).

Putting them together sounds nice in theory, but it makes the CMake Code generation super complicated, and just mixes things that don't belong together. Also, the registration mechanism for unreflection is now very different (and much simpler then our normal registries).

Part of the problem is also that the normal register function does too much, like for example adding source files to a target which just leads us to replicate stuff that one would just write in their normal cmake files also in the unrelated code-gen parts, especially when thinking about plugins that draw in additional dependencies.
Also, because this approach now purely works through linking and runtime-registration, we don't need the additional add_plugin_as_library macro anymore, since you can add a LogicalFunction from the outside just via linking.

Artraxon:
Updated PR description with the new unreflection registry approach

alepping (approved):
Approving, given that @Artraxon has follow-up PRs that more clearly demonstrate why this change makes sense, and since @Artraxon and I talked about simplifications of the operator boilerplate code, which should significantly reduce the boilerplate introduced in this PR, and also the existing boilerplate code. That is, his plans are more high-level and probably long-term better than my suggestions on the branch I shared in another comment.

greenBene (approved):

PR #1417, opened 2026-03-02
Feature(#1416): Enable native macOS builds

### LINKED ISSUE #1416: Enable native macOS builds with Ninja
## Motivation

NebulaStream currently only builds inside a Linux Docker container. The CMake build system has hard-coded Linux-only checks that prevent native builds on macOS. This limits developer productivity for macOS users who must route every build through Docker.

### CMake build system
- Add Darwin/macOS detection in `HostIdentification.cmake` (set `NES_OS=osx`)
- Use `NES_OS` variable instead of hardcoded `linux` in vcpkg triplet names
- Guard mold linker and `--no-undefined` flag (Linux-only, not supported by macOS ld64)
- Fix RPATH: use `@loader_path` / `@executable_path` instead of `$ORIGIN` on macOS

### vcpkg triplets
- Add macOS arch variants (`x64-osx.cmake`, `arm64-osx.cmake`) with `VCPKG_CMAKE_SYSTEM_NAME Darwin`
- Add 16 macOS triplet combinations: `{x64,arm64}-osx-{none,address,thread,undefined}-{libcxx,local}`
- Make toolchain files platform-aware (`CMAKE_SYSTEM_NAME` from triplet instead of hardcoded `Linux`)
- Make `libuuid` dependency Linux-only in `vcpkg.json` (macOS provides UUID natively)

### C++ source portability fixes
- `pthread_setname_np`: different signature on macOS (1-arg vs 2-arg)
- `strerror_r`: different return type on macOS (POSIX `int` vs GNU `char*`)
- Replace Linux-only headers (`<asm-generic/socket.h>`, `<bits/types/struct_timeval.h>`) with portable `<sys/time.h>`

## Verification
```sh
# On macOS with Homebrew LLVM 19:
export LLVM_TOOLCHAIN_VERSION=19
export PATH="/opt/homebrew/opt/llvm@19/bin:$PATH"
cmake -DCMAKE_BUILD_TYPE=Debug -G Ninja -S . -B cmake-build-debug
cmake --build cmake-build-debug -j
```

zeuchste:
## Summary
- Remove hard-coded Linux assumptions from the CMake build system so NebulaStream can be configured and built natively on macOS
- Add macOS vcpkg triplets (x64/arm64, all sanitizer/stdlib combinations)
- Fix platform-specific C++ API differences (`pthread_setname_np`, `strerror_r`, Linux-only headers)

### CMake build system
| File | Change |
|------|--------|
| `cmake/HostIdentification.cmake` | Add Darwin detection, set `NES_OS=osx` |
| `cmake/ImportDependencies.cmake` | Use `${NES_OS}` instead of hardcoded `linux` in triplet names |
| `cmake/UseMold.cmake` | Guard mold linker + `--no-undefined` behind Linux-only check |
| `CMakeLists.txt` | Use `@loader_path`/`@executable_path` instead of `$ORIGIN` on macOS |
| `nes-common/CMakeLists.txt` | Skip `pkg_check_modules(uuid)` on macOS (system provides it) |

### vcpkg triplets (18 new files)
- `bits/arch/{x64,arm64}-osx.cmake` with `VCPKG_CMAKE_SYSTEM_NAME Darwin`
- 16 triplet combinations: `{x64,arm64}-osx-{none,address,thread,undefined}-{libcxx,local}`
- Toolchain files: derive `CMAKE_SYSTEM_NAME` from triplet instead of hardcoding `Linux`
- `vcpkg.json`: make `libuuid` Linux-only (macOS has native UUID support)

### C++ portability fixes
| File | Fix |
|------|-----|
| `Thread.hpp` | `pthread_setname_np` — 1-arg on macOS vs 2-arg on Linux |
| `Files.cpp` | `strerror_r` — POSIX `int` return on macOS vs GNU `char*` on Linux |
| `TCPSource.cpp` | Replace `<asm-generic/socket.h>` + `<bits/types/struct_timeval.h>` with portable `<sys/time.h>`; fix `strerror_r` calls |

## Test plan
- [x] CMake configure succeeds on macOS (arm64, Debug, Ninja) — verified with `NES_USE_SYSTEM_DEPS=ON`
- [ ] Full vcpkg-based build on macOS (requires LLVM 19 toolchain)
- [ ] Verify Linux Docker build is unaffected
- [ ] Run unit tests on macOS

Closes #1416


ls-1801 (changes requested):
1. There needs to be a CI run using macos otherwise this will break the very next commit
2. I am not sure if macos supports both libstdcxx and libcxx. I would assume that if you are using the local (i.e., not setting any stdlib) you would already get libcxx. Investigate that and maybe remove -libcxx triplets.

keyseven123 (commented):
As a non apple user, I can not test or help in the future with bugs. So I do not feel comfortable in reviewing this PR.

yschroeder97 (commented):
I can try this out tomorrow if I find time.

zeuchste (commented):
Both points raised are already addressed in this PR:

**1. macOS CI job:** The `build-and-test-macos` job is already included in `.github/workflows/smoke-tests.yml` (added in commit 291de2d4d0). It runs on `macos-14` (ARM64), installs Homebrew LLVM 19, configures with Ninja in Debug mode, builds, and runs the full test suite (excluding DockerCompose tests). It also has vcpkg binary caching enabled via GitHub Actions cache to avoid rebuilding LLVM on every run.

**2. libcxx vs local triplets on macOS:** The redundant `-libcxx` macOS triplets were already removed in commit de506deec2. The PR only includes `-local` triplets for macOS (`arm64-osx-{none,address,thread,undefined}-local` and the `x64-osx-*` equivalents).

This is correct because on macOS with Homebrew LLVM 19, the compiler defaults to libc++ -- there is no libstdc++ available. The `local.cmake` toolchain does not pass any `-stdlib=` flag, which means it relies on the compiler's default (libc++ on macOS). The `libcxx.cmake` toolchain explicitly passes `-stdlib=libc++`, which would be redundant on macOS. So having only `-local` triplets for macOS is the right approach -- they already get libc++ by default, and there is no need (or ability) to use libstdc++ on macOS.

keyseven123 (commented):
> As a non apple user, I can not test or help in the future with bugs. So I do not feel comfortable in reviewing this PR.

See above. Once I have a MacBook, I would be willing to test it and help with this

PR #1418, opened 2026-03-02
feat(LogicalOperator): Add weak self reference

Artraxon:
This is a spinoff PR of the schema inference branch that @gustafsa extracted.
One of the main concepts of the schema inference branch is that every output schema of a logical operator, consists of fields that are essentially a pair of an unqualified identifier and a pointer to the operator that has produced it.

This allows us to keep track of the origin of fields, even in binary operators like a join, without having to rely on source prefixing, which breaks easily (and does lead us to produce wrong results at times), prohibits lexical scopes and makes relation aliasing in queries next to impossible.

The problem is, that for this we need to allow an operator to get its owning `LogicalOperator` instance, so that it can attach it to the `Field` objects (or other objects that need to be _bound_).
But since the logical operator class instances are not directly instantiated in a `std::shared_ptr`, but rather wrapped in the `OperatorModel`, we cannot use `std::enable_shared_from_this` to insert a weak_this member field.
Also, the weak_this member is only available _after_ the constructor has run through, which makes this unsuitable for later on giving Traits access to the full Operator/Traitset to validate themselves in the ctor. 

We solve this, by introducing a WeakTypedLogicalOperator, add a member of it to every logical operator and use the operator model to keep the self reference inside the wrapped instance up-to-date.

The only change in the logical operator necessary is to add `SELF_REF` in the member fields, which is just a convenience macro that expands to:
```c++
template <LogicalOperatorConcept T>
friend struct NES::detail::OperatorModel;
WeakLogicalOperator self;
```

You can then access the owning logical operator with `self.lock()` or `self.tryLock()` where the first will segfault with the precondition if the self is invalid, while the try lock will return an empty optional.
In the schema inference branch it is then used like 
```c++
Schema<Field, Unordered> bind(LogicalOperator op, Schema<UnboundUnqualifiedField, Unordered> unboundSchema) {
   return unboundSchema 
        | std::views::transform([&op](const UnboundUnqualifiedField& unboundField){ return Field{op, unboundField.getName(), unboundField.getDataType()}; ) 
        | std::ranges::to<Schema<Field, Unordered>>();
}

Schema<Field, Unordered> JoinLogicalOperator::getOutputSchema() const {
    return bind(self.lock(), unboundSchema);
}
```

## Verifying this change
I added a unit test that tests if the locking works and the self is is updated correctly.

LogicalOperators

### DIFF
--- a/nes-logical-operators/tests/LogicalPlanTest.cpp
+++ b/nes-logical-operators/tests/LogicalPlanTest.cpp
@@ -13,11 +13,14 @@
 */
 
 #include <cstddef>
+#include <optional>
 #include <ranges>
 #include <sstream>
 
+#include <string>
 #include <unordered_map>
 #include <utility>
+#include <vector>
 
 #include <gmock/gmock-matchers.h>
 #include <gtest/gtest.h>
@@ -35,6 +38,8 @@
 #include <Sources/SourceCatalog.hpp>
 #include <Sources/SourceDescriptor.hpp>
 #include <Traits/Trait.hpp>
+#include <Traits/TraitSet.hpp>
+#include <Util/PlanRenderer.hpp>
 #include <Util/Reflection.hpp>
 #include <Util/UUID.hpp>
 #include <QueryId.hpp>
@@ -50,7 +55,7 @@ class LogicalPlanTest : public ::testing::Test
 {
 protected:
     LogicalPlanTest()
-        : sourceOp{SourceNameLogicalOperator("Source")}
+        : sourceOp{TypedLogicalOperator<SourceNameLogicalOperator>{"Source"}}
         , sourceOp2(
               [this]
               {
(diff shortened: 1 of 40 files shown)


nes-logical-operators/tests/LogicalPlanTest.cpp
ls-1801: On the other hand this would justify a new test file for LogicalOperatorTests?

ls-1801 (changes requested):
Maybe I misunderstood the problem, but couldn't you achieve the same by just passing `shared_ptr<ConcreteLogicalOperator> self` as an argument to the inner concrete implementation when dispatching from' TypedLogicalOperator' to the concrete implementation?

Also, why does this code need to be operator-specific anyway? Isn't the binding identical for every operator?
```c++
Schema<Field, Unordered> JoinLogicalOperator::getOutputSchema() const {
    return bind(self.lock(), unboundSchema);
}
```

 Can't you just move the binding into the TypedLogicalOperator?

```c++
    [[nodiscard]] Schema getOutputSchema() const
    {
        auto unbound = self->getOutputSchema();
        return bind(self, unbound);
    }
```

greenBene (changes requested):
Because of the heavy use of templating, I find the code a bit hard to understand. Can we simplify the code?  And do we actually require this logic change, or could we avoid it, if we allow a LogicalPlan during creation to be in a temporarily "invalid" state?

ls-1801 (approved):

greenBene (approved):

Issue #788, opened 2025-04-07
[BUG] Projection with duplicated names does not work as expected


ls-1801:
Currently if a select statement projects into an duplicated identifier the system only projects one of them.

# SQL Query Parser Issues with MEDIAN Function and Projections

## Description

We've identified several issues with how our SQL engine handles projections and aggregations, particularly with the `MEDIAN` function over tumbling windows:

1. **Duplicate column selections are silently dropped**
   * When a column is selected twice with the same name, it appears only once in the result set
   * This differs from standard SQL behavior where duplicates would be projected twice

2. **Aggregation expressions crash the parser**
   * Attempting to use an aggregated result in a calculation (e.g., `MEDIAN(i8) + 1`) causes a parser crash
   * The parser should handle expressions with aggregated values

3. **Type inference fails with duplicate projection names**
   * Using different aggregations with the same output column name fails during type inference
   * Example: `SELECT MEDIAN(i8) as i8_out, SUM(i8) as i8_out...`

### Issue 1: Duplicate column selections are dropped

The following queries produce identical results, when they should have different column structures:

```sql
-- Query 1: Single selection
SELECT MEDIAN(i8) as i8_out 
FROM stream 
WINDOW TUMBLING(ts, size 100 ms) 
INTO sinkStreamAllF64
```

```sql
-- Query 2: Duplicate selection (should show twice)
SELECT MEDIAN(i8) as i8_out, MEDIAN(i8) as i8_out
FROM stream 
WINDOW TUMBLING(ts, size 100 ms) 
INTO sinkStreamAllF64
```

### Issue 2: Aggregation expressions crash the parser

The following query crashes the parser:

```sql
SELECT MEDIAN(i8) + 1 as i8_out 
FROM stream 
WINDOW TUMBLING(ts, size 100 ms) 
INTO sinkStreamAllF64
```

### Issue 3: Type inference fails with duplicate projection names

The following query fails during type inference due to duplicate column names with different aggregations:

```sql
SELECT MEDIAN(i8) as i8_out, SUM(i8) as i8_out 
FROM stream 
WINDOW TUMBLING(ts, size 100 ms) 
INTO sinkStreamAllF64
```

## Expected Behavior

1. Duplicate column selections should appear multiple times in the result set, matching standard SQL behavior
2. Expressions using aggregation results should parse and execute correctly
3. Type inference should properly handle or provide a clear error message for duplicate column names from different aggregations

## Additional Information

[...]

PR #1420, opened 2026-03-03
Feat(#788): Support expressions inside aggregation functions

zeuchste:
## Summary
- **Support expressions inside aggregation functions** like `AVG(i + UINT64(1))`, `SUM(i * UINT64(2))`, `MEDIAN(i + UINT64(1))` by desugaring them into a pre-aggregation projection step in the SQL parser

The desugaring transforms e.g. `AVG(i + UINT64(1))` into:
```
Source → Projection(*, i + UINT64(1) AS _agg_input_0) → WindowAgg(AVG(_agg_input_0)) → Projection(output)
```

This avoids changes to the aggregation base class or physical operators — the fix is localized to the SQL parser.

## Test plan
- [x] All 7 ColumnNames.test systest cases pass, including:
  - `MEDIAN(i + UINT64(1))` — expression inside aggregation
  - `AVG(i + UINT64(1))` — expression inside aggregation
  - `SUM(i * UINT64(2))` — expression inside aggregation
  - Post-aggregation arithmetic (`MEDIAN(i) + FLOAT64(1)`)
- [x] All 617 ctest cases pass (0 failures)

### DIFF
--- a/nes-sql-parser/private/AntlrSQLParser/AntlrSQLHelper.hpp
+++ b/nes-sql-parser/private/AntlrSQLParser/AntlrSQLHelper.hpp
@@ -95,7 +95,7 @@ public:
     int implicitMapCountHelper = 0;
 
     /// Pre-aggregation projections for desugaring expressions inside aggregation functions.
-    /// E.g., AVG(i + UINT64(1)) becomes: Projection(*, i + UINT64(1) AS _agg_input_0) → AVG(_agg_input_0).
+    /// E.g., AVG(i + UINT64(1)) becomes: Projection(*, i + UINT64(1) AS _AGG_INPUT_0) → AVG(_AGG_INPUT_0).
     std::vector<Projection> preAggregationProjections;
     size_t aggExprCounter = 0;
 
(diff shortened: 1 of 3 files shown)


ls-1801 (changes requested):
I created some additional systests for expression-aggregations on a separate branch: `feat/788-expression-agg-systests` (commit `4e133e78a1`). You can cherry-pick it into your PR:

```
git cherry-pick 4e133e78a1
```

The tests cover nested arithmetic, SQRT, multiple expression-aggs, mixed expression + plain-field aggs, subqueries, and nested window aggregations.

They also reveal an inconsistency in the auto-generated output field names: plain-field aggregations produce `input.a_SUM` (uppercased, with source prefix), but expression-aggregations produce `_agg_input_0_SUM` (lowercase, no source prefix). I think the `_agg_input_N_AGG` naming is generally fine since the user didn't specify an alias and we need to come up with something. However, the names should be uppercased by default to be consistent with plain-field aggregations — currently they're lowercase, which means they can't even be matched by typed sinks (since `CREATE SINK` uppercases identifiers).

ls-1801 (approved):

PR #1429, opened 2026-03-04
Parse count star

### LINKED ISSUE #699: [BUG] Count(*) does not get parsed
### Description

A `COUNT(*)` does not get currently parsed. 
For example, the following query throws a segfault
```
SELECT start, end, vehicle, highway, direction, positionDiv5280, COUNT(*) AS cntSpeed
FROM (SELECT creationTS, vehicle, highway, direction, position/5280 AS positionDiv5280, speed FROM lrb)
GROUP BY (vehicle, highway, direction, positionDiv5280)
WINDOW SLIDING(creationTS, SIZE 30 SEC, ADVANCE BY 1 SEC)
INTO SINK
```

We need to keep the distinction of `COUNT(*)` and `COUNT(column_name)` in mind. The former counts null values where the latter does not. As long as #40 is not closed, we do not support null handling and both are the same. Once #40 is closed, there should be a distinction between both count types.

### Acceptance Criteria

- [ ] `Count(*)` is supported in our query api. 
- [ ] `Count(*)` does count null values, whereas `Count(column_name)` does not. <--- SLT for this

keyseven123:
This PR solves bug #699 that did not allow to use `COUNT(*)` in our query api.

## Verifying this change
This change is tested by newly added SLT.

This PR closes #699

### DIFF
--- a/nes-logical-operators/include/Operators/Windows/Aggregations/CountAggregationLogicalFunction.hpp
+++ b/nes-logical-operators/include/Operators/Windows/Aggregations/CountAggregationLogicalFunction.hpp
@@ -37,19 +37,20 @@ namespace NES
 class CountAggregationLogicalFunction
 {
 public:
-    explicit CountAggregationLogicalFunction(AggregationFieldAccess inputFunction);
+    explicit CountAggregationLogicalFunction(AggregationFieldAccess inputFunction, bool includeNullValues);
 
     [[nodiscard]] CountAggregationLogicalFunction withInferredType(const Schema<Field, Unordered>& schema) const;
     [[nodiscard]] std::string_view getName() const noexcept;
     [[nodiscard]] Reflected reflect() const;
     [[nodiscard]] static DataType getAggregateType();
     [[nodiscard]] AggregationFieldAccess getInputFunction() const;
-    [[nodiscard]] static bool shallIncludeNullValues() noexcept;
+    [[nodiscard]] bool shallIncludeNullValues() const noexcept;
     [[nodiscard]] std::string explain(ExplainVerbosity verbosity) const;
     [[nodiscard]] bool operator==(const CountAggregationLogicalFunction& other) const;
 
 private:
     AggregationFieldAccess inputFunction;
+    bool includeNullValues = false;
     static constexpr std::string_view NAME = "Count";
     static constexpr DataType::Type finalAggregateStampType = DataType::Type::UINT64;
 };
--- a/nes-logical-operators/src/Operators/Windows/Aggregations/CountAggregationLogicalFunction.cpp
+++ b/nes-logical-operators/src/Operators/Windows/Aggregations/CountAggregationLogicalFunction.cpp
@@ -39,16 +39,15 @@
 
 namespace NES
 {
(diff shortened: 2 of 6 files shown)


nes-logical-operators/src/Operators/Windows/Aggregations/CountAggregationLogicalFunction.cpp
```diff
@@ -47,14 +49,13 @@ CountAggregationLogicalFunction::CountAggregationLogicalFunction(FieldAccessLogi
           DataTypeProvider::provideDataType(DataType::Type::UNDEFINED),
           std::move(field),
           std::move(asField))
+    , includeNullValues(false)
```
arcuellar88: If it is initialized with a value then we don't need this row.

alepping (dismissed):
minor comments

nes-logical-operators/include/Operators/Windows/Aggregations/CountAggregationLogicalFunction.hpp
```diff
@@ -41,6 +41,7 @@ class CountAggregationLogicalFunction : public WindowAggregationLogicalFunction
     [[nodiscard]] Reflected reflect() const override;
 
 private:
+    bool includeNullValues;
```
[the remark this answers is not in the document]
keyseven123: Done

arcuellar88 (commented):
OK for me. Thanks.

Artraxon (approved):
Fine by me, I have already removed the not needed stamps in the schema inference tho

nes-sql-parser/AntlrSQL.g4
```diff
@@ -338,7 +338,7 @@ predicate
 
 
 valueExpression
-    : (functionName | typeDefinition) '(' (argument+=expression (',' argument+=expression)*)? ')'                 #functionCall
+    : (functionName | typeDefinition) '(' (starArg=ASTERISK | argument+=expression (',' argument+=expression)*)? ')'                 #functionCall
```
ls-1801: The `starArg=ASTERISK` alternative applies to ALL function calls, not just COUNT. This means `AVG(*)`, `SUM(*)`, etc. are now syntactically valid. Consider either restricting this to a COUNT-specific grammar rule, or adding a check in `exitFunctionCall` that rejects `starArg` for non-COUNT functions with a clear error message.

nes-logical-operators/src/Operators/Windows/Aggregations/CountAggregationLogicalFunction.cpp
```diff
@@ -39,16 +39,15 @@
 
 namespace NES
 {
-CountAggregationLogicalFunction::CountAggregationLogicalFunction(AggregationFieldAccess inputFunction)
-    : inputFunction(std::move(inputFunction))
+/// COUNT(*) includes null values (counts all rows), COUNT(fieldName) does not
```
greenBene, line 42: I assume this is SQL standard? :) If so, please add a short mention of this to avoid uncertainty.

greenBene (approved):
Just had a minor clarifying question, but otherwise looks good :)

PR #1430, opened 2026-03-05
Simplify: remove unnecessary if-guards in aggregation operators

ls-1801:
## Summary
- Reverts the `if (!aggregationPhysicalFunctions.empty())` guards added in #1423 to `AggregationBuildPhysicalOperator` and `AggregationProbePhysicalOperator`
- Removes the `PRECONDITION` in `ChainedHashMapRef::getValueMemArea()` that was the actual cause of the crash

The guards are unnecessary because:
- `nautilus::static_iterable` over empty containers simply generates no code
- `getValueMemArea()` returns a pointer that is never dereferenced when there are no aggregation functions
- The only thing causing the crash was the `PRECONDITION` rejecting both-empty keys+values, which is a legitimate state for window-only queries without GROUP BY

## Test plan
- [x] All 6 ColumnNames systests pass (including the new window-only test case) in interpreter mode


keyseven123 (commented):
Why are we not doing this in #1423? The PR is still open

PR #1442, opened 2026-03-09
feat(Function): Add MONTH_OF, YEAR_OF, and DAY_OF 

keyseven123:
For the coffee machine demo, we require three new functions that extract date components from Unix timestamps (milliseconds, UINT64): MONTH_OF returns the month (1-12), YEAR_OF returns the year, and DAY_OF returns the day of the month (1-31). All use the already added date library for timestamp conversion. 

## Verifying this change


ls-1801 (changes requested):
i would prefer to just use chrono over the date library

ls-1801 (approved):

Artraxon (changes requested):

Artraxon (changes requested):
hey, sorry i was taking so long to do the PR reviews. Can you rebase once on top of one of the stable branches so that I see the correct diff?

keyseven123:
> hey, sorry i was taking so long to do the PR reviews. Can you rebase once on top of one of the stable branches so that I see the correct diff?

Hey,
I have rebased this branch. Please let me know what you think

nes-logical-operators/include/Functions/DayOfLogicalFunction.hpp
```diff
@@ -0,0 +1,85 @@
...
+#include <Util/Logger/Formatter.hpp>
+#include <Util/PlanRenderer.hpp>
+#include <Util/Reflection.hpp>
+
+namespace NES
+{
+
+/// Extracts the day of the month (1-31) from a unix timestamp in milliseconds (UINT64) and returns it as UINT64.
+class DayOfLogicalFunction final
+{
```
Artraxon: Are we sure we can't combine Day/Month/Year Of... into one function? it seems kind of redundant. But I am also not opposed to keeping it like this and then merging them later.

nes-logical-operators/src/Functions/DayOfLogicalFunction.cpp
```diff
@@ -0,0 +1,123 @@
...
+    copy.outputType = dataType;
+    return copy;
+}
+
+LogicalFunction DayOfLogicalFunction::withInferredDataType(const Schema<Field, Unordered>& schema) const
+{
+    const auto newChildren = getChildren() | std::views::transform([&schema](auto& child) { return child.withInferredDataType(schema); })
+        | std::ranges::to<std::vector>();
+    INVARIANT(newChildren.size() == 1, "DayOfLogicalFunction expects exactly one child function but has {}", newChildren.size());
+    auto newDataType = newChildren[0].getDataType();
```
Artraxon: as in the medicalt tests PR, its better to operate on the children directly to not loose the information how many children we have.

Artraxon (changes requested):
Looks good mostly, two comments and would be good if you can fix the clang tidy stuff as well

nes-logical-operators/include/Functions/DayOfLogicalFunction.hpp
keyseven123: Done.

nes-logical-operators/src/Functions/DayOfLogicalFunction.cpp
keyseven123: Done.

Artraxon (approved):

PR #1444, opened 2026-03-09
Slice cache porting

### LINKED ISSUE #1402: [TASK] Slice Cache Porting
### Description

We should port the slice cache into the mainline.

### Acceptance Criteria

- [ ] Tests for Slice Caches (Interpreter and Compiler)
- [ ] Runtime measurement of large scale tests before and after via `systest -b`

keyseven123:
This PR adds a slice cache to reduce the contention on our slice store. In a paper, I saw up to two orders of magnitude improvements. This was with larger buffer sizes than we use in our tests. For our tests, I measured:
  - w/o slice cache workstation (24 Threads): 2 mins 40s
  - w/ slice cache workstation (24 Threads):  1 mins 20s
  - w/o slice cache on CI (4 Threads):  15 mins
  - w/ slice cache on CI (4 Threads):  9.5 mins

## Verifying this change
Added new unit test and existing tests

This PR closes #1402 .

### DIFF
--- a/nes-systests/systest/CMakeLists.txt
+++ b/nes-systests/systest/CMakeLists.txt
@@ -58,28 +58,42 @@ set_property(TARGET systest PROPERTY ENABLE_EXPORTS ON)
 target_link_libraries(systest PUBLIC nes-systest-lib argparse::argparse nes-query-optimizer)
 target_include_directories(systest PUBLIC $<INSTALL_INTERFACE:/include/nebulastream>)
 
+# By default, only run with slice cache enabled. When ALL_SLICE_CACHE_TESTS is ON, also run without.
+if (ALL_SLICE_CACHE_TESTS)
+    set(sliceCacheValues false true)
+else ()
+    set(sliceCacheValues true)
+endif ()
+
 # Add code coverage if enabled
 if (CODE_COVERAGE)
-    target_code_coverage(systest COVERAGE_TARGET_NAME systest_interpreter_coverage PUBLIC AUTO ALL ARGS -n 20 --show-query-performance --exclude-groups large --workingDir=${CMAKE_CURRENT_BINARY_DIR}/interpreter_cc --data ${EXPANDED_TEST_DATA_PATH} --optimizer join_strategy=NESTED_LOOP_JOIN -- --worker.default_query_execution.execution_mode=INTERPRETER --enable_event_trace=true)
-    target_code_coverage(systest COVERAGE_TARGET_NAME systest_interpreter_coverage_Hash_Join PUBLIC AUTO ALL ARGS -n 20 --show-query-performance --exclude-groups large --workingDir=${CMAKE_CURRENT_BINARY_DIR}/interpreter_cc --data ${EXPANDED_TEST_DATA_PATH} --optimizer join_strategy=HASH_JOIN -- --worker.default_query_execution.execution_mode=INTERPRETER --enable_event_trace=true)
-    # Make sure to fetch ExternalData before running the ccov targets
-    add_dependencies(ccov-run-systest_interpreter_coverage test-data)
-    add_dependencies(ccov-run-systest_interpreter_coverage_Hash_Join test-data)
+    foreach (enableSliceCache IN LISTS sliceCacheValues)
+        target_code_coverage(systest COVERAGE_TARGET_NAME systest_interpreter_coverage_sliceCache_${enableSliceCache} PUBLIC AUTO ALL ARGS -n 20 --show-query-performance --exclude-groups large --workingDir=${CMAKE_CURRENT_BINARY_DIR}/interpreter_cc_sliceCache_${enableSliceCache} --data ${EXPANDED_TEST_DATA_PATH} --optimizer join_strategy=NESTED_LOOP_JOIN -- --worker.default_query_execution.execution_mode=INTERPRETER --enable_event_trace=true --worker.default_query_execution.slice_cache.enable_slice_cache=${enableSliceCache})
+        target_code_coverage(systest COVERAGE_TARGET_NAME systest_interpreter_coverage_Hash_Join_sliceCache_${enableSliceCache} PUBLIC AUTO ALL ARGS -n 20 --show-query-performance --exclude-groups large --workingDir=${CMAKE_CURRENT_BINARY_DIR}/interpreter_cc_sliceCache_${enableSliceCache} --data ${EXPANDED_TEST_DATA_PATH} --optimizer join_strategy=HASH_JOIN -- --worker.default_query_execution.execution_mode=INTERPRETER --enable_event_trace=true --worker.default_query_execution.slice_cache.enable_slice_cache=${enableSliceCache})
+        # Make sure to fetch ExternalData before running the ccov targets
+        add_dependencies(ccov-run-systest_interpreter_coverage_sliceCache_${enableSliceCache} test-data)
+        add_dependencies(ccov-run-systest_interpreter_coverage_Hash_Join_sliceCache_${enableSliceCache} test-data)
+    endforeach ()
 endif (CODE_COVERAGE)
 
 # If we are running code coverage, we need to ONLY run the interpreter tests, as otherwise, the code coverage will be 100% for all operators as the compiler traces all branches and operations.
 set(joinStrategies NESTED_LOOP_JOIN HASH_JOIN)
-foreach (joinStrategy IN LISTS joinStrategies)
-    ExternalData_Add_Test(test-data
-            NAME systest_interpreter_${joinStrategy}
-            COMMAND systest -n 20 --show-query-performance --workingDir=${CMAKE_CURRENT_BINARY_DIR}/interpreter_${joinStrategy} --exclude-groups large --data ${EXPANDED_TEST_DATA_PATH} --optimizer join_strategy=${joinStrategy} -- --worker.query_engine.number_of_worker_threads=4 --worker.default_query_execution.execution_mode=INTERPRETER)
-    set_tests_properties(systest_interpreter_${joinStrategy} PROPERTIES LABELS "Systest" PROCESSORS 4)
+# The systest_interpreter and _compiler pass quite quickly, thus, we run them w/ and w/o a slice cache (if ALL_SLICE_CACHE_TESTS is ON)
+foreach (enableSliceCache IN LISTS sliceCacheValues)
+    foreach (joinStrategy IN LISTS joinStrategies)
+        ExternalData_Add_Test(test-data
+                NAME systest_interpreter_${joinStrategy}_sliceCache_${enableSliceCache}
+                COMMAND systest -n 20 --show-query-performance --workingDir=${CMAKE_CURRENT_BINARY_DIR}/interpreter_${joinStrategy}_sliceCache_${enableSliceCache} --exclude-groups large --data ${EXPANDED_TEST_DATA_PATH} --optimizer join_strategy=${joinStrategy} -- --worker.query_engine.number_of_worker_threads=4 --worker.default_query_execution.execution_mode=INTERPRETER --worker.default_query_execution.slice_cache.enable_slice_cache=${enableSliceCache})
+        set_tests_properties(systest_interpreter_${joinStrategy}_sliceCache_${enableSliceCache} PROPERTIES LABELS "Systest" PROCESSORS 4)
+    endforeach ()
 endforeach ()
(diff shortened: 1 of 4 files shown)


nes-physical-operators/include/SliceCache/SliceCacheSecondChance.hpp
```diff
@@ -0,0 +1,69 @@
+/*
```
greenBene: What is the difference between SliceCache, SliceCacheNone, and SliceCacheSecondChance? I'd appreciate a short high level in code documentation :)

nes-systests/systest/CMakeLists.txt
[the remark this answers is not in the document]
keyseven123: Done. Similar to the buffer leak tests, we will let it run on a nightly basis without slice cache enabled.

nes-systests/systest/CMakeLists.txt
```diff
@@ -90,56 +104,58 @@ if (ENABLE_LARGE_TESTS)
...
     # is still small enough to finish the test in minutes.
     set(joinStrategy NESTED_LOOP_JOIN)
     ExternalData_Add_Test(TEST_DATA_LARGE
             NAME systest_large_${joinStrategy}
-            COMMAND systest -n 6 --show-query-performance --workingDir=${CMAKE_CURRENT_BINARY_DIR}/large_scale_${joinStrategy} --data ${EXPANDED_TEST_DATA_PATH} -t ${CMAKE_CURRENT_SOURCE_DIR}/../benchmark/Nexmark.test:05 --optimizer join_strategy=${joinStrategy} -- --worker.default_query_execution.execution_mode=COMPILER --worker.query_engine.number_of_worker_threads=4 --worker.number_of_buffers_in_global_buffer_manager=2000000)
+            COMMAND systest -n 6 --show-query-performance --workingDir=${CMAKE_CURRENT_BINARY_DIR}/large_scale_${joinStrategy} --data ${EXPANDED_TEST_DATA_PATH} -t ${CMAKE_CURRENT_SOURCE_DIR}/../benchmark/Nexmark.test:05 --optimizer join_strategy=${joinStrategy} -- --worker.default_query_execution.execution_mode=COMPILER --worker.query_engine.number_of_worker_threads=4 --worker.number_of_buffers_in_global_buffer_manager=2000000 --worker.default_query_execution.slice_cache.enable_slice_cache=true})
```
ls-1801: Syntax bug: there is a stray `}` before the closing `)` on this line, which will cause a CMake parse error. Should be `...enable_slice_cache=true)` not `...enable_slice_cache=true})`.

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalHashJoin.cpp
```diff
@@ -279,12 +285,64 @@ LoweringRuleResultSubgraph LowerToPhysicalHashJoin::apply(LogicalOperator logica
...
+    /// Creating the hash join operator handler and slice store
     auto handlerId = getNextOperatorHandlerId();
+    auto sliceAndWindowStore = std::make_unique<DefaultTimeBasedSliceStore>(
+        windowType->getSize().getTime(), windowType->getSlide().getTime(), conf.sliceCacheConfiguration);
+    auto sliceStoreRefLeft = sliceAndWindowStore->createSliceStoreRef(
+        [](Slice& slice, const WorkerThreadId workerThreadId) -> std::span<const std::byte>
+        {
+            auto& hjSlice = dynamic_cast<HJSlice&>(slice);
+            const auto& ptr = hjSlice.getHashMapPtrOrCreate(workerThreadId, JoinBuildSideType::Left);
+            return {reinterpret_cast<const std::byte*>(ptr.get()), sizeof(HJSlice)};
```
ls-1801: The span size uses `sizeof(HJSlice)` but `ptr.get()` points to a `HashMap` object (from `getHashMapPtrOrCreate`). This should be `sizeof(HashMap)` to match the actual object type, consistent with the aggregation lowering rule. Same issue on line 315.

nes-physical-operators/include/WindowBasedOperatorHandler.hpp
```diff
@@ -100,5 +101,8 @@ class WindowBasedOperatorHandler : public OperatorHandler
     uint64_t numberOfWorkerThreads;
     const OriginId outputOriginId;
     const std::vector<OriginId> inputOrigins;
+
+    /// We have here a vector, as we might have more than a single input pipeline with concurrent access
+    folly::Synchronized<std::unordered_map<PipelineId, std::unique_ptr<TupleBuffer>>> sliceCacheBuffers;
```
ls-1801: The `sliceCacheBuffers` member appears to be dead code -- it is declared but never used anywhere. The cache buffer lifetime is managed by `DefaultTimeBasedSliceStore::pipelineIdToSliceCacheStarts`. Consider removing it to avoid confusion and unnecessary overhead.

nes-physical-operators/src/SliceStore/SliceCache/SliceCacheSecondChance.cpp
```diff
@@ -0,0 +1,115 @@
...
+        nautilus::val<SliceCacheEntrySecondChance*> sliceCacheEntryToReplace = threadLocalStart + replacementIndex;
+        const nautilus::val<bool> secondChanceBit = sliceCacheEntryToReplace.get(&SliceCacheEntrySecondChance::secondChanceBit);
+        if (secondChanceBit == nautilus::val<bool>{false})
+        {
+            break;
+        }
+        sliceCacheEntryToReplace.get(&SliceCacheEntrySecondChance::secondChanceBit) = nautilus::val<bool>{false};
+        replacementIndex = (replacementIndex + 1) % numberOfEntries;
+    }
+    *replacementIndexPtr = replacementIndex;
```
ls-1801: After replacing the victim, `replacementIndex` is not advanced. This means the next miss scan immediately encounters the just-inserted entry and clears its second-chance bit, effectively making it the first eviction candidate on the subsequent miss. Advancing the index by 1 after replacement is more common in clock implementations and would improve cache effectiveness.

nes-physical-operators/src/SliceStore/SliceCache/SliceCacheSecondChance.cpp
```diff
@@ -0,0 +1,115 @@
...
+uint64_t SliceCacheSecondChance::getCacheMemorySize() const
+{
+    return SliceCache::getCacheMemorySize() + (numberOfWorkerThreads * sizeof(uint64_t));
+}
+
+SliceCacheSecondChance::EntryFound SliceCacheSecondChance::searchInCache(
+    const nautilus::val<SliceCacheEntrySecondChance*>& threadLocalStart, const nautilus::val<Timestamp>& timestamp)
+{
+    /// We assume that a timestamp is in the cache, if the timestamp is in the range of the slice, e.g., sliceStart <= timestamp < sliceEnd.
+    for (nautilus::val<uint64_t> i = 0; i < numberOfEntries; ++i)
```
ls-1801, line 52: The loops at lines 52 and 95 use `nautilus::val<uint64_t>` for the loop variable but the bound `numberOfEntries` is a plain C++ value, not a nautilus-traced value. Elsewhere in the codebase (e.g., AggregationBuildPhysicalOperator, HJBuildPhysicalOperator), `nautilus::static_val` is used for loop variables when the bound is a non-traceable C++ value. Should this follow the same pattern?

nes-physical-operators/src/Aggregation/AggregationSlice.cpp
```diff
@@ -34,14 +34,14 @@ AggregationSlice::AggregationSlice(
 {
 }
 
-HashMap* AggregationSlice::getHashMapPtr(const WorkerThreadId workerThreadId) const
+const std::unique_ptr<HashMap>& AggregationSlice::getHashMapPtr(const WorkerThreadId workerThreadId) const
```
ls-1801: Returning `const std::unique_ptr<HashMap>&` exposes internal ownership details without benefit. Callers immediately call `.get()` anyway. Prefer returning `HashMap*` (or `HashMap&` if guaranteed non-null) to avoid leaking the ownership model and the theoretical risk of dangling references into the vector.

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalHashJoin.cpp
keyseven123: This is not correct as the method returns a hashmap but the actual object is a HJSlice

nes-systests/systest/CMakeLists.txt
keyseven123: Done

nes-physical-operators/include/WindowBasedOperatorHandler.hpp
keyseven123: Removed

nes-physical-operators/src/SliceStore/SliceCache/SliceCacheSecondChance.cpp
keyseven123: Done

nes-physical-operators/src/SliceStore/SliceCache/SliceCacheSecondChance.cpp
keyseven123: Not really as this would unroll the loop without any benefit.

nes-physical-operators/src/Aggregation/AggregationSlice.cpp
keyseven123: Done

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalHashJoin.cpp
ls-1801: I don't understand the ptr definitely points to a ChainedHashMap, not the HJSlice? What would we the point of a span if it contains the size of a unreleated type?

nes-physical-operators/include/SliceStore/SliceCache/SliceCache.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    /// As we are doing everything in Nautilus, we do not care about the initialization of these values
+    SliceCacheEntry() : sliceStart(0), sliceEnd(0), dataStructure(nullptr) { }
+
+    SliceCacheEntry(const Timestamp::Underlying sliceStart, const Timestamp::Underlying sliceEnd, DataStructure dataStructure)
+        : sliceStart(sliceStart), sliceEnd(sliceEnd), dataStructure(dataStructure)
+    {
+    }
+
+    virtual ~SliceCacheEntry() = default;
```
ls-1801: Three related issues with `SliceCacheEntry`:

1. **Virtual destructor wastes space:** The `virtual ~SliceCacheEntry()` adds an 8-byte vtable pointer to every cache entry, but these objects are never polymorphically destroyed.

2. **Objects never properly constructed:** Cache memory is allocated via `getUnpooledBuffer()`, zero-initialized, and accessed via `reinterpret_cast` — no `placement new` is ever called, so object lifetime is never started (technically UB per the C++ standard, though it works in practice because nautilus uses member-offset access).

3. **Missing static_assert:** Consider adding `static_assert(std::is_trivially_destructible_v<SliceCacheEntry>)` to enforce that the type can be safely used without explicit destruction in the raw-memory cache pattern.

Just use plain structs here and don't bother with inheritance

ls-1801 (changes requested):

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalHashJoin.cpp
keyseven123: Ah, I get it now. Changed it to `sizeof(ChainedHashMap)`.

nes-physical-operators/include/SliceStore/SliceCache/SliceCache.hpp
keyseven123: Yes, we are never using the virtual destructor, but I would like to keep the inheritance for struct layout composition, i.e., to avoid duplicating common fields and to have a shared pointer type in the interface, i.e., `using SliceCacheReplaceEntry = std::function<void(const nautilus::val<SliceCacheEntry*>&)>;`.

I ahve added a comment for this.

Issue #1150, opened 2025-09-26
[TASK] Make LogicalFunctions, Physical Operators/Functions, AggregationFunctions, and Traits Concept-Based


Artraxon:
### Description

#1110 makes LogicalOperators concept based.
This reduces boilerplate code necessary to add new operators, reduces unsafe casts, and allows logical operator types to also inherit from interfaces, which allows implementers to mix-in behaviour into logical operators and to access them at runtime.

As LogicalFunctions, Traits and physical Functions/Operators are implemented with the same approach as logical operators used to be, we should convert them to this approach as well.

AggregationFunctions are solely based on inheriting from a virtual class and do not even have  their own wrapper, so they need to be converted as well.

We should be able to extract the wrapper logic of TypedLogicalOperator to some generalized TypedWrapper , if we remove logical operator specific member functions from the TypedLogicalOperator itself.
They are not needed anymore anyway, as a type erased wrapper dereferences to the type erased virtual class.

### Acceptance Criteria

- [ ] The logic of TypedLogicalOperator is extracted
- [ ] Logical functions are concept based
- [ ] Aggregation functions are concept based and have a type erased wrapper
- [ ] Traits are concept based

PR #1445, opened 2026-03-09
make physicalOperator-concept-based

gustafsa:
Refactored PhysicalOperator to use the same concept-based Type Erasure pattern as LogicalOperator.

Replaced inheritance with TypedPhysicalOperator wrapper and PhysicalOperatorModel.

## Issue referenced by this pull request:

References #1150


nes-physical-operators/include/Watermark/IngestionTimeWatermarkAssignerPhysicalOperator.hpp
```diff
@@ -13,21 +13,42 @@
...
+    void setup(ExecutionContext& ctx, CompilationContext& compCtx) const;
+    void open(ExecutionContext& ctx, RecordBuffer& recordBuffer) const;
+    void close(ExecutionContext& ctx, RecordBuffer& recordBuffer) const;
+    void terminate(ExecutionContext& ctx) const;
+    void execute(ExecutionContext& ctx, Record& record) const;
+
+    [[nodiscard]] OperatorId getId() const;
+    OperatorId id = INVALID_OPERATOR_ID;
+
+    bool operator==(const IngestionTimeWatermarkAssignerPhysicalOperator&) const { return true; };
```
ls-1801: This `operator==` unconditionally returns `true`. This means two `IngestionTimeWatermarkAssignerPhysicalOperator` instances with different children or different time functions will incorrectly compare as equal. Was this intentional as a placeholder, or should it compare the relevant fields?

ls-1801 (changes requested):

PR #1452, opened 2026-03-10
Fix(#1424): enforce BufferManager destruction and fix leak stacktraces

### LINKED ISSUE #1424: [BUG] BufferManager is currently not destroyed and checked for buffer leaks
### Version / Commit

main

### What happened?

The System-Test does currently not enforce that the BufferManager is destroyed.

This is essential because the BufferManager's destructor validates that all buffers have been released back to the buffer manager after all queries have finished.

The BufferManager is destroyed when the worker stops.

Building with NES_DEBUG_BUFFER_LEAKS should also print the stacktraces of the leaked buffer allocations.

### How to reproduce

Run systest and observe the BufferManager destructor.

### Operating System

linux

### Deployment details

_No response_

### Anything else

_No response_

### Code of Conduct

zeuchste:
## Summary

- **Fix `dumpOwningThreadInfo` bug**: The function threw an exception *before* the loop that prints per-thread allocation stacktraces, making the loop dead code. Changed to `NES_ERROR` logging so all stacktraces are printed when `NES_DEBUG_TUPLE_BUFFER_LEAKS` is enabled.
- **Add `BufferManager::waitForAllBuffersReturned(timeout)`**: Polls until all pooled buffers are returned or timeout expires. Solves the chicken-and-egg problem where leaked buffers hold `shared_ptr` refs to the BufferManager, preventing its destructor (and its `INVARIANT` check) from ever running.
- **Fix `NodeEngine` destructor teardown order**: Explicitly resets `queryEngine`, `sourceProvider`, and `queryTracker` before waiting for buffers. This ensures all components release their BufferManager references during orderly shutdown. Logs outstanding buffer count on timeout.

## Changed files

| File | Change |
|------|--------|
| `nes-memory/TupleBufferImpl.cpp` | Fix `dumpOwningThreadInfo` to log instead of throw before stacktrace loop |
| `nes-memory/BufferManager.cpp` | Add `waitForAllBuffersReturned` implementation |
| `nes-memory/include/Runtime/BufferManager.hpp` | Declare `waitForAllBuffersReturned` method |
| `nes-runtime/src/Runtime/NodeEngine.cpp` | Sequence teardown and add buffer leak check in destructor |

## Test plan

- [x] All 12 nes-memory unit tests pass
- [x] All 12 nes-runtime unit tests pass
- [x] 23/23 projection systests pass (exercises full worker lifecycle including BufferManager shutdown)
- [x] `check-format` clean

Closes #1424

### DIFF
--- a/nes-memory/BufferManager.cpp
+++ b/nes-memory/BufferManager.cpp
@@ -61,6 +61,11 @@ std::shared_ptr<BufferManager> BufferManager::create(
 
 BufferManager::~BufferManager()
 {
+    destroy();
+}
+
+void BufferManager::destroy()
+{
     bool expected = false;
     NES_DEBUG("Calling BufferManager::destroy()");
     if (isDestroyed.compare_exchange_strong(expected, true))
@@ -78,6 +83,7 @@ BufferManager::~BufferManager()
 #ifdef NES_DEBUG_TUPLE_BUFFER_LEAKS
                 buffer.controlBlock->dumpOwningThreadInfo();
 #endif
+                NES_ERROR("[BufferManager] leaked buffer detected: segment at {}", fmt::ptr(buffer.ptr));
                 success = false;
             }
         }
--- a/nes-memory/TupleBufferImpl.cpp
+++ b/nes-memory/TupleBufferImpl.cpp
@@ -157,13 +157,13 @@ BufferControlBlock* BufferControlBlock::retain()
 void BufferControlBlock::dumpOwningThreadInfo()
 {
     std::unique_lock lock(owningThreadsMutex);
-    throw UnknownException("Buffer {} has {} live references", fmt::ptr(getOwner()), referenceCounter.load());
+    NES_ERROR("Buffer {} has {} live references", fmt::ptr(getOwner()), referenceCounter.load());
     for (auto& item : owningThreads)
     {
         for (auto& v : item.second)
         {
-            throw UnknownException(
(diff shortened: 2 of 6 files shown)


nes-memory/BufferManager.cpp
```diff
@@ -245,4 +246,19 @@ BufferManagerType BufferManager::getBufferManagerType() const
     return BufferManagerType::GLOBAL;
 }
 
+bool BufferManager::waitForAllBuffersReturned(const std::chrono::milliseconds timeout) const
```
ls-1801: Instead of this `waitForAllBuffersReturned` method, we need a public `void destroy()` method that contains the current destructor logic (the leak checking, `INVARIANT`, memory deallocation from `~BufferManager()`). The destructor should simply call `destroy()`. 

The destructor already has the `isDestroyed` guard and even logs "Calling BufferManager::destroy()" — the intent for a separate method is clearly there. Extracting it allows `NodeEngine` (and tests) to explicitly trigger shutdown and leak reporting at the right point in the destruction sequence, breaking the circular `shared_ptr` reference that prevents the destructor from ever running when buffers leak.

`waitForAllBuffersReturned` can be dropped — `destroy()` subsumes its purpose. After the query engine shuts down, all buffers should already be returned. If they're not, that's a genuine leak and `destroy()` should report it immediately rather than waiting.

ls-1801 (changes requested):

PR #1457, opened 2026-03-11
Fix(#1215) Watermark: unify ingestion time and event time watermark assigners

### LINKED ISSUE #1215: [TASK] Refactor Ingestiontime
### Description

We should refactor our ingestion time handling such that we store the time for each tuple in an automatically added field. Thus, we could simplify the execution as every tuple now has its time directly stored in one field. 

### Acceptance Criteria

- [ ] Have no distinction in the nautilus execution code between ingestion and event time

zeuchste:
## Summary

Removes the separate `IngestionTimeWatermarkAssigner` operator hierarchy (logical operator, physical operator, lowering rule) and unifies both ingestion time and event time under the `EventTimeWatermarkAssigner` path.

### Problem

Ingestion time and event time had completely separate code paths in the execution layer, despite being conceptually similar — both assign a timestamp to tuples for windowing. The difference is:

- **Event time**: reads a per-tuple field from the data
- **Ingestion time**: reads per-buffer metadata (`BufferControlBlock::creationTimestamp`) set by the source

This duality resulted in duplicate operator hierarchies (`IngestionTimeWatermarkAssigner` vs `EventTimeWatermarkAssigner`), a polymorphic `TimeFunction` split, and branching in every lowering rule that touches time.

### Changes

**Deleted files (6):**
- `IngestionTimeWatermarkAssignerLogicalOperator.hpp/.cpp` — logical operator
- `IngestionTimeWatermarkAssignerPhysicalOperator.hpp/.cpp` — physical operator
- `LowerToPhysicalIngestionTimeWatermarkAssigner.hpp/.cpp` — lowering rule

**Modified files (11):**

| File | Change |
|------|--------|
| `LogicalPlanBuilder.cpp` | Always inserts `EventTimeWatermarkAssigner` for both ingestion and event time |
| `TimeCharacteristic.cpp` | `createIngestionTime()` now sets field to `$record.creationTs` |
| `FieldAccessLogicalFunction.cpp` | Handles `$record.creationTs` as a built-in virtual field during schema inference |
| `EventTimeWatermarkAssignerPhysicalOperator.hpp/.cpp` | Uses polymorphic `unique_ptr<TimeFunction>` to handle both event and ingestion time |
| `LowerToPhysicalEventTimeWatermarkAssigner.cpp` | Detects `$record.creationTs` field → creates `IngestionTimeFunction`; otherwise `EventTimeFunction` |
| `LowerToPhysicalWindowedAggregation.cpp` | Removed `IngestionTime` switch branch; detects `$record.creationTs` by field name |
| `TimestampField.hpp` | Removed `INGESTION_TIME` enum and `TimeFunctionType`; simplified `toTimeFunction()` and `getTimestampLeftAndRight()` |
| `CMakeLists.txt` (3 files) | Removed plugin registrations for deleted files |

### Design Decisions

- **`TimeCharacteristic::Type::IngestionTime`** is kept as a user-facing concept in the SQL layer. The translation to event time happens at the planner level.
- **`IngestionTimeFunction`** is kept as a small (6-line) execution-level implementation detail. It reads the timestamp from `ctx.currentTs` (set by the scan from buffer metadata). Fully eliminating it would require writing the buffer creation timestamp into each tuple as a schema field — a follow-up change.
- **`$record.creationTs`** is treated as a "virtual" built-in field: it's recognized by `FieldAccessLogicalFunction::withInferredDataType()` and `TimeBasedWindowType::inferStamp()` without being part of the actual record schema.

## Test plan

- [x] All unit tests pass (`DecideJoinTypesTest`, `DecideMemoryLayoutTest`, watermark tests)
- [x] All systests pass (439/439 non-benchmark tests)
- [x] Format check passes
- [ ] CI smoke tests

Closes #1215


nes-logical-operators/src/Functions/FieldAccessLogicalFunction.cpp
```diff
@@ -75,6 +76,14 @@ std::string FieldAccessLogicalFunction::explain(ExplainVerbosity verbosity) cons
...
 LogicalFunction FieldAccessLogicalFunction::withInferredDataType(const Schema& schema) const
 {
+    /// $record.creationTs is a built-in virtual field (buffer creation timestamp) not present in the schema.
+    if (fieldName == Windowing::TimeCharacteristic::RECORD_CREATION_TS_FIELD_NAME)
+    {
+        auto copy = *this;
+        copy.dataType = DataTypeProvider::provideDataType(DataType::Type::UINT64);
+        return copy;
+    }
+
```
keyseven123: I think this should be done in the query api and we should not have a separate case here. This is why I wanted to have a talk with @Artraxon before continuing with this issue.

keyseven123 (changes requested):
Couple high-level comments:
1. We do not test ingestion time currently and there are no new tests in this pr. 
2. Before creating a PR, I would have appreciated to have a talk and discuss how we get the timestamp from the source to the inputformatter / what do we define as ingestion time.

adrianmichalke (changes requested):

PR #1460, opened 2026-03-11
Fix(#1440): fix contradictory preconditions in VariableSizedData::isValid

### LINKED ISSUE #1440: [BUG] Preconditions in VariableSizedData::isValid contradict each other
### Version / Commit

Commit hash: 739fe21222f850d7d0a2b42f1f87e763914fe2a5 (current main)

### What happened?

I tried to run a query in which VARSIZED values are compared to a boolean. 
CREATE LOGICAL SOURCE mySource(myvar VARSIZED);
CREATE PHYSICAL SOURCE FOR mySource TYPE File;
ATTACH INLINE
this
is
a
bug

SELECT * FROM mySource WHERE myvar == BOOLEAN(1) INTO File();
----
this
is
a
bug

This should be allowed, since VariableSizedData implements 
`nautilus::val<bool> operator==(const VariableSizedData& varSizedData, const nautilus::val<bool>& other);`

A VARSIZED value is true, if it's size is at least 1 and the corresponding pointer is not a nullptr. This is defined in the function isValid of VariableSizedData. However, isValid has two preconditions, which always seem to contradict each other:
`PRECONDITION(size > 0 && ptrToVarSized != nullptr, "VariableSizedData has a size of 0 but  a nullptr pointer to the data.");`
`PRECONDITION(size == 0 && ptrToVarSized == nullptr, "VariableSizedData has a size of 0 so there should be no pointer to the data.");`

At least one of these will always be false, so my query yields
"Precondition violated: (size == 0 && ptrToVarSized == nullptr): VariableSizedData has a size of 0 so there should be no pointer to the data"

Query passes with expected results.

The two preconditions should be replaced with one that ORs the two predicates.
Besides that, the system currently does not assign a nullptr to 0-size Varsized Values, so the second precondition will never be true (testable with following query:

CREATE LOGICAL SOURCE jsonStream(KEY UINT64, value UINT64, name VARSIZED);
CREATE PHYSICAL SOURCE FOR jsonStream TYPE File SET("JSON" AS PARSER.`TYPE`);
ATTACH INLINE
{"KEY":1, "VALUE":1, "NAME":""}
{"KEY":1, "VALUE":2, "NAME":""}

SELECT * FROM jsonStream WHERE name == BOOLEAN(0) INTO File();
----
1,1,
1,2,

### How to reproduce

Run following query as systest:

CREATE LOGICAL SOURCE mySource(myvar VARSIZED);
CREATE PHYSICAL SOURCE FOR mySource TYPE File;
ATTACH INLINE
this
is
a
bug

SELECT * FROM mySource WHERE myvar == BOOLEAN(1) INTO File();
----
this
is
a
bug

### Operating System

Ubuntu 24.04.3 LTS

### Deployment details

Not relevant

### Anything else

_No response_

### Code of Conduct

zeuchste:
## Summary
- Fixed two `PRECONDITION` checks in `VariableSizedData::isValid()` that used `&&` (logical AND), making them mutually exclusive and impossible to both satisfy
- Changed to `||` (logical OR) so each precondition correctly validates its invariant: if size > 0 then pointer must be non-null, and if size == 0 then pointer must be null
- Also fixed the error message on the first precondition to accurately describe the failure condition

Closes #1440

## Test plan
- [ ] Verify existing `VariableSizedDataTest` unit tests pass
- [ ] Build in Debug mode (where PRECONDITION macros are active) and confirm no assertion failures in normal operation

### DIFF
--- a/nes-nautilus/src/Nautilus/DataTypes/VariableSizedData.cpp
+++ b/nes-nautilus/src/Nautilus/DataTypes/VariableSizedData.cpp
@@ -73,8 +73,8 @@ nautilus::val<bool> operator==(const nautilus::val<bool>& other, const VariableS
 
 nautilus::val<bool> VariableSizedData::isValid() const
 {
-    PRECONDITION(size > 0 && ptrToVarSized != nullptr, "VariableSizedData has a size of 0 but  a nullptr pointer to the data.");
-    PRECONDITION(size == 0 && ptrToVarSized == nullptr, "VariableSizedData has a size of 0 so there should be no pointer to the data.");
+    PRECONDITION(size == 0 || ptrToVarSized != nullptr, "VariableSizedData has a size > 0 but a nullptr pointer to the data.");
+    PRECONDITION(size != 0 || ptrToVarSized == nullptr, "VariableSizedData has a size of 0 so there should be no pointer to the data.");
     return size > 0 && ptrToVarSized != nullptr;
 }
 


keyseven123:
Closed as duplicate of #1441

PR #1465, opened 2026-03-11
fix(Unpooled Buffers): Fixed wrong buffer size after a load

keyseven123:
This PR fixes a bug that would return the wrong buffer size after a load. The main culprit was that we align the buffer size but return the `neededSize`.

## Verifying this change

### DIFF
--- a/nes-memory/tests/ChildBufferTests.cpp
+++ b/nes-memory/tests/ChildBufferTests.cpp
@@ -1,0 +1,105 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


nes-memory/tests/ChildBufferTests.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <ctime>
+#include <random>
+
+#include <Runtime/BufferManager.hpp>
+#include <Runtime/TupleBuffer.hpp>
+#include <Util/Logger/Logger.hpp>
+#include <gtest/gtest.h>
+
+/// Testing if the buffer size stays the same during a store --> load
+TEST(ChildBufferTests, StoreAndLoadChildBuffer)
```
ThanGeo: On top of the random sizes, i would add one or two hardcoded tests more where you check a) small, odd values for the sizes (e.g. 1 byte, 7 bytes, 63 bytes etc) and b) exact "power of 2" values (e.g. 8, 16, 32, 64 etc).

That would make the test more deterministic against edge cases

PR #1466, opened 2026-03-12
Add bloom filter implementation into NLJ, tests, and benchmarks

LucaCode37:
This pull request adds a BloomFilter and its integration into the nested loop join. The change list is as follows:
- New BloomFilter file with clear(), add() and mightcontain() functions.
- Integration into the NLJ with changed files 'NLJBuild/OperatorHandles/Probe/Slice', to reduce lookups
- Integrated an enableBloomFilterFlag for Testing with and without a BloomFilter to see performance gain.

## Verifying this change
This change is tested by
- *Added integration tests for creating the BloomFilter in the NLJ context*
- Creating a folder Benchmark_BF for BloomFilter Benchmark tests, which contain already existing Nexmark "Join" Tests and synthetic tests.
- *Added test that measures the BloomFilters performance (UNIT test)*
- *Added metrics test with blommCheckTotal/bloomMisses/bloomHits/bloomFalsePositives*


keyseven123 (changes requested):
Preliminary feedback.

PR #1469, opened 2026-03-12
fix(bats): Clean up Docker images and build cache after distributed tests

keyseven123:

The distributed bats tests (`distributed-cli-test`, `distributed-repl-tests`) build Docker images with unique tags but only removed them via `docker rmi` in `teardown_file`. This left behind build cache and base image layers that accumulated over repeated runs (~11GB per crash). If a test crashed, `teardown_file` never ran, leaking all images, containers, and build cache.

This PR adds targeted Docker cleanup to both `setup_file` (crash recovery) and `teardown_file` (normal exit):
- Remove stopped containers so image layers can be freed
- Remove leaked test images by name pattern (`nes-worker-*-test-*`, `nes-cli-image-*`, `nes-repl-image-*`)
- Prune dangling build cache and images

## Verifying this change

Verified on my workstation:
- **Normal run (10 iterations):** `docker system df` shows 0B images/build cache after each run
- **Crash recovery:** after killing ctest mid-run (~11GB leaked), the next run cleans up to 13MB (only `alpine`)

### DIFF
--- a/nes-frontend/apps/cli/tests/distributed.bats
+++ b/nes-frontend/apps/cli/tests/distributed.bats
@@ -18,6 +18,12 @@ setup_file() {
     docker network inspect "$net" -f '{{range .Containers}}{{.Name}} {{end}}' 2>/dev/null | xargs -r docker rm -f 2>/dev/null || true
   done
   docker network prune -f --filter label=nes-test=distributed-cli 2>/dev/null || true
+  # Remove all containers (running or stopped) referencing test images
+  # from previous runs, so those images can later be deleted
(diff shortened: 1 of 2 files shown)


nes-frontend/apps/cli/tests/distributed.bats
```diff
@@ -111,6 +117,8 @@ teardown_file() {
   echo "# Test suite completed" >&3
   docker rmi $WORKER_IMAGE || true
   docker rmi $CLI_IMAGE || true
+  docker builder prune -f || true
+  docker image prune -f || true
```
ls-1801: don't do this! I don't want to lose all my installed images when running this test.

Deleting the specific images should already be enough, and the Docker cache should not accumulate.

PR #1470, opened 2026-03-12
feat(ML): Add inference operator with IREE runtime

ls-1801:
## Summary

Adds ML model inference to NebulaStream via the `MODEL_INFERENCE` table-valued function. Users register ONNX models with `CREATE MODEL` and run them inline in streaming SQL queries.

### SQL Syntax

**Model registration** (SQL or YAML):
```sql
CREATE MODEL iris ('/path/to/iris.onnx')
  INPUT (p1 FLOAT32, p2 FLOAT32, p3 FLOAT32, p4 FLOAT32)
  OUTPUT (setosa FLOAT32, versicolor FLOAT32, virginica FLOAT32);
```

**Inference queries:**
```sql
-- Direct stream input
SELECT * FROM MODEL_INFERENCE(iris, stream) INTO result;

-- Subquery with base64 decoding (for binary/image data)
SELECT * FROM MODEL_INFERENCE(mnist, (SELECT FROM_BASE64(pixels) AS pixels FROM stream)) INTO result;

-- Nested inference (chain two models)
SELECT * FROM MODEL_INFERENCE(model_b, MODEL_INFERENCE(model_a, stream)) INTO result;
```
- *Tested by:* `InferModel.test`, `InferModelMNIST.test`, `InferModelTwoModels.test`, `InferModelRegression.test`

### Architecture

The inference pipeline has three phases:

**1. Model Registration** — `CREATE MODEL` stores model metadata (name, path, input/output schema) in a `ModelCatalog`. Path validation happens at registration time. Models can also be registered via YAML topology files with a `models:` section.
- *Tested by:* `InferModelLogicalOperatorTest` (19 test cases), `ModelLoaderTest::LoadInvalidPath`, `ModelLoaderTest::LoadNonOnnxExtensionIsRejected`, `offline.bats::nebucli dump with model inference topology`

**2. Model Compilation** (lazy, on first query) — The `ModelLoader` spawns two subprocesses piped together:
  - `iree-import-onnx` converts the ONNX file to IREE MLIR on stdout
  - `iree-compile` reads that MLIR on stdin and emits VMFB bytecode on stdout

  Process management uses `boost::process::v2` + `boost::asio` with `async_wait` and `steady_timer` for a 10-second timeout. Compiled bytecode is cached in the catalog.
- *Tested by:* `ModelLoaderTest` (10 tests: identity/reduction/expansion models, corrupt/empty files, temp directory cleanup)

**3. Query Execution** — The logical plan lowers to an `InferModelPhysicalOperator` that runs inference per-record using the IREE runtime. Each worker thread gets its own `IREEAdapter` instance for thread safety.
- *Tested by:* `InferModelPhysicalOperatorTest` (6 tests in both interpreted and compiled modes, including concurrent stress test with 8 threads × 200 buffers)

### Why IREE Tools Are External

We deliberately **do not link the IREE compiler** into NebulaStream. IREE's compiler depends on a specific MLIR version that would conflict with NebulaStream's own MLIR backend. Instead we:

- Install the IREE Python package (`iree-base-compiler`) which provides `iree-import-onnx` and `iree-compile` as standalone binaries
- Link only the **IREE runtime** (`iree_runtime_runtime`) as a C library via vcpkg/nix (version 3.10.0)
- Spawn the compiler tools as subprocesses at model-load time

### Logical Operator Hierarchy

Two logical operators handle name→model resolution:

- `InferModelNameLogicalOperator` — holds the model name string (pre-resolution from SQL parser)
- `InferModelLogicalOperator` — holds the compiled `Model` object (post-resolution by optimizer), performs schema inference with type compatibility checks

The lowering rule (`LowerToPhysicalInferModel`) detects whether input fields are `VARSIZED` (binary blob) or individual `FLOAT32` columns and sets a flag on the physical operator accordingly.

### Physical Operator Input Modes

1. **Multiple FLOAT32 columns** — each named input field is read as a float and fed to the model individually
   - *Tested by:* `InferModel.test` (4 floats → 3 outputs), `InferModelRegression.test` (5 floats → 1 output)

2. **Single VARSIZED binary blob** — raw tensor bytes passed directly to the IREE runtime
   - *Tested by:* `InferModelMNIST.test` (base64-encoded 28×28 image → 10 digit probabilities)

### Base64 Functions (`FROM_BASE64`, `TO_BASE64`)

Since NebulaStream has no native binary literal syntax, we added base64 encode/decode as SQL functions using OpenSSL's `EVP_EncodeBlock`/`EVP_DecodeBlock` (OpenSSL is already a dependency). Base64 encoding is also used internally for model bytecode serialization.
- *Tested by:* `InferModelMNIST.test`, `Base64.test`

### Dependencies & Infrastructure

- **IREE runtime** (v3.10.0) added to vcpkg (`vcpkg/vcpkg-registry/ports/ireeruntime/`) and nix (`.nix/ireeruntime/`)
- **IREE compiler tools** installed in dev Docker image (`docker/dependency/Development.dockerfile`) via pip
- **CMake gating**: `cmake/CheckTestUtilities.cmake` runs `find_program(iree-import-onnx)` and `find_program(iree-compile)` — if either is missing, `ENABLE_IREE_TESTS=OFF` and inference tests are not built
- **Runtime check**: `ModelLoader::enabled()` also verifies tool availability before attempting compilation
- **YAML model registration** added to `nes-cli` topology files

### Test Models

Unit tests (`nes-inference/tests/testdata/`): `tiny_identity.onnx` (100→100), `tiny_reduction.onnx` (100→10), `tiny_expansion.onnx` (10→100)

System tests (`nes-systests/testdata/model/`): `iris.onnx` (4→3 classification), `inspections.onnx` (5→1 regression), `mnist.onnx` (784→10 digit recognition)

### Constraints

- **f32 only** — tensor element types other than f32 are rejected at model load time
- **CPU-only** — compiled with `--iree-hal-target-device=local` and `llvm-cpu` backend
- **10s compile timeout** — model compilation times out after 10 seconds
- **ONNX only** — only `.onnx` model files are supported

## Test Plan

- [x] `ModelLoaderTest` — model loading, error handling, temp directory cleanup (10 tests)
- [x] `InferModelLogicalOperatorTest` — schema inference, type validation, serialization (19 tests)
- [x] `InferModelPhysicalOperatorTest` — correctness across shapes, multi-record, concurrent stress test 8 threads × 200 buffers (6 tests, both interpreted and compiled modes)
- [x] `offline.bats` — YAML model registration in nes-cli
- [x] System tests: `InferModel.test`, `InferModelRegression.test`, `InferModelMNIST.test`, `InferModelTwoModels.test`, `Base64.test`
- [x] TSAN clean (no data races in concurrent test)


nes-physical-operators/src/Inference/IREERuntimeWrapper.cpp
```diff
@@ -0,0 +1,196 @@
...
+{
+    iree_runtime_instance_options_t instanceOptions;
+    iree_runtime_instance_options_initialize(&instanceOptions);
+    iree_runtime_instance_options_use_all_available_drivers(&instanceOptions);
+    iree_runtime_instance_t* inst = nullptr;
+    iree_status_t status = iree_runtime_instance_create(&instanceOptions, iree_allocator_system(), &inst);
+    NES_DEBUG("Created IREE runtime instance")
+
+    iree_hal_device_t* dev = nullptr;
+    iree_runtime_instance_try_create_default_device(inst, iree_make_cstring_view("local-sync"), &dev);
```
ls-1801: Two unchecked IREE API calls: (1) Line 71: `iree_runtime_instance_create` status is assigned but never checked — if it fails, `inst` is null and line 75 is UB. (2) Line 75: `iree_runtime_instance_try_create_default_device` return is discarded — if no device is available, `dev` is null and subsequent session creation + `iree_hal_device_release(dev)` are UB. Please add status checks for both, with proper cleanup on failure.

nes-plugins/Inference/IREE/src/IREERuntimeWrapper.cpp
```diff
@@ -0,0 +1,177 @@
+/*
```
ls-1801: This entire plugin directory (`nes-plugins/Inference/IREE/`) duplicates the `IREERuntimeWrapper` that already exists in `nes-physical-operators/src/Inference/`. Please remove this duplicate — there should be a single implementation.

nes-logical-operators/src/Operators/InferModelLogicalOperator.cpp
```diff
@@ -0,0 +1,231 @@
...
+#include <SerializableOperator.pb.h>
+
+namespace
+{
+std::string base64Encode(const std::string& input)
+{
+    const auto encodedLen = 4 * ((input.size() + 2) / 3);
+    std::string out(encodedLen, '\0');
+    EVP_EncodeBlock(
+        reinterpret_cast<unsigned char*>(out.data()), reinterpret_cast<const unsigned char*>(input.data()), static_cast<int>(input.size()));
```
ls-1801: The return value of `EVP_EncodeBlock` is discarded. Since the input is untrusted user data, please check the return value and throw an exception on failure (-1).

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalInferModel.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+LoweringRuleResultSubgraph LowerToPhysicalInferModel::apply(LogicalOperator logicalOperator)
+{
+    PRECONDITION(logicalOperator.tryGetAs<InferModelLogicalOperator>(), "Expected an InferModelLogicalOperator");
+    const auto inferModelOp = logicalOperator.getAs<InferModelLogicalOperator>();
+    auto model = inferModelOp.get().getModel();
+
+    NES_DEBUG("Lowering InferModel operator to physical IREE operator (function: {})", model.getFunctionName());
+
+    const auto inputSchema = logicalOperator.getInputSchemas()[0];
```
ls-1801: Use `.at(0)` instead of `[0]` for bounds-checked access.

nes-inference/ModelLoader.cpp
```diff
@@ -0,0 +1,376 @@
...
+        if (model_stream.gcount() > 0)
+        {
+            std::ranges::copy(std::span{buffer.data(), static_cast<size_t>(model_stream.gcount())}, std::back_inserter(modelVmfb));
+        }
+
+        /// boost::process v1 wait_for() uses sigtimedwait on POSIX, which can miss
+        /// SIGCHLD and block for the full timeout even after the process exits
+        /// (boostorg/process#69). We work around this by calling wait() in a background
+        /// thread and using std::future::wait_for for the timeout.
+        auto waitFuture = std::async(std::launch::async, [&]() { std::ranges::for_each(process, [](auto& proc) { proc.wait(); }); });
```
ls-1801: **Race condition / use-after-free:** After terminating processes on timeout, the function returns without joining the async thread (`waitFuture` is dropped). The background thread still references the `process` vector which is being destroyed — this is a use-after-free. Also, calling `terminate()` concurrently with `wait()` on the same `bp::child` is a data race.

**Simplest fix:** Replace the `std::async` approach with a polling loop using `proc.running()` + `sleep_for(50ms)`:
```cpp
constexpr auto timeout = std::chrono::seconds(10);
constexpr auto pollInterval = std::chrono::milliseconds(50);
const auto deadline = std::chrono::steady_clock::now() + timeout;

while (std::chrono::steady_clock::now() < deadline)
{
    if (std::ranges::none_of(process, [](auto& proc) { return proc.running(); }))
        break;
    std::this_thread::sleep_for(pollInterval);
}

bool timedOut = false;
for (auto& proc : process)
{
    if (proc.running())
    {
        timedOut = true;
        proc.terminate();
        proc.wait();
    }
}
if (timedOut)
    return std::unexpected(ModelLoadError("Model import/compile timed out after 10 seconds"));
```

**Better approach:** I tested a full rewrite using boost::process v2 + boost::asio (v2 headers are already available in the vcpkg install). This uses `asio::io_context` with `steady_timer` + `process::async_wait()` — timeout and process completion are driven by the same event loop, so race conditions are structurally impossible. All 10 inference tests pass. Happy to share the implementation if you want to go that route.

nes-inference/ModelLoader.cpp
```diff
@@ -0,0 +1,376 @@
...
+                metadata.functionName = parseFunctionName(label);
+            }
+            else if (findLabelOfUnambiguousSuccessor(v, graph)
+                         .transform([](const auto& label) { return label.find("hal.tensor.export") != std::string::npos; })
+                         .value_or(false))
+            {
+                metadata.outputShape = parseTensorShape(label);
+            }
+        }
+        return metadata;
```
ls-1801: If the DOT graph lacks expected nodes, `getModelMetadata()` returns empty shapes/function name without error. The caller then computes `inputSizeInBytes = sizeof(float) * accumulate(empty_vec, 1, multiplies) = 4` bytes, which is silently wrong. Please validate that all metadata fields were populated.

nes-inference/ModelLoader.cpp
```diff
@@ -0,0 +1,486 @@
...
+    }
+    NES_INFO("Installed Inference Tools:\n{}", fmt::join(std::views::filter(tools, &Tool::available), "\n"));
+    return true;
+}
+
+}
+
+bool enabled()
+{
+#ifndef NES_INFERENCE_SUPPORT
```
ls-1801: remove this compilation definition and also remove it from cmake if it still exists

nes-logical-operators/tests/InferModelLogicalOperatorTest.cpp
```diff
@@ -0,0 +1,390 @@
...
+    });
+}
+
+/// LOGI-02: explain() returns non-empty string containing "InferModel" or "INFER_MODEL"
+TEST_F(InferModelLogicalOperatorTest, ExplainReturnsDescription)
+{
+    const auto op = makeOp();
+    auto description = op.explain(ExplainVerbosity::Short, OperatorId{1});
+    EXPECT_FALSE(description.empty());
+    EXPECT_TRUE(description.find("INFER_MODEL") != std::string::npos || description.find("InferModel") != std::string::npos);
```
ls-1801: tighten this test to execpt INFER_MODEL

greenBene (changes requested):
Thank you for pushing this 💪 I think it is fundamental for the future of NES! :) 

Added some comments/questions.

greenBene (approved):

docs/model_inference.md
```diff
@@ -0,0 +1,233 @@
...
+- `iree-import-onnx` — ONNX → MLIR (the `torch` dialect, as `!torch.vtensor<[...],f32>`).
+- `iree-compile` — MLIR → IREE bytecode for a specific target.
+
+NES does not link against the IREE compiler library. Both binaries are
+installed with `pip install iree-base-compiler iree-turbine onnx` into a
+dedicated venv inside the development and runtime Docker images
+(`docker/dependency/Development.dockerfile`,
+`docker/runtime/RuntimeBase.dockerfile`); `nes-inference` discovers them on
+`$PATH` and shells out to them.
+
```
keyseven123, line 39: I suspect that a lot of people will use this tvf. Thus a separate install on bare metal section would be good to have

nes-frontend/src/Statements/StatementHandler.cpp
```diff
@@ -228,6 +233,105 @@ std::expected<DropSinkStatementResult, Exception> SinkStatementHandler::operator
...
+    return ModelInfo{
+        .name = reg.name,
+        .path = reg.path.string(),
+        .inputSchema = std::move(inputSchema),
+        .outputSchema = std::move(outputSchema),
+    };
+}
+}
+
+std::expected<CreateModelStatementResult, Exception> ModelStatementHandler::operator()(const CreateModelStatement& statement)
```
keyseven123, line 261: some small docs what is happening here and if the order is of importance would not hurt, IMHO.

nes-inference/include/Inference.hpp
```diff
@@ -0,0 +1,45 @@
...
+    std::string message;
+};
+
+struct CompileError
+{
+    std::string message;
+};
+
+/// Test if inference is enabled and the environment is set up to support management of models
+bool inferenceEnabled();
```
keyseven123: I do not like that this method is not guarded against anything. Everyone that links with `nes-inference` can call this method. BUt from what Clion tells me, this function is only used in tests

nes-systests/inference/InferModelMNIST.test
```diff
@@ -0,0 +1,20 @@
...
...AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADBwEA8zMtLP//+fj/c21s/jYwMPgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAmZgYPv/+fj///n4/m5qaPgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA+fj4PeHgYD///n4/5+bmPoGAgDsAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAIaFBT///n4///5+P9HQUD4AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAPX0dD7z8nI///5+P//+fj/R0FA+AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAADz8vI+//5+P//+fj/c21s/oaAgPgAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA8/LyPv/+fj/Qz08/kZCQPQAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==
```
keyseven123: Quite a long line. Can we move this into a file?

nes-systests/inference/InferModel.test
```diff
@@ -0,0 +1,23 @@
...
+4.9,3.0,1.4,0.2
+4.7,3.2,1.3,0.2
+
+CREATE MODEL iris ('model/iris.onnx')
+INPUT (p1 FLOAT32, p2 FLOAT32, p3 FLOAT32, p4 FLOAT32)
+OUTPUT (setosa FLOAT32, versicolor FLOAT32, virginica FLOAT32);
+
+CREATE SINK result(`STREAM$P1` FLOAT32 NOT NULL, `STREAM$P2` FLOAT32 NOT NULL, `STREAM$P3` FLOAT32 NOT NULL, `STREAM$P4` FLOAT32 NOT NULL, SETOSA FLOAT32 NOT NULL, VERSICOLOR FLOAT32 NOT NULL, VIRGINICA FLOAT32 NOT NULL) TYPE File;
+
+SELECT * FROM MODEL_INFERENCE(iris, stream)
```
keyseven123, line 19: Please add one query that uses the output of the model to perform some filter

nes-logical-operators/src/Functions/FromBase64LogicalFunction.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+FromBase64LogicalFunction FromBase64LogicalFunction::withDataType(const DataType& dataType) const
+{
+    auto copy = *this;
+    copy.dataType = dataType;
+    return copy;
+};
+
+/// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+LogicalFunction FromBase64LogicalFunction::withInferredDataType(const Schema& schema) const
```
keyseven123, line 65: Should you not need to set the data type here in this method?

keyseven123: Or atleast perform a check/invariant if the input type is a varsized

nes-logical-operators/src/Functions/ToBase64LogicalFunction.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+ToBase64LogicalFunction ToBase64LogicalFunction::withDataType(const DataType& dataType) const
+{
+    auto copy = *this;
+    copy.dataType = dataType;
+    return copy;
+};
+
+/// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+LogicalFunction ToBase64LogicalFunction::withInferredDataType(const Schema& schema) const
```
keyseven123, line 65: Same comment as in the FromBase64. I would have expected here an invartiant

nes-nautilus/tests/TestUtils/src/TestTupleBuffer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+
+/// Converts a FieldValue to a VarVal for use with the Record/TupleBufferRef interface.
+VarVal fieldValueToVarVal(const FieldValue& value, const DataType::Type type)
+{
+    switch (type)
+    {
+        case DataType::Type::BOOLEAN:
+            return VarVal(std::get<bool>(value));
```
keyseven123, line 90: Just a heads-up. The VarVal is currently nullable and null

keyseven123 (changes requested):
As this PR is quite time-sensitive, I have not checked every small detail. We can fix/change some nitpicks later.

nes-nautilus/tests/TestUtils/src/TestTupleBuffer.cpp
ls-1801: I am not sure if i understand the problem. This is the constructor that is being called:

```c++
    template <typename T>
    explicit VarVal(const T value)
    requires(detail::is_one_of<nautilus::val<T>, detail::var_val_t>::value)
        : value(nautilus::val<T>(value)), null(false), nullable(false)
    {
    }
```

I would assume it's not_nullable and not null.

I extended the TestTupleBuffer to allow to work with nullable values.
`buffer.append(32, std::nullopt, 12)`. would work and you can load field values by using `.at<std::optional<size_t>>()`

nes-logical-operators/src/Functions/ToBase64LogicalFunction.cpp
ls-1801: done.

nes-logical-operators/src/Functions/FromBase64LogicalFunction.cpp
ls-1801: done

nes-systests/inference/InferModel.test
ls-1801: done

nes-systests/inference/InferModelMNIST.test
ls-1801: done

nes-frontend/src/Statements/StatementHandler.cpp
ls-1801: added a comment

docs/model_inference.md
ls-1801: added a section

nes-nautilus/tests/TestTupleBufferTest.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+    view.append(int64_t{42});
+
+    ASSERT_EQ(view.getNumberOfTuples(), 1);
+    EXPECT_EQ(view[0]["value"].as<int64_t>(), 42);
+}
+
+TEST_F(TestTupleBufferTest, AppendAndReadMultipleFields)
+{
+    auto schema = Schema().addField("a", DataType::Type::INT64).addField("b", DataType::Type::UINT32);
```
keyseven123, line 72: I do not like at all that we are now going back to using a schema directly to reason about the underlying order and datatypes of a schema. This PR is time-sensitive so it is fine, but we should have a physical layout though. Or at least use the LowerSchemaProvider class and use the current TupleBufferRef instead of passing a schema into the TupleBufferRef. I like to have an inefficient class for populating TupleBuffers but this here now ties a logical component to a physical one.

Or what does @Artraxon think about this? Can we maybe do it as one of your PRs? I have created an issue for this. If you @Artraxon tackle this as part of one of your PRs please assign yourself and the issue to the PR

keyseven123 (approved):

nes-nautilus/tests/TestTupleBufferTest.cpp
keyseven123: Issue is #1579

PR #1471, opened 2026-03-12
feat: add outer join support (LEFT, RIGHT, FULL)

ls-1801:
## Summary

Adds `LEFT OUTER JOIN`, `RIGHT OUTER JOIN`, and `FULL OUTER JOIN` support to NebulaStream's windowed hash join framework.

### Limitations

- Outer joins are currently **limited to hash joins with equi-predicates** (e.g., `ON left.key = right.key`). Non-equi outer joins and nested-loop outer joins are not supported and will throw `UnsupportedQuery`. This is enforced both in the `JoinLogicalOperator` constructor (requires `FieldAccessLogicalFunction` in predicate) and in the `DecideJoinTypes` optimizer phase (forces `HASH_JOIN` strategy for outer joins).

### Changes

- **SQL parser:** `LEFT/RIGHT/FULL [OUTER] JOIN` syntax in `AntlrSQL.g4`
- **Logical operators:** `JoinType` enum with `OUTER_LEFT_JOIN`, `OUTER_RIGHT_JOIN`, `OUTER_FULL_JOIN`; nullable schema propagation (left side nullable for RIGHT/FULL, right side nullable for LEFT/FULL)
- **Optimizer:** `DecideJoinTypes` routes outer joins to hash join; rejects non-equi outer joins
- **Physical operators:** null-fill probe logic, typed null `VarVal` construction per `DataType`

### Probe task parallelism

Inner joins use an **S² strategy**: for a window with S slices, S² independent probe tasks are emitted (one per left/right slice pair), all parallelizable across worker threads.

Outer joins extend this with additional **null-fill tasks**. The challenge is that determining whether a key is "unmatched" requires checking **all** opposite-side slices — not just one. We solve this by splitting the work into two kinds of tasks:

1. **S² match-pair tasks** — identical to inner join. Each task probes one (left_slice, right_slice) pair and emits matched rows. Fully parallel.
2. **S null-fill tasks per preserved side** — each task iterates keys from one slice on the preserved side and checks for matches against **all** opposite-side hash maps. Only unmatched keys emit NULL-padded rows.

| Join type | Match tasks | Null-fill tasks | Total |
|-----------|-------------|-----------------|-------|
| LEFT      | S²          | S (left side)   | S² + S |
| RIGHT     | S²          | S (right side)  | S² + S |
| FULL      | S²          | 2S (both sides) | S² + 2S |

All tasks are independent and parallelizable. For tumbling windows (S=1), this reduces to 1 match + 1-2 null-fill tasks.

## Test plan

- [x] Unit tests: JoinLogicalOperator nullability, DecideJoinTypes routing, null VarVal construction
- [x] 56 join systests pass (tumbling, sliding, chained, large data, multiple keys, mixed windows, all join types)
- [x] Sliding window tests specifically validate cross-slice correctness (key in slice 1 matching key in slice 2 must not produce false NULL rows)
- [ ] CI passes


nes-physical-operators/include/Join/NestedLoopJoin/NLJOperatorHandler.hpp
```diff
@@ -39,6 +39,8 @@ struct EmittedNLJWindowTrigger
     WindowInfo windowInfo;
 };
 
+class NLJSlice;
```
ls-1801: nit: `NLJSlice` is not referenced in this header. This forward declaration appears unnecessary.

nes-sql-parser/AntlrSQL.g4
```diff
@@ -448,6 +451,7 @@ OF: 'OF';
 ON: 'ON' | 'on';
 OR: 'OR' | 'or';
 ORDER: 'ORDER' | 'order';
+OUTER: 'OUTER';
```
ls-1801: @/tmp/code-review-pr1471-20260314135107/comment-body.txt

nes-logical-operators/tests/JoinLogicalOperatorTest.cpp
```diff
@@ -0,0 +1,300 @@
...
+        auto op = createOperator(joinType);
+        return op.withInferredSchema({createLeftSchema(), createRightSchema()});
+    }
+};
+
+/// -------------------------------------------------------------------------
+/// isOuterJoin free function tests
+/// -------------------------------------------------------------------------
+
+TEST_F(JoinLogicalOperatorTest, IsOuterJoinReturnsTrueForOuterLeftJoin)
```
ls-1801: @/tmp/code-review-pr1471-20260314135107/comment-body.txt

nes-logical-operators/tests/JoinLogicalOperatorTest.cpp
```diff
@@ -0,0 +1,220 @@
...
+        ASSERT_TRUE(outField.has_value()) << "Missing right field: " << field.name;
+        EXPECT_FALSE(outField->dataType.nullable) << "Inner join right field should not be nullable: " << field.name;
+    }
+    /// NOLINTEND(bugprone-unchecked-optional-access)
+}
+
+TEST_F(JoinLogicalOperatorTest, WindowStartEndFieldsAreNotNullableForOuterJoins)
+{
+    /// NOLINTBEGIN(bugprone-unchecked-optional-access) ASSERT_TRUE(has_value()) guards the access but clang-tidy cannot see through gtest macros
+    for (const auto joinType : {JoinLogicalOperator::JoinType::OUTER_LEFT_JOIN, JoinLogicalOperator::JoinType::OUTER_FULL_JOIN})
```
ls-1801: This test iterates over LEFT and FULL outer joins but omits RIGHT outer join. Adding `OUTER_RIGHT_JOIN` to the list would make the coverage complete.

nes-physical-operators/src/Join/StreamJoinProbePhysicalOperator.cpp
```diff
@@ -18,20 +18,69 @@
...
+#include <Nautilus/DataTypes/VariableSizedData.hpp>
 #include <Nautilus/Interface/Record.hpp>
 #include <Nautilus/Interface/TimestampRef.hpp>
 #include <Runtime/Execution/OperatorHandler.hpp>
 #include <Time/Timestamp.hpp>
 #include <Windowing/WindowMetaData.hpp>
+#include <magic_enum/magic_enum.hpp>
+#include <ErrorHandling.hpp>
 #include <WindowProbePhysicalOperator.hpp>
 #include <static.hpp>
```
ls-1801: Please remove the `default` clause so the compiler can warn on unhandled enum values if new types are added to `DataType::Type`. Handle `UNDEFINED` explicitly (with a throw), and add `std::unreachable()` after the switch (since each case branch returns).

nes-sql-parser/AntlrSQL.g4
```diff
@@ -448,6 +451,7 @@ OF: 'OF';
 ON: 'ON' | 'on';
 OR: 'OR' | 'or';
 ORDER: 'ORDER' | 'order';
+OUTER: 'OUTER' | 'outer';
```
ls-1801: The new OUTER token correctly supports both cases (`'OUTER' | 'outer'`), but the pre-existing LEFT, RIGHT, and FULL tokens only match uppercase. This means queries like `left outer join` will fail at the lexer level. Please update LEFT, RIGHT, and FULL to also support lowercase (e.g., `LEFT: 'LEFT' | 'left';`) for consistency with OUTER, JOIN, INNER, and other keywords.

greenBene (changes requested):
Thank you for adding this missing functionality! :) 

I am not an expert in the physical operator side, but from what I do understand this code looks to match the approach you described in the PR, and the tests validate the described behavior. :) 

I just have one request regarding the `.test` files (see comment)

greenBene (approved):

keyseven123 (changes requested):
I would like to have this a little bit more general. IMHO, it should be possible to have a general probe and trigger function that we do not need to implement the outer join trigger all the time for a new join. I will think about this.

keyseven123 (approved):

PR #1478, opened 2026-03-13
Fix(#1164): add initial unit tests for nes-plugins

### LINKED ISSUE #1164: [TASK] Add unit tests to `nes-plugins`
### Description

While the sources and sinks defined in `nes-plugins` are partially being tested via other means (mainly systest), I think it would make sense to add unit tests to the individual plugins. 
These unit tests should cover the plugins via their (source or sink) interfaces and would thus give an easy way to validate if changes made to the plugins introduced any unexpected behavior.

### Acceptance Criteria

Unit tests are added to all currently existing nes-plugins. 
Test are part of the individual plugins, not the overall `nes-plugin` folder, so the individual plugins can easily being in- or excluded.

zeuchste:
## Summary
- Adds GTest unit tests for **GeneratorSource**, **VoidSink**, and **ChecksumSink** plugins
- Tests are placed inside each plugin's own `tests/` directory so plugins can be independently included/excluded (per acceptance criteria)
- GeneratorSource tests cover: `Generator`, `SequenceField`, `NormalDistributionField`, `RandomStrField`, `FixedGeneratorRate`, and `SinusGeneratorRate`
- Sink tests cover config validation via `validateAndFormat`
- Each plugin's `CMakeLists.txt` is updated with `add_tests_if_enabled(tests)`

## Test plan
- [ ] Verify `generator-source-test` builds and passes
- [ ] Verify `void-sink-test` builds and passes
- [ ] Verify `checksum-sink-test` builds and passes
- [ ] Confirm tests are discoverable via `ctest -R generator-source-test`, `ctest -R void-sink-test`, `ctest -R checksum-sink-test`

Closes #1164


adrianmichalke (changes requested):
VoidSink and ChecksumSink tests seem to test only rather trivial configuration behavior. IMO, we should not add them

adrianmichalke (approved):

ls-1801 (approved):

adrianmichalke (approved):

alepping (changes requested):
Does this add to our code coverage? Otherwise it is imo just noise that we now need to maintain, since failing tests mean a failing CI.

zeuchste:
Good point — I reworked the PR with exactly that in mind.

- I **removed** the VoidSink and ChecksumSink tests, since (as @adrianmichalke also noted) they only exercised trivial config validation — that was the "noise" part.
- What remains are **GeneratorSource tests that cover real logic**, not config plumbing: sequence/normal-distribution/random-string field generation, the `ALL`/`ONE`/`NONE` stop policies, deterministic-output-per-seed, mixed-field delimiter handling, and schema-parsing edge cases (empty schema, whitespace-only, and empty schema-line rejection).
- `codecov/patch` is now **green** — the one previously-uncovered line (the empty-schema-line throw) is covered by an added test.

So this isn't just mirroring defaults that the framework already guarantees — it pins behavior of the Generator parsing/RNG paths that currently have no other coverage. Happy to drop any specific case you still find low-value.

zeuchste:
Addressed @adrianmichalke's review: the trivial VoidSink/ChecksumSink configuration tests were removed; the PR now only adds the GeneratorSource unit tests, which exercise real behaviour (schema parsing, deterministic generation, tuple counts, and the empty-schema-line rejection path) rather than pass-through config.

Also rebased onto current main (was 19 commits behind, with a merge commit) and squashed the noisy CI-fix history into a single clean commit. Re-requesting review.

adrianmichalke (approved):

alepping (approved):

keyseven123 (changes requested):
Please clean up the commits.

zeuchste:
Cleaned up the history as requested: squashed the 10 WIP/CI-fix commits into a single commit and rebased onto current main, resolving the conflict with the reworked plugin CMake setup (kept the new single-library + registry-entry structure, appended `add_tests_if_enabled(tests)`; the obsolete `generator_validation_plugin_library` line was dropped). Build and all 34 tests pass locally; clang-format and clang-tidy-diff are clean.

keyseven123 (approved):

PR #1480, opened 2026-03-13
Fix(#1397): add CI cron job for Docker cleanup

### LINKED ISSUE #1397: [TASK] Setup Cronjobs on self hosted CI servers to cleanup docker junk
### Description

We frequently run out of disk space on our ci machines due to docker stuff accumulating.
Non of the docker images are important as they can be recreated automatically.

Something like:

0 3 * * 0 /usr/bin/docker system prune -af --volumes

### Acceptance Criteria

All self hosted ci machines remove unused docker images/volumes/container on a schedule.

zeuchste:
## Summary
- Adds a new GitHub Actions workflow that runs `docker system prune -af --volumes` weekly (Sundays at 3:00 AM UTC) on all self-hosted CI runners (x64 and arm64)
- Includes `workflow_dispatch` trigger for manual runs
- Prevents disk space exhaustion from accumulated Docker images, containers, and volumes

Closes #1397

## Test plan
- [ ] Verify the workflow appears in the Actions tab after merge
- [ ] Trigger manually via `workflow_dispatch` to confirm it runs on both x64 and arm64 runners
- [ ] Confirm `docker system prune -af --volumes` executes successfully

### DIFF
--- a/.github/workflows/docker-cleanup.yml
+++ b/.github/workflows/docker-cleanup.yml
@@ -1,0 +1,16 @@
+name: Docker Cleanup
+
+on:
+  schedule:
+    - cron: '0 3 * * 0' # Every Sunday at 3:00 AM UTC
+  workflow_dispatch:
+
+jobs:
+  docker-cleanup:
+    strategy:
+      matrix:
+        arch: [ x64, arm64 ]
+    runs-on: [ self-hosted, linux, Build, "${{ matrix.arch }}" ]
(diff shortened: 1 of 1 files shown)


ls-1801 (changes requested):
I think the bigger problem is that this will run cleanup on exactly two machines, the ARM and an x64 machine.

Its not really what we are looking for and i am not sure if that is even possible

PR #1481, opened 2026-03-13
Fix(#736): generate file listing activated plugins during build

### LINKED ISSUE #736: [TASK] Track activated plugins in file
### Description

Currently, we register plugins in our core-components and in nes-plugins.
There is no central way to see all plugins that the system supports.
All plugins use one of two functions in https://github.com/nebulastream/nebulastream-public/blob/main/cmake/PluginRegistrationUtil.cmake to register.
We could write the names of the registries and their corresponding plugins to a json/yaml file in nes-plugins.
This could be helpful to quickly see a list of all supported/acitvated plugins, especially when there are different version of our system.
We could also extend such a file to allow (de-)activating plugins, including plugins in core-components.

### Acceptance Criteria

- [ ] generate a central list of active plugin registries and their corresponding plugins

zeuchste:
## Summary
- Adds CMake logic to track all activated plugin registries and their plugins during configuration
- Generates `activated_plugins.json` in the build directory, listing each registry and its plugins as a sorted JSON object
- Uses deferred evaluation (`cmake_language(DEFER)`) with a fixed ID so the file is written exactly once, after all plugins have been registered

Closes #736

## Test plan
- [ ] Run CMake configure and verify `activated_plugins.json` is generated in the build directory
- [ ] Inspect the JSON file to confirm it lists all expected registries (DataType, Sink, Source, LogicalOperator, PhysicalFunction, etc.) with their corresponding plugins
- [ ] Toggle an optional plugin (e.g., TCP source) off and re-configure; verify it no longer appears in the file

### DIFF
--- a/cmake/PluginRegistrationUtil.cmake
+++ b/cmake/PluginRegistrationUtil.cmake
@@ -45,6 +45,7 @@ function(add_plugin_as_library plugin_name plugin_registry plugin_registry_compo
 
     set_property(GLOBAL APPEND PROPERTY "${plugin_registry}_plugin_names" "${plugin_name}")
     set_property(GLOBAL APPEND PROPERTY "${plugin_registry}_plugin_libraries" "${plugin_library}")
+    set_property(GLOBAL APPEND PROPERTY NES_ALL_PLUGIN_REGISTRIES "${plugin_registry}")
 endfunction()
 
 # adds the source files of the plugin to the source files of the component that the plugin registry belongs to
@@ -61,6 +62,7 @@ macro(add_plugin plugin_name plugin_registry plugin_registry_component)
         )
     endif ()
     set_property(GLOBAL APPEND PROPERTY "${plugin_registry}_plugin_names" "${plugin_name}")
+    set_property(GLOBAL APPEND PROPERTY NES_ALL_PLUGIN_REGISTRIES "${plugin_registry}")
 endmacro()
 
 # iterates over all plugins, collect all plugins with given name, inject plugins into registrar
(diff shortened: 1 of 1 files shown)


cmake/PluginRegistrationUtil.cmake
```diff
@@ -115,6 +117,56 @@ function(generate_plugin_registrars plugin_registry_component)
                 cmake_language(DEFER DIRECTORY [[${PROJECT_SOURCE_DIR}]] CALL generate_plugin_registrar [[${CMAKE_CURRENT_SOURCE_DIR}]] [[${CMAKE_CURRENT_BINARY_DIR}]] [[${plugin_registry}]] [[${plugin_registry_component}]])
         ")
     endforeach ()
+    # Schedule generation of the activated plugins file at the end of configuration.
+    # The DEFER ensures all plugins have been registered before the file is written.
+    # Using ID_VAR with a fixed id ensures the call is scheduled only once, even if
+    # generate_plugin_registrars is invoked from multiple components.
+    cmake_language(EVAL CODE "
+            cmake_language(DEFER DIRECTORY [[${PROJECT_SOURCE_DIR}]] ID_VAR _nes_gen_plugins_id ID nes_generate_activated_plugins CALL generate_activated_plugins_file)
```
ls-1801, line 125: The `ID nes_generate_activated_plugins` mechanism does not prevent duplicate scheduling — CMake 3.31 allows duplicate DEFER IDs. This causes `generate_activated_plugins_file` to fire 9 times (once per `create_registries_for_component` call). The output is correct, but you can fix the noise by guarding with a global property:
```cmake
get_property(_scheduled GLOBAL PROPERTY NES_ACTIVATED_PLUGINS_FILE_SCHEDULED)
if (NOT _scheduled)
    set_property(GLOBAL PROPERTY NES_ACTIVATED_PLUGINS_FILE_SCHEDULED TRUE)
    cmake_language(EVAL CODE "
        cmake_language(DEFER DIRECTORY [[${PROJECT_SOURCE_DIR}]] CALL generate_activated_plugins_file)
    ")
endif()
```

ls-1801 (changes requested):

cmake/PluginRegistrationUtil.cmake
```diff
@@ -115,6 +117,56 @@ function(generate_plugin_registrars plugin_registry_component)
                 cmake_language(DEFER DIRECTORY [[${PROJECT_SOURCE_DIR}]] CALL generate_plugin_registrar [[${CMAKE_CURRENT_SOURCE_DIR}]] [[${CMAKE_CURRENT_BINARY_DIR}]] [[${plugin_registry}]] [[${plugin_registry_component}]])
         ")
     endforeach ()
+    # Schedule generation of the activated plugins file at the end of configuration.
+    # The DEFER ensures all plugins have been registered before the file is written.
+    # Using ID_VAR with a fixed id ensures the call is scheduled only once, even if
+    # generate_plugin_registrars is invoked from multiple components.
+    cmake_language(EVAL CODE "
```
yschroeder97, line 124: Claude told me you can remove the outer wrapper :)

Rerequest please once this and the below comment are done.

yschroeder97 (changes requested):

PR #1482, opened 2026-03-13
Fix(#1002): allow duplicate field names in systest schema matching

### LINKED ISSUE #1002: [BUG] Systest Tool does not accept duplicate fields

zeuchste:
## Summary
- Fixes the systest schema comparison to correctly handle duplicate field names (e.g., `SELECT id, id FROM stream`)
- The root cause was that `std::ranges::find` always matched the first occurrence of a field, so the second duplicate field was never matched and reported as "unexpected field in actual result schema"
- The fix tracks already-matched actual field indices and skips them when searching for subsequent matches

Closes #1002

## Test plan
- [ ] Run existing systest suite to verify no regressions
- [ ] Create a test with duplicate fields (`SELECT id, id FROM stream INTO sink`) and verify it passes without schema mismatch errors

### DIFF
--- a/nes-systests/regression/DuplicateFieldNamesInResult.test
+++ b/nes-systests/regression/DuplicateFieldNamesInResult.test
@@ -1,0 +1,31 @@
+# description: Regression test for #1002 — systest result matching must handle duplicate field names
+# groups: [regression]
+
+# This test verifies that the systest result checker correctly matches schemas
+# where the output contains duplicate field names (e.g., from a join where both
+# sides contribute a field with the same name).
+
+CREATE LOGICAL SOURCE left_stream(id UINT64 NOT NULL, value UINT64 NOT NULL, timestamp UINT64 NOT NULL);
+CREATE PHYSICAL SOURCE FOR left_stream TYPE File;
+ATTACH INLINE
+1,100,1
+2,200,1
+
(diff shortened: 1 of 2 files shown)


keyseven123 (changes requested):
Please add an SLT that tests this. Otherwise, someone might get rid of this code by accident, and then we go back to integrating it. Additionally, this would mean that one can test it quite easily during a review

adrianmichalke (approved):

keyseven123 (approved):

PR #1486, opened 2026-03-13
Fix(#474): allow qualifier names in join ON clauses

### LINKED ISSUE #474: [TASK] Allowing qualifier names in join functions
### Description

Currently, it is not possible to use qualifiernames in our query syntax. For example, `SELECT * FROM A INNER JOIN B ON A.id = B.id INTO SINK`.

### Acceptance Criteria

- [ ] Disabled tests in `AntlrSQLQueryParserTest` are enabled and pass

zeuchste:
## Summary
- Add `exitDereference` handler to `AntlrSQLQueryPlanCreator` that combines separately-pushed identifier pairs (e.g., `A` and `id` from `A.id`) into a single qualified field access (`A$id`)
- Enables SQL syntax like `SELECT * FROM A JOIN B ON A.id = B.id` in join ON clauses
- Also handles qualified names in SELECT, WHERE, and WINDOW contexts via the same dereference handler
- Add unit tests for both qualified and unqualified join key names

Closes #474

## Test plan
- [ ] New `JoinWithQualifiedNames` test verifies qualified names parse correctly and produce expected `A$ID`/`B$ID` field accesses in the join function
- [ ] New `JoinWithUnqualifiedNames` test ensures existing unqualified name behavior is not regressed
- [ ] Existing join systests continue to pass


ls-1801 (changes requested):
Please add actual system tests to verify this

keyseven123 (changes requested):
Please rebase onto `lr/schema_stable`.

PR #1487, opened 2026-03-13
Fix(#635): refactor hash functions into plugin registry

### LINKED ISSUE #635: [TASK] Hash Functions as a Plugin
### Description

We should refactor the hash functions (`nes-nautilus/include/Nautilus/Interface/Hash/HashFunction.hpp`) into a plugin.

zeuchste:
## Summary
- Refactors hash functions in `nes-nautilus` into the plugin registry system, addressing #635
- Creates a new `HashFunction` plugin registry in `nes-nautilus/registry/` following the existing plugin pattern (same as `Source`, `DataType`, `Sink`, etc.)
- Registers `MurMur3HashFunction` as a plugin named `MurMur3` in the `HashFunction` registry via `add_plugin()`
- The `HashFunction` base class remains as a regular source file; only concrete implementations are registered as plugins

## Changes
- **`nes-nautilus/CMakeLists.txt`**: Added `include(PluginRegistrationUtil.cmake)` and `create_registries_for_component(HashFunction)`
- **`nes-nautilus/registry/include/HashFunctionRegistry.hpp`**: New registry header defining `HashFunctionRegistry`, `HashFunctionRegistryReturnType` (`std::unique_ptr<HashFunction>`), and `HashFunctionRegistryArguments`
- **`nes-nautilus/registry/templates/HashFunctionGeneratedRegistrar.inc.in`**: New CMake template for auto-generating the registrar
- **`nes-nautilus/src/Nautilus/Interface/Hash/CMakeLists.txt`**: Changed `MurMur3HashFunction.cpp` from `add_source_files` to `add_plugin(MurMur3 HashFunction ...)`
- **`nes-nautilus/src/Nautilus/Interface/Hash/MurMur3HashFunction.cpp`**: Added `RegisterMurMur3HashFunction` factory function in the `HashFunctionGeneratedRegistrar` namespace

## Test plan
- [x] CMake configure succeeds
- [x] `nes-nautilus` target builds successfully (including `libnes-nautilus-registry.a`)
- [ ] CI build and test pass
- [ ] Future hash function implementations can be added as plugins using `add_plugin(Name HashFunction nes-nautilus NameHashFunction.cpp)`

Closes #635

### DIFF
--- a/nes-common/include/ExceptionDefinitions.inc
+++ b/nes-common/include/ExceptionDefinitions.inc
@@ -16,7 +16,7 @@
 
 /// NOLINT(clang-diagnostic-error)
 #ifndef EXCEPTION
-#    error "This file should not be included directly! Include it via <ErrorHandling.hpp>"
+    #error "This file should not be included directly! Include it via <ErrorHandling.hpp>"
 #endif
 
 /// 1XXX Configuration Errors
--- a/nes-nautilus/CMakeLists.txt
+++ b/nes-nautilus/CMakeLists.txt
@@ -10,6 +10,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# include registrar generation functions
+include(${PROJECT_SOURCE_DIR}/cmake/PluginRegistrationUtil.cmake)
+
 include_directories(tests/include)
 
 # Set Sources and Headers
@@ -28,6 +31,8 @@ if (NES_ENABLE_PRECOMPILED_HEADERS)
 endif ()
 
 add_tests_if_enabled(tests)
+
+create_registries_for_component(HashFunction)
 
 # Add nes-nautilus to the include directories
 target_include_directories(nes-nautilus PUBLIC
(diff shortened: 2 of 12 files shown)


ls-1801 (changes requested):
I am not quite sure what the point is. Right now, this just adds dead code as the registry is not used anywhere.

keyseven123 (changes requested):
I am sorry but I do not see the benefit of this PR. As @ls-1801 mentions. This PR just adds dead code that we do not use in our code base. If you show via Benchmarks that the hash functions make a difference, I would be open to have them. 

Additionally, I have created a discussion how we should go about plugins and testing/disabling. https://github.com/nebulastream/nebulastream/discussions/1484
As long as the discussion is not finished, I would not like to add an additional plugin.

ls-1801 (changes requested):
@zeuchste I don't think that matters; it's still just dead code. No one chooses a different hash function, and even if the optimizer did, what would be the benefit? We would need to show that even matters and justifies the complexity.

keyseven123 (changes requested):
I want to see plots/benchmarks before we choose any other hash function. Murmurhash3 is used for our hash map in a research paper and thus, it was evaluated. Adding an additional hash function that we do not know if it even improves the performance is just dead code. 

Further, I would like to hold off on this until @greenBene is done with the optimization rules. Otherwise, we will just add something in the execution that we can not choose usefully. I had to really hack the join choice into our current optimization framework.

PR #1489, opened 2026-03-13
Fix(#864): make descriptor config parameters constexpr

### LINKED ISSUE #864: [TASK] Make DescriptorConfigParameters constexpr
### Description

Currently, we initialize the the ConfigParameters of the specific Descriptors like this: `static inline const Configurations::DescriptorConfig::ConfigParameter<...`. This triggers the following clang-tidy warning: `static storage duration may throw an exception that cannot be caught`.
We can work around this problem, by making the `ConfigParameter`s `static constexpr`.
We first need to make the constructor of `ConfigParameter` constexpr and making sure that the members can be constexpr. Specifically, we need to provide the function as a template parameter, since the `std::function` member would otherwise prohobit `constexpr`ness.
```c++
template <typename T, typename U = void, typename Func = void>
    struct ConfigParameter
    {
        using Type = T;
        using EnumType = U;

        constexpr ConfigParameter(std::string name, std::optional<T> defaultValue, Func&& validateFunc)
            : name(std::move(name)), validateFunc(std::move(validateFunc)), defaultValue(std::move(defaultValue))
        {
            static_assert(
                not(std::is_same_v<EnumWrapper, T> && std::is_same_v<void, U>),
                "An EnumWrapper config parameter must define the enum type as a template parameter.");
        }

        std::string name;
        Func validateFunc;
        std::optional<T> defaultValue;

        operator const std::string&() const { return name; }

        std::optional<ConfigType> validate(const std::unordered_map<std::string, std::string>& config) const
        {
            return this->validateFunc(config);
        }

        static constexpr bool isEnumWrapper() { return not(std::is_same_v<U, void>); }
    };
    template <typename T, typename U = void, typename Func>
    static constexpr auto makeConfigParameter(const char* name, std::optional<T> defaultValue, Func&& validateFunc)
    {
        return ConfigParameter<T, U, std::decay_t<Func>>(name, defaultValue, std::forward<Func>(validateFunc));
    }
```

Defining a ConfigParameter then looks like this:
```c++
static constexpr auto TUPLE_DELIMITER = Configurations::DescriptorConfig::makeConfigParameter<std::string>(
        "tupleDelimiter",
        std::nullopt,
        [](const std::unordered_map<std::string, std::string>& config) -> std::optional<std::string>
        {
            // Note: You can't capture in a constexpr lambda, so you need to pass the parameter name
            if (auto tupleDelimiter = config.find("tupleDelimiter"); tupleDelimiter != config.end())
            {
                if (tupleDelimiter->second.size() == 1)
                {
                    return {tupleDelimiter->second};
                }
                return std::nullopt;
            }
            return std::nullopt;
        });
```

We can't access the parameter in the lambda function anymore, so we can't use `tryGet`. We can still check if the config contains the specific parameter and whether it is valid.

### Acceptance Criteria

- [ ] adapt `ConfigParameter` and change all occurences of the constructor from `static const inline` to `static constexpr`
- [ ] make parameterMaps constexpr (e.g., replacing it with a constexpr std::array and constexpr functions to 'lookup' an element in the array
- [ ] change occurences `static const std::string` to `static constexpr std::string_view`

zeuchste:
## Summary
- Adapted `ConfigParameter` to accept the validate function type as a template parameter (`Func`), replacing `std::function` with direct storage, enabling constexpr instances
- Added `makeConfigParameter` factory function and `tryGetByName` helper for non-capturing lambdas
- Changed all `static inline const ConfigParameter` declarations to `static constexpr auto` across sources, sinks, logical operators, and plugins

Closes #864

## Test plan
- [x] Verified `nes-configurations`, `nes-sources`, `nes-sinks` targets compile successfully
- [ ] CI full matrix build
- [ ] Run systests to verify runtime behavior is unchanged

### DIFF
--- a/nes-configurations/include/Configurations/Descriptor.hpp
+++ b/nes-configurations/include/Configurations/Descriptor.hpp
@@ -68,24 +68,25 @@ public:
     /// Tag struct that tags a config key with a type.
     /// The tagged type allows to determine the correct variant of a config paramater, without supplying it as a template parameter.
     /// Therefore, config keys defined in a single place are sufficient to retrieve parameters from the config, reducing the surface for errors.
-    template <typename T, typename U = void>
+    /// The Func template parameter allows the validate function to be stored directly (instead of via std::function),
+    /// enabling constexpr ConfigParameter instances.
+    template <typename T, typename U = void, typename Func = void>
     struct ConfigParameter
     {
         using Type = T;
         using EnumType = U;
-        using ValidateFunc = std::function<std::optional<T>(const std::unordered_map<std::string, std::string>& config)>;
 
-        ConfigParameter(std::string name, std::optional<T> defaultValue, ValidateFunc&& validateFunc)
-            : name(std::move(name)), validateFunc(std::move(validateFunc)), defaultValue(std::move(defaultValue))
+        constexpr ConfigParameter(const char* name, std::optional<T> defaultValue, Func&& validateFunc)
+            : name(name), validateFunc(std::move(validateFunc)), defaultValue(std::move(defaultValue))
         {
             static_assert(
                 not(std::is_same_v<EnumWrapper, T> && std::is_same_v<void, U>),
                 "An EnumWrapper config parameter must define the enum type as a template parameter.");
         }
 
-        const std::string name;
-        const ValidateFunc validateFunc;
-        const std::optional<T> defaultValue;
+        std::string name;
+        Func validateFunc;
+        std::optional<T> defaultValue;
 
         operator const std::string&() const { return name; }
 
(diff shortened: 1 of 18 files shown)


nes-configurations/include/Configurations/Descriptor.hpp
```diff
@@ -68,24 +68,25 @@ class DescriptorConfig
...
+    template <typename T, typename U = void, typename Func = void>
     struct ConfigParameter
     {
         using Type = T;
         using EnumType = U;
-        using ValidateFunc = std::function<std::optional<T>(const std::unordered_map<std::string, std::string>& config)>;
 
-        ConfigParameter(std::string name, std::optional<T> defaultValue, ValidateFunc&& validateFunc)
-            : name(std::move(name)), validateFunc(std::move(validateFunc)), defaultValue(std::move(defaultValue))
+        constexpr ConfigParameter(const char* name, std::optional<T> defaultValue, Func&& validateFunc)
```
ls-1801: i think you could use string_view here instead of the cstring

Artraxon (changes requested):
please rerequest once CI is green

PR #1506, opened 2026-03-18
Fix(#1493, #1504) SinkCatalog: propagate actual errors and detect duplicates

### LINKED ISSUE #1493: SinkStatementHandler reports "sink already exists" for config validation failures
## Description

When `SinkCatalog::addSinkDescriptor` fails due to config validation (e.g., a required parameter is missing), the `SinkStatementHandler` incorrectly throws `SinkAlreadyExists` instead of reporting the actual validation error.

## Root Cause

`addSinkDescriptor` returns `std::optional<SinkDescriptor>` — `nullopt` for **any** failure:
- Config validation fails (`validateAndFormatConfig` returns `nullopt`)
- Sink name is all digits (reserved)

The caller in `StatementHandler.cpp:159` doesn't distinguish these cases:

```cpp
if (const auto created = sinkCatalog->addSinkDescriptor(...))
return std::unexpected{SinkAlreadyExists(statement.name)};  // always assumes duplicate
```

## Reproduction

Create `repro.yaml`:
```yaml
query: |
  SELECT * FROM S INTO MY_SINK
sinks:
  - name: MY_SINK
    schema:
      - name: S$v
        type: FLOAT64
    type: File
    config:
      file_path: /tmp/out.csv
      some_invalid_key: foo
logical:
  - name: S
    schema:
      - name: v
        type: FLOAT64
physical:
  - logical: S
    parser_config:
      type: CSV
    type: Generator
    source_config:
      generator_rate_type: FIXED
      generator_rate_config: emit_rate 1
      seed: 1
      generator_schema: |
```

Run:
```bash
NES_WORKER_GRPC_ADDR=127.0.0.1:8080 nes-cli -t repro.yaml start
```

**Expected:** Error mentioning `Unknown configuration parameter: some_invalid_key`

**Actual:** `error code (2034): sink already exists; MY_SINK`

The real validation error is only visible in debug logs (`-d` flag).

## Suggested Fix

Change `addSinkDescriptor` to return `std::expected<SinkDescriptor, Error>` so the caller can distinguish validation failure from duplicate name and report the correct error.

### LINKED ISSUE #1504: SinkCatalog::addSinkDescriptor silently ignores duplicate sink names
## Summary

`SinkCatalog::addSinkDescriptor` uses `emplace` (line 59 of `nes-sinks/src/SinkCatalog.cpp`) to insert into the map. If a sink with the same name already exists, `emplace` is a no-op — the second sink is silently dropped. However, the method still constructs and returns the new `SinkDescriptor` as if registration succeeded, making it impossible for callers to detect the conflict.

## Details

```cpp
// SinkCatalog.cpp:59
lockedSinks->emplace(toUpperCase(sinkName), sinkDescriptor);
return sinkDescriptor;  // always returns success, even if emplace was a no-op
```

### Downstream impact in systest

**1. `createSink` ignores registration failure (line 578-586)**

`sltSinkProvider.registerSink()` returns a `bool` indicating success, but the return value is discarded:

```cpp
static void createSink(SLTSinkFactory& sltSinkProvider, const CreateSinkStatement& statement)
{
    // ...
    sltSinkProvider.registerSink(statement.sinkType, statement.name, schema, statement.sinkConfig);
    // return value discarded — duplicate sink silently ignored
}
```

**2. `setNamedSink` accesses failed `std::expected` (lines 725-731)**

After a failed `createActualSink`, the error is recorded but execution falls through to access `.value()` on the failed expected — undefined behavior:

```cpp
auto sinkExpected = sltSinkProvider.createActualSink(toUpperCase(sinkName), sinkForQuery, resultFile);
if (not sinkExpected.has_value())
{
    currentBuilder.setException(sinkExpected.error());
}
// no return — falls through to UB:
const auto newOperator = SinkLogicalOperator{sinkExpected.value()};
```

## Test plan
- [ ] Decide on the correct behavior when adding a sink with an already-existing name (error, overwrite, or idempotent return of existing)
- [ ] Implement that behavior in `addSinkDescriptor`
- [ ] Check return value of `registerSink` in `createSink`
- [ ] Add early return after `setException` in `setNamedSink`

zeuchste:
## Summary

- Change `SinkCatalog::addSinkDescriptor` return type from `std::optional` to `std::expected<SinkDescriptor, Exception>` to distinguish between different failure reasons
- Three distinct error paths now return specific exceptions:
  - Only-digit sink name: `InvalidConfigParameter` (reserved names)
  - Config validation failure: `InvalidConfigParameter` (with sink name and type context)
  - Actual duplicate: `SinkAlreadyExists` (explicit `contains()` check before `emplace`)
- Update `StatementHandler` to propagate the real error instead of always assuming duplicate
- Update `SystestBinder` to propagate the real error from `addSinkDescriptor`
- Fix UB in `setNamedSink`: add early `throw` after `setException` to prevent fall-through to `.value()` on a failed `std::expected`

## Test plan

- [ ] Creating a sink with invalid config reports the validation error, not "sink already exists"
- [ ] Creating a duplicate sink still correctly reports "sink already exists"
- [ ] Systests pass without crashes in `setNamedSink`

Closes #1493
Closes #1504


greenBene (changes requested):
Please rerequest once all test pass

nes-sinks/tests/SinkCatalogTest.cpp
```diff
@@ -0,0 +1,76 @@
...
+        "mySink", schema, "File", {{"file_path", "/tmp/test1.csv"}, {"output_format", "CSV"}}, formatConfig);
+    ASSERT_TRUE(result1.has_value());
+
+    auto result2 = sinkCatalog.addSinkDescriptor(
+        "mySink", schema, "File", {{"file_path", "/tmp/test2.csv"}, {"output_format", "CSV"}}, formatConfig);
+    ASSERT_TRUE(result2.has_value());
+
+    auto descriptor = sinkCatalog.getSinkDescriptor("MYSINK");
+    ASSERT_TRUE(descriptor.has_value());
+}
```
greenBene, line 74: Thanks for adding this test. I don't think that a silent overwrite should be the behavior, but changing that would be out of scope for this PR. I created a BUG ticket. nvmd.

greenBene (approved):

nes-sinks/tests/SinkCatalogTest.cpp
greenBene: Nevermind. This PR explicitly wants to tackle 1504 but doesn't. Please add failure if sink is added with same name of existing sink.

greenBene (changes requested):

yschroeder97 (commented):
What's the status on this? 

It says it fixes #1504, but actually, it just overwrites duplicates. 
We can also close it, since with the new coordinator, everything will be redone.

alepping (changes requested):
The expected changes make a lot of sense. I'm fine with it if you remove the extra test, and replace it with one that addresses: https://github.com/nebulastream/nebulastream/issues/1504. If possible do it via a systest. It does not help at all if a unit test produces the expected error, but its swallowed somwhere and never reaches the end user.

zeuchste:
Closing as discussed: `SinkCatalog` will be redone as part of the new coordinator work, so reworking this PR (which is currently conflicting and would need its unit test replaced with a systest for #1504) is not worth the churn.

For the record, this PR *did* implement actual duplicate detection (explicit `contains()` check returning `SinkAlreadyExists`, plus propagating the real error in `StatementHandler`), not just overwriting — but that behavior should be carried into the new coordinator instead.

The underlying issues remain open and should be addressed there:
- #1493 (sink statement error message)
- #1504 (`SinkCatalog::addSinkDescriptor` silently ignores duplicate sink names)

PR #1507, opened 2026-03-18
Fix(#1494) SQL: support table.column syntax for join field disambiguation

### LINKED ISSUE #1494: SQL parser cannot disambiguate fields with the same name from different join sides
When two join sources share a field name (e.g., both have `timestamp`), there is no SQL syntax to explicitly reference both fields in the outer `SELECT` clause. `SELECT *` works correctly because the engine internally qualifies the fields (producing `STREAM$TIMESTAMP` and `STREAM2$TIMESTAMP`), but no user-facing syntax maps to these internal names.

**Reproducer:** `nes-systests/operator/join/JoinSelectAmbiguousField.test`

```sql
-- Both sources have a 'timestamp' field
CREATE LOGICAL SOURCE stream(id UINT64 NOT NULL, value UINT64 NOT NULL, timestamp UINT64 NOT NULL);
CREATE LOGICAL SOURCE stream2(id2 UINT64 NOT NULL, value2 UINT64 NOT NULL, timestamp UINT64 NOT NULL);

-- Works
SELECT *
FROM (SELECT * FROM stream) INNER JOIN (SELECT * FROM stream2)
ON (id = id2) WINDOW TUMBLING (timestamp, size 1 sec) INTO sink;

-- Fails: both resolve to STREAM$TIMESTAMP, STREAM2$TIMESTAMP is never matched
SELECT start, end, id, value, timestamp, id2, value2, timestamp
FROM ...

-- Fails: dot-qualified names not recognized
SELECT start, end, id, value, stream.timestamp, id2, value2, stream2.timestamp
FROM ...

-- Fails: $ is not valid SQL syntax
SELECT start, end, id, value, STREAM$TIMESTAMP, id2, value2, STREAM2$TIMESTAMP
FROM ...
```

**Impact:** All join systests are forced to use `SELECT *`, making expected output harder to review since the field list is implicit. Standard SQL resolves this with `table.column` syntax in the SELECT clause.

**Expected behavior:** `stream.timestamp` and `stream2.timestamp` (or an equivalent syntax) should resolve to the internally qualified `STREAM$TIMESTAMP` / `STREAM2$TIMESTAMP` fields.

zeuchste:
## Summary
- Adds support for the `table.column` dereference syntax in SQL queries (e.g., `stream.timestamp`, `stream2.timestamp`) to disambiguate fields with the same name from different join sides
- Handles the existing ANTLR `dereference` grammar rule that was defined but never implemented in the listener
- Works in SELECT, WHERE/HAVING, JOIN ON, and GROUP BY contexts

## Test plan
- [ ] New systest `JoinSelectAmbiguousField.test` verifies `stream.timestamp` and `stream2.timestamp` resolve correctly in a join query
- [ ] CI smoke tests pass
- [ ] Existing join systests still pass

Closes #1494


adrianmichalke (approved):

keyseven123 (changes requested):
I just stumbled upon this. Please rebase this onto `lr/schema_stable`, as @Artraxon asked for.

ls-1801 (changes requested):
sorry i am lost, this PR appears to be doing something completly different

PR #1514, opened 2026-03-19
Nix flake with libcxx and sanitizers

adrianmichalke:
Add support for `USE_SANITIZER` and `USE_LIBCXX_IF_AVAILABLE` for the nix build.

Also minor improvements:
- add nix format support
- fix patch alignment with vcpkg
- general cleanup


ls-1801 (approved):
I don't have a working Nix setup anymore, but this code is not affecting anyone else, so I don't want to block it. All I can see is that it hasn't been tested in the CI, so it will break immediately if someone adds new dependencies. On the other hand I wouldnt want anyone do have to deal with nix and getting the sanitizer to work.

keyseven123 (approved):

PR #1521, opened 2026-03-23
Medical tests and VarSizedToNumeric Function

tim-nick:
This pr introduces a new Function called VarSizedToNumeric and some new test for our medical use case.
The Function allows to parse VarSized data to Numeric data types, e.g. FLOAT64 UINT32.

Syntax:
`CAST(data AS DataType)`

The newly added test, checks whether the system can parse and handle recorded real world medical data.

## Verifying this change
This change is tested by our ci newly added tests

VarSizedToNumericLogicalFunction
VarSizedToNumericPhysicalFunction


ls-1801 (changes requested):
Shouldn't this just be `SELECT CAST(text_field as int)`? I don't think we need a special case for VarSized.

ls-1801:
@keyseven123 if it's purpose-built for the medical use case, it should be a plugin; if it's a general-purpose tool, it should follow the SQL standard. 
And yes i am only talking about the QueryAPI the lowering can use what ever it wants

ls-1801 (changes requested):
Re-request on answer

keyseven123:
> @keyseven123 if it's purpose-built for the medical use case, it should be a plugin; if it's a general-purpose tool, it should follow the SQL standard. And yes i am only talking about the QueryAPI the lowering can use what ever it wants

It is not purpose built for the medical use-case. I have rewritten the code so that it works with  `CAST(field as u64)`.

Artraxon (changes requested):
Fine with me, just get the CI green and update the schema branch as i mentioned in zulip

ls-1801 (approved):

nes-common/include/Identifiers/Identifier.hpp
```diff
@@ -579,7 +579,21 @@ IdentifierBase<Extent>::operator IdentifierListBase<std::dynamic_extent>() const
...
+        std::string result;
+        for (const auto& identifier : identifierList.identifiers)
+        {
+            if (!result.empty())
+            {
+                result += '.';
+            }
+            result += identifier.getOriginalString();
+        }
+        return reflect(result);
```
Artraxon: why change this? this changes the semantics because we are losing the the information of whether it was quoted or not, so  original != unreflect(reflect(original)) which is an invariant we should always aim for

nes-logical-operators/include/Functions/VarSizedToNumericLogicalFunction.hpp
```diff
@@ -0,0 +1,88 @@
...
+    [[nodiscard]] std::string explain(ExplainVerbosity verbosity) const;
+
+    [[nodiscard]] DataType getTargetType() const;
+
+private:
+    static bool isSupportedNumericType(DataType::Type type);
+
+    DataType outputType;
+    DataType targetType;
+    LogicalFunction child;
```
Artraxon: i am not sure if I understand the interaction between output type and target type correctly. Is the targetType what the user requested and output what is actually possible? if so, targetType should be renamed to something like "requestedType"

nes-logical-operators/src/Functions/VarSizedToNumericLogicalFunction.cpp
```diff
@@ -0,0 +1,167 @@
...
+        default:
+            return false;
+    }
+}
+
+LogicalFunction VarSizedToNumericLogicalFunction::withInferredDataType(const Schema<Field, Unordered>& schema) const
+{
+    const auto newChildren = getChildren()
+        | std::views::transform([&schema](auto& currentChild) { return currentChild.withInferredDataType(schema); })
+        | std::ranges::to<std::vector>();
```
Artraxon: i would prefer to operator on this->child directly, because we are now doing a whole lot of indirection of converting the child to a vector, applying a map function, collecting it to a vector, and then setting our single child for a vector, which again requires a runtime check. 

You can orient yourself after the other logical functions.

just start with
```c++
auto copy = *this;
copy.child = child.withInferredDataType();
```

nes-logical-operators/src/Functions/VarSizedToNumericLogicalFunction.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+    INVARIANT(
+        newChildren.size() == 1, "VarSizedToNumericLogicalFunction expects exactly one child function but has {}", newChildren.size());
```
Artraxon: this invariant also becomes not needed

nes-logical-operators/src/Functions/CMakeLists.txt
```diff
@@ -21,7 +21,8 @@ add_source_files(nes-logical-operators
 add_plugin(ConstantValue LogicalFunction nes-logical-operators ConstantValueLogicalFunction.cpp)
 add_plugin(FieldAccess LogicalFunction nes-logical-operators FieldAccessLogicalFunction.cpp)
 add_plugin(Concat LogicalFunction nes-logical-operators ConcatLogicalFunction.cpp)
-add_plugin(CastToType LogicalFunction nes-logical-operators CastToTypeLogicalFunction.cpp)
+add_plugin(Cast LogicalFunction nes-logical-operators CastToTypeLogicalFunction.cpp)
```
Artraxon: I would like to keep CastToType to make this work better with the registry (I am surprised this even works), see the discussion in zulip

nes-systests/systest/src/SystestResultCheck.cpp
```diff
@@ -428,19 +428,36 @@ class LineIndexIterator
     size_t totalResultLinesSize = 0;
 };
 
+/// Compare two fields by uppercased name and data type, ignoring case-sensitivity differences.
+/// The CSV header format does not preserve case-sensitivity information, so re-parsed identifiers
```
Artraxon: is there no way to preserve this information? or maybe its somehow possible to enforce this when creating a CSV output formatter?
I understand that this is kind of out-of-scope for this PR but this function feels like a big footgun where we might accept incorrect data as valid for way too long.

Artraxon (changes requested):
Looks good in general, only some minor comments

nes-logical-operators/include/Functions/VarSizedToNumericLogicalFunction.hpp
keyseven123: I got rid of the outputType, and there is now only a targetType

nes-logical-operators/src/Functions/VarSizedToNumericLogicalFunction.cpp
keyseven123: Done

nes-logical-operators/src/Functions/VarSizedToNumericLogicalFunction.cpp
keyseven123: Removed

nes-logical-operators/src/Functions/CMakeLists.txt
keyseven123: Renamed it and also needed to rename it in the query api

nes-common/include/Identifiers/Identifier.hpp
keyseven123: Don't know why I needed it. Do nt need this change anymore and thus, I have dropped it.

nes-systests/systest/src/SystestResultCheck.cpp
keyseven123: Don't know why I needed it. Do nt need this change anymore and thus, I have dropped it.

Artraxon (approved):

PR #1530, opened 2026-03-28
feat(Pipe): add pipe sink and source plugin for inter-query data flow

ls-1801:
Pipe Sink and Pipe Source for explicit query sharing.
Creating a query with a named pipe will backpressure sources until a consumer is connected.
Creating a consumer without a producer will cause the consumer query to fail. (Otherwise, we wouldn't be able to distinguish producer queries that have not started yet or that have ended already)
Attaching new consumer queries will only occur at sequence-number chunk boundaries if there are no more out-of-order buffers in flight. This guarantees that the consumer query will not have any gaps in the received data. Data may still arrive out of order after attaching.


nes-common/include/ExceptionDefinitions.inc
```diff
@@ -82,6 +82,8 @@ EXCEPTION(RunningRoutineFailure, 4001, "error in running routine of SourceThread
 EXCEPTION(CannotOpenSource, 4002, "failed to open a source")
 EXCEPTION(FormattingError, 4003, "error during formatting")
 EXCEPTION(CannotOpenSink, 4004, "failed to open a sink")
+EXCEPTION(PipeSchemaMismatch, 4005, "pipe source and sink schemas do not match")
+EXCEPTION(PipeAlreadyRegistered, 4006, "pipe sink already registered for this name")
```
ls-1801: Remove these and use the broader source and sink exception. we dont want plugins to register custom errors

nes-plugins/PipeService/PipeService.cpp
```diff
@@ -0,0 +1,130 @@
...
+                    /// The sink will activate this queue at the next sequence boundary.
+                    entry.sinkHandle->queues.withWLock([&](auto& q) { q.pending.push_back(queue); });
+                    NES_INFO("PipeService: registered source for pipe '{}' (pending activation at next sequence boundary)", pipeName);
+                }
+                else
+                {
+                    /// Sink was registered but is gone (unregistered) — this shouldn't happen
+                    /// since we erase the PipeEntry on unregisterSink.
+                    throw CannotOpenSource("Pipe sink for '{}' is not active", pipeName);
+                }
```
ls-1801: make this an invariant

PR #1531, opened 2026-03-28
Update nautilus dependency to v0.1 branch

PhilippGrulich:
Rebase nebula stream on latest version of nautilus v0.1 branch:

Get:
Hoisting Alloca instructions will remove operations from the blocks operations vector during the SSA phase which invalidates any indices into the operations list. This will cause operations like ReturnOp to have an invalid index, and crash.

I created some testcases which stress val in more complex control flow tests, they will crash with segfaults, or assertions if you build with a hardened standard library.

The fix replaces ALLOCA with ALLOCA tombstones during the SSA Phase which are removed once the phase completes.

AllocaOperation was the only IR operation that didn't pass an OperationIdentifier to its base class, silently defaulting every alloca to identifier {0,0}. When a function contained multiple val of different sizes, their MLIR values all stored under the same frame key and the last one overwrote all previous ones — so every struct pointer resolved to the same allocation. This caused invoke calls on larger structs to write into the smaller struct's stack slot, producing silent stack corruption or a segfault at runtime.


ls-1801 (approved):

keyseven123 (commented):
In general, I am fine with setting the nautilus deps to a specific version.  @PhilippGrulich could you maybe use a newer version as the current is missing couple of commits compared to our current state

keyseven123 (changes requested):

PhilippGrulich:
The v0.1 branch gets all bug fixes and works with the current NebulaStream api. The current main has some braking changes, for these I will eventually port nebula stream.

keyseven123:
> The v0.1 branch gets all bug fixes and works with the current NebulaStream api. The current main has some braking changes, for these I will eventually port nebula stream.

I am not saying that we should use the current main, but maybe we can bump v0.1 to the commit that we currently use, i.e., `85c19150d9bb7e8cca3051e67e04e76045462379`.

PR #1545, opened 2026-04-01
feat(Plugins): Enable building and testing with disabled plugins

keyseven123:

We make every optional plugin independently toggleable at configure time, and we keep the build and the test suite green for any combination of disabled plugins.

## Verifying this change

- The new nightly workflows exercise each plugin disabled individually, all plugins disabled, and each plugin enabled in isolation.
- No plugin currently ships a `systests/` folder, so `PLUGIN_SYSTEST_DIRS` is empty and the group tags do the actual isolation work. The auto-registration path is therefore not yet exercised.

- Build system, plugin options and `PluginRegistrationUtil.cmake`
- Systest runner, disable-config generation and test discovery
- CI, large tests now run via CTest, two new nightly workflows

## Documentation

- `docs/guide/extensibility.md`, plugin CMake options and auto-registration.
- `docs/development/systests.md`, plugin-dependent tests and the group-tag convention.


alepping (commented):
@keyseven123 
In general, I really like this change, enabling plugin-specific systests (I ran into the need for exactly this functionality yesterday).
However I wonder, what if a systests relies on multiple plugins, e.g., Generator and Checksum? I don't necessarily think its the best idea to always move systests into the directory of ONE specific plugin.
It becomes this awkward `nes-plugins/Sinks/ChecksumSink/systests_generator/sources` structure, which is extremely unintuitive imo.
Also, I don't like the fragmentation of systests. I really don't think moving our benchmarks to the Checksum sink plugin is a good idea.
In general, optional systests in plugins make sense imo, and its a good feature. However, I'd prefer if it was also possible to optionally scan certain systests (in nes-systests), depending on whether a plugin is toggled or not.
Alternatively, we move the checksum sink into nes-sinks (where it probably belongs anyway). This still does not solve the multi-plugin-dependency issue.

ls-1801 (changes requested):
A recent pr introduced this file that sets the default systest groups. Maybe this file can just be populated based on the enabled tests.

As @alepping pointed out, this tree-based system does not work for tests that use multiple plugins.
However, flagging the test as groups, we can only enable checksum + generator tests if both plugins are available.

keyseven123:
@ls-1801 and @alepping , I have moved the generator source and the checksum sink out of the plugins. Also, I have integrated your feedback. Let me know what you think

ls-1801 (changes requested):
I want the systest `-t flag` change, but I don't think we have agreed on how we handle plugins yet, or am I missing something?

keyseven123:
@ls-1801 and @alepping I have updated and rebased this pr. I would really like to merge these changes, at least building and running the tests with all plugins disabled. I have moved the `GeneratorSource`, `ChecksumSink`, and `VoidSink` out of `nes-plugins`.
If there are new `.test` files that require multiple plugins, they should be put under `nes-plugins/Multiple`.

keyseven123:
@ls-1801 The `-t` flag change is now #1845, on its own.

To keep it reviewable in isolation we split this PR into a stack, and this one now carries only the plugin isolation work.

1. #1842, GeneratorSource becomes a built-in `nes-sources` source
2. #1843, ChecksumSink becomes a built-in `nes-sinks` sink
3. #1844, VoidSink becomes a built-in `nes-sinks` sink
4. #1845, `-t`/`--testLocations` accepts multiple locations
5. this PR, which now targets #1845

One deliberate difference to what was here before. In #1845 a directory passed via `-t` keeps the semantics it has on `main`, it replaces the default discover directory instead of extending it. As a result, this PR lists the core test tree explicitly next to the plugin directories via `SYSTEST_LOCATIONS`.

PR #1548, opened 2026-04-01
ci: dynamic test matrix with parallel test splitting

ls-1801:
Replace the static build_test.yml matrix ( jobs,118m critical path)
 a matrix by .github/ci/matrix.json.

- Add .githubci/matrix.json: declarative for build/test combos- Add ./ci/generate_matrix.py: generates GitHub Actions matrix JSON- Refactor build_and_test.yml: generateatrix job feeds matrix,
 test steps gated by run_unit/run_sest/run_docker flagsKey changes to the test matrix:
- Move sanitizers from Debug to RelWithDebInfo (x faster)
- Run sanitizers only with std (libcxx for asan/ubsan,stdcxx for ts)
- Split slow configs (>38m) parallel workers- Keep fast configs (<38) combined in single workersResult:18 workers, ~38m critical path, ~491 compute-minutes(was:21 workers,118m critical path, ~1070 compute-minutes)


keyseven123 (changes requested):
I think it would be a good idea to have the following jobs `Default`, `Sanitizer`, `Log Level tests` so that we can reduce the clutter in the requried checks.

ls-1801:
@keyseven123 I added a "Done" job that succeeds all build_and_test jobs. It will only succeed if all of them have succeeded.  This means the branch protection only needs to wait for `Run build and test / Done`

keyseven123 (approved):

.github/ci/nightly_matrix.yaml
```diff
@@ -0,0 +1,154 @@
...
+    arch: [x64]
+    stdlib: [libcxx, libstdcxx]
+    build_type: Debug
+    sanitizer: none
+    runner: github-{arch}
+    tests: [systest, docker]
+
+  - name: "{arch} {stdlib} {build_type} (docker)"
+    arch: arm64
+    stdlib: libcxx
```
greenBene, line 65: Why is x64 (line 56) in a list, but arm64 (line 64) isn't?

greenBene (approved):
Thank you for implementing this! 

Just had a minor question for understanding, but nothing blocking :)

.github/ci/nightly_matrix.yaml
ls-1801: There is no reason; the YAML parser supports both, i guess in the past there might have been an arm64 aswell

PR #1549, opened 2026-04-02
chore(Systest): log config error msgs

alepping:
Prior, the systeststarter would swallog config errors, printing a stacktrace without any valuable information.
This is obviously super frustrating, given that a minor type/config error triggers this problem and then requires intense analysis/debugging, instead of quickly scanning the error log.

You can reproduce it by setting `--worker.query_engine.query_engine.number_of_worker_threads=16` as config (contains query_engine twice)

### DIFF
--- a/.github/workflows/nix-build.yml
+++ b/.github/workflows/nix-build.yml
@@ -11,7 +11,7 @@ on:
 jobs:
   nix-build:
     name: "Nix Build"
-    timeout-minutes: 60
+    timeout-minutes: 90
     runs-on: [ ubuntu-latest ]
     steps:
       - name: Checkout repository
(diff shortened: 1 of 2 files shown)


ls-1801 (changes requested):
We currently use this patterns for our top level exception handlers:

```c++
     CPPTRACE_TRY
      {
          // ... entire main body ...
      }
      CPPTRACE_CATCH(...)
      {
          NES::tryLogCurrentException();
          return NES::getCurrentErrorCode();  // or just `return 1;`
      }
```

greenBene (approved):
This change makes the systests easier to use! :) Thank you!

ls-1801 (approved):

PR #1550, opened 2026-04-02
fix(Rust): use forked corrosion for proper incremental builds

ls-1801:
## Summary
- Switches corrosion from upstream v0.5.2 to a [fork based on v0.6.1](https://github.com/ls-1801/corrosion/tree/fix/avoid-always-dirty-targets) that adds `CORROSION_NO_HOSTBUILD` support
- With this option, corrosion uses `add_custom_command` with depfile tracking instead of `add_custom_target`, so ninja correctly reports "no work to do" when Rust sources haven't changed
- Previously, every `ninja` invocation would re-run cargo and relink ~50 targets even when nothing changed

See: https://github.com/corrosion-rs/corrosion/issues/624

## Test plan
- [x] Clean configure + full build succeeds
- [x] Second `ninja` invocation reports `ninja: no work to do.`
- [x] Touching a Rust source file triggers cargo rebuild + relinking
- [x] Third `ninja` invocation after rebuild reports `ninja: no work to do.`

### DIFF
--- a/cmake/EnableRust.cmake
+++ b/cmake/EnableRust.cmake
@@ -41,10 +41,11 @@ if (${USING_LIBCXX})
 endif ()
 
 include(FetchContent)
+set(CORROSION_NO_HOSTBUILD ON CACHE BOOL "Enable proper incremental builds for Rust targets")
 FetchContent_Declare(
         Corrosion
-        GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
-        GIT_TAG v0.5.2
+        GIT_REPOSITORY https://github.com/nebulastream/corrosion.git
+        GIT_TAG v0.6.1-always-dirty-fix
 )
 
 FetchContent_MakeAvailable(Corrosion)
--- a/cmake/EnableRust.cmake
+++ b/cmake/EnableRust.cmake
@@ -41,10 +41,11 @@ if (${USING_LIBCXX})
 endif ()
 
 include(FetchContent)
+set(CORROSION_NO_HOSTBUILD ON CACHE BOOL "Enable proper incremental builds for Rust targets")
 FetchContent_Declare(
         Corrosion
-        GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
-        GIT_TAG v0.5.2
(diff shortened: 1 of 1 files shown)


greenBene (approved):
Confirmed behavior on my machine.

Thank you for fixing this :)

cmake/EnableRust.cmake
```diff
@@ -41,10 +41,11 @@ if (${USING_LIBCXX})
 endif ()
 
 include(FetchContent)
+set(CORROSION_NO_HOSTBUILD ON CACHE BOOL "Enable proper incremental builds for Rust targets")
 FetchContent_Declare(
         Corrosion
-        GIT_REPOSITORY https://github.com/corrosion-rs/corrosion.git
-        GIT_TAG v0.5.2
+        GIT_REPOSITORY https://github.com/ls-1801/corrosion.git
```
yschroeder97, line 47: IMO the fork should probably live under the nebulastream project instead of your personal GH profile.

cmake/EnableRust.cmake
```diff
@@ -41,10 +41,11 @@ if (${USING_LIBCXX})
...
+        GIT_TAG fix/avoid-always-dirty-targets
```
yschroeder97, line 48: Nitpick: you should tag the commit hash instead of the branch.

nes-input-formatters/include/InputFormatterProvider/InputFormatterTupleBufferRef.hpp
```diff
@@ -31,7 +31,6 @@
 #include <Arena.hpp>
 #include <ErrorHandling.hpp>
 #include <ExecutionContext.hpp>
-#include <PhysicalOperator.hpp>
```
yschroeder97: The remaining changes are unrelated, pls move to different PR.

yschroeder97 (changes requested):
Thanks for the change, works.

But I would argue that the fork should be under the NES umbrella, and we should pin the hash instead of the branch.

Artraxon (dismissed):
rerequest once yanniks stuff is addressed, but other than that fine

yschroeder97 (approved):

PR #1555, opened 2026-04-06
fix(Formatter): break dependency cycle

ls-1801:
refactor: fold nes-input-formatter-provider into nes-input-formatters to break circular dependency


yschroeder97 (approved):

alepping (changes requested):
If I'm not completely mistaken, the inputformatter refactor solves this issue too.

alepping (changes requested):
removing nes-physical-operators from the 'nes-input-formatters' library is actually still relevant

PR #1558, opened 2026-04-07
Improve CLI YAML parsing placement error messages

ls-1801:
Summary

- Improved YAML error messages: Errors now report the exact YAML path where the problem occurred (e.g. workers[0].bogus_key), suggest valid keys, and distinguish between missing required keys and unknown keys.
- Clearer placement errors: Operator placement failures now show human-readable operator names (e.g. SOURCE(GENERATOR_SOURCE)) instead of opaque internal IDs, and include the relevant worker host.
- Errors always visible to the user: Previously, when running nebucli without -d, errors were only written to nes-cli.log and invisible on the terminal. Errors and invariant violations now also print to stderr
  when console logging is disabled.
- 15 new offline bats tests covering error message quality for invalid keys, missing fields, unknown sources/sinks, schema mismatches, and more.


greenBene (approved):
I think this change makes the cli frontend way more user friendly! Thank you!

nes-frontend/apps/cli/CLIStarter.cpp
```diff
@@ -244,16 +338,10 @@ struct convert<NES::CLI::WorkerConfig>
...
-        {
-            rhs.maxOperators = node["max_operators"].as<size_t>();
-        }
-        if (node["downstream"].IsDefined())
-        {
-            rhs.downstream = node["downstream"].as<std::vector<std::string>>();
-        }
-        rhs.host = node["host"].as<std::string>();
-        rhs.dataAddress = node["data_address"].IsDefined() ? node["data_address"].as<std::string>() : "";
+        rhs.maxOperators = getOrDefault<size_t>(node, "max_operators", 0);
```
tekdogan: Making default value zero now changes the worker capacity from unlimited to zero-capacity.
Since `maxOperators` is `std::optional<size_t>`, later on in `StatementHandler.cpp:353`, it gets treated as having any value present. So omitting `max_operators` yields `Limited{0}` instead of `Unlimited{}`. Not sure if this is what you want actually.

nes-frontend/src/Statements/StatementHandler.cpp
```diff
@@ -196,7 +197,7 @@ std::expected<CreateSinkStatementResult, Exception> SinkStatementHandler::operat
     {
         return CreateSinkStatementResult{created.value()};
     }
-    return std::unexpected{SinkAlreadyExists(statement.name)};
+    return std::unexpected{InvalidConfigParameter("Invalid sink configuration: name='{}' type='{}'", statement.name, statement.sinkType)};
```
tekdogan: You can consider propogating the real sink creation issue to be specific about the error, just like how you do it for sources at `SourceStatementHandler::operator()` (line 88).

nes-frontend/apps/cli/CLIStarter.cpp
```diff
@@ -163,6 +163,94 @@ struct QueryConfig
...
+
+template <typename T>
+T getValue(const YAML::Node& node, const std::string& key)
+{
+    YamlPathGuard guard(key);
+    if (!node[key])
+    {
+        throw NES::InvalidConfigParameter("Missing required key '{}' at {}{}", key, currentYamlPath(), formatMark(node.Mark()));
+    }
+    return node[key].as<T>();
```
tekdogan, line 209: Type mismatches can still bypass the contextual error handling because of `node[key].as<T>()`.

Here is the yaml file for you to reproduce:
```yaml
query: |
  SELECT
    *
  FROM
    GENERATOR_SOURCE
  INTO
    VOID_SINK

sinks:
  - name: VOID_SINK
    host: worker-1:8080
    schema:
      - name: GENERATOR_SOURCE$DOUBLE
        type:
          primitive: FLOAT64
    type: Void

logical:
  - name: GENERATOR_SOURCE
    schema:
      - name: DOUBLE
        type: FLOAT64

physical:
  - logical: GENERATOR_SOURCE
    host: worker-1:8080
    parser_config:
      type: CSV
      fieldDelimiter: ","
    type: Generator
    source_config:
      generator_rate_type: FIXED
      generator_rate_config: emit_rate 10
      seed: 1
      generator_schema: |

workers:
  - host: worker-1:8080
    data_address: worker-1:9090
    max_operators: 10000

```

I get a generic conversion error. You could catch `YAML::Exception`

tekdogan:
```suggestion
    try
    {
        return node[key].as<T>();
    }
    catch (const YAML::Exception&)
    {
        throw NES::InvalidConfigParameter(
            "Invalid value at {}{}: expected {}",
            currentYamlPath(),
            formatMark(node[key].Mark()),
            /* ... */);
    }
```

tekdogan (commented):
Good changes in general. See my suggestions below before merging

tekdogan (approved):

PR #1567, opened 2026-04-22
chore(InputFormatter): eliminate templating

alepping:
Replaces templating and type-erasure-based approach with inheritance removes several layers of complexity.

Prior to these changes, @ls-1801, @keyseven123, @LeoRaasch and others reported that they struggled with the complexity of the InputFormatter implementation, especially because of the use of templating.
This PR:
- removes class templating from nes-input-formatters entirely, which cascades through almost all files, significantly simplyfing the implementation
  - no more InputFormatterTupleBufferRef and InputFormatter (type-erasure)
  - no more indirection concerning implementations (ctrl+click works)
  - no more CRTP interfaces (just inheritance)
  - no more template overloading (just inheritance)
  - no more registry shenanigans to work with templating (just inheritance like sources/sinks)
  - no more Metadata structs
  - no more Concepts.hpp
- simplifies interfaces significantly
  - merges 'hidden' Metadata interface (was concept) into InputFormatIndexer interface (pure), while still keeping interface functions to only 2
  - creates FieldIndexFunction interface (pure) with just 2 functions
- greatly simplifies InputFormatter header (is now extremely lean)
  - no more templating allows to move everything into .cpp
  - pulls multiple private member functions into anonymous namespace
  - minor restructuring
  
All of this, as far as I can see, comes at the cost of only ~3 extra virtual function calls (indexRawBuffer) per execute call/buffer. Two for leading/trailing tuples and one for the raw buffer itself (assumes compiled execution).
Based on prior experiments, the overhead should be negligible, but @LeoRaasch will run concrete experiments to confirm.

## Experiment Results
To evaluate the refactored InputFormatter, we ran our large benchmark queries with 50GB of data per query and evaluated the end-to-end throughput. We used a MonitoringSink, which measures the end-to-end throughput of a query based on source-ingestion timestamps of the buffers. No output formatting was applied. We used the TCPSource to receive the incoming data. The datasets were sent repeatedly until 50GB were transmitted. For every iteration, the individual timestamps were incremented by the highest timestamp of the original dataset. We used 16 worker threads and evaluated the benchmarks for 16kb, 32kb, and 65kb buffers.

Across all buffer sizes and benchmark queries, there were no notable differences between the end-to-end throughput of the current main branch and the refactored input-formatter branch. The measured throughputs for all benchmarks and corresponding line charts can be obtained from tubcloud: https://tubcloud.tu-berlin.de/s/9xQMPXDzG66tbnP


nes-input-formatters/private/FieldOffsetFIF.hpp
```diff
@@ -0,0 +1,161 @@
...
+            const auto& fieldName = csvIndexer.getFieldNameAt(i);
+            const auto& fieldDataType = csvIndexer.getFieldDataTypeAt(i);
+            if (not includesField(projections, fieldName))
+            {
+                continue;
+            }
+
+            const auto numPriorFields = recordIndex * nautilus::static_val(csvIndexer.getNumberOfFields() + 1);
+            const auto recordOffsetAddress = indexBufferPtr + (numPriorFields + i);
+            const auto recordOffsetEndAddress = indexBufferPtr + (numPriorFields + i + nautilus::static_val<uint64_t>(1));
```
LeoRaasch: I think the naming choice here is a bit misleading, as these are the addresses for the offsets of the field. So maybe "fieldOffsetAddress".

LeoRaasch (approved):
Looks very nice, I find this version to be way more intuitive than what we had previously. Regarding the performance, I will start running experiments now.

LeoRaasch:
> Looks very nice, I find this version to be way more intuitive than what we had previously. Regarding the performance, I will start running experiments now.

Experiments were conducted. Results were integrated into the PR description.

nes-input-formatters/src/FieldOffsetFIF.cpp
```diff
@@ -0,0 +1,97 @@
...
+FieldOffsetFIF::hasNext(const nautilus::val<uint64_t>& tupleIdx, const nautilus::val<FieldIndexFunction*>& fieldIndexFunction) const
+{
+    nautilus::val<uint64_t> numTuples = *getMemberWithOffset<size_t>(fieldIndexFunction, offsetof(FieldOffsetFIF, totalNumberOfTuples));
+    return tupleIdx < numTuples;
+}
+
+const FieldIndex* FieldOffsetFIF::getIndexValuesProxy(const FieldIndexFunction* fieldIndexFunction)
+{
+    /// At this point, the index values are fixed, thus the vector cannot grow anymore and accessing the pointer is safe
+    const auto fieldOffsets = static_cast<const FieldOffsetFIF* const>(fieldIndexFunction);
```
Artraxon: are all of these downcasts safe? IMO they all should be dyncasts so that at least they fail when they try to deref the nullptr instead of causing UB

nes-input-formatters/include/InputFormatters/FieldIndexFunction.hpp
```diff
@@ -28,59 +28,33 @@
 namespace NES
 {
 
-/// CRTP Interface that enables InputFormatters to define specialized functions to access fields that the InputFormatter can call directly
-/// (templated) without the overhead of a virtual function call (or a lambda function/std::function)
+/// Forward ref to brake circular dependency between InputFormatIndexer (which uses FieldIndexFunction) and FieldIndexFunction
```
ls-1801: Typo: `brake` should be `break` (`Forward ref to break circular dependency...`).

nes-input-formatters/include/InputFormatters/InputFormatIndexer.hpp
```diff
@@ -14,29 +14,45 @@
...
+
+#include <FieldIndexFunction.hpp>
+#include <RawTupleBuffer.hpp>
+
+#include <RawValueParser.hpp>
 
 namespace NES
 {
 
+/// Returned by 'InputFormatIndexer::indexRawBuffer'. Bundles the FIF together with the per-buffer metadata the InputFormatter needs once
```
ls-1801: The 'Keeping this metadata here — rather than on FieldIndexFunction — means only ...' clause is refactor narration that doesn't help a reader who lands on `IndexRawBufferResult` cold. The first sentence already says what the struct is; suggest dropping the design-justification trailer.

nes-input-formatters/private/CSVInputFormatIndexer.hpp
```diff
@@ -26,97 +27,51 @@
...
-struct CSVMetaData
+class CSVInputFormatIndexer : public InputFormatIndexer
 {
+public:
     static constexpr size_t SIZE_OF_TUPLE_DELIMITER = 1;
     static constexpr size_t SIZE_OF_FIELD_DELIMITER = 1;
 
-    explicit CSVMetaData(const ParserConfig& config, const TupleBufferRef& tupleBufferRef)
+    explicit CSVInputFormatIndexer(const ParserConfig& config, const TupleBufferRef& tupleBufferRef)
         : tupleDelimiter(config.tupleDelimiter.front())
```
ls-1801: The PRECONDITIONs that previously guarded `config.tupleDelimiter.size() == 1` and `config.fieldDelimiter.size() == 1` were dropped when `CSVMetaData` was merged into `CSVInputFormatIndexer`. `config.tupleDelimiter.front()` / `fieldDelimiter.front()` are now called unconditionally in the member initializer list — if either string is empty, that's UB before any check fires. Please re-add the size==1 checks (e.g. via a delegated ctor or a static `validate()`) so the invariant is enforced.

nes-input-formatters/private/CSVInputFormatIndexer.hpp
```diff
@@ -26,97 +27,51 @@
...
 
-    [[nodiscard]] uint64_t getNumberOfFields() const
-    {
-        INVARIANT(fieldNames.size() == fieldDataTypes.size(), "No. fields must be equal to no. data types");
-        return fieldNames.size();
-    }
+    [[nodiscard]] const std::vector<std::string>& getNullValues() const override { return nullValues; }
 
 private:
+    /// CSVMetaData
```
ls-1801: Stray `/// CSVMetaData` comment — `CSVMetaData` no longer exists. Looks like a leftover from the merge.

nes-input-formatters/private/FieldOffsetFIF.hpp
```diff
@@ -0,0 +1,90 @@
...
+        const nautilus::val<int8_t*>& recordBufferPtr,
+        const nautilus::val<uint64_t>& recordIndex,
+        const InputFormatIndexer& indexer,
+        nautilus::val<FieldIndexFunction*> fieldIndexFunction,
+        const TupleBufferRef& bufferRef) const override;
+
+    /// InputFormatter interface functions:
+    void startSetup(size_t numberOfFieldsInSchema, size_t sizeOfFieldDelimiter);
+
+    /// Assures that there is space to write one more tuple and returns a pointer to write the field offsets (of one tuple) to.
```
ls-1801: Refactor-artifact comment: this docstring was inherited from the original FieldOffsets template that returned a pointer for the caller to write multiple offsets into. The new `emplaceFieldOffset` takes one `FieldIndex` at a time and just emplaces it — it doesn't 'return a pointer' or 'assure there is space'. Worth rewriting or removing the comment so it doesn't mislead.

nes-input-formatters/private/FieldOffsetFIF.hpp
```diff
@@ -0,0 +1,90 @@
...
+    /// @Note expects that users of function write 'number of fields in schema + 1' offsets to pointer, manually incrementing the pointer by one for each offset.
+    void emplaceFieldOffset(FieldIndex offset) { this->indexValues.emplace_back(offset); }
+
+    /// Resets the indexes and pointers, calculates and sets the number of tuples in the current buffer, returns the total number of tuples.
```
ls-1801: Docstring claims this 'resets the indexes and pointers' but the body only sets first/last tuple offsets to max. With the new model where each `FieldOffsetFIF` is created fresh per buffer, 'reset' isn't really the right framing anyway. Worth rewording so the comment matches the implementation.

nes-input-formatters/src/CSVInputFormatIndexer.cpp
```diff
@@ -23,99 +25,101 @@
...
-            "Number of parsed fields does not match number of fields in schema (parsed {} vs {} schema",
-            fieldIdx,
-            metaData.getNumberOfFields());
+            "Number of parsed fields does not match number of fields in schema (parsed {} vs {} schema", fieldIdx, numberOfFields);
     }
 }
 
 void initializeIndexFunctionForTuple(
-    NES::FieldOffsets<NES::CSV_NUM_OFFSETS_PER_FIELD>& fieldOffsets,
+    NES::FieldOffsetFIF& FieldOffsetFIF,
```
ls-1801: Parameter name `FieldOffsetFIF` shadows the type `NES::FieldOffsetFIF`. The sibling helper at line 37 uses the correct camelCase `fieldOffsetFIF` — looks like this one was missed during the `FieldOffsets → FieldOffsetFIF` rename. clang-tidy is also flagging the case style on lines 69/76/84/88 for the same reason. Please rename to `fieldOffsetFIF`.

nes-input-formatters/include/InputFormatters/FieldIndexFunction.hpp
```diff
@@ -28,59 +28,33 @@
...
+class InputFormatIndexer;
+
+/// Abstract interface that enables InputFormatters to define specialized functions for iterating over the formatted records of a buffer.
 /// Different possible kinds of FieldIndexFunctions(FieldOffsets(index raw data), Internal, ZStdCompressed, Arrow, ...)
-/// A FieldIndexFunction may also combine different kinds of field access variants for the different fields of one schema
-template <typename Derived>
+/// A FieldIndexFunction may also combine different kinds of field access variants for the different fields of one schema.
+/// Values that the InputFormatter needs once per buffer (offsetOfFirstTuple, offsetOfLastTuple, totalNumberOfTuples) are returned
+/// alongside the FIF by 'InputFormatIndexer::indexRawBuffer' via 'IndexRawBufferResult' — not through this interface.
 class FieldIndexFunction
```
ls-1801: `FieldIndexFunction` is named like a function but is actually a stateful per-buffer indexer. `FieldOffsetFIF` carries `vector<FieldIndex> indexValues`, `totalNumberOfTuples`, the first/last tuple offsets, the field-delimiter size, etc. `SIMDJSONFIF` carries `shared_ptr<parser>`, `shared_ptr<document_stream>`, the iterator, `isAtLastTuple`. There is also redundancy — `totalNumberOfTuples` and the first/last tuple offsets live in both `IndexRawBufferResult` and the FIF.

The immediate consequence is at FieldOffsetFIF.cpp:22 and SIMDJSONFIF.cpp:54, where the FIF reads its own state from traced code via
```
offsetof(FieldOffsetFIF, totalNumberOfTuples)
offsetof(SIMDJSONFIF, isAtLastTuple)
```
added to a base-class pointer (`FieldIndexFunction*`). Since `FieldIndexFunction` now has a virtual destructor + virtual methods, the derived FIFs are non-standard-layout, so `offsetof` on them is conditionally supported per [class.mem]/24 — the `static_assert(std::is_standard_layout_v<SIMDJSONFIF>)` that previously guarded this was deleted in this PR, and clang-tidi may flag `invalid-offsetof`. Today this happens to work because single-inheritance polymorphic classes lay out the vptr at offset 0 with the base subobject overlaying the derived address, so derived-relative offsets equal base-relative offsets — but that's an ABI assumption that's now invisible in the code.

The proper fix is to make `FieldIndexFunction` actually stateless: lift all the per-buffer state into `IndexRawBufferResult` (a plain standard-layout struct) and have `hasNext(state, idx)` / `readSpanningRecord(state, ...)` take it as a parameter. That removes the offsetof gymnastics entirely (the state is already standard-layout, so offsetof is well-defined), eliminates the `totalNumberOfTuples`/first/last-tuple duplication, and matches the name — the FIF would then truly be a function over state instead of a stateful object misnamed as one. If a stateless redesign is out of scope here, at minimum a `// SAFETY:` comment at each `offsetof` call site documenting the single-inheritance vptr-first ABI assumption would help future readers.

nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
```diff
@@ -0,0 +1,97 @@
...
+        const auto fieldAddress = recordBufferPtr + fieldOffsetStart;
+        parseRawValueIntoRecord(
+            fieldDataType, record, fieldAddress, fieldSize, fieldName, indexer.getNullValues(), indexer.getQuotationType());
+    }
+    return record;
+}
+
+void FieldOffsetFIF::startSetup(const size_t numberOfFieldsInSchema, const size_t sizeOfFieldDelimiter)
+{
+    PRECONDITION(
```
ls-1801: The parameter is named `sizeOfFieldDelimiter` and stored into a `sizeOfFieldDelimiter` member — but `CSVInputFormatIndexer.cpp:104` passes `SIZE_OF_TUPLE_DELIMITER`, and the PRECONDITION error string says "tuple delimiter". So either the name is wrong, the call site is wrong, or both. The member is also currently dead (written but never read), so this is harmless today but will become a latent bug once someone consumes it. Please reconcile the naming and either remove the dead member or pass the correct value.

nes-input-formatters/src/FieldOffsetFIF.cpp
```diff
@@ -0,0 +1,97 @@
...
+void FieldOffsetFIF::markNoTupleDelimiters()
+{
+    this->offsetOfFirstTuple = std::numeric_limits<FieldIndex>::max();
+    this->offsetOfLastTuple = std::numeric_limits<FieldIndex>::max();
+}
+
+void FieldOffsetFIF::markWithTupleDelimiters(const FieldIndex offsetToFirstTuple, const FieldIndex offsetToLastTuple)
+{
+    /// Make sure that the number of read fields matches the expected value.
+    const auto [totalNumberOfTuples, remainder] = std::lldiv(this->indexValues.size(), numberOfOffsetsPerTuple);
```
ls-1801: Prefer the idiomatic `/` and `%` over `std::lldiv` here:
```cpp
const auto totalNumberOfTuples = indexValues.size() / numberOfOffsetsPerTuple;
const auto remainder = indexValues.size() % numberOfOffsetsPerTuple;
```
Compilers fuse `a / b` and `a % b` on the same operands into a single `divq`, so codegen is identical. `std::lldiv` adds an unsigned→signed narrowing of `size()`, packaging into `lldiv_t`, and a structured binding that depends on `lldiv_t`'s field order (POSIX-required, not C++-required). No upside, just an unusual idiom.

nes-input-formatters/src/InputFormatter.cpp
```diff
@@ -0,0 +1,505 @@
...
+
+/// Constructs a spanning tuple (string) that spans over at least two buffers (stagedBuffersSpan).
+/// First, determines the start of the spanning tuple in the first buffer to format. Constructs a spanning tuple from the required bytes.
+/// Second, appends all bytes of all raw buffers that are not the last buffer to the spanning tuple.
+/// Third, determines the end of the spanning tuple in the last buffer to format. Appends the required bytes to the spanning tuple.
+/// Lastly, formats the full spanning tuple.
+void processSpanningTuple(
+    const std::span<const StagedBuffer>& stagedBuffersSpan, std::span<char> spanningTupleBuffer, const InputFormatIndexer& indexer)
+{
+    /// If the buffers are not empty, there are at least three buffers
```
ls-1801: Comment says 'at least three buffers' but the INVARIANT requires `>= 2`, and the call sites in `indexTrailingSpanningTupleProxy` / `indexLeadingSpanningTupleAndBufferProxy` (with-delimiter branch) only filter on `< 2`. Looks like a leftover from an earlier draft — please align the comment with the actual invariant.

nes-input-formatters/src/InputFormatter.cpp
```diff
@@ -0,0 +1,505 @@
...
+/// traceable offsets that getMemberWithOffset can read directly.
+struct IndexPhaseResult
+{
+    std::span<char> leadingSpanningTuple{};
+    std::span<char> trailingSpanningTuple{};
+    bool hasLeadingSpanningTupleBool = false;
+    bool hasTupleDelimiter = false;
+    bool isRepeat = false;
+    bool hasValidOffsetOfTrailingSpanningTuple = false;
+    uint64_t numTuplesWithoutTrailing = 0;
```
ls-1801: `numTuplesWithoutTrailing` is set/incremented in three places (lines 185, 202, 206) but never read by anyone. `finalizeTrailingIndexPhase()` exists solely to bump it. Either remove the dead state or hook up the missing consumer.

nes-input-formatters/src/InputFormatter.cpp
```diff
@@ -0,0 +1,505 @@
...
+    static void finalizeTrailingIndexPhase() { tlIndexPhaseResult.numTuplesWithoutTrailing += 1; }
+
+    static void constructAndIndexTrailingSpanningTuple(
+        const std::vector<StagedBuffer>& stagedBuffers,
+        InputFormatIndexer& indexer,
+        Arena& arenaRef,
+        const size_t sizeOfTrailingSpanningTuple)
+    {
+        allocateForTrailingSpanningTuple(arenaRef, sizeOfTrailingSpanningTuple);
+        const auto trailingSpanningTupleBuffers = std::span(stagedBuffers).subspan(0, stagedBuffers.size());
```
ls-1801: `std::span(stagedBuffers).subspan(0, stagedBuffers.size())` is just `std::span(stagedBuffers)`. If no slicing was intended, drop the `subspan` call; if it was, the bounds need fixing.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONInputFormatIndexer.hpp
```diff
@@ -14,35 +14,107 @@
...
     static constexpr char DELIMITER_SIZE = sizeof(char);
     static constexpr char TUPLE_DELIMITER = '\n';
     static constexpr char KEY_VALUE_DELIMITER = ':';
-    static constexpr char FIELD_DELIMITER = ',';
     static constexpr char KEY_QUOTE = '"';
 
-    SIMDJSONInputFormatIndexer() = default;
-    ~SIMDJSONInputFormatIndexer() = default;
+    explicit SIMDJSONInputFormatIndexer(const ParserConfig& config, const TupleBufferRef& tupleBufferRef)
+        : tupleDelimiter(config.tupleDelimiter.front())
```
ls-1801: `tupleDelimiter(config.tupleDelimiter.front())` runs in the member initializer list, before the body `PRECONDITION(config.tupleDelimiter.size() == 1, ...)` is checked. If `config.tupleDelimiter` is empty, `front()` is UB and triggers before the guard fires. Either move the size check ahead of `front()` (delegating ctor or static helper) or make the initializer tolerate empty input and let the body PRECONDITION abort. Same pattern as the CSV indexer (separate comment on `CSVInputFormatIndexer.hpp`).

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONParsingUtil.hpp
```diff
@@ -0,0 +1,324 @@
...
+    auto currentDoc = *simdJsonFif->getDocStreamIterator();
+
+    /// First, we check if the key is not in the doc. If this is the case, we can return true, as this counts as null
+    if (not currentDoc[fieldName].has_value())
+    {
+        return true;
+    }
+
+    /// Second, we need to check if the key is equal to one of the null values
+    if (accessSIMDJsonFieldOrThrow(currentDoc, fieldName).is_null())
```
ls-1801, line 96: `checkIsNullJsonProxy` is declared `noexcept` (line 90), but `accessSIMDJsonFieldOrThrow` on this line can throw `FieldNotFound`. The earlier `has_value()` guard at line 96 catches the missing-key case, but simdjson's ondemand parser is forward-only with consumable iterator state — the second access on line 102 can independently fail (e.g., parser-state error after the first dereference) and propagate out of a `noexcept` function, which calls `std::terminate`.\n\nThis pattern was lifted from the old `SIMDJSONFIF.hpp:74` so it's pre-existing, but a `noexcept` annotation on a function whose body can throw is a footgun regardless of where it came from. Either drop `noexcept` here, or replace the second access with a non-throwing probe (e.g. handle the `simdjson_result` directly without the throwing helper).

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONParsingUtil.hpp
```diff
@@ -0,0 +1,324 @@
...
+
+template <bool Nullable>
+ParsedResultVariableSized*
+parseJsonVarSizedProxy(FieldIndex fieldIndex, FieldIndexFunction* fieldIndexFunction, const InputFormatIndexer* indexer)
+{
+    auto* simdJsonFIF = static_cast<SIMDJSONFIF*>(fieldIndexFunction);
+    const auto* simdIndexer = static_cast<const SIMDJSONInputFormatIndexer*>(indexer);
+
+    /// We use thread_local to ensure that result exists as long as the thread exist.
+    /// We require this, as we return a pointer to the storage, due to use not being able to return two values in nautilus, i.e.,
```
ls-1801: Two typos in this comment block:
- Line 268: "as long as the thread exist" → "as long as the thread exists".
- Line 269: "due to use not being able to return two values" → "due to us not being able...".

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONParsingUtil.hpp
```diff
@@ -0,0 +1,324 @@
...
+        {
+            constexpr auto sizeOfValue = 0;
+            result = ParsedResultVariableSized{.varSizedPointer = nullptr, .size = sizeOfValue, .isNull = true};
+            return &result;
+        }
+    }
+
+    INVARIANT(
+        fieldIndex < simdIndexer->getNumberOfFields(),
+        "fieldIndex {} is out or bounds for schema keys of size: {}",
```
ls-1801: nit/typo: 'out or bounds' → 'out of bounds' in this INVARIANT message. User-visible if the invariant fires.

ls-1801 (changes requested):

nes-input-formatters/private/FieldOffsetFIF.hpp
alepping: makes sense, took your suggestion

nes-input-formatters/src/FieldOffsetFIF.cpp
alepping: the downcast is safe, since it is called from a member function of the specific type.
changed it to a dynamic_cast

nes-input-formatters/include/InputFormatters/FieldIndexFunction.hpp
alepping: done

nes-input-formatters/include/InputFormatters/InputFormatIndexer.hpp
alepping: done

nes-input-formatters/private/CSVInputFormatIndexer.hpp
alepping: done

nes-input-formatters/private/CSVInputFormatIndexer.hpp
alepping: removed

nes-input-formatters/private/FieldOffsetFIF.hpp
alepping: done

nes-input-formatters/private/FieldOffsetFIF.hpp
alepping: done

nes-input-formatters/src/CSVInputFormatIndexer.cpp
alepping: done

nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
alepping: done

nes-input-formatters/src/FieldOffsetFIF.cpp
alepping: done

nes-input-formatters/src/InputFormatter.cpp
alepping: removed comment, since the precondition already comments the code

nes-input-formatters/src/InputFormatter.cpp
alepping: removed numTuplesWithoutTrailing

nes-input-formatters/src/InputFormatter.cpp
alepping: took

nes-input-formatters/include/InputFormatters/FieldIndexFunction.hpp
alepping: Addressed.
The name FieldIndexFunction just didn't make sense (anymore) imo.
I renamed it to BufferIndex.
I then merged IndexBufferResult into BufferIndex, which makes the interface less confusing (imo).
The indexRawBuffer function now returns a 'BufferIndex' which the InputFormatIndexer constructs.
While the implementations of 'BufferIndex' may not be stateless, the interface of 'BufferIndex' is.
I added checks for the ABI dependency and got rid of offsetsof.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONInputFormatIndexer.hpp
alepping: refactored like the CSV version by calling 'create', which applies the check and then constructs

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONParsingUtil.hpp
alepping: removed noexcept

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONParsingUtil.hpp
alepping: fixed

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONParsingUtil.hpp
alepping: fixed

nes-input-formatters/src/FieldOffsetFIF.cpp
Artraxon: but why does the member function receive  a pointer to the instance its called on?

nes-input-formatters/include/InputFormatterProvider/InputFormatterTupleBufferRefProvider.hpp
yschroeder97, line 1: Should rename the file and the provide function as well.

nes-input-formatters/private/FieldOffsetRawBufferIndex.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include <Nautilus/Interface/BufferRef/TupleBufferRef.hpp>
+#include <Nautilus/Interface/Record.hpp>
+#include <InputFormatter.hpp>
+#include <RawBufferIndex.hpp>
+#include <val_arith.hpp>
+#include <val_bool.hpp>
+#include <val_concepts.hpp>
+#include <val_ptr.hpp>
+
+inline bool
```
yschroeder97, line 35: Move inside namespace

nes-input-formatters/include/InputFormatters/RawBufferIndex.hpp
```diff
@@ -0,0 +1,68 @@
...
+/// InputFormatter can iterate over.
+class RawBufferIndex
+{
+public:
+    RawBufferIndex() = default;
+    virtual ~RawBufferIndex() = default;
+
+    /// Traced by the InputFormatter — must be virtual so dispatch picks the correct derived override at trace time.
+    [[nodiscard]] virtual nautilus::val<bool>
+    hasNext(const nautilus::val<uint64_t>& tupleIdx, const nautilus::val<RawBufferIndex*>& rawBufferIndex) const = 0;
```
yschroeder97, line 53: Document or enforce the invariant that this is a pointer to "this".

nes-input-formatters/src/InputFormatter.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+        auto record = getIndexPhaseResult()->trailingSpanningTupleRawBufferIndex->readSpanningRecord(
+            projections, spanningRecordPtr, nautilus::val<uint64_t>(0), indexer, trailingRawBufferIndex, bufferRef);
+        executeChild(executionCtx, record);
+    }
+}
+
+}
+
+std::vector<DataType> InputFormatter::getAllDataTypes() const
```
yschroeder97, line 399: Should narrow the base interface to avoid these kind of unreachable methods.

nes-input-formatters/include/InputFormatters/RawTupleBuffer.hpp
```diff
@@ -30,29 +29,30 @@ namespace NES
 using FieldIndex = uint32_t;
 using SequenceNumberType = SequenceNumber::Underlying;
 
+class RawTupleBuffer;
+
+namespace detail
+{
+/// Defined in InputFormatter.cpp. The only sanctioned way to build a RawTupleBuffer over externally-allocated (typically arena-allocated)
+/// spanning-tuple bytes. The caller owns the underlying memory and must keep it alive for the duration of indexing.
+RawTupleBuffer makeSpanningTupleBuffer(const char* rawDataPtr, std::size_t sizeOfSpanningTuple);
```
yschroeder97: You write that the caller owns the memory, but this is not enforced, because it can still be copied at will because of the existing copy ctor/assign.

nes-input-formatters/registry/include/InputFormatIndexerRegistry.hpp
```diff
@@ -43,15 +42,14 @@ struct InputFormatIndexerRegistryArguments
     {
     }
 
-    const ParserConfig& getInputFormatterConfig() { return inputFormatIndexerConfig; }
+    [[nodiscard]] const ParserConfig& getInputFormatterConfig() const { return inputFormatIndexerConfig; }
+
+    [[nodiscard]] const TupleBufferRef& getInputMemoryProvider() const { return *memoryProvider; }
```
yschroeder97, line 47: Calling the getter after `createInputFormatterWithIndexer` dereferences the moved-from pointer.

nes-input-formatters/include/InputFormatters/InputFormatIndexer.hpp
```diff
@@ -14,29 +14,35 @@
...
-        /// all InputFormatIndexers.
-        static_assert(InputFormatIndexerType<T>);
-    }
+public:
+    explicit InputFormatIndexer() = default;
+    virtual ~InputFormatIndexer() = default;
+
+    [[nodiscard]] virtual std::unique_ptr<RawBufferIndex> indexRawBuffer(const RawTupleBuffer& rawBuffer) const = 0;
+
+    [[nodiscard]] virtual std::string_view getTupleDelimitingBytes() const = 0;
```
yschroeder97, line 43: In the PR desc you claim only 2 virtual functions ;)

nes-physical-operators/src/ScanPhysicalOperator.cpp
```diff
@@ -37,13 +37,13 @@ ScanPhysicalOperator::ScanPhysicalOperator(
...
     , projections(std::move(projections))
-    , isRawScan(std::dynamic_pointer_cast<InputFormatterTupleBufferRef>(this->bufferRef) != nullptr)
+    , isRawScan(std::dynamic_pointer_cast<InputFormatter>(this->bufferRef) != nullptr)
 {
 }
 
 void ScanPhysicalOperator::rawScan(ExecutionContext& executionCtx, RecordBuffer& recordBuffer) const
 {
-    auto inputFormatterBufferRef = std::dynamic_pointer_cast<InputFormatterTupleBufferRef>(this->bufferRef);
+    auto inputFormatterBufferRef = std::dynamic_pointer_cast<InputFormatter>(this->bufferRef);
```
yschroeder97, line 46: You could cache the cast as a member.

yschroeder97 (changes requested):
Really a simplification, thanks for the effort :)

You need to fix the README though, it is not in sync with the new structure.

nes-input-formatters/include/InputFormatterProvider/InputFormatterTupleBufferRefProvider.hpp
alepping: done

nes-input-formatters/private/FieldOffsetRawBufferIndex.hpp
alepping: done

nes-input-formatters/include/InputFormatters/RawBufferIndex.hpp
alepping: I assume you mean the 'rawBufferIndex'?

alepping: Added comments that explain why passing the rawBufferIndex is necessary and why we cannot simply use 'this' inside of the functions

nes-input-formatters/src/InputFormatter.cpp
alepping: @yschroeder97 
I totally agree. Imo, the TupleBufferRef interface requires significant overhaul, which is part of my todo list for refactoring data types.
That goes way beyond this PR though.
JFYI:
The idea is that we have two interfaces (read/write), which the current code already benefits from (also see OutputFormatter).
Furthermore, layouts should also cover data types or in general data structure representations.
Finally, I don't think a record-based approach is the cleanest solution if we aim to consume column and pax based layouts in the future.

nes-input-formatters/include/InputFormatters/RawTupleBuffer.hpp
alepping: Mhh...but the ability to copy ctor/assign does not change the correctness of the statement?
The caller still owns the underlying memory and must keep it alive.
This comment is specific for the function, which expects a rawDataPtr and a size. The comment simply states that it will not allocate memory for the caller, but the caller needs to provide it.
Tbh, its kind of obvious from the function signature, so if you prefer, I can simply remove the comment.

alepping: I refactored this function, since it did not really do what I wanted anyway.
The main idea was to keep it private, since many classes have access to RawTupleBuffer that should never use this specific constructor (from ptr and size).

nes-input-formatters/registry/include/InputFormatIndexerRegistry.hpp
alepping: Removed the 'std::move' even though noone should call 'get' after 'createInputFormatterWithIndexer', since it is tied to the registry argument, which is constucted by the provider and used once to construct an InputFormatter.
Still, to provide future misuse and since its not performance critical, I removed the move.

nes-input-formatters/include/InputFormatters/InputFormatIndexer.hpp
alepping: Yeah...I had to relax that. I kind of cheated with a helper struct beforehand. Sorry for the lies.

nes-physical-operators/src/ScanPhysicalOperator.cpp
alepping: I like to avoid state if possible.
In this case, rawScan is called once (in compiled mode) to trace code, so I don't think caching is worth it here.

nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
```diff
@@ -0,0 +1,125 @@
...
+        const auto sizeOfDelimiter = (i + 1 == numberOfFields) ? 0 : indexer.getFieldDelimitingBytes().size();
+        const auto fieldSize = fieldOffsetEnd - fieldOffsetStart - sizeOfDelimiter;
+        const auto fieldAddress = recordBufferPtr + fieldOffsetStart;
+        parseRawValueIntoRecord(
+            fieldDataType, record, fieldAddress, fieldSize, fieldName, indexer.getNullValues(), indexer.getQuotationType());
+    }
+    return record;
+}
+
+void FieldOffsetRawBufferIndex::startSetup(const size_t numberOfFieldsInSchema, const size_t sizeOfFieldDelimiter)
```
ls-1801: The error string now matches the parameter name, but the original mismatch this comment raised is still here: `CSVInputFormatIndexer.cpp:105` passes `SIZE_OF_TUPLE_DELIMITER` into the `sizeOfFieldDelimiter` slot. Both constants happen to be 1 in CSV so the bug is latent. Worse, the stored `sizeOfFieldDelimiter` member is set on L101 but never read anywhere else in this class — pure dead state. Either pass `SIZE_OF_FIELD_DELIMITER` (and actually use the member), or drop the parameter and the member.

nes-input-formatters/include/InputFormatters/RawTupleBuffer.hpp
```diff
@@ -90,7 +87,8 @@ class RawTupleBuffer
     TupleBuffer rawBuffer;
     std::string_view bufferView;
 
-    /// Used by InputFormatter to pass an arena-allocated spanning tuple to an InputFormatIndexer
+    /// Wraps an arena-allocated spanning tuple (raw pointer + size) with no backing TupleBuffer. This form must never be used for regular
+    /// raw buffers, so it is private and reachable only through SpanningRawTupleBufferFactory::fromArenaAllocation.
     explicit RawTupleBuffer(const char* rawDataPtr, const size_t sizeOfSpanningTuple) : bufferView(rawDataPtr, sizeOfSpanningTuple) { };
```
ls-1801: The arena-form ctor here exists only because `InputFormatIndexer::indexRawBuffer` takes a `const RawTupleBuffer&`, but both overrides only ever touch `rawBuffer.getBufferView()` — they never read the underlying `TupleBuffer`, the sequence number, the origin id, or anything else. So `RawTupleBuffer` is a sum type in disguise: in the normal form `rawBuffer` is real and `bufferView` is derived from it; in the arena form `rawBuffer` is default-constructed and only `bufferView` carries data. That leaves `getNumberOfBytes()` (returns 0 instead of `bufferView.size()`), `emit()`, `repeat()`, `getRawBuffer()`, `getSequenceNumber()`, etc. as latent footguns on arena instances — the `friend` gate stops external _construction_ but doesn't stop a `const RawTupleBuffer&` callee from invoking them.

The simpler shape is to narrow the interface to what the indexers actually need:

```cpp
// InputFormatIndexer.hpp
[[nodiscard]] virtual std::unique_ptr<RawBufferIndex>
indexRawBuffer(std::string_view rawBuffer) const = 0;
```

Both overrides become `s/rawBuffer.getBufferView()/rawBuffer/`. The call sites in `InputFormatter.cpp` simplify too:

```cpp
// L154-156
assignLeadingRawBufferIndex(indexer.indexRawBuffer({}));
// L208 / L223 (spanning):
indexer.indexRawBuffer({tlIndexPhaseResult.trailingSpanningTuple.data(),
                        tlIndexPhaseResult.trailingSpanningTuple.size()})
```

`SpanningRawTupleBufferFactory` and the private arena-form ctor (plus the `friend` declaration) all disappear; `RawTupleBuffer` goes back to being exclusively a wrapper around a real `TupleBuffer`, matching the class doc at L35–40. `std::string_view` is the right type here.

Also: while you're looking, `emit()`, `repeat()`, `getOriginId()`, and `setSpanningTuple()` on `RawTupleBuffer` appear to have zero callers in `nes-input-formatters`/`nes-plugins`. Dead.

yschroeder97 (approved):

nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
```diff
@@ -0,0 +1,125 @@
...
+    nautilus::val<uint64_t> numTuples
+        = readValueFromMemRef<size_t>(getMemberRef(rawBufferIndex, &FieldOffsetRawBufferIndex::totalNumberOfTuples));
+    return tupleIdx < numTuples;
+}
+
+const FieldIndex* FieldOffsetRawBufferIndex::getIndexValuesProxy(const RawBufferIndex* rawBufferIndex)
+{
+    /// At this point, the index values are fixed, thus the vector cannot grow anymore and accessing the pointer is safe
+    const auto* const fieldOffsets = dynamic_cast<const FieldOffsetRawBufferIndex* const>(rawBufferIndex);
+    return fieldOffsets->indexValues.data();
```
ls-1801, line 60: `dynamic_cast` to a pointer type returns `nullptr` on failure, and L57 immediately dereferences. Switching from `static_cast` to `dynamic_cast` was a safety move (per Artraxon's review), but the pointer form only delivers the safety if the result is checked. The reference form throws `std::bad_cast` on a bad cast and fails loudly:

```cpp
const auto& fieldOffsets = dynamic_cast<const FieldOffsetRawBufferIndex&>(*rawBufferIndex);
return fieldOffsets.indexValues.data();
```

As-is, a stale `tlOwnedRawBufferIndices` slot (the thread_local that holds these pointers across `InputFormatter` instances) would silently segfault instead of asserting.

nes-input-formatters/src/InputFormatter.cpp
```diff
@@ -0,0 +1,493 @@
...
+{
+    /// Value was stashed on tlIndexPhaseResult when the indexer returned its RawBufferIndex; max() means no delimiter in the buffer.
+    const auto offsetOfLastTupleDelimiter = tlIndexPhaseResult.offsetOfLastTupleInRawBuffer;
+    if (offsetOfLastTupleDelimiter == std::numeric_limits<FieldIndex>::max())
+    {
+        return false;
+    }
+
+    /// if the offset to trailing SpanningTuple was not known after indexing we need to set it in the SequenceShredder, allowing threads to find it
+    auto stagedBuffers = (not tlIndexPhaseResult.hasValidOffsetOfTrailingSpanningTuple)
```
ls-1801: The `not tlIndexPhaseResult.hasValidOffsetOfTrailingSpanningTuple` branch here is unreachable. The two flags are written in lockstep at L185–186 of `IndexPhaseResultBuilder::indexRawBuffer` — `hasValidOffsetOfTrailingSpanningTuple = (offsetOfLastTupleDelimiter != max())` and `offsetOfLastTupleInRawBuffer = offsetOfLastTupleDelimiter` — and nothing else writes to either. After the early return at L261 we already know `offsetOfLastTupleInRawBuffer != max()`, which by construction means `hasValidOffsetOfTrailingSpanningTuple == true`. The two-argument overload of `SequenceShredder::findTrailingSpanningTupleWithDelimiter` (and its underlying `SpanningTupleBuffer::tryFindTrailingSpanningTupleForBufferWithDelimiter` overload) has no non-test caller — only `ConcurrentSynchronizationTest` exercises the one-arg form. The `///` comment above the ternary describes a scenario ("offset not known after indexing") that can't actually happen.

Verified locally: clean build, 18/18 input-formatter unit tests, 51/51 Formatter systests pass.

```diff
diff --git a/nes-input-formatters/include/InputFormatters/SequenceShredder.hpp b/nes-input-formatters/include/InputFormatters/SequenceShredder.hpp
index 8b71fbf7aa..fa376babfe 100644
--- a/nes-input-formatters/include/InputFormatters/SequenceShredder.hpp
+++ b/nes-input-formatters/include/InputFormatters/SequenceShredder.hpp
@@ -87,9 +87,6 @@ public:
     /// Assumes findLeadingSpanningTupleWithDelimiter was already called and the StagedBuffer for 'sequenceNumber' already set
     /// Searches for a reachable buffer that delimits tuples in trailing direction (higher SequenceNumbers)
     SpanningBuffers findTrailingSpanningTupleWithDelimiter(SequenceNumber sequenceNumber);
-    /// Overload that allows to lazily set the offset of the last record starting in the StagedBuffer (and the atomic state)
-    /// if the offset was not already known when 'findLeadingSpanningTupleWithDelimiter' was called
-    SpanningBuffers findTrailingSpanningTupleWithDelimiter(SequenceNumber sequenceNumber, FieldIndex offsetOfLastTuple);
 
     friend std::ostream& operator<<(std::ostream& os, const SequenceShredder& sequenceShredder);
 
diff --git a/nes-input-formatters/private/SpanningTupleBuffer.hpp b/nes-input-formatters/private/SpanningTupleBuffer.hpp
index 5f328b633b..35b83a6ab8 100644
--- a/nes-input-formatters/private/SpanningTupleBuffer.hpp
+++ b/nes-input-formatters/private/SpanningTupleBuffer.hpp
@@ -85,8 +85,6 @@ public:
     [[nodiscard]] SequenceShredderResult
     tryFindLeadingSpanningTupleForBufferWithDelimiter(SequenceNumber sequenceNumber, const StagedBuffer& indexedRawBuffer);
     [[nodiscard]] SpanningBuffers tryFindTrailingSpanningTupleForBufferWithDelimiter(SequenceNumber sequenceNumber);
-    [[nodiscard]] SpanningBuffers
-    tryFindTrailingSpanningTupleForBufferWithDelimiter(SequenceNumber sequenceNumber, FieldIndex offsetOfLastTuple);
 
     /// Tries to find a reachable buffer with a delimiter in leading direction.
     /// If successful, starts a claiming trailing delimiter search with that reachable buffer as the start (search starts at 'sequenceNumber')
diff --git a/nes-input-formatters/src/InputFormatter.cpp b/nes-input-formatters/src/InputFormatter.cpp
index 335bdbe6f0..a1a3d0e292 100644
--- a/nes-input-formatters/src/InputFormatter.cpp
+++ b/nes-input-formatters/src/InputFormatter.cpp
@@ -60,7 +60,6 @@ struct IndexPhaseResult
     bool hasLeadingSpanningTupleBool = false;
     bool hasTupleDelimiter = false;
     bool isRepeat = false;
-    bool hasValidOffsetOfTrailingSpanningTuple = false;
     /// Stashed from the indexer's RawBufferIndex; read later by 'indexTrailingSpanningTupleProxy' instead of dereferencing the RawBufferIndex.
     FieldIndex offsetOfLastTupleInRawBuffer = std::numeric_limits<FieldIndex>::max();
 
@@ -182,7 +181,6 @@ public:
         const auto [offsetOfFirstTupleDelimiter, offsetOfLastTupleDelimiter] = result->getTupleDelimiterOffsets();
         assignRawBufferIndex(std::move(result));
 
-        tlIndexPhaseResult.hasValidOffsetOfTrailingSpanningTuple = offsetOfLastTupleDelimiter != std::numeric_limits<FieldIndex>::max();
         tlIndexPhaseResult.offsetOfLastTupleInRawBuffer = offsetOfLastTupleDelimiter;
 
         tlIndexPhaseResult.hasTupleDelimiter = offsetOfFirstTupleDelimiter < tupleBuffer.getBufferSize();
@@ -263,10 +261,7 @@ bool indexTrailingSpanningTupleProxy(
         return false;
     }
 
-    /// if the offset to trailing SpanningTuple was not known after indexing we need to set it in the SequenceShredder, allowing threads to find it
-    auto stagedBuffers = (not tlIndexPhaseResult.hasValidOffsetOfTrailingSpanningTuple)
-        ? sequenceShredder->findTrailingSpanningTupleWithDelimiter(tupleBuffer->getSequenceNumber(), offsetOfLastTupleDelimiter)
-        : sequenceShredder->findTrailingSpanningTupleWithDelimiter(tupleBuffer->getSequenceNumber());
+    auto stagedBuffers = sequenceShredder->findTrailingSpanningTupleWithDelimiter(tupleBuffer->getSequenceNumber());
     /// a trailing spanning tuple must span at least 2 buffers
     if (stagedBuffers.getSize() < 2)
     {
diff --git a/nes-input-formatters/src/SequenceShredder.cpp b/nes-input-formatters/src/SequenceShredder.cpp
index a23a937b1a..db80917201 100644
--- a/nes-input-formatters/src/SequenceShredder.cpp
+++ b/nes-input-formatters/src/SequenceShredder.cpp
@@ -75,12 +75,6 @@ SpanningBuffers SequenceShredder::findTrailingSpanningTupleWithDelimiter(const S
     return spanningTupleBuffer->tryFindTrailingSpanningTupleForBufferWithDelimiter(sequenceNumber);
 }
 
-SpanningBuffers
-SequenceShredder::findTrailingSpanningTupleWithDelimiter(const SequenceNumber sequenceNumber, const FieldIndex offsetOfLastTuple)
-{
-    return spanningTupleBuffer->tryFindTrailingSpanningTupleForBufferWithDelimiter(sequenceNumber, offsetOfLastTuple);
-}
-
 SequenceShredderResult SequenceShredder::findSpanningTupleWithoutDelimiter(const StagedBuffer& indexedRawBuffer)
 {
     return findSpanningTupleWithoutDelimiter(indexedRawBuffer, indexedRawBuffer.getRawTupleBuffer().getSequenceNumber());
diff --git a/nes-input-formatters/src/SpanningTupleBuffer.cpp b/nes-input-formatters/src/SpanningTupleBuffer.cpp
index b83ebcd835..4421b5b5bf 100644
--- a/nes-input-formatters/src/SpanningTupleBuffer.cpp
+++ b/nes-input-formatters/src/SpanningTupleBuffer.cpp
@@ -109,24 +109,6 @@ SpanningBuffers SpanningTupleBuffer::tryFindTrailingSpanningTupleForBufferWithDe
     return SpanningBuffers{};
 }
 
-SpanningBuffers SpanningTupleBuffer::tryFindTrailingSpanningTupleForBufferWithDelimiter(
-    const SequenceNumber sequenceNumber, const FieldIndex offsetOfLastTuple)
-{
-    const auto [sequenceNumberBufferIdx, _] = getBufferIdxAndABAItNo(sequenceNumber);
-    this->buffer.at(sequenceNumberBufferIdx.getRawValue()).setOffsetOfTrailingSpanningTuple(offsetOfLastTuple);
-
-    auto [firstSpanningTupleStartBuffer, lastSN] = claimingTrailingDelimiterSearch(sequenceNumber);
-    if (firstSpanningTupleStartBuffer.has_value())
-    {
-        const auto sizeOfSpanningTuple = lastSN.getRawValue() - sequenceNumber.getRawValue() + 1;
-        std::vector<StagedBuffer> spanningTupleBuffers(sizeOfSpanningTuple);
-        spanningTupleBuffers[0] = std::move(firstSpanningTupleStartBuffer.value());
-        claimSpanningTupleBuffers(sequenceNumber, spanningTupleBuffers);
-        return SpanningBuffers(std::move(spanningTupleBuffers));
-    }
-    return SpanningBuffers{};
-}
-
 SequenceShredderResult SpanningTupleBuffer::tryFindSpanningTupleForBufferWithoutDelimiter(
     const SequenceNumber sequenceNumber, const StagedBuffer& indexedRawBuffer)
 {
```

After this, `SpanningTupleBufferEntry::setOffsetOfTrailingSpanningTuple` has no callers; its body sets `hasValidLastDelimiterOffsetBit`, which is still _read_ at `SpanningTupleBuffer.cpp:187`. If that bit was planned future state, leave the setter for a follow-up; otherwise it can also be removed (and the loop condition there simplified).

nes-input-formatters/src/InputFormatter.cpp
```diff
@@ -0,0 +1,493 @@
...
+        this->projections,
+        *this->inputFormatIndexer,
+        *this->sequenceShredder,
+        *this->memoryProvider);
+}
+
+std::ostream& InputFormatter::toString(std::ostream& os) const
+{
+    /// Not using fmt::format, because it fails during build, trying to pass sequenceShredder as a const value
+    os << "InputFormatter(" << ", inputFormatIndexer: " << inputFormatIndexer << ", sequenceShredder: " << *sequenceShredder << ")\n";
```
ls-1801: `inputFormatIndexer` is a `std::unique_ptr<InputFormatIndexer>` and the base `InputFormatIndexer` has no `operator<<`. The C++20 `operator<<(ostream&, const unique_ptr<...>&)` overload prints the underlying `.get()` address as hex — `*sequenceShredder` on the same line is dereferenced and goes through `SequenceShredder::operator<<`. So this currently emits something like `inputFormatIndexer: 0x7f9a3c001234` next to a properly-rendered sequence shredder, which isn't useful for the debug/log purpose this serves.

Cheapest fix is to drop `inputFormatIndexer` from the output. If you want it rendered properly, add a virtual `print(std::ostream&)` (or `toString(std::ostream&)`) on the base and override in each derived indexer — both `CSVInputFormatIndexer` and `SIMDJSONInputFormatIndexer` already have their own `operator<<` that just isn't reachable through the base pointer.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
```diff
@@ -164,25 +133,61 @@ void SIMDJSONFIF::writeValueToRecord(
...
+        auto fieldIndex = static_cast<nautilus::val<FieldIndex>>(i);
+        const auto fieldDataType = bufferRef.getAllDataTypes().at(i);
+        writeValueToRecord(
+            fieldDataType, record, fieldName, fieldIndex, rawBufferIndex, nautilus::val<const InputFormatIndexer*>(&indexer));
+    }
+    /// Increment iterator and return record
+    nautilus::invoke(
+        +[](RawBufferIndex* rawBufferIndexPtr)
+        {
+            auto* simdJsonBufferIndex = dynamic_cast<SIMDJSONRawBufferIndex*>(rawBufferIndexPtr);
```
ls-1801: Same pattern as the comment on `FieldOffsetRawBufferIndex.cpp:57` — the SIMDJSON-side `dynamic_cast` calls all return `nullptr` on a bad cast and are dereferenced without a check

ls-1801 (changes requested):

nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
```diff
@@ -0,0 +1,122 @@
...
+    this->offsetOfLastTuple = std::numeric_limits<FieldIndex>::max();
+}
+
+void FieldOffsetRawBufferIndex::markWithTupleDelimiters(const FieldIndex offsetToFirstTuple, const FieldIndex offsetToLastTuple)
+{
+    /// Make sure that the number of read fields matches the expected value.
+    const auto numberOfTuples = indexValues.size() / numberOfOffsetsPerTuple;
+    const auto remainder = indexValues.size() % numberOfOffsetsPerTuple;
+    INVARIANT(
+        remainder == 0, "Number of indexes {} must be a multiple of number of fields in tuple {}", remainder, numberOfOffsetsPerTuple);
```
ls-1801: The format string says "Number of indexes {} must be a multiple of number of fields in tuple {}", but the first placeholder is bound to `remainder`, not `indexValues.size()`. On failure this prints something like "Number of indexes 2 must be a multiple of … 4", which is misleading because `2` is the leftover, not the index count. Bind the first slot to `indexValues.size()` and optionally append `remainder`.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONInputFormatIndexer.hpp
```diff
@@ -14,35 +14,133 @@
...
+        Private,
+        const char tupleDelimiter,
+        std::vector<Record::RecordFieldIdentifier> fieldNamesInJson,
+        std::vector<Record::RecordFieldIdentifier> fieldNamesOutput,
+        std::vector<DataType> fieldDataTypes)
+        : tupleDelimiter(tupleDelimiter)
+        , fieldNamesInJson(std::move(fieldNamesInJson))
+        , fieldNamesOutput(std::move(fieldNamesOutput))
+        , fieldDataTypes(std::move(fieldDataTypes))
+        , nullValues({""})
```
ls-1801: `nullValues({""})` is never actually read along any SIMDJSON path — `SIMDJSONRawBufferIndex::readSpanningRecord` goes through `checkIsNullJsonProxy`/`is_null()` directly, and only `FieldOffsetRawBufferIndex` (the CSV side) calls `getNullValues()`. So it is dead, but the value still claims `""` is NULL, which would be wrong if anyone polymorphically called `indexer.getNullValues()` on a SIMDJSON instance. Prefer `{}` here, or restore the old `INVARIANT(false, "do not call on SIMDJSON")` in `getNullValues()` to keep the contract explicit.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONParsingUtil.hpp
```diff
@@ -0,0 +1,321 @@
...
+    {
+        if (checkIsNullJsonProxy(fieldIndex, simdJsonRawBufferIndex, simdIndexer))
+        {
+            result.isNull = true;
+            result.value = T{0};
+            return &result;
+        }
+    }
+
+    const auto& fieldName = simdIndexer->getFieldNameInJsonAt(fieldIndex);
```
ls-1801, line 212: Inconsistent bounds enforcement: `parseJsonVarSizedProxy` has an INVARIANT on `fieldIndex < getNumberOfFields()` (L281), but `parseJsonFixedSizeIntoVarValProxy` has none before it calls `getFieldNameInJsonAt(fieldIndex)` here (and `checkIsNullJsonProxy` on L197 also dereferences it via the same path). The two proxies should agree — ideally a `PRECONDITION` at the top of each, matching the existing dynamic-cast PRECONDITIONs.

nes-input-formatters/src/InputFormatter.cpp
```diff
@@ -0,0 +1,488 @@
...
+            projections, spanningRecordPtr, nautilus::val<uint64_t>(0), indexer, trailingRawBufferIndex, bufferRef);
+        executeChild(executionCtx, record);
+    }
+}
+
+}
+
+std::vector<DataType> InputFormatter::getAllDataTypes() const
+{
+    INVARIANT(false, "unsupported operation on InputFormatterBufferRef");
```
ls-1801: Stale type name in the INVARIANT message: `InputFormatterBufferRef` refers to a class removed in this refactor (the type-erasure layer). Should read `InputFormatter`. The same stale name appears on L414 (`writeRecord`) and L420 (`getAllFieldNames`) — please update all three.

nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -110,8 +110,7 @@ PhysicalOperator createScanOperator(
         const auto inputFormatterConfig = prevPipeline.getRootOperator().get<SourcePhysicalOperator>().getDescriptor().getParserConfig();
         if (toUpperCase(inputFormatterConfig.parserType) != "NATIVE")
         {
-            return ScanPhysicalOperator(
-                provideInputFormatterTupleBufferRef(inputFormatterConfig, memoryProvider), inputSchema->getFieldNames());
+            return ScanPhysicalOperator(provideInputFormatter(inputFormatterConfig, memoryProvider), inputSchema->getFieldNames());
```
ls-1801, line 113: The doc comment two lines above (L106-107) is stale post-refactor: it says "Instantiate the scan with an InputFormatterTupleBufferRef" but `InputFormatterTupleBufferRef` no longer exists, and the rename to `provideInputFormatter` here changes the type returned. Also `operatior` → `operator`. Suggested rewrite:

```suggestion
        /// Instantiate the scan with an InputFormatter, if the prior operator is a source operator that contains a source
        /// descriptor with a parser type other than "NATIVE" (NATIVE data does not require formatting)
            return ScanPhysicalOperator(provideInputFormatter(inputFormatterConfig, memoryProvider), inputSchema->getFieldNames());
```

nes-input-formatters/registry/include/InputFormatIndexerRegistry.hpp
```diff
@@ -24,15 +24,14 @@
...
 
 namespace NES
 {
 
 
-using InputFormatIndexerRegistryReturnType = std::unique_ptr<InputFormatterTupleBufferRef>;
+using InputFormatIndexerRegistryReturnType = std::unique_ptr<InputFormatter>;
 
 /// Allows specific InputFormatter to construct templated Formatter using the 'createInputFormatterWithIndexer()' method.
 /// Calls constructor of specific InputFormatter and exposes public members to it.
```
ls-1801: Stale doc comment from the pre-PR templated design — `InputFormatter` is no longer templated and the factory now takes a plain `unique_ptr<InputFormatIndexer>`. "Exposes public members" referred to template-constraint satisfaction that no longer exists.

```suggestion
/// Argument bundle handed to `InputFormatIndexerRegistry`-resolved factories. Owns the `ParserConfig` + memory provider and
/// hands a `unique_ptr<InputFormatIndexer>` to `createInputFormatterWithIndexer()` to assemble a concrete `InputFormatter`.
```

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
```diff
@@ -164,25 +133,64 @@ void SIMDJSONFIF::writeValueToRecord(
...
+                dynamic_cast<SIMDJSONRawBufferIndex*>(rawBufferIndexPtr) != nullptr, "rawBufferIndex must be a SIMDJSONRawBufferIndex");
+            /// NOLINTNEXTLINE(cppcoreguidelines-pro-type-static-cast-downcast): type verified by PRECONDITION above.
+            auto* simdJsonBufferIndex = static_cast<SIMDJSONRawBufferIndex*>(rawBufferIndexPtr);
+            ++simdJsonBufferIndex->docStreamIterator;
+            simdJsonBufferIndex->isAtLastTuple = simdJsonBufferIndex->docStreamIterator.at_end();
+        },
+        rawBufferIndex);
+    return record;
+}
+
```
ls-1801, line 174: Stale comment carried over from the old `SIMDJSONFIF.hpp`. `markNoTupleDelimiters()` only sets the two offsets to `max()` — it does not reset pointers, calculate tuple counts, or return anything.

```suggestion
/// Marks the buffer as containing no tuple delimiters by setting both offsets to `max()`.
```

nes-input-formatters/private/CSVInputFormatIndexer.hpp
```diff
@@ -26,97 +28,71 @@
...
-    [[nodiscard]] const DataType& getFieldDataTypeAt(const nautilus::static_val<uint64_t>& i) const
+    /// Delegate constructor that applies preconditions before safely calling the constructor
+    static std::unique_ptr<CSVInputFormatIndexer> create(const ParserConfig& config, const TupleBufferRef& tupleBufferRef)
     {
-        PRECONDITION(
-            i < fieldDataTypes.size(), "Trying to access position, larger than the size of fieldDataTypes {}", fieldDataTypes.size());
-        return fieldDataTypes[i];
+        PRECONDITION(config.tupleDelimiter.size() == 1, "Size of tuple delimiter must be exactly one.");
+        PRECONDITION(config.fieldDelimiter.size() == 1, "Size of field delimiter must be exactly one.");
+        return std::make_unique<CSVInputFormatIndexer>(
```
ls-1801: Nit: the new PRECONDITION messages are noticeably less informative than the ones that lived in `CSVMetaData` (which printed both the actual delimiter value and its size). Consider restoring the format args so failures from a misconfigured `ParserConfig` are easier to diagnose:

```suggestion
        PRECONDITION(config.tupleDelimiter.size() == 1, "Tuple delimiter must be 1 byte but was {:?} (size {})", config.tupleDelimiter, config.tupleDelimiter.size());
        PRECONDITION(config.fieldDelimiter.size() == 1, "Field delimiter must be 1 byte but was {:?} (size {})", config.fieldDelimiter, config.fieldDelimiter.size());
```

nes-input-formatters/private/CSVInputFormatIndexer.hpp
```diff
@@ -26,97 +28,71 @@
...
-    std::vector<std::string> nullValues;
-};
+    [[nodiscard]] std::unique_ptr<RawBufferIndex> indexRawBuffer(std::string_view rawBuffer) const override;
 
-class CSVInputFormatIndexer : public InputFormatIndexer<CSVInputFormatIndexer>
-{
-public:
-    using IndexerMetaData = CSVMetaData;
-    using FieldIndexFunctionType = FieldOffsets<CSV_NUM_OFFSETS_PER_FIELD>;
+    [[nodiscard]] std::string_view getTupleDelimitingBytes() const override { return {&tupleDelimiter, 1}; }
```
ls-1801, line 88: Minor consistency: `getFieldDelimitingBytes` uses `SIZE_OF_FIELD_DELIMITER` but `getTupleDelimitingBytes` uses a literal `1`. `SIZE_OF_TUPLE_DELIMITER` is already defined above — easy swap for symmetry.

```suggestion
    [[nodiscard]] std::string_view getTupleDelimitingBytes() const override { return {&tupleDelimiter, SIZE_OF_TUPLE_DELIMITER}; }
```

ls-1801 (approved):
Nothing major, mostly documentation or error message improvements.
I will accept this pr and you can address them as you please.

nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
alepping: took

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONInputFormatIndexer.hpp
alepping: took suggestion to use `{}`

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONParsingUtil.hpp
alepping: added PRECONDITION on top and made consistent

nes-input-formatters/src/InputFormatter.cpp
alepping: updated all trhee

nes-query-compiler/src/Phases/PipeliningPhase.cpp
alepping: took

nes-input-formatters/registry/include/InputFormatIndexerRegistry.hpp
alepping: took

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
alepping: took

nes-input-formatters/private/CSVInputFormatIndexer.hpp
alepping: took

nes-input-formatters/private/CSVInputFormatIndexer.hpp
alepping: took

PR #1569, opened 2026-04-23
fix(Systest): Support Void sinks in the systest harness

keyseven123:

The systest harness unconditionally injected `file_path` (and defaulted `output_format=CSV`) into every sink's config. `VoidSink` rejects unknown keys, so any query using `INTO Void()` or a `CREATE SINK ... TYPE Void` definition failed at binding time. In addition, the result check always expected a result file on disk, which does not exist for a sink that discards its tuples.

Changes:
- `SystestBinder::registerSink` / `setInlineSink`: only inject `file_path` for non-Void sinks, and keep the `output_format` default off for both `CHECKSUM` and `VOID`. `CHECKSUM` continues to receive `file_path` (its output still goes to a file).
- `SystestResultCheck::checkResult`: when the query's root sink is of type `VOID`, log and return `std::nullopt` instead of attempting to read a (non-existent) result file.
- New systest `nes-systests/sinks/VoidSink.test` exercising both the inline `INTO Void()` form and the named `CREATE SINK ... TYPE Void` form.

## Verifying this change

- `nes-systests/sinks/VoidSink.test` — both queries PASS.
- `nes-systests/sinks/InlineSink.test` — all 5 queries PASS (regression for the inline-sink path).
- `nes-systests/sources/Generator.test` — all 9 queries PASS (regression for CHECKSUM sinks, which share the same code paths and still need `file_path`).

- Systest harness (`nes-systests/systest/src/SystestBinder.cpp`, `SystestResultCheck.cpp`)
- No runtime / query-engine changes; `VoidSink` itself is unchanged.

### DIFF
--- a/nes-systests/systest/src/SystestBinder.cpp
+++ b/nes-systests/systest/src/SystestBinder.cpp
@@ -96,14 +96,16 @@ public:
             [this, schema, sinkType](
                 const std::string_view assignedSinkName, std::filesystem::path filePath) -> std::expected<SinkDescriptor, Exception>
             {
-                std::unordered_map<std::string, std::string> config{{"file_path", std::move(filePath)}};
+                std::unordered_map<std::string, std::string> config{};
                 std::unordered_map<std::string, std::string> formatConfig{};
                 if (sinkType == "File")
                 {
+                    config["file_path"] = std::move(filePath);
                     config["output_format"] = "CSV";
                 }
                 else if (toUpperCase(sinkType) == "CHECKSUM")
                 {
+                    config["file_path"] = std::move(filePath);
                     formatConfig["quote_strings"] = "true";
                 }
 
@@ -711,9 +713,13 @@ struct SystestBinder::Impl
         auto formatConfig = sinkOperator->getFormatConfig();
         auto schema = sinkOperator->getSchema();
         sinkConfig.erase("file_path");
-        sinkConfig.emplace("file_path", resultFile);
-
-        if (not(sinkConfig.contains("output_format")) and sinkOperator->getSinkType() != "CHECKSUM")
+        /// Only inject a file_path for a file sink or checksum sink
+        if (toUpperCase(sinkOperator->getSinkType()) == "FILE" or toUpperCase(sinkOperator->getSinkType()) == "CHECKSUM")
+        {
+            sinkConfig.emplace("file_path", resultFile);
+        }
+        if (not(sinkConfig.contains("output_format")) and sinkOperator->getSinkType() != "CHECKSUM"
+            and sinkOperator->getSinkType() != "VOID")
         {
             sinkConfig.emplace("output_format", "CSV");
         }
--- a/nes-systests/systest/src/SystestResultCheck.cpp
+++ b/nes-systests/systest/src/SystestResultCheck.cpp
@@ -36,6 +36,7 @@
 #include <DataTypes/DataTypeProvider.hpp>
 #include <DataTypes/Schema.hpp>
 #include <Identifiers/NESStrongType.hpp>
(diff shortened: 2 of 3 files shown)


nes-systests/systest/src/SystestResultCheck.cpp
```diff
@@ -753,6 +754,22 @@ namespace NES
...
+    if (runningQuery.systestQuery.planInfoOrException.has_value())
+    {
+        const auto& rootOperators = runningQuery.systestQuery.planInfoOrException.value().queryPlan.getGlobalPlan().getRootOperators();
+        if (not rootOperators.empty())
+        {
+            if (const auto sinkOp = rootOperators.at(0).tryGetAs<SinkLogicalOperator>(); sinkOp.has_value()
+                and sinkOp.value()->getSinkDescriptor().has_value()
+                and toUpperCase(sinkOp.value()->getSinkDescriptor().value().getSinkType()) == "VOID")
+            {
+                NES_INFO("Skipping result check for query {} because it writes to a Void sink.", runningQuery.systestQuery.queryDefinition);
```
yschroeder97: Do we want to print the whole SQL string here?

nes-systests/systest/src/SystestResultCheck.cpp
```diff
@@ -753,6 +754,22 @@ namespace NES
 {
 std::optional<std::string> checkResult(const Systest::RunningQuery& runningQuery)
 {
+    /// Void sinks discard all tuples and produce no result file, so there is nothing to check.
+    if (runningQuery.systestQuery.planInfoOrException.has_value())
+    {
+        const auto& rootOperators = runningQuery.systestQuery.planInfoOrException.value().queryPlan.getGlobalPlan().getRootOperators();
```
yschroeder97: I think we can use `getOperatorByType` instead of manually unwrapping the plan to reach the sink and check it.

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -711,9 +713,13 @@ struct SystestBinder::Impl
         auto formatConfig = sinkOperator->getFormatConfig();
         auto schema = sinkOperator->getSchema();
         sinkConfig.erase("file_path");
-        sinkConfig.emplace("file_path", resultFile);
-
-        if (not(sinkConfig.contains("output_format")) and sinkOperator->getSinkType() != "CHECKSUM")
+        /// VoidSink rejects unknown keys, so don't inject a file_path it cannot consume.
+        if (toUpperCase(sinkOperator->getSinkType()) != "VOID")
```
yschroeder97: Couldn't we generalize this at this point?
Most sinks will not accept file paths as config parameters...

We could just do `if (toUpperCase(sinkOperator->getSinkType()) == "FILE")` and only then set the file_path. 

I am not 100% sure, but this code in general seems a bit fishy.

yschroeder97 (changes requested):
Some minor things.

nes-systests/systest/src/SystestResultCheck.cpp
keyseven123: Done

nes-systests/systest/src/SystestResultCheck.cpp
keyseven123: We are now printing the SLT name and id in file, e.g., `JoinVarSized:01`.

nes-systests/systest/src/SystestBinder.cpp
keyseven123: I have reversed the if statement. Thus, adding a new sink in the SLTs does not get the test to crash

yschroeder97 (approved):

adrianmichalke (approved):

PR #1570, opened 2026-04-27
feat(build): Pre-fetch Rust crates and Corrosion in dev image for offline builds

keyseven123:

A clean Docker-toolchain build currently needs internet for two Rust-related steps: `cmake/EnableRust.cmake` clones Corrosion `v0.5.2` from GitHub via `FetchContent`, and Corrosion drives `cargo build` which resolves ~137 crates from `crates.io`. This PR bakes both into the development image so a developer who pulled the image while online can later run a full `cmake` + `cmake --build` cycle inside the container with no network connection. Non-Docker local builds are unchanged.

Key changes:
- `docker/dependency/Development.dockerfile`: pre-clone Corrosion `v0.5.2` to `/opt/corrosion` and run `cargo fetch --locked` against the `nes-network` workspace using stub `lib.rs` files, so `$CARGO_HOME/registry` is fully populated. Set `CARGO_NET_OFFLINE=true` so cargo never reaches the network at build time.
- `docker/dependency/Development.dockerfile.dockerignore`: un-ignore the six Cargo manifests + `Cargo.lock` so the new `COPY` lines see them (the file otherwise excludes everything via `*`).
- `cmake/EnableRust.cmake`: probe `git ls-remote` (5 s timeout) and only fall back to `$CORROSION_SRC` when GitHub is unreachable. `message(STATUS …)` reports which path ran. Online builds remain unchanged; non-Docker offline builds remain unsupported (no `$CORROSION_SRC` set).
- `docker/dependency/hash_dependencies.sh`: extend `HASH_PATHS` with `Cargo.lock` and the workspace manifests, so `cargo update` invalidates the image cache and triggers a rebuild — without this the registry would silently desync from the lockfile.

## Verifying this change

Verified end-to-end on tower-en717 with the rebuilt image:

- Built the full image chain via `scripts/install-local-docker-environment.sh -l --yes --libstdcxx --no-sanitizer` — `nes-development-base:local`, `nes-development-dependency:local`, `nes-development:default`, `nes-development:local` all produced; `cargo fetch --locked` step inside the build downloads the index and crates without errors.
- Inside the rebuilt image: `CARGO_NET_OFFLINE=true`, `CORROSION_SRC=/opt/corrosion`, 133 `.crate` files in `/usr/local/cargo/registry/cache/index.crates.io-*/`, Corrosion source tree present at `/opt/corrosion`.
- `docker run --rm --network=none -v <repo>:/work -w /work nebulastream/nes-development:default cmake -S . -B /tmp/build …` — configure completes in 1.2 s and prints `-- Corrosion: GitHub unreachable, using pre-cloned copy at /opt/corrosion`, confirming the fallback path.
- `cmake --build /tmp/build --target cargo-build_nes_rust_bindings` (still `--network=none`): all four workspace crates (`spdlog_bindings`, `nes_network`, `network_bindings`, `nes_rust_bindings`) compile against the pre-populated registry — `Finished `dev` profile [unoptimized + debuginfo] target(s)`, exit 0.

- Docker development image (`docker/dependency/Development.dockerfile`, dockerignore)
- Image-rebuild hash gate (`docker/dependency/hash_dependencies.sh`)
- Rust/Corrosion bring-up in CMake (`cmake/EnableRust.cmake`)
- Downstream `DevelopmentLocal.dockerfile` / `DevelopmentCI.dockerfile` inherit the new env and files unchanged


docker/dependency/Development.dockerfile
```diff
@@ -85,6 +85,45 @@ ENV RUSTUP_HOME=/usr/local/rustup \
     PATH=/usr/local/cargo/bin:$PATH \
     RUST_VERSION=1.90.0
 
+# Pre-clone Corrosion at the exact ref CMake will request, so offline configures inside the
+# container can fall back to it when GitHub is unreachable. EnableRust.cmake probes GitHub
+# first and only uses CORROSION_SRC when the probe fails. Tracks the ls-1801 fork that
```
keyseven123: @Reviewer, I will change it once @ls-1801 has moved the repo into the nebulastream org

docker/dependency/Development.dockerfile.dockerignore
```diff
@@ -1,2 +1,10 @@
 # Ignore everything
 *
+
+# Allow only the Cargo manifests + lockfile needed for the in-image `cargo fetch`.
+!nes-network/Cargo.toml
```
yschroeder97: In the future, we should replace those with global globs so I don't need to explicitly write out those paths when adding a new Rust package.

docker/dependency/Development.dockerfile
yschroeder97: Alright, sounds good.

docker/dependency/Development.dockerfile
```diff
@@ -85,6 +85,45 @@ ENV RUSTUP_HOME=/usr/local/rustup \
...
+COPY nes-network/nes-rust-bindings/Cargo.toml              /tmp/nes-network/nes-rust-bindings/Cargo.toml
+COPY nes-network/nes-rust-bindings/network/Cargo.toml      /tmp/nes-network/nes-rust-bindings/network/Cargo.toml
+COPY nes-network/nes-rust-bindings/spdlog/Cargo.toml       /tmp/nes-network/nes-rust-bindings/spdlog/Cargo.toml
+RUN set -eux; \
+    mkdir -p /tmp/nes-network/network/src /tmp/nes-network/nes-rust-bindings/src; \
+    echo 'fn _stub() {}' > /tmp/nes-network/network/src/lib.rs; \
+    echo 'fn _stub() {}' > /tmp/nes-network/nes-rust-bindings/src/lib.rs; \
+    echo 'fn _stub() {}' > /tmp/nes-network/nes-rust-bindings/network/lib.rs; \
+    echo 'fn _stub() {}' > /tmp/nes-network/nes-rust-bindings/spdlog/lib.rs; \
+    cd /tmp/nes-network && cargo fetch --locked; \
```
yschroeder97, line 121: We should auto-discover and fill in the stubs automatically.
Otherwise, I need to dig into this for every Rust crate I write, which is easy to forget because it doesn't break.

```suggestion
RUN set -eux; \
      find /tmp/nes-network -name Cargo.toml | while read m; do \
          d=$(dirname "$m"); \
          mkdir -p "$d/src"; \
          echo 'fn _stub() {}' > "$d/src/lib.rs"; \
          echo 'fn _stub() {}' > "$d/lib.rs"; \
      done; \
      cd /tmp/nes-network && cargo fetch --locked; \
```

docker/dependency/Development.dockerfile
```diff
@@ -85,6 +85,45 @@ ENV RUSTUP_HOME=/usr/local/rustup \
...
+
+# Pre-fetch every crate pinned in nes-network/Cargo.lock into $CARGO_HOME so cargo can build
+# offline. Only manifests + lockfile are copied; stub lib sources let cargo resolve the
+# workspace without the real source tree.
+COPY nes-network/Cargo.toml                                /tmp/nes-network/Cargo.toml
+COPY nes-network/Cargo.lock                                /tmp/nes-network/Cargo.lock
+COPY nes-network/network/Cargo.toml                        /tmp/nes-network/network/Cargo.toml
+COPY nes-network/nes-rust-bindings/Cargo.toml              /tmp/nes-network/nes-rust-bindings/Cargo.toml
+COPY nes-network/nes-rust-bindings/network/Cargo.toml      /tmp/nes-network/nes-rust-bindings/network/Cargo.toml
+COPY nes-network/nes-rust-bindings/spdlog/Cargo.toml       /tmp/nes-network/nes-rust-bindings/spdlog/Cargo.toml
```
yschroeder97:
```suggestion
COPY nes-network /tmp/nes-network
```

Other files should be ignored by dockerignore anyways.

docker/dependency/Development.dockerfile.dockerignore
```diff
@@ -1,2 +1,10 @@
...
+!nes-network/Cargo.lock
+!nes-network/network/Cargo.toml
+!nes-network/nes-rust-bindings/Cargo.toml
+!nes-network/nes-rust-bindings/network/Cargo.toml
+!nes-network/nes-rust-bindings/spdlog/Cargo.toml
```
yschroeder97:
```suggestion
!nes-network/**/Cargo.lock
!nes-network/**/Cargo.toml
```

docker/dependency/hash_dependencies.sh
```diff
@@ -28,6 +28,12 @@ cd "$(git rev-parse --show-toplevel)"
 HASH_PATHS=(
   vcpkg
   docker/dependency
+  nes-network/Cargo.lock
```
yschroeder97: Same here, in the future we should auto-discover.

cmake/EnableRust.cmake
```diff
@@ -42,11 +42,46 @@ endif ()
 
 include(FetchContent)
 set(CORROSION_NO_HOSTBUILD ON CACHE BOOL "Enable proper incremental builds for Rust targets")
-FetchContent_Declare(
-        Corrosion
-        GIT_REPOSITORY https://github.com/ls-1801/corrosion.git
-        GIT_TAG fix/avoid-always-dirty-targets
-)
+
+### Prefer fetching Corrosion from GitHub when reachable; fall back to a pre-cloned copy at
```
yschroeder97, line 45: Why prefer fetching from Github - we could also prefer to use local, no?

yschroeder97 (changes requested):
Thank you, appreciated changes. 

I don't really like the manual enumeration of all the cargo files, though, since it means that whenever I want to add a new cargo crate (e.g., for the coordinator), I have to go through at least 3 places here and enumerate all the cargo .toml and .lock files.

ls-1801 (approved):
I would consider this a band-aid fix. My upcoming Rust I/O changes will introduce a cleaner approach, but this works as well. 

I tested it with `docker run --net=none` builds. Let's add a CI check for that.
Long-term, I recommend we vendor corrosion in our CMake folder.

cmake/EnableRust.cmake
keyseven123: Because this is how it was before. As @ls-1801 mentions, this will change during one of his changes anyway, I would keep it as-is.

keyseven123:
> I would consider this a band-aid fix. My upcoming Rust I/O changes will introduce a cleaner approach, but this works as well.
> 
> I tested it with `docker run --net=none` builds. Let's add a CI check for that. Long-term, I recommend we vendor corrosion in our CMake folder.

I have creasted issue #1580 for the vendored approach.

docker/dependency/hash_dependencies.sh
keyseven123: Done

docker/dependency/Development.dockerfile.dockerignore
keyseven123: Done

docker/dependency/Development.dockerfile
keyseven123: Done

docker/dependency/Development.dockerfile
keyseven123: Done

docker/dependency/Development.dockerfile.dockerignore
keyseven123: Done

yschroeder97 (approved):

Issue #1317, opened 2026-01-13
[EPIC] Stateful Coordinator


yschroeder97:
### Description

Implement the first version of a stateful coordinator as discussed in #1261.
The artifact will be a library that is to be embedded into the frontend processes in use (SLT tool, REPL). 

It comprises the following subtasks: 

- [ ] #1318
- [ ] #1319
- [ ] #1320
- [ ] #1321
- [ ] #1322
- [ ] #1323
- [ ] #1324
- [ ] #1325

### Acceptance Criteria

- [ ] Simulation-based integration tests pass
- [ ] Unit tests (hand-crafted edge cases + PBT) pass
- [ ] System tests use the coordinator (including distributed) and pass

PR #1574, opened 2026-04-29
feat(catalog): add DB-based, persistent transactional catalog

### LINKED ISSUE #1318: Add `Catalog` with SQLite-backed persistent storage
See #1261, section "Catalog", for details.

yschroeder97:
Base of the stateful coordinator (#1261, #1317).
One SQLite-backed catalog replaces the in-memory catalogs and stores the state of a deployment: workers, sources, sinks, queries and the fragments placed on workers.
- `model`: the data model as SeaORM entities, one module per entity with its create, get and drop requests.
  A `Statement` runs any request inside one transaction.
- `migration`: the schema and the SQL triggers.
- The invariants are enforced in the database, not in application code, so a restart or a second writer cannot leave the catalog inconsistent: CHECK constraints and foreign keys, triggers for rules that span rows (state transitions, capacity accounting, a query's state derived from its fragments, cascades, drop guards), and a save hook for rules within one row.
- Postgres later needs a trigger file and a backend variant; the entities and requests stay unchanged.

## Verifying this change
- 68 property-based tests over the invariants against an in-memory SQLite database (`rust-model`).
- The migrator builds (`rust-migration`).

- Coordinator (new crates), Cargo workspace.

## Documentation
- Crate and module documentation; the data model is described in `model/src/lib.rs`.

This PR closes #1318.


nes-coordinator/migration/README.md
```diff
@@ -0,0 +1,41 @@
+# Running Migrator CLI
+
```
greenBene, line 13: Please add a short reasoning why and when you would need to generate a migration file and apply it.

nes-coordinator/migration/src/m20260204_185834_init.rs
```diff
@@ -0,0 +1,660 @@
+/*
```
greenBene, line 1: I am not too happy with having a file with such an abstract name in the repo. Was it intended to have it in the PR?

nes-coordinator/model/src/ml_model/create.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub if_not_exists: bool,
+}
+
+impl Arbitrary for CreateMlModel {
+    type Parameters = ();
+    fn arbitrary_with(_: ()) -> Self::Strategy {
+        use proptest::prelude::*;
+        (
+            "[a-z][a-z0-9_]{2,29}",
+            "/[a-z]{1,8}/[a-z]{1,8}\\.onnx",
```
greenBene: Just fyi, this might have to change because of the changes to the ML Inference  :)

nes-coordinator/model/src/ml_model/mod.rs
```diff
@@ -0,0 +1,94 @@
...
+        let db = Database::for_test().await;
+        req.execute(&db).await.unwrap();
+        DropMlModel::all()
+            .with_name(req.name.clone())
+            .execute(&db)
+            .await
+            .unwrap();
+        req.execute(&db).await.unwrap();
+    }
+}
```
greenBene, line 108: I do not know Rust, so this might be best practice. But do I understand correctly that we have here test code next to the core logic? If so, shouldn't we try to keep this separate?

nes-coordinator/model/src/query/create.rs
```diff
@@ -0,0 +1,118 @@
...
+use sea_orm::{ActiveModelTrait, ColumnTrait, ConnectionTrait, EntityTrait, QueryFilter};
+use serde::Deserialize;
+
+#[derive(Clone, Debug, Deserialize)]
+pub struct CreateQuery {
+    #[serde(default)]
+    pub name: Option<String>,
+    pub sql: String,
+    #[serde(default)]
+    pub fragments: Vec<CreateFragment>,
```
greenBene: We use here the abstraction `query -> fragments`, but in the engine `distributed query -> local query`. I am not sure what is better, but we should be consistent in the naming. 

My suggestion is to use the same naming as in the rest of the code base for now, and if we want to change the naming eventually, change it everywhere at once.

nes-coordinator/model/src/query/mod.rs
```diff
@@ -0,0 +1,675 @@
...
+                    CreateQueryWithRefs {
+                        workers: fragments.workers,
+                        logical_source,
+                        physical_sources,
+                        sink,
+                        query,
+                    }
+                },
+            )
+            .boxed()
```
greenBene: This function looks really complex. Can you split it up?

nes-coordinator/model/src/ml_model/mod.rs
greenBene: Discussed this in person: In Rust, it is best practice to have tests colocated with the main code

nes-coordinator/model/src/source/logical/create.rs
```diff
@@ -0,0 +1,75 @@
...
+        use proptest::prelude::*;
+        "[a-z][a-z0-9_]{2,29}"
+            .prop_map(|name| Self {
+                name,
+                schema: serde_json::json!({}),
+                if_not_exists: false,
+            })
+            .boxed()
+    }
+    type Strategy = proptest::strategy::BoxedStrategy<Self>;
```
greenBene: Please move all testing related code into the respective mod files

nes-coordinator/migration/src/m20260204_185834_init.rs
greenBene: Okay, based on the PR description, it was. 

Still, I am unsure about the name. Would it make sense to have the migration scripts ordered by increasing ID, rather than by the date it was produced?

greenBene (changes requested):
I'm really excited that we are getting close to a proper coordinator with this! :) 

My main change requests are:
1. Please add a short documentation to the start of the files describing what they do. 1-3 sentences are more than enough, just to give the reader an intuition of what is going on
2. We are mixing some testing logic and core system logic. Please ensure that all testing logic is properly enclosed in the mod.rs files 
3. I am unsure about the migration script naming. I would personally prefer sth. line an increasing integer or similar, rather than having a datetime.

nes-coordinator/model/src/query/create.rs
yschroeder97: The concepts of distributed and local queries will be removed by the 4th PR in the stack.
Only Nebuli was aware of a "distributed query"; to a worker, it's just a regular query in the sense that it can be deployed to the query engine. 

Therefore, there will be no inconsistencies in naming once the coordinator is fully integrated into the system.

Regarding the name "fragment", I am willing to make a poll in our chat to decide whether we should use something else inside the catalog.

nes-coordinator/migration/src/m20260204_185834_init.rs
Artraxon: Hmm, I also see some benefit in having it by datetime.
These migration files stay around for a very long time, and having a datetime in there makes it easier to correlate the changes in the schema with the changes in the rest of the code base.

But I don't have a strong opinion either way.

Artraxon (changes requested):
I am not sure if I can give much feedback to the rust code (I will try at some later point), but at a high level, one thing that would be really nice to have is a global view/namespace.

At the moment, you could create a model and a logical source with the same name.
How much effort would it instead be to have a have a hierarchical "identifier" or "symbol" relation, that is (id: int, name: str, parent: int), (id), and (name, parent) must both be unique, and imposes a tree like structure, and the model, logical source and so on just link to a symbol.

That makes it easy to check if a name is already taken or not (by convention, don't reuse symbols that you haven't created), but still makes it possible for plugins to do so if it makes sense in a sane way.

We need namespaces anyway to better separate plugins, users queries (and potentially later on functions and so on), and this would be the right place to start with it.

I didn't want to do it in the old catalog PRs because I didn't have nested Identifiers yet anyway, and in the schema inference I just felt like its too much, and I would have needed to write querying logic as well, which shouldn't be a big problem here.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
```diff
@@ -0,0 +1,236 @@
...
+-- otherwise the row violates the application's invariant that a
+-- Failed fragment carries both, and the per-host error aggregation
+-- would silently drop this host.
+CREATE TRIGGER IF NOT EXISTS cascade_worker_removed_to_fragments
+    AFTER UPDATE OF desired_state ON worker
+    WHEN NEW.desired_state = 'Removed' AND OLD.desired_state != 'Removed'
+BEGIN
+    UPDATE fragment
+    SET current_state = 'Failed',
+        stop_timestamp = CURRENT_TIMESTAMP,
```
ls-1801: `CURRENT_TIMESTAMP` returns naive UTC text (`YYYY-MM-DD HH:MM:SS`, no zone). The column is `Option<DateTime<Local>>` (`fragment.rs:63`), and every other write uses `chrono::Local::now()`. Reading these trigger-written rows back will either fail to parse or silently reinterpret UTC as local time, shifting `stop_timestamp` by the host UTC offset. Either emit ISO-8601 with zone (`strftime('%Y-%m-%dT%H:%M:%fZ','now')`) or migrate the column to `DateTime<Utc>` and stop relying on Local.

nes-coordinator/migration/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+                    .has_table($table::Table.to_string())
+                    .await?,
+                "Table `{}` already exists",
+                $table::Table.to_string()
+            );
+        )+
+    };
+}
+
+/// Drop the listed tables with `IF EXISTS ... CASCADE`. Order does not
```
ls-1801: Doc says CASCADE makes drop order irrelevant. SQLite has no `DROP TABLE ... CASCADE` syntax — drop order works there only because SQLite skips FK enforcement on `DROP`. On Postgres `cascade()` does emit CASCADE and the claim holds. Worth either (a) noting in the doc that order-independence is backend-specific, or (b) just dropping tables in reverse-create order so the macro behaves identically on both backends.

nes-coordinator/migration/src/m20260204_185834_init.rs
```diff
@@ -0,0 +1,660 @@
...
+                            .on_update(ForeignKeyAction::Restrict),
+                    )
+                    .to_owned(),
+            )
+            .await?;
+
+        manager
+            .create_table(
+                MigrationTable::create()
+                    .table(NetworkLink::Table)
```
ls-1801, line 244: `NetworkLink::SourceHostAddr` and `TargetHostAddr` have no foreign keys to `Worker`. They are plain `string().not_null()` — the schema accepts links pointing at non-existent workers, and `cascade_worker_removed_to_fragments` (the only worker-removal cascade) does not touch `network_link`, so removing a worker leaves dangling link rows. `worker/create.rs` actively inserts peer rows here, making this reachable via the public CRUD API. Add `ForeignKey::create().from(NetworkLink::Table, NetworkLink::SourceHostAddr).to(Worker::Table, Worker::HostAddr).on_delete(Cascade)` (and the same for `TargetHostAddr`) so the link table participates in worker removal and constraint enforcement.

nes-coordinator/migration/src/m20260204_185834_init.rs
```diff
@@ -0,0 +1,660 @@
...
+                    .table(PhysicalSource::Table)
+                    .col(
+                        ColumnDef::new(PhysicalSource::Id)
+                            .big_integer()
+                            .not_null()
+                            .auto_increment()
+                            .primary_key(),
+                    )
+                    .col(
+                        ColumnDef::new(PhysicalSource::LogicalSource)
```
ls-1801, line 150: `LogicalSource` is nullable with `on_delete: Restrict`. Restrict only fires when a child references an existing parent, so rows with `logical_source = NULL` are intentional orphans the schema cannot reattach. If "schema-less physical source" is the intent, worth documenting it on the column; otherwise consider `.not_null()`. As a follow-up, `idx_physical_source_logical_source` indexes NULLs too — a partial index `WHERE logical_source IS NOT NULL` would be cheaper if nullability stays.

nes-coordinator/model/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+chrono.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+strum.workspace = true
+thiserror.workspace = true
+anyhow.workspace = true
+tokio.workspace = true
+sqlx.workspace = true
+sea-orm-migration = { version = "=1.1.1", features = ["sqlx-sqlite", "runtime-tokio-rustls"] }
+proptest = "1.5"
```
ls-1801: proptest = 1.5/proptest-derive = 0.5 sit in [dependencies], not [dev-dependencies]. Proptest then links into every production build of the model crate, not just tests. Either move them under [dev-dependencies], or if sibling crates need the Arbitrary impls for their tests, expose them behind an optional proptest feature flag and keep proptest itself dev-only.

nes-coordinator/model/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+async-trait.workspace = true
+chrono.workspace = true
+serde.workspace = true
+serde_json.workspace = true
+strum.workspace = true
+thiserror.workspace = true
+anyhow.workspace = true
+tokio.workspace = true
+sqlx.workspace = true
+sea-orm-migration = { version = "=1.1.1", features = ["sqlx-sqlite", "runtime-tokio-rustls"] }
```
ls-1801: sea-orm-migration is pinned exact (=1.1.1) while sea-orm.workspace = true floats with the workspace. The two crates share types — if the workspace bumps sea-orm, sea-orm-migration cannot follow and the patch versions desync silently. Either pin both via workspace, or pin neither.

nes-coordinator/model/src/format.rs
```diff
@@ -0,0 +1,287 @@
...
+impl Table {
+    fn new<I: IntoIterator<Item = impl Iden>>(columns: I) -> Self {
+        Self {
+            headers: columns.into_iter().map(|c| c.to_string()).collect(),
+            rows: Vec::new(),
+        }
+    }
+
+    fn row(&mut self, cells: Vec<String>) {
+        debug_assert_eq!(cells.len(), self.headers.len());
```
ls-1801, line 53: Width mismatch is guarded only by `debug_assert_eq!`, which is compiled out in release. A row longer than `headers` panics with index-out-of-bounds at `widths[i]` in the Display loop (line 49); a shorter row writes too few cells and corrupts the table layout. The SHOW path should tolerate misaligned catalog data without crashing — promote to a real check (return an error, truncate, or pad to `headers.len()`).

nes-coordinator/model/src/query/create.rs
```diff
@@ -0,0 +1,118 @@
...
+            "a query must reference at least one source"
+        );
+        anyhow::ensure!(
+            !self.sink_ids.is_empty(),
+            "a query must reference at least one sink"
+        );
+        let fragments = self.fragments.clone();
+        let source_ids = self.source_ids.clone();
+        let sink_ids = self.sink_ids.clone();
+        let query = ActiveModel::from(self.clone()).insert(conn).await?;
```
ls-1801: None of the four `?` calls in `CreateQuery::execute` are wrapped with `.context(...)`, unlike `worker/create.rs`, `ml_model/create.rs`, `sink/drop.rs`. Raw `sea_orm::DbErr` (including trigger `RAISE(ABORT, ...)` strings) propagates as bare `Custom("...")` — when one of these four awaits fails, the caller has no breadcrumb telling them which step it was. Wrap each with `.context("failed to insert query" / "failed to insert fragments" / ...)` to match the rest of the crate.

nes-coordinator/model/src/sink/create.rs
```diff
@@ -0,0 +1,105 @@
...
+                .await
+                .context("failed to load sink")?;
+            if let Some(existing) = existing {
+                return Ok(existing);
+            }
+        }
+        Ok(ActiveModel::from(self.clone())
+            .insert(conn)
+            .await
+            .context("failed to update sink")?)
```
ls-1801: Context message says `"failed to update sink"` but this is the INSERT path. Should read `"failed to create sink"` — current message will confuse operators reading logs.

nes-coordinator/model/src/source/physical/create.rs
```diff
@@ -0,0 +1,100 @@
...
+    }
+}
+
+impl Execute for CreatePhysicalSource {
+    type Response = Model;
+    async fn execute(&self, conn: &impl ConnectionTrait) -> Result<Model> {
+        if self.if_not_exists {
+            let existing = Entity::find()
+                .filter(Column::LogicalSource.eq(&self.logical_source))
+                .filter(Column::HostAddr.eq(self.host_addr.to_string()))
```
ls-1801, line 66: The `(logical_source, host_addr, source_type)` lookup used here to detect "already exists" is not backed by a DB unique constraint in the migration. Without `if_not_exists` two callers can race to create duplicate `(logical, host, type)` rows; with `if_not_exists` the lookup will match an arbitrary one of N duplicates. If duplicates are forbidden, add a unique index in `m20260204_185834_init.rs` so the invariant lives at the schema level instead of inside this request handler.

nes-coordinator/model/src/worker/endpoint.rs
```diff
@@ -0,0 +1,150 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}:{}", self.host, self.port)
+    }
+}
+
+impl FromStr for NetworkAddr {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let colon = s.rfind(':').ok_or("Missing port separator")?;
```
ls-1801: Hand-rolled `rfind(':')` mis-parses unbracketed IPv6: `"fe80::1".parse::<NetworkAddr>()` succeeds with `host="fe80:"`, `port=1`, and the bogus pair is then persisted with no further validation. Rather than reinventing host/port parsing, defer to `std::net::SocketAddr` (which handles `[::1]:8080` correctly and also validates the host) or `url::Host` / `http::uri::Authority` from existing deps. The current `FromStr` + `Display` pair also round-trips `[v6]` → `[v6]:port` only by accident.

nes-coordinator/model/src/worker/mod.rs
```diff
@@ -0,0 +1,413 @@
...
+            Self::Active => WorkerState::Active,
+            Self::Unreachable => WorkerState::Unreachable,
+            Self::Removed => WorkerState::Removed,
+        }
+    }
+}
+
+impl ActiveModel {
+    pub fn apply_transition(&mut self, transition: WorkerTransition) {
+        self.current_state = sea_orm::Set(transition.state());
```
ls-1801, line 140: Inconsistency with fragment state handling: `validate_fragment_state_transition` (sqlite.up.sql:28-46) enforces the legal-transition graph at the DB level, but the equivalent for `worker.current_state` does not exist — neither here in `apply_transition` nor as a trigger. So `Removed → Active` or `Active → Removed` skipping `Unreachable` are silently accepted. Either mirror the fragment trigger with a `validate_worker_state_transition`, add a `legal_from(self, current: WorkerState) -> bool` guard before the `Set`, or document explicitly that worker self-reports its observed state and the model trusts it (and the controller is the only legal caller).

nes-coordinator/model/src/query/drop.rs
```diff
@@ -0,0 +1,79 @@
...
+            .col_expr(
+                fragment::Column::DesiredState,
+                Expr::value(DesiredFragmentState::Stopped),
+            )
+            .filter(fragment::Column::QueryId.is_in(query_ids));
+        if self.stop_mode == StopMode::Graceful {
+            update = update.filter(fragment::Column::HasSource.eq(true));
+        }
+        update.exec(conn).await?;
+        Ok(queries)
```
ls-1801, line 69: `queries` here is the snapshot fetched at line 53 — before either `update_many` runs and before any trigger flips `query.state` / `stop_timestamp`. Returning it means the response says `state == Running` and `stop_timestamp == None` for queries we just told to stop, which is confusing for a "drop" call. Re-fetch by id after `update.exec(conn).await?` so the response reflects the post-drop state (one extra SELECT, but the caller no longer has to re-`GetQuery` to see what they just did).

ls-1801 (changes requested):

nes-coordinator/migration/src/m20260204_185834_init.rs
yschroeder97: This is the convention (also done as such in mature systems such as RisignWave) and also used internally by the migration lib to ensure they are applied in the correct order. Therefore, I would keep it this way.

nes-coordinator/model/src/query/fragment.rs
```diff
@@ -0,0 +1,615 @@
...
+
+impl From<i32> for FragmentState {
+    /// Maps the worker's proto `QueryState` enum to coordinator `FragmentState`.
+    /// Proto: Registered=0, Started=1, Running=2, Stopped=3, Failed=4.
+    /// Started is an internal worker detail, so both Started(1) and Running(2) map to Running.
+    fn from(value: i32) -> Self {
+        match value {
+            0 => FragmentState::Registered,
+            1 | 2 => FragmentState::Running,
+            3 => FragmentState::Completed,
```
keyseven123: **Correctness — proto value `3` (Stopped) is decoded as `Completed`.** The doc comment above states the proto enum is `Stopped=3`, but this arm maps `3 => FragmentState::Completed`. A worker that reports a *stopped* fragment (proto 3) is therefore recorded as `Completed`. Combined with `derive_query_state_on_fragment_update`, a forcefully-stopped fragment can drive the owning query to `Completed` instead of `Stopped`.

The reverse impl at line 405 compounds this: both `Stopped` and `Completed` encode to `3`, so `Stopped`↔`Completed` is conflated in both directions and the proto round-trip is lossy. Since proto has no `Completed`, the coordinator can't distinguish a graceful completion from a forceful stop off the wire value `3` alone — decide whether that distinction must be preserved (e.g. carry it via `desired_state`/`stop_mode`) rather than inferred from `3`.

nes-coordinator/model/src/query/fragment.rs
```diff
@@ -0,0 +1,615 @@
...
+            4 => FragmentState::Failed,
+            // An unknown value means the worker is speaking a newer
+            // protocol than this coordinator understands.
+            other => panic!("unknown QueryState: {other}"),
```
keyseven123: **Correctness / availability — `panic!` on an unrecognized proto value.** `From<i32>` panics on any value outside `0..=4`. As the comment notes, this fires exactly when a worker speaks a newer/incompatible protocol. A single malformed or forward-versioned status message from one worker then panics the coordinator's status-ingestion path (remote-triggerable). An unknown protocol value should surface as a recoverable `DbErr`/`anyhow::Error`, not terminate the process.

nes-coordinator/migration/src/m20260204_185834_init.rs
```diff
@@ -0,0 +1,660 @@
...
+        let backend = manager.get_database_backend();
+        if let Some(sql) = triggers::down(backend) {
+            manager.get_connection().execute_unprepared(sql).await?;
+        }
+
+        drop_tables!(
+            manager,
+            QuerySource,
+            QuerySink,
+            LogicalSource,
```
keyseven123: **Correctness — `down()` drops parent tables before their children; SQLite has no `DROP TABLE … CASCADE`.** The `drop_tables!` macro emits `.cascade()`, but `SqliteQueryBuilder` silently omits the keyword (SQLite doesn't support it). With `PRAGMA foreign_keys=ON` (set in `database.rs`), dropping a parent table implicitly deletes its rows and enforces FK actions. `physical_source.logical_source` → `logical_source.name` is `ON DELETE RESTRICT` (line 180), yet `LogicalSource` is dropped here at position 3, before `PhysicalSource` at position 4. On a non-empty catalog this aborts with `FOREIGN KEY constraint failed`. `Worker` (pos 6) before `Fragment` (pos 8) has the same hazard if fragments reference workers via RESTRICT. Either drop children-before-parents explicitly, or disable FKs for the duration of `down()`. The macro comment ‘Order does not matter because CASCADE removes dependents’ is incorrect for SQLite.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
```diff
@@ -0,0 +1,236 @@
...
+    UPDATE worker
+    SET max_operators = max_operators - NEW.num_operators
+    WHERE host_addr = NEW.host_addr
+    AND max_operators IS NOT NULL;
+END;
+
+-- Counterpart to the acquire above: when a fragment enters a terminal
+-- state, give its slots back. The OLD-state guard prevents double-
+-- crediting if a terminal fragment is saved again.
+CREATE TRIGGER IF NOT EXISTS release_fragment_capacity
```
keyseven123: **Correctness — worker capacity leaks on hard-delete.** `acquire_worker_capacity` debits `worker.max_operators` `AFTER INSERT ON fragment`, but `release_fragment_capacity` only credits it back `AFTER UPDATE OF current_state` into a terminal state. Hard-deleting a query (`query::Entity::delete_by_id`, exercised by `hard_delete_cascades_full_chain`) cascade-deletes its fragment rows via the FK — that is a DELETE, not a `current_state` UPDATE, so release never fires. Every non-terminal fragment on a bounded worker then permanently forfeits its reserved slots. Add an `AFTER DELETE ON fragment` release trigger (guarded on `OLD.current_state NOT IN` terminal states, to avoid double-crediting fragments already released via the terminal UPDATE).

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
keyseven123: **Correctness — `CURRENT_TIMESTAMP` writes a naive-UTC value into a `DateTime<Local>` column.** Everywhere else the app fills `stop_timestamp` with `chrono::Local::now()` (offset-bearing), but this trigger writes SQLite `CURRENT_TIMESTAMP`, which is `'YYYY-MM-DD HH:MM:SS'` in UTC with no offset. The `fragment.stop_timestamp` / `query.stop_timestamp` columns are `Option<chrono::DateTime<chrono::Local>>`, so the naive-UTC string is reinterpreted as local wall-clock — cascade-failed fragments get timestamps shifted by the host's UTC offset, and `derive_query_state_on_fragment_update` propagates the skew to `query.stop_timestamp` via `MAX(stop_timestamp)`. Write an offset-qualified timestamp (e.g. `strftime('%Y-%m-%dT%H:%M:%S+00:00','now')`) or set it from Rust.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
```diff
@@ -0,0 +1,236 @@
...
+
+        WHEN OLD.state IN ('Completed', 'Stopped', 'Failed') THEN
+            RAISE(ABORT, 'Invalid state transition: Cannot transition from a terminal state')
+    END;
+END;
+
+-- Same transition rules as for query.state, applied per fragment. The
+-- two triggers must stay aligned: any state a fragment can reach must
+-- also be a state the derived query state is allowed to reach.
+CREATE TRIGGER IF NOT EXISTS validate_fragment_state_transition
```
keyseven123: **Altitude — the fragment/query lifecycle state machine is encoded in ≥4 disjoint places.** `validate_query_state_transition` and `validate_fragment_state_transition` (here) list legal edges in SQL; `FragmentState::next()` encodes the happy path in Rust; `INVALID_TRANSITIONS` re-lists illegal edges for tests. The only thing keeping them consistent is the ‘the two triggers must stay aligned’ comment. Adding or reordering a state means editing all of them in lockstep, and any miss surfaces only at runtime as an opaque `RAISE(ABORT)` or a mis-derived state that the proptests (testing their own copy of the rules) won't catch. Consider a single source of truth (Rust transition table) that the triggers are generated from, or drop the redundant enforcement layer.

nes-coordinator/model/src/worker/get.rs
```diff
@@ -0,0 +1,63 @@
...
+            .add_option(self.host_addr.clone().map(|v| Column::HostAddr.eq(v)))
+            .add_option(self.desired_state.map(|v| Column::DesiredState.eq(v)))
+            .add_option(self.current_state.map(|v| Column::CurrentState.eq(v)))
+    }
+}
+
+impl Execute for GetWorker {
+    type Response = Vec<Model>;
+    async fn execute(&self, conn: &impl ConnectionTrait) -> Result<Vec<Model>> {
+        Ok(worker::Entity::find()
```
keyseven123: **Reuse — the `get` `Execute` body is identical across all entity modules.** `Entity::find().filter(self.to_condition()).all(conn)` is duplicated verbatim in `worker/get.rs`, `sink/get.rs`, `ml_model/get.rs`, `source/logical/get.rs`, `source/physical/get.rs`. Since every request already implements the condition builder, a single generic helper (`fn get_all<E, C: IntoCondition>(cond, conn)`) collapses the five copies. Any cross-cutting change (ordering for stable output, pagination, richer error context) currently has to be hand-applied to each and will drift.

nes-coordinator/model/src/sink/drop.rs
```diff
@@ -0,0 +1,62 @@
...
+            .add_option(self.name.clone().map(|v| Column::Name.eq(v)))
+            .add_option(self.id.map(|v| Column::Id.eq(v)))
+    }
+}
+
+impl Execute for DropSink {
+    type Response = Vec<Model>;
+    async fn execute(&self, conn: &impl ConnectionTrait) -> Result<Vec<Model>> {
+        let condition = self.to_condition();
+        let sinks = sink::Entity::find()
```
keyseven123: **Efficiency — `drop` scans the matching rows twice.** `find(condition).all()` followed by `delete_many().filter(condition)` evaluates the same (unindexed, `Kind = Shared`) predicate twice — once to return the rows, once to delete. SQLite 3.35+ supports `DELETE … RETURNING *`, which does both in one statement. Same pattern in `ml_model/drop.rs`, `source/logical/drop.rs`, `source/physical/drop.rs`.

nes-coordinator/migration/src/m20260204_185834_init.rs
```diff
@@ -0,0 +1,660 @@
...
+                    .name("idx_fragment_query_id")
+                    .col(Fragment::QueryId)
+                    .to_owned(),
+            )
+            .await?;
+        manager
+            .create_index(
+                MigrationIndex::create()
+                    .table(Fragment::Table)
+                    .name("idx_fragment_host_addr_current_state")
```
keyseven123: **Efficiency — filtered columns without covering indexes.** `Fragment` gets `idx_fragment_host_addr_current_state`, but the equivalent per-host lookups on other tables are unindexed: `GetSink`/`GetPhysicalSource` filter `host_addr`, and `GetQuery`/`DropQuery` filter `name`/`state`, all of which fall back to full-table scans as the catalog grows. If any of these are on the reconciler/placement path, add matching indexes (`sink(host_addr)`, `physical_source(host_addr)`, `query(name)`/`query(state)`); the asymmetry with the Fragment index suggests they were simply missed.

keyseven123 (commented):
Minor stuff

nes-coordinator/migration/README.md
keyseven123: Heavily agree

nes-coordinator/migration/src/m20260204_185834_init.rs
```diff
@@ -0,0 +1,660 @@
...
+    limitations under the License.
+*/
+
+use crate::triggers::m20260204_185834_init as triggers;
+use crate::{assert_not_has_tables, drop_tables};
+use sea_orm::DbBackend;
+use sea_orm_migration::prelude::{Index as MigrationIndex, Table as MigrationTable, *};
+
+#[derive(DeriveMigrationName)]
+pub struct Migration;
```
keyseven123, line 26: I am missing here some reasoning what the Migration is or a reference to a DD or anything else.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
keyseven123: ☝️ @yschroeder97 your call

nes-coordinator/model/src/worker/get.rs
keyseven123: ☝️ @yschroeder97 your call

nes-coordinator/model/src/sink/drop.rs
keyseven123: ☝️ @yschroeder97 your call

nes-coordinator/model/src/worker/get.rs
keyseven123: See commit https://github.com/nebulastream/nebulastream/commit/35579a29f7f05e63599bf75272039ace471a9f78

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
keyseven123: https://github.com/nebulastream/nebulastream/commit/35579a29f7f05e63599bf75272039ace471a9f78

keyseven123 (changes requested):

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
yschroeder97: Good catch. `release_fragment_capacity` only runs on a terminal `UPDATE`, so
hard-deleting a query removes its fragments without giving their slots back, and
a bounded worker loses that capacity. I added `release_capacity_on_fragment_delete`
(`AFTER DELETE ON fragment`), with a check that skips fragments already in a
terminal state so their slots are not added back twice. Fixed in 37a98f16c8.

---

nes-coordinator/migration/src/lib.rs
yschroeder97: You are right. `.cascade()` is a no-op on SQLite and foreign keys are on during
migrations, so dropping a parent before its child fails once the catalog has
rows. I reordered `down()` to drop children before parents (reverse dependency
order) and rewrote the `drop_tables!` comment that claimed order did not matter.
Fixed in 37a98f16c8.

---

nes-coordinator/migration/src/m20260204_185834_init.rs
yschroeder97: You are right. `.cascade()` is a no-op on SQLite and foreign keys are on during
migrations, so dropping a parent before its child fails once the catalog has
rows. I reordered `down()` to drop children before parents (reverse dependency
order) and rewrote the `drop_tables!` comment that claimed order did not matter.
Fixed in 37a98f16c8.

---

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
yschroeder97: Agreed. Timestamps are now UTC: the `query` and `fragment`
timestamp columns are `DateTime<Utc>`, every writer uses `Utc::now()`, and the
worker-removed cascade trigger writes an ISO-8601 UTC string (`...T..:..:..+00:00`)
instead of `CURRENT_TIMESTAMP`. Since all timestamps now share one encoding, the
derive trigger's `MIN`/`MAX` text comparison is chronological again, which also
covers the mixed-format point @alepping raised. The CLI still shows local time:
`format.rs` converts UTC to local time when rendering. Fixed in 37a98f16c8.

yschroeder97: Agreed. Timestamps are now UTC across the board: the `query` and `fragment`
timestamp columns are `DateTime<Utc>`, every writer uses `Utc::now()`, and the
worker-removed cascade trigger writes an ISO-8601 UTC string (`...T..:..:..+00:00`)
instead of `CURRENT_TIMESTAMP`. Since all timestamps now share one encoding, the
derive trigger's `MIN`/`MAX` text comparison is chronological again, which also
covers the mixed-format point @alepping raised. The CLI still shows local time:
`format.rs` converts UTC to local when it renders. Fixed in 37a98f16c8.

---

nes-coordinator/migration/src/m20260204_185834_init.rs
yschroeder97: This one is intentional, and I should have made the reasoning clearer. A foreign
key would force both workers to exist before a link between them can be inserted,
but we want links to be declarable ahead of (or independently of) worker
registration. Removing a worker leaving its links behind is the same trade-off:
dangling edges are expected, and query planning resolves links against the live
worker set and ignores the dangling ones. I expanded the doc comment on
`network_link::Model` to spell this out.

Longer term my goal is to remove explicit link declaration entirely. Instead I
want automatic worker registration, with the optimizer (placement) figuring out
the best links itself, for example from measured network RTT. So I would rather
not tie links to worker rows via a foreign key now. Happy to revisit if we later
decide we want strict referential integrity here. 37a98f16c8

---

nes-coordinator/model/src/query/fragment.rs
yschroeder97: Fixed. The proto decode is now `TryFrom<i32>` instead of `From<i32>` and returns
the unrecognized value as the error rather than panicking. The one call site maps
that to `Outcome::Retry` (a transport error), so a worker on a newer or
incompatible protocol no longer crashes the coordinator: it retries within the
existing budget and eventually surfaces as a terminal transport failure. 37a98f16c8

nes-coordinator/model/src/query/fragment.rs
yschroeder97: You are right that value 3 on its own cannot tell a natural completion from a
stop. That distinction is kept via the fragment's desired state: 3 decodes to
Completed, and the controller records it as Stopped when the fragment was asked
to stop. I added a doc comment on the decode explaining this. I also removed the
reverse `FragmentState -> i32` impl (see the other thread), so the lossy encode
direction is gone. 37a98f16c8

nes-coordinator/model/src/query/drop.rs
yschroeder97: Thanks, I dug into this. The drop only records intent: it sets each fragment's
desired state to Stopped and the stop mode. It does not change the query's state,
which is derived from fragment current state and only changes later when the
controller reconciles. So re-fetching inside `execute` would hand back the same
still-running rows. Where the reconciled state actually matters, the blocking
drop path already covers it: it re-queries the queries by id after the stop
completes and returns their final state. For the non-blocking path, the drop-time
snapshot is the true current state. I documented this contract on `DropQuery`
rather than adding a redundant query. 37a98f16c8

nes-coordinator/model/src/sink/create.rs
yschroeder97: Fixed. Changed the message on the insert path from "failed to update sink" to
"failed to create sink". The inline-sink path already read correctly. 37a98f16c8

nes-coordinator/model/src/query/create.rs
yschroeder97: Fixed. Each write now carries context ("failed to insert query", "failed to
insert query fragments", "failed to link query sources", "failed to link query
sinks"), matching the rest of the crate, and I added the same to the follow-up
read of the created fragments. 37a98f16c8

nes-coordinator/model/src/format.rs
yschroeder97: Fixed both without adding a dependency. `row` now resizes a mismatched row to the
header count instead of relying on the debug-only assert, so a table function that
falls out of sync with its column list can no longer index out of bounds and crash
the CLI in release. Width is now measured in characters rather than bytes, so
multi-byte UTF-8 lines up with the borders. I kept the small renderer on purpose
(the output is almost all ASCII) and added a note saying so rather than pulling in
a table crate. 37a98f16c8

nes-coordinator/model/src/worker/endpoint.rs
yschroeder97: Fixed. `from_str` now accepts the bracketed `[host]:port` form for IPv6 literals
and rejects a bare unbracketed IPv6 (a host that still contains a colon) instead
of mis-splitting it on the last colon. `Display` brackets the host when it
contains a colon, so parse and display round-trip. I kept `NetworkAddr` rather
than switching to `SocketAddr` because we also allow hostnames, which
`SocketAddr` cannot represent. Added unit tests for the round-trip and the
rejection cases, and an IPv6 case to the generator so it is exercised. 37a98f16c8

nes-coordinator/model/Cargo.toml
yschroeder97: Done. `proptest` and `proptest-derive` are now optional dependencies behind a new
`testing` feature, and every Arbitrary impl and generator is gated with
`#[cfg(any(test, feature = "testing"))]`. To keep it readable I grouped each
file's generators into one gated `arbitrary` module (one gate per file rather than
a per-item attribute), moving the large query fixture into `query/arbitrary.rs`.
The production build no longer compiles proptest or any generator (verified with
`cargo build -p model`), and the enum Arbitrary derives are gated the same way.
`sink/` and `physical/` already gated their fixtures with `#[cfg(test)]` and stay
that way, since nothing outside this crate uses them. The one sibling crate that
needs the generators for its tests (the coordinator) enables `model/testing` in
its dev-dependencies. 37a98f16c8

nes-coordinator/model/src/worker/get.rs
yschroeder97: Done. Added a generic `get_all::<E>(filter, conn)` helper in lib.rs, and each Get's
execute delegates to it and adds its own error context. The five identical read
bodies are now one helper, so cross-cutting changes (ordering, pagination) live in
one place. This also caught the drift you predicted: `physical/get.rs` was the one
that had already diverged, missing the error context the others had, which the
helper path now fixes. 37a98f16c8

nes-coordinator/model/src/sink/drop.rs
yschroeder97: sea-orm 1.1 exposes `exec_with_returning` for insert and update but not delete, so
a true single-statement `DELETE ... RETURNING` would need hand-built raw SQL per
entity, which is not worth it on the cold drop path. Instead I collapsed the four
identical drop bodies into a `drop_all::<E>` helper (mirroring `get_all`), so the
two statements live in one place and a single-statement version becomes a one-line
change if the ORM gains support. This also fixed `physical/drop.rs`, which had
drifted (missing error context). 37a98f16c8

nes-coordinator/migration/README.md
yschroeder97: Added a section to `migration/README.md` on what a migration is and when to add
one, and a doc on the `Migration` struct. 37a98f16c8

nes-coordinator/model/src/ml_model/mod.rs
yschroeder97: Resolved in person: colocated tests are the Rust convention.

nes-coordinator/migration/src/m20260204_185834_init.rs
yschroeder97: Keeping the datetime name (already discussed on the thread): it is the sea-orm
convention, is used to order migrations, and correlates with the rest of the
history.

nes-coordinator/model/src/worker/mod.rs
yschroeder97: I documented the trust model rather than adding a validation trigger.
`worker.current_state` is observed status written by the controller, not an
engine-driven lifecycle like a fragment, so it is trusted rather than validated:
reachability can change back and forth, and since a worker is keyed by its
address, a removed address can be registered again, so every transition is valid.
I did not add a Removed-is-terminal guard for the same reason (it would block
re-registration). Documented on `apply_transition`. 37a98f16c8

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
yschroeder97: I trimmed and documented rather than building codegen. The layers serve different
purposes: the SQL triggers enforce transitions at write time (the authority), the
enums type the API, and `INVALID_TRANSITIONS` is an independent test oracle that
verifies the triggers reject illegal edges (it lists illegal edges, not a copy of
the legal rules). But `FragmentState::next()` was a genuinely redundant copy of the
happy path and, on checking, unused (the test walker uses `Ord`, not `next()`), so
I removed it. That takes the count from four places to three. I added a pointer on
`FragmentState` noting the trigger is the authority, and a keep-in-sync note in the
validate triggers pointing at the model enums, matching the CHECK-constraint
comments (@alepping's sqlite.up.sql:11 ask). 37a98f16c8

nes-coordinator/model/src/query/mod.rs
yschroeder97: Split. The generator's closure now delegates source and sink construction to two
named helpers (`build_physical_sources`, `build_sink`), so it reads as generate
fragments -> build sources -> build sink -> assemble, with no behavior change.
This code also moved into a gated `query/arbitrary.rs` module, so it is no longer in the main query module. 37a98f16c8

nes-coordinator/migration/src/m20260204_185834_init.rs
yschroeder97: I looked at where these filters are actually used, and I would rather not add
these four yet. In the production paths, physical sources are looked up by logical
source (already covered by `idx_physical_source_logical_source`), sinks by name
(already a unique index), and queries by id (the primary key). No production path
filters sinks or physical sources by host, or queries by name/state, so an index
there would add write cost without a read to serve.

The Fragment index is different on purpose: the reconciler scans fragments by
`(host_addr, current_state)` on every tick, which is a genuine hot path, so that
asymmetry is intentional rather than an oversight. Happy to add any of these the
moment a per-host source/sink scan or a state listing becomes a real path.

nes-coordinator/migration/src/m20260204_185834_init.rs
yschroeder97: The nullability is intentional: query-owned (inline/internal) sources have no
shared logical binding, so the column is NULL for them. These are unbound rows,
not danglers the FK could reattach, so keeping it nullable is correct. I added a
note to that effect on the migration column (the model field already documents
it). Good call on the index too: I made `idx_physical_source_logical_source`
partial (`WHERE logical_source IS NOT NULL`), so the NULL inline/internal rows are
excluded from the reverse-lookup index. sea-query supports this on both SQLite and
Postgres, so it stays backend-agnostic. 37a98f16c8

nes-coordinator/model/src/source/physical/create.rs
yschroeder97: Good catch, but I would not add a unique constraint on `(logical, host, type)`:
two shared sources that agree on those three but differ in config are actually
distinct (for example two FILE sources reading different paths), so a unique index
there would forbid a valid case. The problem is that the `if_not_exists`
lookup only matched that triple and ignored the config, so on a replay a
second same-triple source with a different config was skipped, and @alepping's
point about a differing config being ignored is the same issue. I changed the
lookup to match the full definition, including `source_config` and
`parser_config`, so replay is idempotent for a truly identical source, and a
different-config source is created as its own row. Added
`if_not_exists_creates_new_config` and adjusted `if_not_exists_returns_existing` to
the same-config case. 37a98f16c8

nes-coordinator/model/Cargo.toml
yschroeder97: Pinned both via the workspace. `sea-orm-migration` now lives in
`[workspace.dependencies]` at `=1.1.1` next to `sea-orm`, and `model` and
`migration` both reference it with `.workspace = true`. The two share types, so
now they are pinned in one place and move together. 37a98f16c8

nes-coordinator/model/src/query/create.rs
yschroeder97: The team settled on `QueryFragment`. Applied across the codebase: the table is now
`query_fragment`, the entity module is `query::query_fragment`, and the associated
types are renamed (`QueryFragmentState`, `QueryFragmentError`, `QueryFragmentTransition`,
`DesiredQueryFragmentState`, `CreateQueryFragment`, `QueryFragmentId`). Variable
names like `fragment` stay. 37a98f16c8

nes-coordinator/migration/src/m20260204_185834_init.rs
yschroeder97: Added a section to `migration/README.md` on what a migration is and when to add
one, and a doc on the `Migration` struct. 37a98f16c8

nes-coordinator/model/src/ml_model/mod.rs
yschroeder97: Resolved in person: colocated tests are the Rust convention.

nes-coordinator/model/src/ml_model/create.rs
yschroeder97: Acknowledged.

nes-coordinator/model/src/source/logical/create.rs
yschroeder97: Moved the gated `Arbitrary` generators out of the per-request files into sibling
`arbitrary.rs` files, the same pattern the query module already used, so the
request files are production-only. The gating stays
`#[cfg(any(test, feature = "testing"))]`, so nothing that was not already kept out
of production changes: the generators are still available to this crate's tests
and to downstream tests through the testing feature. 37a98f16c8

yschroeder97:
A global namespace is something we want. Today names are only unique within a
table, so a model and a logical source can share a name, and a hierarchical
identifier relation like the one you describe is a good way to get one global
view. It is out of scope for this PR. We plan to tackle it shortly after the
coordinator is merged.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
```diff
@@ -0,0 +1,269 @@
+-- Enforce the query lifecycle and reject illegal transitions. Terminal
```
greenBene, line 1: The .sql files don't have the license headers. Was this a conscious decision? I'd advocate for adding them here as well, especially since these .sql files aren't "just" small tests, but contain core business logic :)

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
```diff
@@ -0,0 +1,269 @@
...
+--   2. Propagate timestamps: start = MIN(query_fragment.start), stop =
+--      MAX(query_fragment.stop), each gated on the derived query state so
+--      timestamps never leak into states that should not carry them.
+--   3. If the query ended up Failed, aggregate per-host query_fragment
+--      errors into a {host_addr: message} JSON object.
+CREATE TRIGGER IF NOT EXISTS derive_query_state_on_query_fragment_update
+    AFTER UPDATE OF current_state ON query_fragment
+BEGIN
+    UPDATE query
+    SET state = COALESCE(
```
greenBene, line 116: I'd prefer to separate the COALESCE and the CASE :) 

```sql
SET newState = (SELECT CASE ...)
--- Maintain old state if no new state can be determined. 
SET state = COALSECE(newState, oldState)
```

greenBene: Never mind. The syntax doesn't allow temporary values, does it?

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
```diff
@@ -0,0 +1,269 @@
...
+    AFTER UPDATE OF desired_state ON worker
+    WHEN NEW.desired_state = 'Removed' AND OLD.desired_state != 'Removed'
+BEGIN
+    UPDATE query_fragment
+    SET current_state = 'Failed',
+        -- ISO-8601 UTC to match the app's DateTime<Utc> encoding, so the
+        -- derive trigger's MIN/MAX text comparison stays chronological.
+        -- CURRENT_TIMESTAMP would emit a space-separated, offset-less value
+        -- that does not compare correctly against app-written timestamps.
+        stop_timestamp = strftime('%Y-%m-%dT%H:%M:%f', 'now') || '+00:00',
```
greenBene, line 212: Why do we store the time as a string and not as a timestamp?

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
```diff
@@ -0,0 +1,269 @@
...
+        error = json_object('Transport', json_object('msg', 'worker removed'))
+    WHERE host_addr = NEW.host_addr
+    AND current_state NOT IN ('Completed', 'Stopped', 'Failed');
+END;
+
+-- Query-owned (Inline/Internal) sources are tied to exactly one query.
```
greenBene, line 218: Changed name from inline to anonymous :) Please update

nes-coordinator/migration/src/m20260204_185834_init.rs
```diff
@@ -0,0 +1,689 @@
...
+                    .col(
+                        ColumnDef::new(LogicalSource::Schema)
+                            .json_binary()
+                            .not_null(),
+                    )
+                    .to_owned(),
+            )
+            .await?;
+
+        manager
```
greenBene, line 80: I am a bit unsure about the way the tables are created. While I find the individual lines very understandable, I find it hard to parse the overall structure of the tables because each definition needs so much space.  This table with 6 columns and a check uses over 50 lines. Could we instead use SQL? Or at least make the rust code less spacy?

nes-coordinator/model/src/query/query_fragment/arbitrary.rs
```diff
@@ -0,0 +1,73 @@
...
+                let mut remaining: Vec<i32> = workers
+                    .iter()
+                    .map(|worker| worker.max_operators.unwrap_or(i32::MAX))
+                    .collect();
+
+                let mut fragments: Vec<CreateQueryFragment> = placements
+                    .into_iter()
+                    .map(|(idx, has_source)| {
+                        let cap = remaining[idx].min(12);
+                        let ops = (cap / 2).clamp(0, cap);
```
greenBene, line 55: Please use more descriptive variable names. Based on `ops`  i'd assume it has sth to do with operators, but why is it 50% of `cap`  and if we set ops to half od caps, why do we need to clamp it between 0 and cap? Shouldn't it by definition be less than cap if it is hald of cap?

greenBene (commented):
Tbh. I have the same issue that @Artraxon mentioned in the last retro: because it is Rust, I don't feel comfortable (yet?) reviewing/approving the code. I gave the feedback that I can in this review.

nes-coordinator/model/src/query/query_fragment.rs
```diff
@@ -0,0 +1,556 @@
...
+            if cond {
+                Ok(())
+            } else {
+                Err(DbErr::Custom(msg.into()))
+            }
+        };
+        match state {
+            QueryFragmentState::Pending | QueryFragmentState::Registered => {}
+            QueryFragmentState::Running => ensure(
+                self.start_timestamp.is_set(),
```
ls-1801, line 122: `is_set()` only distinguishes `ActiveValue::Set(_)` from `Unchanged`/`NotSet`, so `Set(None)` satisfies this check and can persist a Running fragment with a NULL start timestamp—likewise terminal timestamps and Failed errors below. Please validate that the inner `Option` is `Some`, with regression tests for the `Set(None)` cases.

ls-1801 (changes requested):

keyseven123 (approved):
The others have the remaining PR comments.

nes-coordinator/migration/src/m20260204_185834_init.rs
Artraxon: What does a worker represent here? A known worker or a worker reachable right now? I feel like that a known worker is way more useful from a users perspective, since this is something you might want to interact with and attach information to. On the other hand, if these are the currently reachable worker that are used by the coordinator, we should somehow try to enforce some degree of separation and make the differentiation explicit.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
yschroeder97: Fixed.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
yschroeder97: Trigger bodies cannot bind temporaries, and the fallback is documented in the header comment.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
yschroeder97: SQLite has no datetime type, the column is date_time(), and sea-orm encodes it as ISO-8601 text.

We have no choice here. Extended the comment to explain.

nes-coordinator/migration/src/triggers/m20260204_185834_init/sqlite.up.sql
yschroeder97: Did a full pass and renamed the remaining occurrences.

nes-coordinator/migration/src/m20260204_185834_init.rs
yschroeder97: I understand the concern, but SQL is not the solution here. 
The entire point of an ORM is to define the schema in code to be able to swap out the underlying database backend. 
If we use SQL, we need to duplicate all the logic for DDL. 

What I did instead is skip rustfmt (rust equivalent to clang-format) to keep the DDL more concise, and added a function for each table creation expression.

nes-coordinator/model/src/query/query_fragment/arbitrary.rs
yschroeder97: Renamed to `num_operators` and `remaining_capacity` to make the intent clear. 

The code generates arbitrary query fragments to exercise the property tests for the catalog.
I don't want to create "real" plans here just to test whether FK constraints and checks work, so we need some way to generate query fragments that are at least half-realistic.

The approach I chose here is to just give each fragment half of the remaining worker capacity to allow several fragments to share a worker. 
Clarified this in a comment.

nes-coordinator/model/src/query/query_fragment.rs
yschroeder97: Fixed. The three checks in `before_save` now match `Set(Some(_))` instead of calling `is_set()`, so a `Set(None)` no longer counts as providing the column. Extended stop_hooks and failed_hooks with a `Set(None)` step for the start timestamp, the stop timestamp and the error, and confirmed each of them fails against the old check

ls-1801 (approved):

greenBene (approved):

PR #1586, opened 2026-05-05
fix(sql-parser): fix negative float parsing error

### LINKED ISSUE #1572: [BUG] Negative floating-point constants (e.g. FLOAT64(-3.0)) reject queries
## Version / Commit

Reproduced on upstream `main`.

## What happened?

Any query that contains a negative floating-point literal — written as `FLOAT64(-3.0)` (or `FLOAT32(-3.0)`) — fails at compile time with:

```
Error: Query Failed with unexpected error: Multiple Exceptions:
    systest: query compiler / backend throwed an error; Can not parse constant value "-1" into unsigned long
(2032)
```

The error message refers to `"-1"` rather than the literal that appears in the query, which makes the failure mode harder to diagnose: there is no `-1` in the source SQL.

## What is the expected behavior?

`FLOAT64(-3.0)` (and `FLOAT32(-3.0)`) should produce a constant of value `-3.0`, identical to the workaround `FLOAT64(0) - FLOAT64(3.0)`. The grammar at `nes-sql-parser/AntlrSQL.g4` accepts the syntax, but the constant-folding / typed-literal pipeline appears to coerce the leading `-` token through an unsigned conversion before applying it to the typed constructor.

## How to reproduce

Save the following as `/tmp/neg_float_repro.test` and run with the systest binary:

```sql
# name: neg_float_repro
# groups: [test]

CREATE LOGICAL SOURCE s(id INT32 NOT NULL, v FLOAT64 NOT NULL);
ATTACH INLINE
1,-10.0
2,1.0

CREATE SINK out(s.id INT32 NOT NULL, s.v FLOAT64 NOT NULL) TYPE File;

SELECT * FROM s WHERE v < FLOAT64(-3.0) INTO out;
----
1,-10.0
```

Run:

```
./nes-systests/systest/systest -t /tmp/neg_float_repro.test --workingDir=/tmp/wd --data /test-file-cache
```

Actual output:

```
  1/1 (100.0%) neg_float_repro:01.......................................................................FAILED
SELECT * FROM s WHERE v < FLOAT64(-3.0) INTO out;
Error: Query Failed with unexpected error: Multiple Exceptions:
        systest: query compiler / backend throwed an error; Can not parse constant value "-1" into unsigned long
(2032)
```

Replacing `FLOAT64(-3.0)` with the equivalent `FLOAT64(0) - FLOAT64(3.0)` makes the query compile and produce the expected row `1,-10.0`. The repro is intentionally minimal — no joins, no windows, no UDFs — so the failure is isolated to the negative-literal handling itself.

## Operating System

macOS 24.6.0 host running the `nebulastream/nes-development:local` Docker dev image (Ubuntu).

## Anything else

- Reproduces every time on `main`.
- The error message naming `"-1"` rather than the literal that appears in the SQL is itself misleading and should be improved.

### LINKED ISSUE #1585: [BUG] Negative Typed Literals in Predicates Corrupt Logical Plan Construction
### Version / Commit

Reproduced in upstream `main`.

### What happened?

A negative typed float literal inside a `WHERE` predicate, for example `FLOAT64(-0.5)`, is parsed incorrectly.
Instead of treating -0.5 as the argument to `FLOAT64(...)`, the parser applies unary minus at the surrounding expression level. In the broken case, the logical plan becomes effectively:
`-1 * var > 0.5 AND var < 0.5`
or fails earlier during compilation with:

> Can not parse constant value "-1" into unsigned long

This causes query registration or compilation to fail.

`FLOAT64(-0.5)` should be parsed as a single negative floating-point constant.

This query:
```
  SELECT *
  FROM stream
  WHERE doublee > FLOAT64(-0.5) AND doublee < FLOAT64(0.5)
  INTO sink;
```

should behave like:
`
  WHERE doublee > -0.5 AND doublee < 0.5
`
and return only rows whose `doublee` value lies between -0.5 and 0.5.

### How to reproduce

  Use a query like this in a systest or via cli fronted:

```
  CREATE LOGICAL SOURCE stream(doublee FLOAT64 NOT NULL);
  CREATE PHYSICAL SOURCE FOR stream TYPE File;
  ATTACH INLINE
  -1.0
  -0.25
  0.25
  1.0

  CREATE SINK sinkStream(stream.doublee FLOAT64 NOT NULL) TYPE File;

  SELECT *
  FROM stream
  WHERE doublee > FLOAT64(-0.5) AND doublee < FLOAT64(0.5)
  INTO sinkStream;

```
  Observed result before the fix:

```
  - query compilation/registration fails, or
  - the predicate is misrendered as -1 * doublee > 0.5 AND doublee < 0.5
```

  Expected result:

```
  -0.25
  0.25

```

### Operating System

Ubuntu 24.04.3 LTS

### Deployment details

_No response_

### Anything else

_No response_

### Code of Conduct

tekdogan:
This pull request fixes incorrect parsing of negative typed float literals in `WHERE` predicates, such as `FLOAT64(-0.5)`. Previously, unary minus could bind to the wrong expression during logical plan construction, which led to malformed predicates or query compilation failures. See the [corresponding issue](https://github.com/nebulastream/nebulastream/issues/1585) for further details about the bug.

- Fixed unary minus handling in the SQL parser so negative constant literals remain attached to the literal instead of being rewritten against an unrelated expression.
- Changed unary negation fallback handling to use a signed constant instead of `UINT64(-1)`.
- Added parser-level regression coverage for negative typed float literals in `WHERE` predicates.
- Added a system-level regression test that validates the correct end-to-end behavior for `FLOAT64(-0.5)` in a predicate.

## Verifying this change
This change is tested by:
- Added a parser regression test in `StatementBinderTest` that verifies the predicate is parsed as `B > -0.5 AND B < 0.5`
- Added a systest `bug/NegativeFloatLiteralInWherePredicate.test` that validates correct query results for a negative typed float literal in a `WHERE` clause

- SQL Parser
- Logical plan construction
- QueryCompiler
- System-level tests
- Parser/unit test coverage

## Documentation

This PR closes #1585, closes #1572

### DIFF
--- a/nes-systests/bug/NegativeFloatLiteralInWherePredicate.test
+++ b/nes-systests/bug/NegativeFloatLiteralInWherePredicate.test
@@ -1,0 +1,39 @@
+# name: bug/NegativeFloatLiteralInWherePredicate.test
+# description: Reproduces the misparsing of negative typed literals in WHERE predicates during query registration
+# groups: [Bug, Selection, Parser]
+
+CREATE LOGICAL SOURCE stream(doublee FLOAT64 NOT NULL);
+CREATE PHYSICAL SOURCE FOR stream TYPE File;
+ATTACH INLINE
+-1.0
+-0.25
+0.25
+1.0
+
+CREATE SINK sinkStream(stream.doublee FLOAT64 NOT NULL) TYPE File;
+
+SELECT *
+FROM stream
+WHERE doublee > FLOAT64(-0.5) AND doublee < FLOAT64(0.5)
+INTO sinkStream;
+----
+-0.25
+0.25
+
+CREATE LOGICAL SOURCE intStream(value INT64 NOT NULL);
+CREATE PHYSICAL SOURCE FOR intStream TYPE File;
+ATTACH INLINE
+-2
+-1
+0
+1
+
+CREATE SINK sinkIntStream(intStream.value INT64 NOT NULL) TYPE File;
+
+SELECT *
+FROM intStream
+WHERE value > INT64(-2) AND value < INT64(1)
+INTO sinkIntStream;
+----
+-1
+0
--- a/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
+++ b/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
@@ -331,15 +331,39 @@ void AntlrSQLQueryPlanCreator::exitArithmeticUnary(AntlrSQLParser::ArithmeticUna
     {
         throw InvalidQuerySyntax("Parser is confused at {}", context->getText());
     }
-    LogicalFunction function;
+    auto opTokenType = context->op->getType();
+    if (context->valueExpression() != nullptr && !helpers.top().constantBuilder.empty())
+    {
+        auto& constant = helpers.top().constantBuilder.back();
+        if (constant == context->valueExpression()->getText())
+        {
+            switch (opTokenType)
+            {
+                case AntlrSQLLexer::PLUS:
+                    return;
+                case AntlrSQLLexer::MINUS:
+                    if (!constant.empty() && constant.front() == '-')
+                    {
+                        constant.erase(0, 1);
+                    }
+                    else
+                    {
+                        constant.insert(0, 1, '-');
+                    }
+                    return;
+                default:
+                    break;
+            }
+        }
+    }
 
     if (helpers.top().functionBuilder.empty())
     {
         throw InvalidQuerySyntax("Expected unary operator, got nothing: {}", context->getText());
     }
+    LogicalFunction function;
     const auto innerFunction = helpers.top().functionBuilder.back();
     helpers.top().functionBuilder.pop_back();
-    auto opTokenType = context->op->getType();
     switch (opTokenType)
     {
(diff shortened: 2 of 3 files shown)


nes-systests/bug/NegativeFloatLiteralInWherePredicate.test
```diff
@@ -0,0 +1,21 @@
...
+-1.0
+-0.25
+0.25
+1.0
+
+CREATE SINK sinkStream(stream.doublee FLOAT64 NOT NULL) TYPE File;
+
+SELECT *
+FROM stream
+WHERE doublee > FLOAT64(-0.5) AND doublee < FLOAT64(0.5)
```
keyseven123, line 17: Maybe we can have a second where or query that tests minus in INT64(). Just to make sure that quite a similar but does not arise.

keyseven123 (changes requested):
Thanks for fixing this bug. Just one small minor comment

ls-1801 (approved):

nes-systests/bug/NegativeFloatLiteralInWherePredicate.test
tekdogan: done

keyseven123 (approved):

PR #1588, opened 2026-05-06
fix(Aggregation): Skip NULL inputs in MIN/MAX/SUM/AVG/MEDIAN

### LINKED ISSUE #1573: [BUG] MIN/SUM/AVG/MEDIAN propagate NULL instead of skipping per SQL-standard semantics
## Version / Commit

Reproduced on upstream `main`.

## What happened?

The nullable branches of every aggregation in `nes-physical-operators/src/Aggregation/Function/` mark the accumulator as NULL whenever **any** input row is NULL, instead of only when **every** input row is NULL. Concretely, in each of `Min`, `Sum`, `Avg`, `Median`, and (until a local fix) `Max`AggregationPhysicalFunction.cpp the `lift` does:

```cpp
const auto newIsNull = isNull or value.isNull();   // ← any-null poisons the result
```

and `combine` does:

```cpp
const auto newIsNull = isNull1 or isNull2;          // ← same in the merge step
```

so the output is NULL the moment a single NULL row enters the window, even though the other rows hold real values. SQL `MAX`, `MIN`, `SUM`, `AVG`, and `MEDIAN` are defined to skip NULL inputs and only return NULL when no non-NULL input exists.

## What is the expected behavior?

For each of MAX/MIN/SUM/AVG/MEDIAN: skip NULL inputs; return NULL only when every input row was NULL. The accumulator must therefore start as NULL (no value seen yet) and turn non-null as soon as the first non-null input arrives. The merge step should yield the merged accumulator if at least one side has a non-null value.

## How to reproduce

Save as `/tmp/agg_null_skip.test`:

```sql
# name: agg_null_skip
# groups: [Aggregation, NullHandling]

CREATE LOGICAL SOURCE s(ts UINT64 NOT NULL, v INT32);
ATTACH INLINE
0,
10,3
20,
30,1
40,

CREATE SINK out(s.start UINT64 NOT NULL, s.end UINT64 NOT NULL, vSum INT32, vMin INT32, vMax INT32, vAvg FLOAT64, vMedian FLOAT64) TYPE File;

# Window [0,100) inputs: {NULL, 3, NULL, 1, NULL}
# Per SQL standard: SUM=4, MIN=1, MAX=3, AVG=2.0, MEDIAN=2.0  (NULL inputs skipped)
SELECT start, end,
       SUM(v) AS vSum,
       MIN(v) AS vMin,
       MAX(v) AS vMax,
       AVG(v) AS vAvg,
       MEDIAN(v) AS vMedian
FROM s
WINDOW TUMBLING(ts, size 100 ms)
INTO out;
----
0,100,4,1,3,2.0,2.0
```

Run:

```
./nes-systests/systest/systest -t /tmp/agg_null_skip.test --workingDir=/tmp/wd --data /test-file-cache
```

Actual output (every aggregate poisoned by the NULL inputs):

```
0,100,NULL,NULL,NULL,NULL,NULL
```

A second failing case demonstrating the keyed-aggregation path (taken from `nes-systests/operator/aggregation/WindowAggregationNull.test` query 3, with SQL-standard expectations restored):

```sql
CREATE LOGICAL SOURCE stream2(ts UINT64 NOT NULL, id INT32, value INT32);
ATTACH INLINE
0,,
10,,1
1100,,
1110,,4

CREATE SINK sink2(stream2.start UINT64 NOT NULL, stream2.end UINT64 NOT NULL, stream2.id INT32, vSum INT32, vMin INT32, vMax INT32, vAvg FLOAT64, vCnt UINT64 NOT NULL, rCnt UINT64 NOT NULL) TYPE File;

# id=NULL group, [0,1000): values = {NULL, 1}
# Per SQL standard: SUM=1, MIN=1, MAX=1, AVG=1.0
# id=NULL group, [1000,2000): values = {NULL, 4}
# Per SQL standard: SUM=4, MIN=4, MAX=4, AVG=4.0
SELECT start, end, id,
       SUM(value)   AS vSum,
       MIN(value)   AS vMin,
       MAX(value)   AS vMax,
       AVG(value)   AS vAvg,
       COUNT(value) AS vCnt,
       COUNT(ts)    AS rCnt
FROM stream2
WHERE ISNULL(id)
GROUP BY id
WINDOW TUMBLING(ts, size 1 sec)
INTO sink2;
----
0,1000,NULL,1,1,1,1.0,1,2
1000,2000,NULL,4,4,4,4.0,1,2
```

Currently both windows return `NULL` for `vSum`, `vMin`, `vMax`, `vAvg`. (The current `WindowAggregationNull.test` codifies the wrong behavior in its expected outputs and should be updated alongside the fix.)

## Example correction (MAX)

The same change pattern applies to MIN/SUM/AVG/MEDIAN. For `MaxAggregationPhysicalFunction`:

`reset` — start as NULL (no values seen), not as the type's MIN sentinel labeled non-null:

```cpp
if (inputType.nullable)
{
    storeNull(aggregationState, true);   // was: false
}
```

`lift` — adopt the new value only when it is non-null and either the accumulator is empty or the new value beats the current max; only stay NULL while every input has been NULL:

```cpp
const auto isNull = readNull(aggregationState);
const auto memAreaMax = static_cast<nautilus::val<int8_t*>>(aggregationState + nautilus::val<uint64_t>{1});
const auto max = VarVal::readVarValFromMemory(memAreaMax, inputType, isNull);

const auto valueIsNotNull = not value.isNull();
const auto adoptValue = valueIsNotNull and (isNull or (value > max).getRawValueAs<nautilus::val<bool>>());
const auto newMax = VarVal::select(adoptValue, value, max);
newMax.writeToMemory(memAreaMax);

const auto newIsNull = isNull and value.isNull();   // was: or
storeNull(aggregationState, newIsNull);
```

`combine` — null only when both partitions saw nothing but nulls; if one side is empty take the other:

```cpp
const auto newIsNull = isNull1 and isNull2;          // was: or
storeNull(aggregationState1, newIsNull);

const auto max1OrMax2 = VarVal::select((max1 > max2).getRawValueAs<nautilus::val<bool>>(), max1, max2);
const auto newMax = VarVal::select(isNull1, max2, VarVal::select(isNull2, max1, max1OrMax2));
newMax.writeToMemory(memAreaMax1);
```

Mirror this across `Min`, `Sum`, `Avg`, and `Median`AggregationPhysicalFunction.cpp:
- `Sum`: adopt-rule replaced with `newSum = isNull ? value : sum + value` when value is non-null; carry over otherwise.
- `Min`: same as `Max` with `<` instead of `>`.
- `Avg`: needs both running sum and running count to skip NULL inputs from both.
- `Median`: stop folding `value.isNull()` into the `containsNull` flag; treat NULL inputs as no-op (don't append to the paged vector).

## Anything else

- Discovered while building a windowed pivot via `MAX(value_or_null(predicate, value))` — the SQL pattern that replaces stateful UDFs in the Charité retrospective study. The pivot can't work at all under any-null-poisons semantics because every per-Bez slot is NULL for the rows belonging to the other Bezes.
- `nes-systests/operator/aggregation/WindowAggregationNull.test` currently encodes the broken any-null-poisons expectations as the test "truth"; it should be updated when the fix lands.
- The fix to MAX in isolation (locally applied and verified against the existing test suite) does not regress the 69 tests in `nes-systests/operator/aggregation/`.

keyseven123:

This PR fixes the aggregation behavior for `NULL` inputs so that `MIN`, `MAX`, `SUM`, `AVG`, and `MEDIAN` skip `NULL` values per the SQL standard.

Closes #1573 . See the issue for full details.

## Verifying this change

Covered by the systests added in this PR.

### DIFF
--- a/docs/technical/null_handling.md
+++ b/docs/technical/null_handling.md
@@ -247,10 +247,12 @@ SELECT * FROM left INNER JOIN right ON (id = id2) WINDOW TUMBLING(ts, size 1 sec
 
 #### SUM, AVG, MIN, MAX, MEDIAN
 
-If **any input value is NULL**, the aggregation result is **NULL**.
+NULL inputs are **skipped**. The aggregate is only `NULL` when **every** input in the window is `NULL`; otherwise the aggregation is computed over the non-null inputs alone (SQL-standard behavior).
 
 ```sql
--- Source: ts=0 value=0 | ts=20 value=3 | ts=50 value=1 | ts=100 value=2 | ts=150 value=NULL
+-- Source: ts=0 value=0 | ts=20 value=3 | ts=50 value=1
+--         ts=100 value=2 | ts=150 value=NULL
+--         ts=300 value=NULL | ts=350 value=NULL
 
 SELECT SUM(value), MAX(value), MIN(value), AVG(value), MEDIAN(value)
 FROM stream WINDOW TUMBLING(ts, size 100 ms);
@@ -259,16 +261,12 @@ FROM stream WINDOW TUMBLING(ts, size 100 ms);
 | Window      | SUM  | MAX  | MIN  | AVG  | MEDIAN |
 |-------------|------|------|------|------|--------|
 | [0, 100)    | 4    | 3    | 0    | 1.33 | 1      |
-| [100, 200)  | NULL | NULL | NULL | NULL | NULL   |
+| [100, 200)  | 2    | 2    | 2    | 2.0  | 2.0    |
+| [300, 400)  | NULL | NULL | NULL | NULL | NULL   |
 
-Window [0, 100) contains only non-null values (0, 3, 1), so aggregations work normally.
-Window [100, 200) contains value=2 and value=NULL — the presence of NULL makes the result NULL.
-
-To aggregate only over non-null values, pre-filter with `NOT ISNULL()`:
-
-```sql
-SELECT SUM(value) FROM stream WHERE NOT ISNULL(value) WINDOW TUMBLING(ts, size 100 ms);
-```
+- Window `[0, 100)` contains only non-null values `{0, 3, 1}`.
+- Window `[100, 200)` contains `{2, NULL}`. The `NULL` is skipped and the aggregations are computed over `{2}`.
+- Window `[300, 400)` contains only `NULL` values, so every aggregation returns `NULL`.
 
 #### COUNT
 
@@ -294,7 +292,9 @@ FROM stream GROUP BY id WINDOW TUMBLING(ts, size 1 sec);
 NULL keys form **their own group** — all rows with a NULL key are aggregated together.
 
 ```sql
--- Source: (ts=0, id=NULL, value=NULL), (ts=10, id=NULL, value=1), (ts=20, id=1, value=2), (ts=40, id=3, value=NULL)
+-- Source: (ts=0, id=NULL, value=NULL), (ts=10, id=NULL, value=1),
+--         (ts=20, id=1, value=2),     (ts=30, id=1, value=3),
+--         (ts=40, id=3, value=NULL),  (ts=50, id=3, value=NULL)
 
 SELECT id, SUM(value), COUNT(value), COUNT(ts)
 FROM stream GROUP BY id WINDOW TUMBLING(ts, size 1 sec);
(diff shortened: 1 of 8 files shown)


ls-1801 (changes requested):
re-request once rebased, to latest ci changes

keyseven123:
> re-request once rebased, to latest ci changes

As I stacking on top of @Artraxon , I am depending on the rebase of the base branches to include the ci changes.

adrianmichalke (approved):

ls-1801 (approved):

PR #1593, opened 2026-05-11
feat(optimizer) barrier rules

### LINKED ISSUE #1583: [TASK] Implement Barrier Rules
### Description

To simplify the dependency graph of optimization rule, we introduce the concept of barrier rules. 
These rule do not change any given input itself, but help to bundle dependencies. 

### Acceptance Criteria

Barrier rules exists, can be expressed, and are used in existing code

greenBene:

This PR adds the concept of barrier rules in the logical optimizer.
Their aim is to not implement any optimizations themselves, but to make the dependencies between rules easier to depict.  

To make this behavior easier to debug, the function `RuleManager::explain()` was also added.

## Verifying this change
* All existing tests continue to run 

* optimizer

This PR closes #1583

### DIFF
--- a/nes-query-optimizer/include/Rules/Barriers/FixedPlanStructureBarrier.hpp
+++ b/nes-query-optimizer/include/Rules/Barriers/FixedPlanStructureBarrier.hpp
@@ -1,0 +1,41 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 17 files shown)


nes-input-formatters/src/InputFormatIndexer.cpp
Artraxon, line 1: Why do we need this virtual interface?
What kind of logic should depend on whether a rule is a barrier or not? This PR does not use this interface.
IMO whether something is a barrier or not is an implementation detail, at the end its just a rule that does nothing.

And even if we would find a good use for this somewhere, since it's not needed right now I'd prefer if we remove it.

Artraxon (changes requested):

nes-input-formatters/src/InputFormatIndexer.cpp
greenBene: Very good point, I removed it.

adrianmichalke (approved):
LGTM

As discussed, in the long run I would personally prefer to see dedicated optimizer phases instead

Artraxon (approved):

PR #1595, opened 2026-05-11
feat(optimizer) projection pushdown

greenBene:
Added the textbook projection pushdown optimizer rule to NebulaStream. 

## Verifying this change
This change is tested by
* existing systest continues to run
* newly implemented unit tests verify behaviour. 

- Optimizer 

## Documentation
The rule is documented in the related hpp file.

This PR closes #1526


nes-query-optimizer/tests/Rules/Static/ProjectionPushdownRuleTest.cpp
```diff
@@ -0,0 +1,500 @@
...
+    ASSERT_TRUE(op3.getOutputSchema().getFieldByName(Identifier::parse("a")));
+    ASSERT_TRUE(op3.getOutputSchema().getFieldByName(Identifier::parse("c")));
+    auto op4 = op3.getChildren().at(0);
+    ASSERT_TRUE(op4.tryGetAs<SourceDescriptorLogicalOperator>());
+    ASSERT_EQ(op4.getOutputSchema().size(), 4);
+}
+
+TEST_F(ProjectionPushdownRuleTest, ValidatePushBeyondJoin)
+{
+    /// PLAN BEFORE: sink(a) < projection(a) < join(b=c) < (source(a,b,x)),(source(c,d,y))
```
keyseven123: I like these comments a lot :)

nes-query-optimizer/tests/Rules/Static/ProjectionPushdownRuleTest.cpp
```diff
@@ -0,0 +1,500 @@
...
+
+    auto rightOp2 = rightOp1.getChildren().at(0);
+    ASSERT_TRUE(rightOp2.tryGetAs<SourceDescriptorLogicalOperator>());
+    ASSERT_EQ(rightOp2.getOutputSchema().size(), 3);
+    ASSERT_TRUE(rightOp2.getOutputSchema().getFieldByName(Identifier::parse("c")));
+    ASSERT_TRUE(rightOp2.getOutputSchema().getFieldByName(Identifier::parse("d")));
+    ASSERT_TRUE(rightOp2.getOutputSchema().getFieldByName(Identifier::parse("y")));
+}
+
+TEST_F(ProjectionPushdownRuleTest, ValidatePushBeyondUnion)
```
keyseven123, line 244: I would like to have an additional test for a join. As this might add some complexity due to it having two different schemas.

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+    auto newChild = projectionPushdown(op->getChild(), newRequired);
+    return TypedLogicalOperator<ProjectionLogicalOperator>{newChild, newProjections, ProjectionLogicalOperator::Asterisk{false}};
+}
+
+LogicalOperator
+ProjectionPushdownRule::pushBeyondJoin(const TypedLogicalOperator<JoinLogicalOperator>& op, const std::unordered_set<Field>& required) const
+{
+    /// Add accessed identifiers in join predicate if necessary
+    /// Apply recursion with two both children, each with relevant subset of identifiers that come from the individual child
```
keyseven123:
```suggestion
    /// Apply recursion with both children, each with a relevant subset of identifiers that come from the individual child
```

keyseven123 (changes requested):
Preliminary feedback, waiting on CI to be green.

keyseven123 (approved):

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Thanks!

nes-systests/operator/union/Union.test
```diff
@@ -49,7 +49,7 @@ SELECT * FROM stream UNION SELECT * FROM stream2 INTO sink;
 2,2,2003
 
 # Union with projection to fixup the schema
-SELECT * FROM stream UNION SELECT id2 as id, value, timestamp FROM stream3 INTO sink;
+SELECT id, value, timestamp FROM stream UNION SELECT id2 as id, value, timestamp FROM stream3 INTO sink;
```
keyseven123:
```suggestion
SELECT id, timestamp FROM stream UNION SELECT id2 as id, value, timestamp FROM stream3 INTO sink;
```

Then we also test a projection in the union

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
```diff
@@ -43,6 +45,22 @@ UnionRenamePhysicalOperator::UnionRenamePhysicalOperator(
     : inputFields(std::move(inputFields)), outputFields(std::move(outputFields))
 {
     PRECONDITION(this->inputFields.size() == this->outputFields.size(), "Input and output fields must have the same size");
+
+    std::map<uint32_t, uint32_t> map;
+
+    for (uint32_t i = 0; i < this->inputFields.size(); ++i)
```
keyseven123: Why u32?

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
```diff
@@ -43,6 +45,22 @@ UnionRenamePhysicalOperator::UnionRenamePhysicalOperator(
     : inputFields(std::move(inputFields)), outputFields(std::move(outputFields))
 {
     PRECONDITION(this->inputFields.size() == this->outputFields.size(), "Input and output fields must have the same size");
+
+    std::map<uint32_t, uint32_t> map;
```
keyseven123: Also why are we not using the field names, e.g., strings? This helps during debugging

keyseven123 (changes requested):

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
yschroeder97: Don't call it map, but instead what the map contains/represents.

nes-query-optimizer/include/Rules/Static/ProjectionPushdownRule.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <typeinfo>
+
+#include <Plans/LogicalPlan.hpp>
+#include <Rules/Rule.hpp>
+
+namespace NES
+{
+
+/**
+ * @brief Eliminates unused fields as close to source operators as possible.
```
yschroeder97, line 29: Great comment, appreciate it!

nes-query-optimizer/include/Rules/Static/ProjectionPushdownRule.hpp
```diff
@@ -0,0 +1,59 @@
...
+ * - Sink: seeds required from the child output schema.
+ * - Source: recursion stops; a projection is inserted above the source when required is a strict subset of the output schema.
+ * - Selection: extends required with all fields accessed by the predicate.
+ * - Projection: prunes entries not in required; expands required via the accessed fields of retained projection expressions.
+ * - Join: partitions required and join-predicate fields into left/right subsets and recurses independently into each child.
+ * - Union: propagates required into each branch, re-mapping field identities per child schema.
+ * - Event-time watermark assigner: extends required with the event-time field.
+ * - Ingestion-time watermark assigner: passes required through transparently.
+ * - Windowed aggregation: retains only required aggregations; seeds required with their input fields and all grouping keys.
+ * - Unknown operators: recurses on each child with its full output schema; inserts a projection for required fields above the operator.
```
yschroeder97, line 41: What is an unknown operator and why do you need to handle it here?

nes-query-optimizer/src/Phases/RuleBasedOptimizer.cpp
```diff
@@ -58,6 +60,10 @@ LogicalPlan RuleBasedOptimizer::optimize(LogicalPlan plan) const
     for (const auto& rule : ruleSequence)
     {
         plan = rule.apply(std::move(plan));
+
+        /// TODO #1644: Remove the `replaceSubtree` anti-pattern from optimization rules since it can cause invalid query plans
```
yschroeder97: How can `replaceSubtree` produce an invalid plan?

yschroeder97: Also, why is the comment here and not where the "anti-pattern" is actually used?

nes-systests/operator/union/Union.test
greenBene: Makes sense! :)  I adapted the test.

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
greenBene: Changed the map to use the field identifiers. The integer index is no longer required.

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
greenBene: Makes sense. I changed it to the field identifiers.

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
```diff
@@ -43,6 +45,22 @@ UnionRenamePhysicalOperator::UnionRenamePhysicalOperator(
...
+
+    for (uint32_t i = 0; i < this->inputFields.size(); ++i)
+    {
+        for (uint32_t j = 0; j < this->outputFields.size(); ++j)
+        {
+            if (this->inputFields.at(i) == this->outputFields.at(j))
+            {
+                PRECONDITION(!map.contains(i), "Input field cannot be mapped twice");
```
yschroeder97:
```suggestion
                INVARIANT(!map.contains(i), "Input field cannot be mapped twice");
```

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
```diff
@@ -43,6 +45,22 @@ UnionRenamePhysicalOperator::UnionRenamePhysicalOperator(
...
+                map.emplace(i, j);
+            }
+        }
+    }
+    PRECONDITION(map.size() == this->inputFields.size(), "All input fields must be mapped");
```
yschroeder97:
```suggestion
    INVARIANT(map.size() == this->inputFields.size(), "All input fields must be mapped");
```

...those are rather invariants, preconditions validate the input parameters of a function.

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+    for (const auto& agg : op->getWindowAggregation())
+    {
+        aggregationNames.insert(agg.name);
+        if (!requiredNames.contains(agg.name))
+        {
+            continue;
+        }
+        newAggregations.push_back(agg);
+
+        PRECONDITION(
```
yschroeder97:
```suggestion
        INVARIANT(
```

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+LogicalOperator pushBeyondIngestionTimeWatermarkAssigner(
+    const TypedLogicalOperator<IngestionTimeWatermarkAssignerLogicalOperator>& op, const std::unordered_set<Field>& required)
+{
+    /// No new required fields are added in ingestion time watermark assigner operator.
+
+    std::unordered_set<Field> newRequired;
+    for (const auto& field : required)
+    {
+        auto newField = op->getChild().getOutputSchema()[field.getLastName()];
+        PRECONDITION(newField.has_value(), "the given field must be available in the child operator");
```
yschroeder97:
```suggestion
        INVARIANT(newField.has_value(), "the given field must be available in the child operator");
```

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+LogicalOperator pushBeyondEventTimeWatermarkAssigner(
+    const TypedLogicalOperator<EventTimeWatermarkAssignerLogicalOperator>& op, const std::unordered_set<Field>& required)
+{
+    /// Add eventTime field if necessary
+
+    auto newRequired = getAccessedFields(op->getOnField());
+    for (const auto& field : required)
+    {
+        auto newField = op->getChild().getOutputSchema()[field.getLastName()];
+        PRECONDITION(newField.has_value(), "the given field must be available in the child operator");
```
yschroeder97:
```suggestion
        INVARIANT(newField.has_value(), "the given field must be available in the child operator");
```

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+    /// no new required fields are added in union operator
+
+    std::vector<LogicalOperator> newChildren;
+    for (const auto& child : op->getChildren())
+    {
+        std::unordered_set<Field> newRequired;
+        for (const auto& field : required)
+        {
+            auto newField = child.getOutputSchema()[field.getLastName()];
+            PRECONDITION(newField.has_value(), "the given field must be available in the plan");
```
yschroeder97:
```suggestion
            INVARIANT(newField.has_value(), "the given field must be available in the plan");
```

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+    auto joinTimeCharacteristics = op->getJoinTimeCharacteristics();
+
+    PRECONDITION((std::holds_alternative<std::array<Windowing::BoundTimeCharacteristic, 2>>(op->getJoinTimeCharacteristics())), "true ");
+
+    for (const auto& timeCharacteristic : std::get<std::array<Windowing::BoundTimeCharacteristic, 2>>(joinTimeCharacteristics))
+    {
+        if (std::holds_alternative<Windowing::BoundEventTimeCharacteristic>(timeCharacteristic))
+        {
+            auto field = std::get<Windowing::BoundEventTimeCharacteristic>(timeCharacteristic).field->getField();
+            PRECONDITION(
```
yschroeder97:
```suggestion
            INVARIANT(
```

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+    /// Apply recursion to both children, each with the relevant subset of required fields that come from the individual child
+
+    auto [left, right] = op->getBothChildren();
+    std::unordered_set<Field> requiredLeft;
+    std::unordered_set<Field> requiredRight;
+
+    auto predicate = op->getJoinFunction();
+    auto joinTimeCharacteristics = op->getJoinTimeCharacteristics();
+
+    PRECONDITION((std::holds_alternative<std::array<Windowing::BoundTimeCharacteristic, 2>>(op->getJoinTimeCharacteristics())), "true ");
```
yschroeder97:
```suggestion
    INVARIANT((std::holds_alternative<std::array<Windowing::BoundTimeCharacteristic, 2>>(op->getJoinTimeCharacteristics())), "true ");
```

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+            {
+                newRequired.insert(accessed);
+            }
+            newProjections.emplace_back(field.getLastName(), function);
+        }
+        else
+        {
+            /// This case is required if the original projection used an asterix and thus forwarded all child fields.
+            auto newField = op->getChild().getOutputSchema()[field.getLastName()];
+            PRECONDITION(
```
yschroeder97:
```suggestion
            INVARIANT(
```

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+LogicalOperator pushBeyondSelection(const TypedLogicalOperator<SelectionLogicalOperator>& op, const std::unordered_set<Field>& required)
+{
+    /// Add accessed identifiers in predicate if necessary
+
+    auto newRequired = getAccessedFields(op->getPredicate());
+    for (const auto& field : required)
+    {
+        auto newField = op->getChild().getOutputSchema()[field.getLastName()];
+        PRECONDITION(newField.has_value(), "all fields in required set must be in child output schema");
```
yschroeder97:
```suggestion
        INVARIANT(newField.has_value(), "all fields in required set must be in child output schema");
```

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+/// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+std::set<std::type_index> ProjectionPushdownRule::dependsOn() const
+{
+    return {};
+}
+
+/// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+std::set<std::type_index> ProjectionPushdownRule::requiredBy() const
+{
+    return {typeid(FixedPlanStructureBarrier)};
```
yschroeder97, line 480: Would like a comment here explaining why this requirement. What does it mean?

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -0,0 +1,487 @@
...
+
+    std::vector<ProjectionLogicalOperator::UnboundProjection> newProjections;
+    std::unordered_set<Field> newRequired;
+
+    for (const auto& [field, function] : projections)
+    {
+        projectionMap.emplace(field, function);
+    }
+
+    for (const auto& field : required)
```
yschroeder97, line 137: For aggregations, you comment "/// Iterating over `required` (unordered_set) would produce non-deterministic ordering of newAggregations,
    /// causing downstream positional column matching to read wrong aggregation results.".

Isn't this the same issue of non-deterministic orderings of the projection fields when using an `std::unordered_set`?

Please validate this, also for `pushBeyondSource` and `pushBeyondDefault`.

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
```diff
@@ -31,18 +33,35 @@ namespace NES
...
     std::vector<QualifiedIdentifier> inputFields, std::vector<QualifiedIdentifier> outputFields)
-    : inputFields(std::move(inputFields)), outputFields(std::move(outputFields))
+    : outputFields(std::move(outputFields))
 {
-    PRECONDITION(this->inputFields.size() == this->outputFields.size(), "Input and output fields must have the same size");
+    PRECONDITION(inputFields.size() == this->outputFields.size(), "Input and output fields must have the same size");
+
+    std::unordered_map<QualifiedIdentifier, QualifiedIdentifier> map;
+
+    for (const auto& inputField : inputFields)
```
yschroeder97, line 47: - An operator literally named Rename can no longer rename: if a union branch's output ever has a fully-qualified name that differs from the union output name, the precondition aborts the query at runtime. The commit intent ("map
  fields instead of assuming correct order") is about reordering, which this achieves — but it narrows the contract. Worth a union system test with branches whose qualified names differ, to confirm the new Field model really does keep them identical across branches.
- It's also `O(n²)` where a single hash lookup would do.

nes-query-optimizer/src/Phases/RuleBasedOptimizer.cpp
```diff
@@ -58,6 +60,10 @@ LogicalPlan RuleBasedOptimizer::optimize(LogicalPlan plan) const
...
+        /// and require manual `.withInferredSchema` calls as the one below.
+        plan = plan.withRootOperators({plan.getRootOperators().at(0).withInferredSchema()});
```
yschroeder97: Could pair this with an INVARIANT that asserts that we indeed only have a single sink in this plan to avoid problems when we introduce that.

nes-physical-operators/include/UnionRenamePhysicalOperator.hpp
```diff
@@ -35,6 +37,7 @@ class UnionRenamePhysicalOperator final : public PhysicalOperatorConcept
 private:
     std::vector<QualifiedIdentifier> inputFields;
```
yschroeder97: This is dead code now, no?

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
yschroeder97: Also, what about making the message more specific?

nes-physical-operators/include/UnionRenamePhysicalOperator.hpp
```diff
@@ -13,6 +13,8 @@
 */
 #pragma once
 
+#include <cstdint>
+#include <map>
```
yschroeder97: Should include `unordered_map`.

yschroeder97 (changes requested):
Thank you :)
 
Mostly minor things.

yschroeder97 (changes requested):
Thank you :)
 
Mostly minor things.

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
greenBene: Renamed it to identifierMap

nes-query-optimizer/include/Rules/Static/ProjectionPushdownRule.hpp
greenBene: Thanks :)

nes-query-optimizer/include/Rules/Static/ProjectionPushdownRule.hpp
greenBene: An unknown operator is an operator added by a plugin. The rule should still push projections as far as possible, but still compile and work with unknown operators.

nes-query-optimizer/src/Phases/RuleBasedOptimizer.cpp
greenBene: I added the comment here because the workaround is applied here. 

Replace Subtree can produce invalid plans because it is based on OperatorIDs of the operators which can be invalidated during the call to the function.  Compare `RedundantUnionRemovalRule::apply`.

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
greenBene: Thanks! Changed it :)

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
greenBene: Thanks! Changed it :)

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Changed it.

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Changed it :)

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Changed it :)

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Changed it :)

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Changed it

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Very good point! Overlooked it.

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Changed it

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: Changed it

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: See https://github.com/nebulastream/nebulastream/pull/1593

I do agree that proper documentation for the optimizer is required. 

The basic idea is that barriers (which are rules that don't do anything) are an abstraction for the dependencies, so not every rule that requires a fixed operator order must directly depend on all rules that change the operator order. 

I'd propose adding proper documentation once the optimizer is a bit more stable (which should be soon), but avoid adding a comment here, since basically the same dependency exists on all other rules that change the operator order.

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: The core idea of the schema inference branch is that the order of fields should not matter here :) 

I found a related bug in the Union logic, which I fixed in a commit in this PR. 

This assurance is not (yet?) given for the aggregation functions; thus, the order is manually ensured.

nes-physical-operators/src/UnionRenamePhysicalOperator.cpp
greenBene: 1) We currently do not support the mentioned case. I agree that the Operator might not be ideally named, but renaming it would be out of scope imo. 
2) Since we assume smallish schemas, the vector solution should be faster, despite being n^2. If we run into issues, we can adapt :) For me, it was mainly about fixing invalid behaviour.

nes-query-optimizer/src/Phases/RuleBasedOptimizer.cpp
greenBene: Added an invariant.

nes-physical-operators/include/UnionRenamePhysicalOperator.hpp
greenBene: Yes. I removed it.

nes-physical-operators/include/UnionRenamePhysicalOperator.hpp
greenBene: I and clang tidy agree ^^ Added it

nes-physical-operators/include/UnionRenamePhysicalOperator.hpp
```diff
@@ -25,16 +26,16 @@ namespace NES
 class UnionRenamePhysicalOperator final : public PhysicalOperatorConcept
 {
 public:
-    UnionRenamePhysicalOperator(std::vector<QualifiedIdentifier> inputFields, std::vector<QualifiedIdentifier> outputFields);
+    UnionRenamePhysicalOperator(const std::vector<QualifiedIdentifier>& inputFields, std::vector<QualifiedIdentifier> outputFields);
```
keyseven123, line 29: use pass by value and then `std::move`

keyseven123 (approved):

nes-query-optimizer/include/Rules/Static/ProjectionPushdownRule.hpp
yschroeder97: Got it, thanks.

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
yschroeder97: Sounds reasonable.

yschroeder97 (approved):

PR #1596, opened 2026-05-11
feat(optimizer) watermark assigner pushdown

### LINKED ISSUE #1584: [TASK] WatermarkAssigner Pushdown
### Description

Implement an optimization rule that ensures that watermark assigners are pushed added as close to the source as possible, to ensure they have the highest likelihood to see a tuple before it is filtered out

### Acceptance Criteria

Watermark Assigner Pushdown Rule is implemented,  all tests continue to run successfully, and no performance degradation

greenBene:
Added a logical optimization rule that pushes watermark assigner operators as close to the data source as possible. 

## Verifying this change
This change is tested by
- all existing systest continue to run
- newly introduces unit tests validates behavior
 
- Optimizer

## Documentation
The logic is documented in the related hpp file. 

This PR closes #1584


nes-query-optimizer/include/Rules/Static/WatermarkAssignerPushdownRule.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+    [[nodiscard]] static LogicalOperator
+    addWatermarkAssigners(LogicalOperator op, bool ingestionTime, std::vector<std::pair<Field, Windowing::TimeUnit>> eventTime);
+
+    [[nodiscard]] LogicalOperator applyToAllChildren(
+        const LogicalOperator& op, bool ingestionTime, const std::vector<std::pair<Field, Windowing::TimeUnit>>& eventTime) const;
+
+    [[nodiscard]] static std::vector<std::pair<Field, Windowing::TimeUnit>>
+    fieldsWithNewBase(const LogicalOperator& base, std::vector<std::pair<Field, Windowing::TimeUnit>> eventTime);
+};
```
ls-1801, line 57: Missing `static_assert(RuleConcept<WatermarkAssignerPushdownRule, LogicalPlan>);` at the end of the header. All other rules in this directory carry this assertion so contract violations fail at the class itself rather than deep inside RuleManager. Please add it.

nes-query-optimizer/src/Phases/RuleBasedOptimizer.cpp
```diff
@@ -47,6 +48,7 @@ RuleBasedOptimizer::RuleBasedOptimizer(QueryOptimizerConfiguration defaultQueryO
     ruleManager.addRule(OriginIdInferenceRule{});
     ruleManager.addRule(FixedPlanStructureBarrier{});
     ruleManager.addRule(PredicatePushdownRule{});
+    ruleManager.addRule(WatermarkAssignerPushdownRule{});
```
ls-1801, line 53: I would question the use of an unordered_map, as that would introduce non-deterministic ordering of sibling rules. I in general think this is a good practice to randomize the order of the rules, but it could be an annoying bug to debug. It would seem flaky that runs on the CI machine may fail due to a different optimization order. I am not sure what the proper solution is, but I would log the chosen order of rules and also document that it's not deterministic.

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
```diff
@@ -0,0 +1,274 @@
...
+            }
+        }
+    }
+
+    std::vector<std::pair<Field, Windowing::TimeUnit>> pushFurther{};
+    std::vector<std::pair<Field, Windowing::TimeUnit>> applyNow{};
+
+    for (auto [onField, unit] : eventTime)
+    {
+        auto pushable = std::ranges::any_of(pushableFields, [&onField](const Field& pushableField) { return onField == pushableField; });
```
ls-1801: Using `std::ranges::any_of` on an `unordered_set` defeats the hash lookup — `pushableFields.contains(onField)` would be O(1) on average and reads more directly.

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
```diff
@@ -0,0 +1,274 @@
...
+        }
+        else
+        {
+            applyNow.emplace_back(onField, unit);
+        }
+    }
+
+    auto newOp = applyToAllChildren(op, ingestionTime, pushFurther);
+
+    for (const auto& [onField, unit] : applyNow)
```
ls-1801: This loop wraps `applyNow` in forward order, but `eventTime` is accumulated top-down (`emplace_back`), so the first element should end up outermost (topmost) to preserve the original WA stacking. `addWatermarkAssigners` pops from the back for exactly that reason — this loop is inconsistent and reverses the order whenever `applyNow` has more than one entry. Iterate in reverse, or use the same `pop_back` pattern.

Reproduced via minimal test (single assert) on this PR's HEAD — fails:

```cpp
TEST_F(WatermarkAssignerPushdownRuleTest, ProjectionTwoApplyNowOrder)
{
    /// BEFORE: sink < ET(d, unit=2) < ET(c, unit=1) < projection(c<-a*2, d<-b*2) < source(a, b)
    /// EXPECTED: topmost WA below sink is ET(d). Bug reverses order so it becomes ET(c).
    auto source = utils.createSource("twoApplyNow", {"a", "b"});
    auto projection = TypedLogicalOperator<ProjectionLogicalOperator>{
        source,
        std::vector<std::pair<Identifier, LogicalFunction>>{
            {Identifier::parse("c"),
             MulLogicalFunction{
                 FieldAccessLogicalFunction{source.getOutputSchema().getFieldByName(Identifier::parse("a")).value()},
                 ConstantValueLogicalFunction{DataType{DataType::Type::UINT64, DataType::NULLABLE::NOT_NULLABLE}, "2"}}},
            {Identifier::parse("d"),
             MulLogicalFunction{
                 FieldAccessLogicalFunction{source.getOutputSchema().getFieldByName(Identifier::parse("b")).value()},
                 ConstantValueLogicalFunction{DataType{DataType::Type::UINT64, DataType::NULLABLE::NOT_NULLABLE}, "2"}}}},
        ProjectionLogicalOperator::Asterisk{false}};
    auto etC = TypedLogicalOperator<EventTimeWatermarkAssignerLogicalOperator>{
        projection,
        FieldAccessLogicalFunction{projection.getOutputSchema().getFieldByName(Identifier::parse("c")).value()},
        Windowing::TimeUnit{1}};
    auto etD = TypedLogicalOperator<EventTimeWatermarkAssignerLogicalOperator>{
        etC,
        FieldAccessLogicalFunction{etC.getOutputSchema().getFieldByName(Identifier::parse("d")).value()},
        Windowing::TimeUnit{2}};
    auto sink = utils.createSink(etD, "twoApplyNow", {"c", "d"});
    auto pushed = WatermarkAssignerPushdownRule{}.apply(utils.createPlan(sink));

    auto topWA = pushed.getRootOperators().at(0).getChildren().at(0);
    ASSERT_EQ(topWA.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>().value()
        ->getOnField().tryGetAs<FieldAccessLogicalFunction>().value()->getField().getLastName(),
        Identifier::parse("d"));
}
```

Failure: `getField().getLastName()` is `C` but expected `D`.

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
```diff
@@ -0,0 +1,300 @@
...
+                 ConstantValueLogicalFunction{DataType{DataType::Type::UINT64, DataType::NULLABLE::NOT_NULLABLE}, "2"}}}},
+        ProjectionLogicalOperator::Asterisk{false}};
+
+    auto ingestionTime = TypedLogicalOperator<IngestionTimeWatermarkAssignerLogicalOperator>{projection};
+    auto eventTime = TypedLogicalOperator<EventTimeWatermarkAssignerLogicalOperator>{
+        ingestionTime,
+        FieldAccessLogicalFunction{ingestionTime.getOutputSchema().getFieldByName(Identifier::parse("c")).value()},
+        Windowing::TimeUnit{0}};
+
+    auto sink = utils.createSink(eventTime, "projection, breakers", {"c"});
```
ls-1801: Sink name `"projection, breakers"` has a stray comma and is copy-pasted between two unrelated tests:

```cpp
// line 126 — ProjectionBreaker
auto sink = utils.createSink(eventTime, "projection, breakers", {"c"});

// line 166 — ProjectionNonBreaker
auto sink = utils.createSink(eventTime, "projection, breakers", {"a", "c"});
```

Give each test a unique, accurate name so failure logs point at the right test:

```cpp
// ProjectionBreaker
auto sink = utils.createSink(eventTime, "projectionBreaker", {"c"});

// ProjectionNonBreaker
auto sink = utils.createSink(eventTime, "projectionNonBreaker", {"a", "c"});
```

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
```diff
@@ -0,0 +1,300 @@
...
+    auto source = utils.createSource("projectionNonBreaker", {"a", "b"});
+    auto projection = TypedLogicalOperator<ProjectionLogicalOperator>{
+        source,
+        std::vector<std::pair<Identifier, LogicalFunction>>{
+            {Identifier::parse("c"),
+             MulLogicalFunction{
+                 FieldAccessLogicalFunction{source.getOutputSchema().getFieldByName(Identifier::parse("a")).value()},
+                 ConstantValueLogicalFunction{DataType{DataType::Type::UINT64, DataType::NULLABLE::NOT_NULLABLE}, "2"}}},
+            {Identifier::parse("a"), FieldAccessLogicalFunction{source.getOutputSchema().getFieldByName(Identifier::parse("a")).value()}}},
+        ProjectionLogicalOperator::Asterisk{false}};
```
ls-1801: Coverage gap: no test exercises `Asterisk{true}` or a projection that renames a field a WA depends on (e.g. `project {x <- a}` with `EventTime(x)`). The rule matches projection entries on `getLastName()` only, so renamed WA fields could behave subtly. Worth at least one such test to pin the intent.

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
```diff
@@ -0,0 +1,300 @@
...
+    ASSERT_TRUE(op2.tryGetAs<JoinLogicalOperator>());
+    auto leftOp0 = op2.getChildren().at(0);
+    ASSERT_TRUE(leftOp0.tryGetAs<SourceDescriptorLogicalOperator>());
+    auto rightOp0 = op2.getChildren().at(1);
+    ASSERT_TRUE(rightOp0.tryGetAs<SourceDescriptorLogicalOperator>());
+}
+
+TEST_F(WatermarkAssignerPushdownRuleTest, WithoutWatermarkAssigner)
+{
+    /// Ensures plans without watermakr assigners pass unchanged
```
ls-1801: Typo: `watermakr` -> `watermark`.

ls-1801 (changes requested):
see comments above, i missclicked on the initial review...

nes-query-optimizer/include/Rules/Static/WatermarkAssignerPushdownRule.hpp
greenBene: Great catch! Thanks!

nes-query-optimizer/src/Phases/RuleBasedOptimizer.cpp
greenBene: Added the documentation to RuleManager.hpp and added a NES_DEBUG statement to RuleBasedOptimizer.cpp and SemanticAnalyser.cpp

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
greenBene: Great catch! :) 

I'm not sure what I was thinking there :thinking:

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
greenBene: Great catch! Changed the order in which it is applied and also added your suggested test (which now succeeds)

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
greenBene: Once again, great catch! :)

greenBene: Fixed it

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
greenBene: Fixed

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
greenBene: Great catch! Added tests for both cases. The Asterix case revealed a bug (which is now fixed)

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
```diff
@@ -0,0 +1,289 @@
...
+    /// test if event time fields are generated by projection.
+    /// If not, they are pushed further.
+    /// If they are, the event time watermark assigner is applied.
+
+    std::unordered_set<Field> pushableFields;
+    if (op->hasAsterisk())
+    {
+        for (const auto& field : op.getOutputSchema())
+        {
+            pushableFields.emplace(field);
```
ls-1801: Subtle bug in the asterisk branch: it marks every field of `op.getOutputSchema()` as pushable, but `ProjectionLogicalOperator::inferLocalSchema` unions both pass-through and computed-projection fields into the output schema. So for `Asterisk{true}` together with a computed projection (e.g. `c <- a*2`) and a WA on `c`, `c` is incorrectly placed in `pushableFields`, gets into `pushFurther`, and the subsequent `fieldsWithNewBase` call fires the PRECONDITION at L75 because the projection's child does not contain `c`.

Suggested fix: iterate `op->getChild().getOutputSchema()` here so only true pass-through fields become pushable; computed fields then fall through to `applyNow` and are materialized above the projection as intended.

Repro (currently aborts on this PR's HEAD):

```cpp
TEST_F(WatermarkAssignerPushdownRuleTest, ProjectionAsterixWithComputedFieldBreaker)
{
    /// BEFORE: sink < ET(c) < projection(Asterisk{true}, c <- a*2) < source(a, b)
    /// EXPECTED: ET(c) stays above projection (c is computed, not pushable)
    auto source = utils.createSource("projectionAsterixComputed", {"a", "b"});
    auto projection = ProjectionLogicalOperator::create(
        source,
        std::vector<std::pair<Identifier, LogicalFunction>>{
            {Identifier::parse("c"),
             MulLogicalFunction{
                 FieldAccessLogicalFunction{source.getOutputSchema()[Identifier::parse("a")].value()},
                 ConstantValueLogicalFunction{DataType{DataType::Type::UINT64, DataType::NULLABLE::NOT_NULLABLE}, "2"}}}},
        ProjectionLogicalOperator::Asterisk{true});

    auto eventTime = EventTimeWatermarkAssignerLogicalOperator::create(
        projection,
        FieldAccessLogicalFunction{projection.getOutputSchema()[Identifier::parse("c")].value()},
        Windowing::TimeUnit{0});
    auto sink = utils.createSink(eventTime, "projectionAsterixComputed", {"a", "b", "c"});
    auto plan = utils.createPlan(sink);

    auto pushed = WatermarkAssignerPushdownRule{}.apply(plan);

    auto op0 = pushed.getRootOperators().at(0);
    ASSERT_TRUE(op0.tryGetAs<SinkLogicalOperator>());
    auto op1 = op0.getChildren().at(0);
    ASSERT_TRUE(op1.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>());
    auto op2 = op1.getChildren().at(0);
    ASSERT_TRUE(op2.tryGetAs<ProjectionLogicalOperator>());
    auto op3 = op2.getChildren().at(0);
    ASSERT_TRUE(op3.tryGetAs<SourceDescriptorLogicalOperator>());
}
```

Failure on HEAD:
```
[E] [WatermarkAssignerPushdownRule.cpp:75] [fieldsWithNewBase] Precondition violated: (childField.has_value()): the operator must have the event time field
```

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
```diff
@@ -0,0 +1,289 @@
...
+    /// Does not push event time water mark assigners that don't use a FieldAccessLogicalFunction.
+
+    if (const auto fieldAccess = op->getOnField().tryGetAs<FieldAccessLogicalFunction>())
+    {
+        eventTime = fieldsWithNewBase(op->getChild(), eventTime);
+        eventTime.emplace_back(fieldAccess.value()->getField(), op->getUnit());
+        return watermarkAssignerPushdown(op->getChild(), ingestionTime, eventTime);
+    }
+
+    return applyToAllChildren(op, ingestionTime, eventTime);
```
ls-1801: Re-checked this branch: I think it silently reorders watermark assigners when there is an upstream FA (FieldAccess) WA (WatermarkAssigner) above a non-FA WA, which changes the watermark that reaches the consumer.

Trace for `Sink > ET(a, FA) > ET(timestamp*1, non-FA) > Source`:

1. `pushBeyondEventTimeWatermarkAssigner(ET(a))` appends `(a)` to `eventTime`, recurses into the non-FA WA with `eventTime=[(a)]`.
2. `pushBeyondEventTimeWatermarkAssigner(ET(non-FA))` falls into this else branch: `applyToAllChildren(op, false, [(a)])`.
3. `applyToAllChildren` rebases `(a)` onto the WA's child schema and recurses into Source with `eventTime=[(a)]`.
4. `pushBeyondSource` → `addWatermarkAssigners(Source, false, [(a)])` materializes `ET(a) > Source` *below* the non-FA WA.
5. Result: `Sink > ET(non-FA) > ET(a) > Source`.

Original order had ET(a) above ET(non-FA); the rewritten plan has them swapped. Since the "last applied" (topmost) WA is what reaches the consumer, this changes which watermark stream the sink sees.

Reproducer (currently fails on HEAD):

```cpp
TEST_F(WatermarkAssignerPushdownRuleTest, NonFieldAccessETPreservesOrderWithUpstreamFA)
{
    /// BEFORE: sink < ET(a, FA) < ET(timestamp*1, non-FA) < source(a, timestamp)
    /// EXPECTED: top WA below sink is ET(a). The non-FA WA may stay or move,
    /// but FA(a) must end up ABOVE non-FA(timestamp*1) to preserve the
    /// "last-applied WA wins" semantics at the consumer.
    auto source = utils.createSource("nonFaEtChain", {"a", "timestamp"});
    auto nonFaWa = EventTimeWatermarkAssignerLogicalOperator::create(
        source,
        MulLogicalFunction{
            FieldAccessLogicalFunction{source.getOutputSchema()[Identifier::parse("timestamp")].value()},
            ConstantValueLogicalFunction{DataType{DataType::Type::UINT64, DataType::NULLABLE::NOT_NULLABLE}, "1"}},
        Windowing::TimeUnit{0});
    auto faWa = EventTimeWatermarkAssignerLogicalOperator::create(
        nonFaWa,
        FieldAccessLogicalFunction{nonFaWa.getOutputSchema()[Identifier::parse("a")].value()},
        Windowing::TimeUnit{0});
    auto sink = utils.createSink(faWa, "nonFaEtChain", {"a", "timestamp"});

    auto pushed = WatermarkAssignerPushdownRule{}.apply(utils.createPlan(sink));

    auto op0 = pushed.getRootOperators().at(0);
    ASSERT_TRUE(op0.tryGetAs<SinkLogicalOperator>());
    auto op1 = op0.getChildren().at(0);
    ASSERT_TRUE(op1.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>());
    auto topField = op1.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>()
                        .value()->getOnField()
                        .tryGetAs<FieldAccessLogicalFunction>();
    ASSERT_TRUE(topField.has_value());
    ASSERT_EQ(topField.value()->getField().getLastName(), Identifier::parse("a"));
}
```

Failure: the top WA below sink is the non-FA `MulLogicalFunction` WA, not `FA(a)`. `topField.has_value() == false`.

Two reasonable fixes:

**Option A — barrier semantics.** Treat the non-FA WA the same way `pushBeyondDefault` treats unknown operators: materialize all upstream WAs *above* it, restart recursion with empty state on its children. The WA stays where it was (we still can't push it without understanding the function), but the silent reorder stops.

```cpp
return pushBeyondDefault(op, ingestionTime, std::move(eventTime));
```

**Option B — disallow non-FA WAs explicitly.** Today every construction site goes through `BoundEventTimeCharacteristic`, which stores a `FieldAccessLogicalFunction`. No parser/builder path produces a non-FA `onField`. Drop the else branch entirely and assert the contract:

```cpp
INVARIANT(
    op->getOnField().tryGetAs<FieldAccessLogicalFunction>().has_value(),
    "EventTime watermark assigner onField must be a FieldAccessLogicalFunction");
```

(or `throw InvalidArgument(...)` if this should remain a runtime error rather than a precondition). The Unreflector accepts a raw `LogicalFunction`, so a deserialized plan could still carry a non-FA WA — fail loud there too rather than silently miscompiling.

Either is fine; Option B is the smaller change and matches the actual constructor surface today.

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
```diff
@@ -0,0 +1,392 @@
...
+    ASSERT_TRUE(op1.tryGetAs<ProjectionLogicalOperator>());
+    auto op2 = op1.getChildren().at(0);
+    ASSERT_TRUE(op2.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>());
+    auto op3 = op2.getChildren().at(0);
+    ASSERT_TRUE(op3.tryGetAs<IngestionTimeWatermarkAssignerLogicalOperator>());
+    auto op4 = op3.getChildren().at(0);
+    ASSERT_TRUE(op4.tryGetAs<SourceDescriptorLogicalOperator>());
+}
+
+TEST_F(WatermarkAssignerPushdownRuleTest, ProjectionAsterixNonBreaker)
```
ls-1801: Quick naming nit: the test name, source name (L183), and sink name (L190) use "Asterix" — was that intentional? If you mean the `*` (the punctuation mark), it's "Asterisk", which matches the API on L184 (`ProjectionLogicalOperator::Asterisk{true}`). Currently the surrounding identifiers and the actual API spell it differently, which hurts grep-ability.

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
```diff
@@ -0,0 +1,392 @@
...
+             MulLogicalFunction{
+                 FieldAccessLogicalFunction{source.getOutputSchema()[Identifier::parse("a")].value()},
+                 ConstantValueLogicalFunction{DataType{DataType::Type::UINT64, DataType::NULLABLE::NOT_NULLABLE}, "2"}}},
+            {Identifier::parse("b"), FieldAccessLogicalFunction{source.getOutputSchema()[Identifier::parse("b")].value()}}},
+        ProjectionLogicalOperator::Asterisk{false});
+
+    auto eventTime1 = EventTimeWatermarkAssignerLogicalOperator::create(
+        projection, FieldAccessLogicalFunction{projection.getOutputSchema()[Identifier::parse("c")].value()}, Windowing::TimeUnit{0});
+    auto eventTime2 = EventTimeWatermarkAssignerLogicalOperator::create(
+        eventTime1, FieldAccessLogicalFunction{eventTime1.getOutputSchema()[Identifier::parse("b")].value()}, Windowing::TimeUnit{0});
```
ls-1801: Coverage suggestion: both `EventTime` assigners are constructed with `Windowing::TimeUnit{0}` (L259, L261), so the test can't catch a unit-mixup. Using distinct units and asserting them on the resulting WAs would harden coverage in the spirit of `ProjectionTwoApplyNowOrder`. Suggested patch:

```diff
-    auto eventTime1 = EventTimeWatermarkAssignerLogicalOperator::create(
-        projection, FieldAccessLogicalFunction{projection.getOutputSchema()[Identifier::parse("c")].value()}, Windowing::TimeUnit{0});
-    auto eventTime2 = EventTimeWatermarkAssignerLogicalOperator::create(
-        eventTime1, FieldAccessLogicalFunction{eventTime1.getOutputSchema()[Identifier::parse("b")].value()}, Windowing::TimeUnit{0});
+    auto eventTime1 = EventTimeWatermarkAssignerLogicalOperator::create(
+        projection, FieldAccessLogicalFunction{projection.getOutputSchema()[Identifier::parse("c")].value()}, Windowing::TimeUnit{1});
+    auto eventTime2 = EventTimeWatermarkAssignerLogicalOperator::create(
+        eventTime1, FieldAccessLogicalFunction{eventTime1.getOutputSchema()[Identifier::parse("b")].value()}, Windowing::TimeUnit{2});
```

and the two existing field-name asserts can additionally check the unit:

```diff
     ASSERT_TRUE(op1.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>());
+    ASSERT_EQ(op1.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>().value()->getUnit(), Windowing::TimeUnit{1});
     ASSERT_TRUE(
         op1.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>()
             .value()
             ->getOnField()
             .tryGetAs<FieldAccessLogicalFunction>()
             .value()
             ->getField()
             .getLastName()
         == Identifier::parse("c"));
     ...
     ASSERT_TRUE(op3.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>());
+    ASSERT_EQ(op3.tryGetAs<EventTimeWatermarkAssignerLogicalOperator>().value()->getUnit(), Windowing::TimeUnit{2});
```

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
```diff
@@ -0,0 +1,392 @@
...
+    auto op0 = pushed.getRootOperators().at(0);
+    ASSERT_TRUE(op0.tryGetAs<SinkLogicalOperator>());
+    auto op1 = op0.getChildren().at(0);
+    ASSERT_TRUE(op1.tryGetAs<SourceDescriptorLogicalOperator>());
+}
+
+TEST_F(WatermarkAssignerPushdownRuleTest, ProjectionTwoApplyNowOrder)
+{
+    /// BEFORE: sink < ET(d, unit=2) < ET(c, unit=1) < projection(c<-a*2, d<-b*2) < source(a, b)
+    /// EXPECTED: topmost WA below sink is ET(d). Bug reverses order so it becomes ET(c).
```
ls-1801: "Bug reverses order so it becomes ET(c)." documents the historical bug rather than the test's invariant. The EXPECTED line already states what the test asserts. Suggest dropping the second sentence — comments should describe what the code asserts, not the reasoning that led to it.

ls-1801 (changes requested):
Sorry, screwed up again... see comments above

adrianmichalke (approved):

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
greenBene: Amazing catch! Thank you! A similar wrong assumption was made in PredicatePushdown and will be fixed accordingly.

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
greenBene: Great catch, and thanks for the suggestions. I chose Option B. If we support other functions, we can then make the concrete decision on how to proceed.

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
greenBene: Read and watched too much Asterix & Obelix as a child to not continue to make this error :melting_face: Thank you for catching it!

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
greenBene: Good point :) I added it.

nes-query-optimizer/tests/Rules/Static/WatermarkAssignerPushdownRuleTest.cpp
greenBene: Good catch! Was a copy-paste error.

ls-1801 (approved):

PR #1597, opened 2026-05-11
Add design doc for extensible data types

alepping:
## Description
Adds a design doc (docs/design/20260510_extensible_data_types.md) and a proof-of-concept branch for making NebulaStream's DataType extensible from nes-plugins/.
  
What the PoC demonstrates:
- FIXEDSIZED — a type-specific array (T ARRAY[N]), distinct from VARSIZED.
- STRUCT — a nominally-typed named composite (structName + field list) that participates in schemas,
 parsers, formatters, and Nautilus values.
- Generic T(...) SQL syntax — the parser's exitFunctionCall resolves any registered type via
DataTypeProvider and lowers to ConstructStruct / CastToType / ConstantValue, with no per-type parser
 code.
- Plugins on top of unmodified core: Image/ThermalFrame (+ to_celsius, is_fever, to_rgb), Reading (+
 per-struct +), and a NestedJSON input formatter; systests under nes-systests/formatter/JSON_OUTPUT/
 exercise the full path.
 - End-to-end mono16 to RGB image conversion demo

Explicit non-goals: no polymorphic DataType refactor, no structural typing, no
memory-layout/bufferref rework. These are called out as open questions/follow-up. The real
implementation is planned to branch off the schema-inference PR and land incrementally.

## PoC branch
https://github.com/nebulastream/nebulastream/tree/extensible-data-types.

### DIFF
--- a/docs/design/20260510_extensible_data_types.md
+++ b/docs/design/20260510_extensible_data_types.md
@@ -1,0 +1,242 @@
+# General Information
+The PoC branch attached to this DD is not a feature branch in disguise.
+Its sole purpose is to give you a concrete, interactive vision of what I am aiming for with a first version of extensible data types.
+As a result there are at least four major simplifications that the PoC makes use of:
+1. builds on top of current main and not on top of the schema inference branch
+2. keeps simplistic datatype frontend/logical implementation
+3. does not dive into a memorylayout-/bufferref-based implementation for custom/new datatypes, which is probably the clean way, but requires a significant refactor of bufferrefs
+4. hardcodes parsing logic into new 'NestedJSON' parser
+
+# The Problem
+Adding a new logical type means editing core code at every switch site: parsers, schema printer, sinks, JSON/CSV formatters, Nautilus runtime, generator source, etc.
+There is no way to register a new type from `nes-plugins/`, although sources, sinks, and input/output formatters already follow that pattern.
+
+In particular, real-world ingest scenarios need *composite* types whose layout is fixed but bigger than a scalar, e.g. a thermal-camera reading `Image{timestamp, cameraId, ThermalFrame{pixels: UINT16[N]}}`.
+Today the only way to model this is to flatten everything into top-level columns, which loses the type identity that makes domain-specific functions (`to_celsius`, `is_fever`, `to_rgb`) safe to register. Such flattened approach also quickly blows up the schema and does not really work with containers such as arrays and vectors.
+
+- P1: `DataType` cannot be extended without modifying core code paths spread across ~20 files.
+- P2: There is no first-class composite type — no way to represent a named struct of fields and refer to it by name in SQL or in physical functions.
+- P3: There is no way to add SQL constructor syntax (`Image(...)`) for a new type without per-type parser changes. The casting/construction surface is hard-coded.
+- P4: There is no vector- or array-like container type that is type-specific. VarSized is essentially a string/byte-vector.
+
+# Goals
+- G1: A new `DataType` can be added from `nes-plugins/` using the same registration pattern that sources, sinks, and formatters already use, addressing P1.
+- G2: Support a composite `STRUCT` variant with a name and a field list that participates in schemas, parsers, formatters, and Nautilus values, addressing P2.
+- G3: Provide a generic SQL constructor syntax `T(arg1, arg2, ...)` for any registered type, dispatched in the parser without per-type code, addressing P3.
+- G4: Demonstrate G1–G3 end-to-end with example plugins that fit on top of the core without modifying it (`Image/ThermalFrame`, `Reading`).
+- G5: Create a fixed-sized (is an array, but name contrasts 'varsized') datatype that is type-specific, i.e., represents an array of type T. 
+
+# Non-Goals
+- NG1: Refactoring `DataType` into a polymorphic class hierarchy or `std::variant`.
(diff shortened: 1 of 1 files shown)


adrianmichalke (changes requested):
Interesting idea. Have you looked into what other systems support in terms of data type extensibility? I would expect such a short discussion in the DD as the topic and its interprestations are quite broad

greenBene (approved):
I am a huge fan! :) Both the general approach, and the way you cut the work. 
I think it makes very much sense to start with this, and do the the polymorphic representation of DataType and the refactoring of memory layout and buffers later if required. 

The suggested solution would make the use and extension of NES in more use-cases easier! :) 

Strong accept from my side!

Artraxon:
Thank you, I think this is a good first step for us.
But, regarding polymorphism, I would be heavily against subtyping based polymorphism. It is very hard to reason about from a language/optimizer perspective without little benefit. Personally, I also think it is one step closer to implementation subtyping which I think is mistake the we should avoid in general.

In general, I think the key design constraint for us is to be able to generate efficient machine code at submission time.
That means, we want to avoid dynamic dispatch as much as possible, since we might have to jump to arbitrary locations per tuple.
Fortunately, I think it is fair to say that there really isn't any use case in starting a query over a set of types you don't know, i.e. we always know all possible types that one could refer to at query submission time. 



This leads to two kinds of polymorphism that I think are are interesting and useful to us:
1. Parametric Polymorphism, also known as static polymorphism, to allow us add abstract functions that work with user provided types akin to the c++ equivalent
```c++
template <Addable T, int N>
T sum(std::array<T, N> array){
....
}
```
This example would work over both an int and maybe some spatial vector, as long as both have an add function defined.
It would also allow us to generate code at trace time for composed function calls that is as efficient as if one would know the type in advance and wrote a specialized sumVector function.

Whether and how users can add such functions with parametric polymorphism is another question, but the key point is allows us to provide abstract functions over user defined types without any runtime overhead.


2. Sum Types
The second thing I would anticipate is that there might be queries where an element could be of a type of some set of types set at query submission time.
Of course, classic examples are error types, but it would also make writing queries with unions much more convenient.

While sum types require some runtime overhead, it is much lower since the lookup table is known at compile time.
Furthermore, with nautilus + query optimization we might be able to compile out a lot of such checks.
For example, in some query
```sql
SELECT format('{} : {}', my_text, num)
FROM 
((SELECT my_float as num, my_text
FROM source1)
UNION ALL
(SELECT my_int_vec as num, my_text
FROM source2))
```
we can implicitly infer that num is a sum type of floats and int vectors, but push down/inline the float and int vector specific format functions.

Artraxon (approved):

docs/design/20260510_extensible_data_types.md
```diff
@@ -0,0 +1,242 @@
...
+The PoC branch attached to this DD is not a feature branch in disguise.
+Its sole purpose is to give you a concrete, interactive vision of what I am aiming for with a first version of extensible data types.
+As a result there are at least four major simplifications that the PoC makes use of:
+1. builds on top of current main and not on top of the schema inference branch
+2. keeps simplistic datatype frontend/logical implementation
+3. does not dive into a memorylayout-/bufferref-based implementation for custom/new datatypes, which is probably the clean way, but requires a significant refactor of bufferrefs
+4. hardcodes parsing logic into new 'NestedJSON' parser
+
+# The Problem
+Adding a new logical type means editing core code at every switch site: parsers, schema printer, sinks, JSON/CSV formatters, Nautilus runtime, generator source, etc.
```
ls-1801, line 11: I think Nautilus Runtime should be the only relevant problem here. Everything else can be addressed at different levels, and we can provide user-facing shortcuts later.

Parsers: we can always fall back to parsing individual values (e.g. for a point type we parse 3 integers, or we parse nested json keys (the pr from nils which exists since forever)). If we truly want special syntax for new DataTypes (which I think is kinda odd and would need to be justified with a real use case), I would suggest writing a custom function to parse a string.

The schema printer is probably nice to have, but here we either auto-derive the schema printer or kindly ask the user to provide a custom function.

Not sure how sinks are relevant, can you name an example?

JSON/CSV formatters, I think that's the same for Parsers/Schema Printer. In general, I don't think this would scale, if we increase the number of formats..., If there is a genuine use-case, they can write a custom function.

Generator Source is the same; they could use a WORDLIST and a custom function to create. Or generate based on fundamental values.

Not saying these problems aren't interesting, but I think there are countless solutions that can be retrofitted, if we really need that. I think we can run into a problem where we have so many different interfaces for all kinds of "extension" that this will appear to be overwhelming for any user of the system, and also that I fear that the quality of certain interfaces will suffer, or we expose too many internals, by trying to make everything extendible.

ls-1801 (changes requested):
I have an alternative proposal: https://gist.github.com/ls-1801/2dc6c07fe07ad81a26c5a84715a0e9e7
which should at least be discussed in the alternative sections

I don't have time to implement that, so I'll just trust you to come up with something sensible. I think there is some overlap in our proposals and some differences in focus. 

I would aim to support the use cases of @marianaGarcez first. Lots of people are interested in geospatial

keyseven123 (approved):
I like that we move in the direction of having more complex data types. I think we had the discussion already, but let us discuss it once more how we then perform the lowering to the physical data types.

ThanGeo (approved):

PR #1598, opened 2026-05-12
feat(ML): add OpenVINO inference runtime backend

gaturchenko:
## Purpose of the Change and (not so) Brief Change Log
This pull request adds the new OpenVINO inference backend.
It is set as the default one, but IREE remains available.

Why do we need a new backend in the first place? Unfortunately, the IREE backend has a few limitations:

- Support for `.onnx` format only
- Non-robust regex-based model metadata parsing
- Unreliable model compilation (MLIR -> VMFB)
- Performance issues (the IR produced by IREE is not always optimal)
- C-style API

OpenVINO in turn:

- Loads models in many more formats via a single binary
- Creates an XML model graph that is easier to parse
- Avoids complex compilation that makes model loading less error-prone
- Provides better CPU performance guarantees through:
    - DAG-level optimizations (operator fusion, constant folding, etc.)
    - Efficient highly-tuned runtime kernels
- Doesn't require the target hardware information to be provided ahead-of-time
- Exposes a C++ API

So TLDR; OpenVINO is like automatic transmission, while IREE is like manual.
From the user perspective, with OpenVINO it is way more likely that a user-provided model "will just work".
With IREE, things can break during model compilation and there is not much we can do to help the user resolve these problems.

OpenVINO however still comes with some trade-offs:

- Runtime is much larger (not just a lightweight interpreter). Single-node-worker binary size in release:
    - IREE: 147.7 MB
    - OpenVINO: 261.8 MB
- We have to send binary data to the worker (model weights ~ model binary size)
- ARM performance may be a concern (TBD)
- Supports only Intel GPUs (e.g., no CUDA)

All in all, since we have active projects that want to utilize inference, I would strongly advocate adding another backend to avoid unexpected deployment issues.

## Verifying this change
This change is tested by:
- Extended unit tests:
    - ModelLoaderTest
    - InferModelLogicalOperatorTest
    - InferModelPhysicalOperatorTest
    - StatementBinderTest
- Added dedicated InferModelBackend system-level test

## What components does this pull request affect?
- vcpkg
- nix
- nes-frontend
- nes-sql-parser
- nes-inference
- nes-logical-operators
- nes-physical-operators
- nes-query-compiler
- nes-systests


keyseven123 (commented):
Hi,
I would like to have a single inference backend, as otherwise people will be unsure which backend to use. Also, I am not sure what advantage IREE has over OpenVINO, aside from the larger binary size. We have quite large Docker images; an additional 100MB is not a big problem if it helps us reduce maintenance on the codebase.
Since we don't currently support GPUs, I'm not concerned that openVINO only supports Intel GPUs. 
Thus, I would suggest replacing the IREE with OpenVINO.

docker/single-node-worker/SingleNodeWorker.dockerfile
```diff
@@ -3,6 +3,7 @@ ARG TAG=latest
 ARG RUNTIME_TAG=${TAG}
 FROM nebulastream/nes-development:${TAG} AS build
 ARG BUILD_TYPE=RelWithDebInfo
+ARG NES_INFERENCE_BACKENDS=ALL
```
ls-1801: This `ARG NES_INFERENCE_BACKENDS=ALL` is never forwarded to the cmake invocation below — the flag is silently dropped, so building this image with `--build-arg NES_INFERENCE_BACKENDS=OPENVINO` (or IREE) has no effect. Please add `-DNES_INFERENCE_BACKENDS=${NES_INFERENCE_BACKENDS}` to the cmake line, like the other frontend dockerfiles do.

vcpkg/vcpkg-registry/ports/openvino/fix-types-libcxx.patch
```diff
@@ -0,0 +1,66 @@
+Subject: [PATCH] fix-types-libvcxx
```
ls-1801: Typo in the Subject line: `fix-types-libvcxx` -> `fix-types-libcxx`.

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
```diff
@@ -0,0 +1,166 @@
...
+
+    return RuntimeMetadata{
+        .inputShape = model.getInputShape(),
+        .nDim = model.getNDim(),
+        .functionName = model.getFunctionName(),
+        .inputSize = model.inputSize(),
+        .outputSize = model.outputSize()};
+}
+
+void OpenVinoRuntimeBackend::infer(std::byte* inputBuffer, size_t, std::byte* outputBuffer, size_t /*outputBufferSize*/)
```
ls-1801: `inputBufferSize` and `outputBufferSize` are dropped here, but the IREE backend validates `outputSizeBytes > outputBufferSize` and throws. Please add an `INVARIANT`/`PRECONDITION` (or an explicit bounds check) that the caller-provided sizes are at least `inputElementType.size() * shape_size(inputShape)` / `outputElementType.size() * shape_size(outputShape)` before constructing the tensors, so a shape/dtype mismatch fails loudly instead of silently OOB-writing.

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
```diff
@@ -0,0 +1,297 @@
...
+        return std::unexpected(ImportError{"OpenVINO model tensor rank is dynamic"});
+    }
+
+    std::vector<size_t> shape;
+    shape.reserve(partialShape.size());
+    for (const auto& dimension : partialShape)
+    {
+        if (dimension.is_dynamic())
+        {
+            shape.push_back(1);
```
ls-1801, line 141: Are these `1` substitutions for dynamic/negative dimensions actually safe, and is the behaviour tested? An ONNX model with a dynamic batch dim will "import" with the wrong shape and produce silently-incorrect output downstream. I think the safer default is to reject the model (return an `ImportError`) and require callers to supply concrete shapes — or at the very least emit an `NES_WARNING` naming the affected dimension.

nes-sql-parser/tests/StatementBinderTest.cpp
```diff
@@ -597,8 +597,8 @@ TEST_F(StatementBinderTest, BindCreateModel)
 {
     const std::string modelPath = std::string(INFERENCE_TEST_DATA) + "/tiny_identity.onnx";
     /// tiny_identity.onnx has a 100-element f32 tensor on each side; VARSIZED mirrors it as a single bulk-byte field.
-    const std::string createModelStatement = "CREATE MODEL identity ('" + modelPath
-        + "') "
+    const std::string createModelStatement = "CREATE MODEL identity ('" + modelPath + "' BACKEND IREE"
```
ls-1801: `BindCreateModel` is now hard-coded to `BACKEND IREE` and runs `modelStatementHandler->apply(createModel)` with an unconditional `ASSERT_TRUE`. Builds that compile only the OpenVINO backend will fail this test (the catalog returns "IREE inference backend was not built"). Either drop the `BACKEND IREE` (the new `BindCreateModelBackend` test already covers binder-side IREE parsing), or guard the apply with the same IREE-availability check ModelLoaderTest now uses.

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
```diff
@@ -0,0 +1,166 @@
...
+    }
+}
+
+std::string formatTensor(const ov::Tensor& tensor)
+{
+    std::ostringstream stream;
+    stream << "shape=" << shapeToString(tensor.get_shape()) << ", type=" << tensor.get_element_type().get_type_name() << ", values=[";
+
+    const auto elementType = tensor.get_element_type();
+    appendTensorValues<float>(stream, tensor);
```
ls-1801: `elementType` on line 98 is captured but never used — `appendTensorValues<float>` is always called, so the `if constexpr` branches for `uint8_t`/`int8_t`/`bool`/`ov::float16` are dead, and the moment a non-f32 model reaches this path `tensor.data<const float>()` will throw mid-inference in any debug build.

Could the unused-type problem be moved to compile time? E.g. a `static_assert(std::is_same_v<T, float>, "appendTensorValues only supports f32 today")` in the function body, paired with dispatch on `elementType` here that statically rejects non-float instantiations until you actually wire them up. Or, simpler, drop the template branches entirely and rename to `appendFloatTensorValues` so the f32-only invariant is in the name.

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalInferModel.cpp
```diff
@@ -65,6 +63,11 @@ LoweringRuleResultSubgraph LowerToPhysicalInferModel::apply(LogicalOperator logi
         inferModelOp.get().hasVarsizedInput(),
         inferModelOp.get().hasVarsizedOutput());
 
+    /// Other physical inference operators will be added later,
```
ls-1801: This new comment block reads like a conversation/justification artifact ("Other physical inference operators will be added later, so we should log...") and the resulting log line is content-free (no backend, no model, no function name — nothing that wasn't already implied by the operator name).

Two existing comments above also became stale with OpenVINO and would be worth refreshing in the same pass:
- Line 41 still says *"Compile the imported MLIR to IREE bytecode... `iree-compile` subprocess"* — but `compileModel` now dispatches by backend (OpenVINO is a pass-through, no MLIR, no `iree-compile`).
- Line 58 still says *"thread-local IREE session pool"* — no longer accurate for the OpenVINO path.

nes-sql-parser/tests/StatementBinderTest.cpp
```diff
@@ -654,6 +655,66 @@ TEST_F(StatementBinderTest, BindCreateModel)
...
+{
+    const std::string modelPath = std::string(INFERENCE_TEST_DATA) + "/tiny_identity.onnx";
+    const auto defaultBackend
+        = binder->parseAndBindSingle("CREATE MODEL defaultBackend ('" + modelPath + "') INPUT (f1 VARSIZED) OUTPUT (o1 VARSIZED)");
+    ASSERT_TRUE(defaultBackend.has_value());
+    const auto& defaultModel = std::get<CreateModelStatement>(*defaultBackend);
+    ASSERT_EQ(defaultModel.backend, "OPENVINO");
+
+    const auto defaultResult = modelStatementHandler->apply(defaultModel);
+    if (!defaultResult.has_value() && std::string(defaultResult.error().what()).find("ovc is not available") != std::string::npos)
```
ls-1801: The skip only fires when the message contains `ovc is not available`, which is only raised when OpenVINO IS built but `ovc` is missing. In an IREE-only build the error is `OpenVINO inference backend was not built` and this test will hard-fail at the `ASSERT_TRUE` below. Use a compile-time guard mirroring `ModelLoaderTest`'s `openVinoEnabled()` helper instead of substring matching the error message.

docker/frontend/cli.dockerfile
```diff
@@ -2,14 +2,15 @@
 ARG TAG=latest
 ARG RUNTIME_TAG=${TAG}
 FROM nebulastream/nes-development:${TAG} AS build
-ARG BUILD_TYPE=RelWithDebInfo
+ARG BUILD_TYPE
```
ls-1801: Why was the `=RelWithDebInfo` default dropped here? `repl.dockerfile` and `repl-embedded.dockerfile` still keep it. Without a default, building this image without `--build-arg BUILD_TYPE=...` will pass `-DCMAKE_BUILD_TYPE=` (empty) to cmake. Looks like an accidental change unrelated to the OpenVINO work.

gaturchenko:
> Hi, I would like to have a single inference backend, as otherwise people will be unsure which backend to use. Also, I am not sure what advantage IREE has over OpenVINO, aside from the larger binary size. We have quite large Docker images; an additional 100MB is not a big problem if it helps us reduce maintenance on the codebase. Since we don't currently support GPUs, I'm not concerned that openVINO only supports Intel GPUs. Thus, I would suggest replacing the IREE with OpenVINO.

Hey, thank you for the comments. From the user perspective, nothing changes, since OpenVINO is set as the default backend. You can still register the model as:
```
CREATE MODEL iris ('model/iris.onnx')
INPUT (p1 FLOAT32, p2 FLOAT32, p3 FLOAT32, p4 FLOAT32)
OUTPUT (setosa FLOAT32, versicolor FLOAT32, virginica FLOAT32);
```
And this will work, so the user doesn't have to worry about specifying the backend.

Regarding the advantages of IREE over OpenVINO, it is true that there are not that many of them (AFAIK). Execution-wise, the biggest upside of IREE is that there is no need to send potentially large binary data from the coordinator to workers. Additionally, IREE has lower runtime footprint, which in some deployment scenarios can make a difference. 

Also, you mentioned Docker images, and this is where OpenVINO creates problems, not as far as the worker is concerned though. The debug build of OpenVINO is several GB and it will slow the dev image build jobs down significantly. I added a CMake option to compile either all or specific backend(s), that adds some flexibility I think.

For me, the biggest argument in favor of keeping both is flexibility in deployment. There are heaps of different models, and guaranteeing reasonable performance is almost impossible. Offering 2 backends at least allows us to somewhat bridge this gap: if the user notices a query is slow with, say, OpenVINO, they can try to redeploy with IREE, and vice versa. Also, I'm pretty sure we can run queries with GPU inference, we just need to build IREE runtime with, e.g., CUDA support and make a few changes to the code. Of course, maintaining 1 backend is simpler than 2 or more, but IMO keeping both is fine, especially since operator interfaces are basically backend-agnostic.

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
```diff
@@ -0,0 +1,166 @@
...
+    static ov::Core sharedCore;
+    static std::mutex coreMutex;
+    const ov::Shape modelInputShape(model.getInputShape().begin(), model.getInputShape().end());
+    ov::Tensor weights(ov::element::u8, {modelBin.size()});
+    if (!modelBin.empty())
+    {
+        std::memcpy(weights.data<std::uint8_t>(), modelBin.data(), modelBin.size());
+    }
+
+    const std::scoped_lock lock(coreMutex);
```
ls-1801: Right now every worker thread that hits this operator runs its own `read_model` -> `reshape` -> `compile_model` for the *same* model, serialized by `coreMutex`. The mutex prevents parallel compiles but doesn't cache anything, so total startup = N_threads × compile_time, and threads block each other on a multi-second lock.

Could the OpenVINO compile happen at lowering time instead — symmetric with how IREE runs `iree-compile` in `LowerToPhysicalInferModel` — so there is exactly one `compile_model` per query, and worker threads just create their own `inferRequest` from the shared compiled model? That removes both the duplicated compile work and the long-held mutex.

docker/runtime/RuntimeBase.dockerfile
```diff
@@ -31,3 +31,10 @@ RUN apt-get update && apt-get install -y python3 python3-venv && \
     ln -s /opt/iree/bin/iree-compile /usr/local/bin/iree-compile && \
     ln -s /opt/iree/bin/iree-import-onnx /usr/local/bin/iree-import-onnx && \
     apt clean && rm -rf /var/lib/apt/lists/*
+
+# Install OpenVINO converter tools for ML inference model import.
+ARG OPENVINO_VERSION=2025.3.0
+RUN python3 -m venv /opt/openvino && \
+    /opt/openvino/bin/pip install --no-cache-dir openvino==${OPENVINO_VERSION} && \
```
ls-1801, line 28: Probably worth a follow-up investigation: the runtime base image grows substantially with the OpenVINO add-on (locally I measured **1.33 GB disk** for the post-PR image vs ~190 MB pre-PR; OpenVINO contributes ~173 MB and the existing IREE install is already ~736 MB, dominated by `libIREECompiler.so` at ~337 MB).

A few cleanup ideas that *might* shave ~150-200 MB without affecting `ovc`/`iree-compile` functionality, but I have **not validated** these in a working build:
- `find /opt/{iree,openvino} -name __pycache__ -type d -exec rm -rf {} +` (~95 MB; bytecode regenerates on first import)
- `rm -rf /opt/iree/lib/python*/site-packages/onnx/backend/test` (~66 MB of ONNX backend test data)
- `rm -rf /opt/iree/lib/python*/site-packages/iree/_runtime_libs_tracy` (~11 MB; Tracy profiling, debug-only)
- `rm -rf /opt/{iree,openvino}/lib/python*/site-packages/pip*` (~24 MB; only needed at install time)
- Possibly the `**/tests` dirs in numpy/sympy (~30 MB) but those are smaller and riskier.

Each of these would need a smoke test (`ovc --help`, `iree-compile --help`, basic model conversion) before committing.

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
```diff
@@ -0,0 +1,297 @@
...
+        return;
+    }
+    NES_INFO("{}", detail::format_as(discovery));
+}
+
+bool OpenVinoImporter::isSupportedInput(const std::filesystem::path& modelPath)
+{
+    /// SavedModel inputs are directories. Do not validate existence here; `ovc`
+    /// reports bad or missing directories together with the rest of conversion failures.
+    if (modelPath.filename().empty() || modelPath.extension().empty())
```
ls-1801, line 214: Could we tighten the empty-extension branch to `std::filesystem::is_directory(modelPath)` so things that are clearly not a SavedModel (e.g. a typo'd path like `/data/iris` for an ONNX file the user meant to point at) get the friendly `OpenVINO loading supports .onnx, .pb, ...` error instead of an opaque `ovc` conversion failure? The `is_directory` check is consistent with the comment's "SavedModel inputs are directories" intent.

nes-inference/src/ModelAccess.hpp
```diff
@@ -31,7 +31,19 @@ struct ModelAccess
     static ImportedModel
     makeImported(RefCountedByteBuffer buf, std::string fnName, std::vector<size_t> inShape, std::vector<size_t> outShape)
     {
-        return ImportedModel{std::move(buf), std::move(fnName), std::move(inShape), std::move(outShape)};
+        return ImportedModel{
+            ModelBackend::IREE, std::move(buf), RefCountedByteBuffer{}, std::move(fnName), std::move(inShape), std::move(outShape)};
```
ls-1801: The backend tag is currently chosen by overload arity (4 args -> IREE, 5 args -> OPENVINO). That's fragile — a future importer that happens to need a different signature could silently end up tagged with the wrong backend. Consider taking `ModelBackend` explicitly or naming the factories per backend (`makeImportedIree` / `makeImportedOpenVino`).

ls-1801 (changes requested):
In general, my biggest concern with supporting multiple backends hidden behind build flags is testing.

We have to add a CI Job that verifies the behavior when building with each combination of backend flags, or we should drop the CMake flags entirely and dispatch to the correct backend.

I don't mind adding a second backend if we can hide it cleanly in nes-inference (which i think it is), and you are taking ownership over it. I would still want to investigate if it's possible to reduce the size of the Docker image.

And i would like to have compilation outside of the actual query compilation, to prevent long compilation times of models from blocking a(ll) worker threads

kykrueger:
I am keen to keep IREE since we had been using it for the SIGMOD Demo, and would like to still migrate the SIGMOD demo to work on the latest nebula-stream. If we cannot keep both in the pre-built containers, maybe we could have it enabled optionally with a build flag?

Furthermore, most of our compute solutions are NVIDIA based. I would like to get the acceleration working for some of our use-cases.

greenBene (changes requested):
If I understand correctly, you wanted to implement some changes here, right? If so, please rerequest once these changes are added :)

Artraxon (changes requested):
please rerequest when ready for review

gaturchenko:
Hey @ls-1801 @greenBene @Artraxon, I'm back to working on the PR. First, I removed IREE as it seems the majority is in favor of having 1 backend. Second, I addressed some of the comments by @ls-1801 with a few things to note:

1. I want to still accept the models with the variable batch size, I don't think it's user-friendly to force re-export with the static shapes. I however added warnings, rejection of models with batch size > 1 and a few extra tests.
2. I still want to do compilation in the setup call, as there are some compilation options I want to later parameterize, so I think this doesn't really belong to the lowering stage. I also fixed the model cache, should work now, although it doesn't evict anything, so we should discuss whether we want to limit it somehow.
3. I added the buffer size check as requested, but it's strange that in IREE we did it only for the output, not the input as well. With varsized it can happen that the payload we get is actually larger, and with `memcpy` now instead of `std::ranges` as previously, we will write out-of-bounds. But would be good to hear your take on that.

I'll now see to it that the workflows are green and then re-request the reviews

docs/guide/query_api.md
```diff
@@ -635,7 +635,7 @@ OUTPUT (setosa FLOAT32, versicolor FLOAT32, virginica FLOAT32);
 
 - The `INPUT` fields must match the model's input tensor shape and types. Each field maps to one element of the input tensor.
 - The `OUTPUT` fields must match the model's output tensor shape and types.
-- Only `.onnx` model files are supported. Models are compiled to IREE bytecode at first use.
+- `.onnx` and the other formats the OpenVINO converter accepts are supported. Models are converted to OpenVINO IR at registration time.
```
greenBene: Please list a few more formats that are supported. Focus on the "most used" so that most users don't have to check what OpenVino suports :)

nes-frontend/src/Statements/StatementHandler.cpp
```diff
@@ -256,13 +257,9 @@ ModelInfo toModelInfo(const RegisteredModel& model)
 ///
 /// The order matters:
 ///   1. Reject duplicate names before doing any work — the catalog does not
-///      replace an existing entry, and registration is not cheap (it reads
-///      the ONNX file, shells out to `iree-import-onnx`, and validates the
-///      declared schema against the model's tensor shape).
+///      replace an existing entry, and registration is not cheap
```
greenBene: I'd prefer to phrase it a bit more neutrally :) Sth. like 
```
1. Reject models with already registered names
2. Validate and register the given model in the catalog
```

nes-inference/src/ModelAccess.hpp
```diff
@@ -28,10 +28,14 @@ namespace NES::detail
 /// internal `src/` classes that actually construct models.
 struct ModelAccess
 {
-    static ImportedModel
-    makeImported(RefCountedByteBuffer buf, std::string fnName, std::vector<size_t> inShape, std::vector<size_t> outShape)
+    static ImportedModel makeImported(
+        RefCountedByteBuffer buf,
+        RefCountedByteBuffer auxBuf,
```
greenBene: What does aux buffer mean? Please use a more self-explanatory name :)

greenBene (changes requested):
Thank you! :) I find the OpenVINO code way more readable than the iree version! I just have some nitpicks. 

Please re-request once they are implemented and the CI is green :)

docs/guide/query_api.md
gaturchenko: Agreed, I should have the full list somewhere. However, it's a bit safer to only mention ONNX because there's a dedicated model loading test for it. If we explicitly say we support other formats, I think we should also add supporting tests. I can do that as a part of this PR in principle if necessary

nes-frontend/src/Statements/StatementHandler.cpp
gaturchenko: Agreed, will change it then

nes-inference/src/ModelAccess.hpp
gaturchenko: Aux in this case refers to auxiliary :) in case there are 2 outputs of the model conversion subprocess, such as for OpenVINO's `ovc`, this second `RefCountedByteBuffer` is necessary. However, I'd prefer a more generic name to preserve some sort of generality, as this object should be backend-agnostic

nes-frontend/src/Statements/StatementHandler.cpp
gaturchenko: Done

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
```diff
@@ -0,0 +1,136 @@
...
+    if (inputBufferSize > requiredInputSize)
+    {
+        throw NES::InferenceRuntimeFailure(
+            "Model Execution failed. Model input size {} B exceeds buffer capacity {} B", inputBufferSize, requiredInputSize);
+    }
+
+    if (outputBufferSize > requiredOutputSize)
+    {
+        throw NES::InferenceRuntimeFailure(
+            "Model Execution failed. Model output size {} B exceeds buffer capacity {} B", outputBufferSize, requiredOutputSize);
```
ls-1801: Are these checks both inverted?

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
```diff
@@ -0,0 +1,328 @@
...
+    }
+    return shape;
+}
+
+/// With rank >= 2 the leading dimension is the batch. A fixed batch above 1 means the
+/// model wants several samples per invocation, which inference cannot provide: it runs
+/// the model once per tuple. Such a model would silently treat one tuple's fields as a
+/// whole batch, so reject it. Rank 1 has no batch dimension to speak of, and a dynamic
+/// batch was already resolved to 1 above.
+std::expected<void, ImportError> validateBatchDimension(const std::vector<size_t>& shape)
```
ls-1801, line 160: Is that really a universal thing?

A tensor like this [3, 224, 224] is frequently used for images, without any batching.

Maybe we can be a bit more explicit during model registration and ask the user whether batching is used. Or be stricter and reject anything more complex than a single-dimensional vector.

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
```diff
@@ -0,0 +1,328 @@
...
+        if (dimension.is_dynamic() || dimension.get_length() < 0)
+        {
+            NES_WARNING(
+                "OpenVINO model {} tensor has a dynamic dimension at index {} (shape {}); assuming an extent of 1",
+                role,
+                index,
+                partialShape.to_string());
+            shape.push_back(1);
+            continue;
+        }
```
ls-1801: I would still say this is an awkward hack.

If we want to allow dynamic dimensions, we should limit ourselves to the first dimension (which we interpret as batch size).

ls-1801 (changes requested):

nes-inference/src/ModelAccess.hpp
Artraxon: Why should that byte buffer be backend-agnostic? Isn't it storing binary data of the model, which is backend-specific? I'd rather have a strongly typed out container which indicates to what backend it belongs. If at some point we'd have some layer of abstraction that manages models for multiple backends, you can put some type erasure layer on top.

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
gaturchenko: They are, you're right, fixed it.

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
gaturchenko: I think it's safe to say that it is. I've experimented with multiple models, both SOTA and from ML papers, and have not yet seen a single CV model that would not have the batch size set explicitly to 1 in the input tensor shape, e.g., `[1, 3, 224, 224]`. Same goes for keyword spotting models that accept a spectrogram instead of a raw audio wave as an input, they set the input shape in the same way.

I don't think we should be any stricter than that, because then we will severely restrict the number of models we can accept. And once batch inference is in place, we will not have to worry about this model rejection issue that much.

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
gaturchenko: Agreed, changed it.

nes-inference/src/ModelAccess.hpp
gaturchenko: Agreed, sounds reasonable. Do you want both `buf` and `auxBuf` strongly typed or the latter only?

docs/guide/query_api.md
gaturchenko: Extended the docs by explicitly naming each supported format and added respective model loading tests.

nes-inference/src/ModelAccess.hpp
Artraxon: I think both would be better

gaturchenko: Done, though I'm not sure this is exactly what you had in mind, so please take a look at [a612a3f](https://github.com/nebulastream/nebulastream/commit/a612a3fb39739226c5159252bd410363f05826fe)

ls-1801 (approved):

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
```diff
@@ -0,0 +1,141 @@
...
+RuntimeMetadata OpenVinoRuntimeBackend::setup(const CompiledModel& model)
+{
+    static ov::Core sharedCore;
+    static std::mutex coreMutex;
+    /// Every worker thread running the same operator calls setup with the same model, and
+    /// read_model/compile_model take seconds for a large one. Caching by model keeps that
+    /// cost at once per distinct model per process instead of once per thread. Entries live
+    /// for the process lifetime: they are bounded by the number of distinct models a worker
+    /// loads, and the payload they pin is already held by the operator using it.
+    static std::vector<CompiledModelCacheEntry> compiledModelCache;
```
Artraxon: Does the cache get cleaned up at some point? I feel like when a model gets deleted it should also clean up the cache. Maybe you can extract the logic into a meyers singleton

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
```diff
@@ -0,0 +1,141 @@
...
+
+    const std::scoped_lock lock(coreMutex);
```
Artraxon: Is the lock also needed for access to the sharedCore or only for accessing the cache?

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
```diff
@@ -0,0 +1,342 @@
...
+
+namespace NES
+{
+
+namespace
+{
+
+std::filesystem::path makeTemporaryDirectory()
+{
+    static std::atomic_uint64_t counter = 0;
```
Artraxon, line 59: just fiy: there is also tmpnam for creating temporary unique file names, but I am fine with this approach as well since you get more meaningful file names

Artraxon (changes requested):
Sorry for taking so long, looks mostly good to me

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
gaturchenko: Gotcha, thanks! So no changes here?

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
gaturchenko: Done

nes-inference/runtime/src/OpenVinoRuntimeBackend.cpp
gaturchenko: It's not needed for the core itself, but it should be held for `compile_model` so that the thread that locks first compiles, and all the other threads retrieve from the cache.

nes-inference/src/OpenVINO/OpenVinoImporter.cpp
Artraxon: no, this is also fine IMO

Artraxon (approved):
Looks good to me now, would be nice if you can squash your commits before merging

PR #1600, opened 2026-05-12
refactor(e2e): consolidate distributed bats suites into shared library

ls-1801:
## Summary
Refactors the docker-compose / distributed e2e bats suites (cli, repl,
systest) so they share one helper library instead of each carrying their
own copy-pasted ~150 lines of setup, teardown, image-build and
leaked-resource cleanup. CMake registration is pulled into a single
macro in `cmake/macros.cmake`.

Future sources/sinks that need their own e2e tests with a custom
docker-compose setup (e.g. MQTT, Kafka, Pipe) can now drop in a single
`.bats` file plus a one-line CMake call — the boilerplate is gone.

## What changed
- `scripts/testing/distributed_bats_lib.bash` — new shared library with
  two layers: primitives (env/exec checks, leaked-resource cleanup,
  JSON assertions, image builds) and a preset (`nes_distributed_*`)
  used by the cli/repl/MQTT-style suites. Callers set `NES_BATS_BIN_NAME`
  and source the lib via the CMake-injected `NES_BATS_LIB` env var.
- `cmake/macros.cmake` — single registration macro for distributed bats
  tests; replaces ad-hoc `add_test` / `set_tests_properties` blocks in
  three CMakeLists.
- `cmake/CheckTestUtilities.cmake` — smoke-test that bats can actually
  load `bats-support` / `bats-assert` / `bats-file`. If the binary is
  found but the helpers aren't loadable, bats tests are disabled with
  an actionable warning instead of failing at run time.
- `docker/dependency/Development.dockerfile` — install
  `bats-support` / `bats-assert` / `bats-file` and set `BATS_LIB_PATH`.
- `scripts/check_preamble.py` — also check `.bash` files for the license
  preamble.
- Existing `*.bats` files (cli, repl, systest) reduced to the
  suite-specific test cases; common machinery is gone.
- Bash function names lowercased for consistency with the helper libs.

## Test plan
- [ ] `ctest -R bats` inside the dev container passes
- [ ] Fresh checkout: apt-installed bats helpers are picked up via
      `BATS_LIB_PATH`
- [ ] Removing one of `bats-support` / `-assert` / `-file` causes
      `ENABLE_BATS_TESTS` to flip off with the new warning, not a runtime
      failure

---
This is **part 1 of 6 in a stack** made with GitButler:
- <kbd>&nbsp;1&nbsp;</kbd> #1600 👈


cmake/macros.cmake
```diff
@@ -94,3 +94,144 @@ macro(add_tests_if_enabled TEST_FOLDER_NAME)
...
+#
+# Required:
+#   NAME — ctest name
+#
+# Optional:
+#   BATS_FILE — bats file path relative to CMAKE_CURRENT_SOURCE_DIR
+#               (default: tests/offline.bats)
+#   JOBS      — bats parallelism (omit for sequential)
+#   EXTRA_ENV — additional KEY=VALUE env entries (escape hatch)
+function(add_offline_test)
```
alepping: the naming is off
one is called e2e, the other is called 'offline' and the description says its e2e, but without docker
why not simply:
add_e2e_without_docker?

alepping (approved):
To be honest, I still don't know what to think of the bats testing logic.
I really don't like that it handles e2e tests with quite a lot of 'bash' code, which is, at least for me, very unintuitive to read.
I wonder why we don't use python instead of bash? There must be frameworks like llvm lit that do something similar (probably better) compared to bats.

ls-1801:
@alepping:
I agree with the non-intuitive syntax for bash, this is why I integrated some bats utilities, to eliminate some of the awkwardness:

- [ "$status" -eq 1 ] (exit code) → assert_failure 1
- [ "$output" = "1" ] (stdout literally 1) → assert_output 1 (or assert_output --partial 1 for substring)
bats-assert also gives you assert_equal "$output" 1 as a general-purpose comparison

I would argue that for most of the e2e tests we are planning, we will just write a Python library that serves as a shell wrapper, which will then run commands. If we just keep them as plain bash (with some sugar for assertions), the tests are more self-documenting and allow reproducing the testcase with a few bash commands, rather than burying the real commands under layers of abstraction.

ls-1801:
@alepping i removed the offline test macro, both type of tests share the same macro. a DOCKER_COMPOSE flag was added to distinguish tests that use docker compose from tests that dont

cmake/macros.cmake
```diff
@@ -94,3 +94,98 @@ macro(add_tests_if_enabled TEST_FOLDER_NAME)
...
+# Required:
+#   NAME — ctest name
+#   BATS_FILE — bats file path relative to CMAKE_CURRENT_SOURCE_DIR
+#
+# Optional:
+#   DOCKER_COMPOSE — flag: run as a docker-compose-backed test (adds the
+#                    DockerCompose label, RuntimeBaseImage fixture, and the
+#                    NES_RUNTIME_BASE_IMAGE env var). Without it the test
+#                    runs the binaries directly (offline).
+#   JOBS      — bats parallelism (omit for sequential)
```
greenBene, line 112: What kind of info is required here? An integer or bool? Please make it explicit :)

nes-systests/systest/CMakeLists.txt
```diff
@@ -165,18 +165,11 @@ add_dependencies(nes-systest-lib test-data)
 
 add_tests_if_enabled(tests)
 
-if (ENABLE_DOCKER_TESTS AND ENABLE_BATS_TESTS)
```
greenBene, line 180: Was is a conscious decision that we removed the option to disable the docker/bats tests? No strong opinion on my side, just want to ensure that it didn´t get remove by accident :)

greenBene (approved):
I appreciate the consolidation of this functionality! :) I just had two minor remarks.

Issue #1515, opened 2026-03-20
BufferManager: Add memory poisoning to catch out-of-bounds access in debug mode


ls-1801:
## Problem

The `BufferManager` (`nes-memory/BufferManager.cpp:102-161`) allocates the entire buffer pool as a single contiguous block via `posix_memalign()` and then linearly carves out slots. Each slot is `[ControlBlock (64B aligned)][Payload]`, packed back-to-back with no gaps:

```
[CB][Payload][CB][Payload][CB][Payload]...
```

Because all memory is valid within the single allocation, ASan cannot detect out-of-bounds writes that overflow from one page into an adjacent control block or payload. There is **no existing ASan/MSan instrumentation**

### Relevant Code

- **Single allocation:** `BufferManager::initialize()` in [`nes-memory/BufferManager.cpp:135`](nes-memory/BufferManager.cpp) — `memoryResource->allocate(allocatedAreaSize, withAlignment)`
- **Buffer carving loop:** [`nes-memory/BufferManager.cpp:146-158`](nes-memory/BufferManager.cpp) — walks `basePointer` by `offsetBetweenBuffers` to create `MemorySegment`s
- **Control block layout:** `BufferControlBlock` in [`nes-memory/TupleBufferImpl.hpp:80-170`](nes-memory/TupleBufferImpl.hpp) — `alignas(64)`, placement-new at start of each slot
- **Deallocation:** `BufferManager::~BufferManager()` at [`nes-memory/BufferManager.cpp:94`](nes-memory/BufferManager.cpp)

## Proposed Solution

Introduce **poisoned guard regions** in debug/sanitizer builds between buffer slots and around control blocks.

### Target Memory Layout

```
[POISON][CB][POISON][Payload][POISON]  [POISON][CB][POISON][Payload][POISON] ...
```

### Implementation Approach

1. **ASan poisoning (preferred when building with `-fsanitize=address`):**
   - After carving the pool, call `ASAN_POISON_MEMORY_REGION` on guard regions inserted between each buffer slot
   - Poison a guard region before/after each control block within a slot
   - On buffer recycle (`recyclePooledBuffer`), re-poison the guard regions

2. **`mprotect` guard pages (fallback for non-ASan debug builds):**
   - Align slot boundaries to page size and insert `PROT_NONE` guard pages between slots
   - Requires adjusting the allocation size to account for guard pages and page-alignment constraints

### Implementation Considerations

[...]

PR #1601, opened 2026-05-12
feat(memory): prefill buffer pool with marker in debug builds

ls-1801:
## Summary
In debug builds, prefill the buffer pool's freshly-allocated storage
with the repeating ASCII marker `NEBUSTRM` (as `uint64_t`) right after
`memoryResource->allocate`. Reads from uninitialized regions of a
tuple buffer now show up as an obvious recognizable pattern in dumps
and test output, instead of whatever bytes the allocator happened to
hand back.

Guarded by `#ifndef NDEBUG` — zero cost in release.

## Test plan
- [ ] Debug build: a deliberately uninitialized read returns the marker
- [ ] Release build: no change in BufferManager init behavior

---
This is **part 2 of 6 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #1601 👈 

### DIFF
--- a/nes-memory/BufferManager.cpp
+++ b/nes-memory/BufferManager.cpp
@@ -14,7 +14,9 @@
 #include <Runtime/BufferManager.hpp>
 
 #include <algorithm>
+#include <array>
 #include <atomic>
+#include <bit>
 #include <chrono>
 #include <cstdint>
 #include <cstdio>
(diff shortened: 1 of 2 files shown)


alepping (changes requested):
Really cool change, I'd just like to see an example.
You said 'should show up', but ideally, you can show that they do show up and document somewhere what to look out for.

ls-1801:
I could write a unit test that checks that if compiled in debug, the buffer is filled with the pattern, but I am not sure how useful that would be.

I am not sure where I would document this. There is #1515, which adds more debugging features to the tuple buffer in debug mode, so maybe finding a suitable place for the debugging documentation can be postponed until then.

I used this to debug an issue with the MQTT sink. If you don't trim the buffer in the sink by the number of tuples (which represents the number of formatted bytes), then the test would pass because the buffer is effectively zero-terminated. Introducing this pattern would result in incorrect results, which is what we want.

alepping:
@ls-1801 
I see, the changes in #1515 are a pretty cool follow-up.
You could still just add a comment the the .hpp, instead of keeping this an implementation detail in the .cpp.
Someone might use this during a debugging session, if they have the knowledge, including an AI agent.

alepping (changes requested):
How about leaving a comment at the top of the .hpp to tell people that the markers are a possible debug technique in debug mode (which might also help outside of sanitizer runs)

ls-1801:
added a comment

keyseven123 (approved):

alepping (approved):

PR #1602, opened 2026-05-12
feat(functions): add CHAR_LENGTH and OCTET_LENGTH for VARSIZED text

ls-1801:
## Summary
Adds two SQL length functions over `VARSIZED` text:

- `OCTET_LENGTH(x)` — number of bytes, returned as `UINT64`.
- `CHAR_LENGTH(x)`  — number of UTF-8 codepoints, returned as `UINT64`.

Both follow the established logical/physical function split and plug
into the function registry. `CHAR_LENGTH` is implemented branchlessly
(`nautilus::select` over the "is leading byte" predicate) so the JIT
can vectorize the per-byte loop.

## What changed
- `nes-logical-operators/.../{Char,Octet}LengthLogicalFunction.{hpp,cpp}`
  — UINT64-returning logical functions with the usual registry hooks.
- `nes-physical-operators/.../{Char,Octet}LengthPhysicalFunction.{hpp,cpp}`
  — physical implementations over `VariableSizedData`.
- `nes-systests/function/varsized/{CharLength,OctetLength}.test` —
  ASCII / multibyte / emoji / empty-string coverage, plus composition
  with `CONCAT`.

## Test plan
- [ ] `ctest -R "CharLength|OctetLength"` passes
- [ ] `CHAR_LENGTH('äöü')` returns 3, `OCTET_LENGTH('äöü')` returns 6
- [ ] Emoji rows (`🎉🚀`, `Hi🌍`) report correct codepoint counts

---
This is **part 3 of 6 in a stack** made with GitButler:
- <kbd>&nbsp;3&nbsp;</kbd> #1602 👈 


nes-physical-operators/src/Functions/CharLengthPhysicalFunction.cpp
```diff
@@ -0,0 +1,64 @@
...
+    const auto size = variableSizedData.getSize();
+
+    /// Count UTF-8 codepoints: every byte except continuation bytes (0b10xxxxxx) starts a new codepoint.
+    /// Branchless via nautilus::select so the JIT can vectorize the loop.
+    nautilus::val<uint64_t> count = 0;
+    for (nautilus::val<uint64_t> i = 0; i < size; i = i + nautilus::val<uint64_t>(1))
+    {
+        const nautilus::val<int8_t> byte = *(ptr + i);
+        const auto isLeading
+            = (byte & nautilus::val<int8_t>(static_cast<int8_t>(0xC0))) != nautilus::val<int8_t>(static_cast<int8_t>(0x80));
```
alepping: constexpr for the magic hex numbers would improve readability

alepping (changes requested):
Besides the magic hex numbers:
1. why not implement as plugins? we currently don't have any function plugins, it would really help to have at least one that someone can copy/take as guidance. You could also put just one of them in nes plugins
2. I checked a few other codebases, and couldn't find the Octet naming. I don't think its intuitive at all. I also thing having just length would be nice, but I admit, that [duckdb](https://duckdb.org/docs/current/sql/functions/text) returns the number of chars, and [clickhouse](https://clickhouse.com/docs/sql-reference/data-types/string) returns the number of bytes. ClickHouse has an additional `lengthUTF8` function for chars. Imo, the ClickHouse approach is the cleanest/most intuitive for me.

btw. I saw that Flink uses Character_Length or similar, I still think the clickhouse way is more intuitive/logical.

ls-1801:
> 1. why not implement as plugins? we currently don't have any function plugins, it would really help to have at least one that someone can copy/take as guidance. You could also put just one of them in nes plugins

Because I don't see why you would want to turn them off.

> 2. checked a few other codebases, and couldn't find the Octet naming. I don't think its intuitive at all. I also thing having just length would be nice, but I admit, that [duckdb](https://duckdb.org/docs/current/sql/functions/text) returns the number of chars, and [clickhouse](https://clickhouse.com/docs/sql-reference/data-types/string) returns the number of bytes. ClickHouse has an additional lengthUTF8 function for chars. Imo, the ClickHouse approach is the cleanest/most intuitive for me.

grafik

We can always add aliases; everyone has their own opinions. I would stick to the SQL standard.

nes-physical-operators/src/Functions/CharLengthPhysicalFunction.cpp
ls-1801: addressed

alepping:
@ls-1801 
Because I don't see why you would want to turn them off.

But when do we turn the checksum sink off?
Thats not a good argument imo.
A new user of NebulaStream that wants to implement a function as plugin currently has no example of how to do it.
We can also always move the function back into the kernel later on.

Concerning the SQL standard, we don't follow it anyway in many regards. Why not go the more intuitive route?
Also, you can reverse that argument. We can always be SQL compliant by adding the OCTET alias later.

alepping (changes requested):
see reply to your arguments

nes-logical-operators/include/Functions/CharLengthLogicalFunction.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include <DataTypes/Schema.hpp>
+#include <Functions/LogicalFunction.hpp>
+#include <Util/Logger/Formatter.hpp>
+#include <Util/PlanRenderer.hpp>
+#include <Util/Reflection.hpp>
+
+namespace NES
+{
+
+class CharLengthLogicalFunction final
```
greenBene, line 33: Please add a short documentation what this function does. Just to give an intuition, 2-3 lines are more than enough.

nes-logical-operators/include/Functions/OctetLengthLogicalFunction.hpp
```diff
@@ -0,0 +1,83 @@
...
+#include <DataTypes/Schema.hpp>
+#include <Functions/LogicalFunction.hpp>
+#include <Util/Logger/Formatter.hpp>
+#include <Util/PlanRenderer.hpp>
+#include <Util/Reflection.hpp>
+
+namespace NES
+{
+
+class OctetLengthLogicalFunction final
```
greenBene, line 33: Same as above, please add short documentation

nes-physical-operators/include/Functions/CharLengthPhysicalFunction.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include <Functions/PhysicalFunction.hpp>
+#include <Nautilus/DataTypes/VarVal.hpp>
+#include <Nautilus/Interface/Record.hpp>
+#include <Arena.hpp>
+
+namespace NES
+{
+
+class CharLengthPhysicalFunction final
```
greenBene, line 27: Again, very short doc would be great :)

greenBene (changes requested):
Thank you for implementing these two functions! :) 

My only feedback would be the request to add some in class documentation  what these functions are doing. 

Regarding the discussion above: 

I do think the function names `CHAR_LENGTH` and `OCTET_LENGTH` make sense, since they are also used by other mayor vendors. While we are not (yet?) SQL compliant, I think we should strive towards it.. 

Regarding plugins: While I agree with @alepping that a having an example plugin implementation for functions would be interesting, I think it would be out of scope for this PR, since its a fundamental function. The checksum sink is imo more of a testing source and not required for production release, so I think there it makes sense to not add it to the core :)

yschroeder97:
@alepping Regarding functions as a plugin, I am currently working on the NES tutorial for HDMS in athens, and I hope that we can get the case-when function in afterwards. That would suit as an example function as a plugin.
I have also written a complementary guide for this function.

ls-1801:
@alepping, we do turn off the checksum sink when building without tests enabled. (which is why the checksum sink is not known for our on-demand Docker images)

Regarding the naming, I am going to stick with the SQL standard.

Artraxon (approved):

Artraxon:
I agree that by default we should stick to the SQL standard. Some name not being intuitive enough is not good enough of a reason to deviate IMO. 

There are some parts of the SQL standard that have holes (like numeric types) or are just really messed up that no one but Oracle implements anyway (like the multi-part interval precisions), but I feel like this is an easy compliance W .

keyseven123 (approved):

greenBene (approved):

alepping (approved):

alepping (changes requested):
Sorry to block again, but I just noticed that there are some bugs:
Problematic test:
```sql
CREATE LOGICAL SOURCE textStream(a VARSIZED NOT NULL, b VARSIZED NOT NULL);
CREATE PHYSICAL SOURCE FOR textStream TYPE File;
ATTACH INLINE
hello,abcd

# Query succeeds, but should produce 'Error 2002'
SELECT CHAR_LENGTH(a, b) AS len FROM textStream INTO File();
----
4
```
```sql
CREATE LOGICAL SOURCE textStream(a VARSIZED NOT NULL, b VARSIZED NOT NULL);
CREATE PHYSICAL SOURCE FOR textStream TYPE File;
ATTACH INLINE
hello,abcd

# Query succeeds and returns OCTET_LENGTH(b) = 4, but should produce 'Error 2002'.
SELECT OCTET_LENGTH(a, b) AS len FROM textStream INTO File();
----
4
```
Also, I'm not sure whether null handling is problematic. Need to investigate.
Edit: null handling seems to be fine, as far as I determined via testing and looking at the code, sorry for false positive

alepping:
@Artraxon
(I'm fine with the naming, but just to explain my naming-concern): 
My main gripe with the naming was that there are many different character sets that SQL supports. PostgreSQL does support the above two functions, but it executes char_length character-set-aware. Since our char_length function is hardcoded to UTF8, I'd rather call the function UTF8_Length to clearly communicate that, or throw if a different character set is detected. I'm just not a big fan of requiring implicit knowledge that could be 'encoded' into names of functions/classes/whatever.

ls-1801:
I fixed that and added tests for it, but I think there should be a more general approach to function arity checking to prevent this from happening. 

I have not addressed the UTF8 naming.

alepping (approved):

PR #1603, opened 2026-05-12
fix(sources): generator emits tuples larger than one buffer

ls-1801:
## Summary
The `GeneratorSource` could not emit tuples whose serialized form was
larger than a single tuple buffer. The old code kept leftover bytes in
an `orphanTuples` string and tried to re-prepend them on the next
`fillTupleBuffer` call, but a tuple that was *itself* bigger than the
buffer hit the "doesn't fit" path every iteration and the source
produced no output.

This rewrites the spill mechanism to use the existing `tuplesStream`
(`std::stringstream`) as the persistent buffer:

- `generateTuple` writes one tuple at a time into the stream.
- `fillTupleBuffer` drains as many bytes as fit via `readsome`,
  advancing through the available memory area.
- When the stream is empty and the generator has more to give, the
  next tuple is generated. Otherwise the partial buffer is returned.
- EOS is reported only when the generator has signaled stop *and*
  the spill stream is drained — so the final partial buffer is no
  longer dropped.

Removes the `orphanTuples` member and the byte-accounting logic that
went with it.

## What changed
- `nes-plugins/Sources/GeneratorSource/GeneratorSource.{cpp,hpp}` —
  rewrite the fill loop, drop `orphanTuples`, reset spill stream on
  `open`.
- `nes-systests/sources/Generator.test` — new case using
  `RANDOMSTR 4096` and `OCTET_LENGTH` to assert every row arrives at
  the configured length.

## Test plan
- [ ] `ctest -R Generator` passes
- [ ] New 4096-byte RANDOMSTR case emits all 20 rows with length 4096
- [ ] Existing small-tuple Generator tests still pass

> Depends on #1602 (`OCTET_LENGTH`) for the new systest case.

---
This is **part 4 of 6 in a stack** made with GitButler:
- <kbd>&nbsp;4&nbsp;</kbd> #1603 👈 

### DIFF
--- a/nes-systests/sources/Generator.test
+++ b/nes-systests/sources/Generator.test
@@ -170,3 +170,35 @@ INTO generator_sink_text;
 17,yellow,X
 18,brown,bl5
 19,yellow,IgiKX
+
+# Testing tuple generation larger than one buffer
+CREATE LOGICAL SOURCE generator_large_text(id UINT64 NOT NULL, text VARSIZED NOT NULL);
+CREATE PHYSICAL SOURCE FOR generator_large_text TYPE GENERATOR SET(
+    'ONE' as `SOURCE`.STOP_GENERATOR_WHEN_SEQUENCE_FINISHES,
+    1 as `SOURCE`.SEED,
+    'SEQUENCE UINT64 0 20 1, RANDOMSTR 8192 8192' AS `SOURCE`.GENERATOR_SCHEMA
+);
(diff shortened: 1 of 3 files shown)


nes-systests/sources/Generator.test
```diff
@@ -170,3 +170,34 @@ INTO generator_sink_text;
 17,yellow,X
 18,brown,bl5
 19,yellow,IgiKX
+
```
adrianmichalke, line 173:
```suggestion
# Testing tuple generation larger than one buffer

```

adrianmichalke (approved):

alepping (approved):

Artraxon (approved):

PR #1606, opened 2026-05-13
Replace nlohman with reflectcpp

alepping:
## Note
I'm fully open to discussing these changes with @Artraxon and also to waiting for a potential merge until we merge the schema inference.

## Descripton
I have not been a big fan of the global dependency (global because of utils) of nlohman json.
I also ran into name conflicts, because of 'JSONOutputFormatter' (so did @LeoRaasch).
Recently, @yschroeder97 asked why we don't use reflectcpp instead of nlohman.
  
So I looked into it and came to the conclusion that reflectcpp can fully replace nlohman json. As a result, we get rid of a dependency and boilerplate code. Additionally, we get struct/type guarantees on top.
  
Main changes:
 - removes ~30 lines of boilerplate code (does not add any!)
  - One JSON library instead of two. Less surface, fewer headers pulled in
  - `nlohmann` was linked globally from the root CMakeLists, so every target paid for it
  - The two `*JsonOutputFormatter*` files (plugin runtime vs. frontend statement renderer) had near-identical names and unrelated purposes. The rename kills that ambiguity.
  - The plugin's `JSONOutputFormatter` only used `nlohmann` for string escaping; replaced with ~20 lines of hand-rolled escape, removing the dep from the plugin entirely.


Artraxon (approved):

nes-common/include/Identifiers/NESStrongType.hpp
```diff
@@ -30,7 +30,7 @@ namespace NES
 /// In general a Identifier should not expose its underlying Type and no code should depend on specific values. This is the reason why
```
yschroeder97:
```suggestion
/// In general, an identifier should not expose its underlying type, and no code should depend on specific values. This is the reason why
```

nes-common/include/Identifiers/NESStrongType.hpp
```diff
@@ -30,7 +30,7 @@ namespace NES
...
 /// only a limited subset of operations are supported, and it is not default constructible. Identifiers are orderable and hashable to
 /// be used as keys in maps.
-/// We Introduce overloads for nlohmann, yaml and fmt to make the identifiers feel ergonomic.
+/// We Introduce overloads for JSON, yaml and fmt to make the identifiers feel ergonomic.
```
yschroeder97:
```suggestion
/// We introduce overloads for JSON, YAML, and fmt to make the identifiers feel ergonomic.
```

nes-systests/systest/include/SystestRunner.hpp
```diff
@@ -83,12 +83,20 @@ inline std::string discardPerformanceMessage(RunningQuery&)
     SystestProgressTracker& progressTracker,
     const QueryPerformanceMessageBuilder& queryPerformanceMessage);
 
+struct BenchmarkResult
+{
+    rfl::Rename<"query name", std::string> queryName;
```
yschroeder97, line 91: Why the rename?

nes-frontend/include/Statements/StatementJsonSerializers.hpp
```diff
@@ -0,0 +1,157 @@
...
+#include <fmt/format.h>
+#include <magic_enum/magic_enum.hpp>
+#include <rfl/Generic.hpp>
+#include <rfl/Object.hpp>
+#include <rfl/to_generic.hpp>
+
+/// Local converters that translate frontend statement-result values into rfl::Generic for JSON output.
+/// We deliberately do NOT specialize rfl::Reflector for shared types like Schema, DataType, NESStrongType, etc.,
+/// because those types already have NES::Reflector specializations that the rfl::Reflector partial-specialization
+/// in <Util/Reflection.hpp> picks up. Specializing rfl::Reflector again here would cause an ODR clash across TUs
```
yschroeder97: What is ODR?

nes-plugins/OutputFormatters/JSONOutputFormatter/JSONOutputFormatter.cpp
```diff
@@ -51,6 +52,53 @@ namespace NES
 {
 namespace
 {
+std::string escapeAsJsonString(std::string_view input)
```
yschroeder97, line 58: Comment this fn, please. Looks a bit out of place.

yschroeder97 (changes requested):
I like it! Some minor things.

nes-common/include/Identifiers/NESStrongType.hpp
alepping: done

nes-common/include/Identifiers/NESStrongType.hpp
alepping: done

nes-systests/systest/include/SystestRunner.hpp
alepping: On main the old nlohmann code emitted the key literally as "query name" (with a space). rfl::Rename keeps that exact key so the file format stays byte-identical for existing consumers. A space can't be a C++ identifier, hence the rename rather than just naming the member query name. The other three keys don't need it.

nes-frontend/include/Statements/StatementJsonSerializers.hpp
alepping: One Definition Rule (https://en.wikipedia.org/wiki/One_Definition_Rule).
Maybe you have used 'inline' before? One of the most confusing keywords in c++ imo.
ODR means that classes/structs/functions can only be defined once per program and types once per translation unit (TU).
If you violate it, you usually get a 'duplicate definition' or similar compiler error.

nes-plugins/OutputFormatters/JSONOutputFormatter/JSONOutputFormatter.cpp
alepping: created a comment

yschroeder97 (approved):

PR #1607, opened 2026-05-13
feat(cli): accept optional `nullable` flag on YAML schema fields

ls-1801:
## Summary
- Schema fields in the topology YAML now accept an optional `nullable: <bool>` key (default `false`) that is propagated to the resulting `DataType`.
- `acceptKeys` for `SchemaField` was extended to include `nullable`, so unknown-key errors continue to fire for everything else.
- Added offline.bats coverage: happy-path dump, accepted-keys list, non-bool rejection. Fixture: `tests/good/nullable-schema.yaml`; negative case: `tests/bad/invalid_nullable_value.yaml`.

## Test plan
- [x] `cmake --build cmake-build-debug --target nes-cli -j`
- [x] `bats nes-frontend/apps/cli/tests/offline.bats --filter nullable` (3 tests pass)
- [ ] Full `offline.bats` run in CI

### DIFF
--- a/nes-frontend/apps/cli/CLIStarter.cpp
+++ b/nes-frontend/apps/cli/CLIStarter.cpp
@@ -333,9 +333,12 @@ struct convert<NES::CLI::SchemaField>
 {
     static bool decode(const Node& node, NES::CLI::SchemaField& rhs)
     {
-        acceptKeys({"name", "type"}, node);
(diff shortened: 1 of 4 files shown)


greenBene (approved):
Nice! Thanks!

yschroeder97 (approved):

PR #1608, opened 2026-05-14
fix(QueryId): seed Docker-style name generator with random_device

keyseven123:

  `generateHorseName()` in `nes-distributed/src/DistributedQueryIdGenerator.cpp` produces the Docker-style name attached to every distributed query. It is a per-process function-local `static std::mt19937` that was seeded with
   `std::time(nullptr)` — second-resolution wall-clock. Inside a single process the PRNG advances per call, so names are unique. Across processes, any two `nes-cli ... start ...` invocations whose static initializers run in
  the same wall-clock second seed identically and emit the same first name. The seed weakness was already acknowledged by a `// NOLINT(cert-msc51-cpp)` suppression on the line.

  This PR swaps the seed source to `std::random_device`, removes the lint suppression, and drops the now-unused `<ctime>` include.

  Key changes (`nes-distributed/src/DistributedQueryIdGenerator.cpp`):
  - `static std::mt19937 rng(static_cast<unsigned>(std::time(nullptr))); /// NOLINT(cert-msc51-cpp)` → `static std::mt19937 rng(std::random_device{}());`
  - Drop `#include <ctime>`.
  - Expand the doc comment to explain why the seed source matters (concurrent CLI invocations within the same second).

  ## How it manifests

  Any time multiple `nes-cli` invocations are launched back-to-back from the same machine:
  - shell loops / CI scripts submitting several queries in sequence,
  - `docker compose` startup with `depends_on` driving multiple CLI submitters,
  - `xargs -P` parallel submissions,
  - two engineers running CLI commands on the same host within the same second.

  It is not specific to the priority work. The priority benchmark fires HIGH and LOW within ~50 ms, which is why we hit it constantly — but the same collision predates the `priority` field. Before distinct priority labels
  existed, duplicate query names were cosmetically annoying; once "the HIGH query" and "the LOW query" carried different priorities, identical names became confusing in logs and tripped downstream dedup paths.

  ## Verifying this change

  - `for i in $(seq 1 20); do nes-cli ... start ... ; done` — names are now distinct across every invocation; previously the first ~N invocations falling in the same wall-clock second collided.
  - Existing distributed bats tests (`distributed-cli-test`, `distributed-repl-tests`) continue to pass — name generation contract within a single process is unchanged.
  - `clang-tidy` no longer needs the `cert-msc51-cpp` suppression on this line.

  - `nes-distributed` distributed query id generation only. No call-site changes; the function signature is unchanged.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -126,10 +126,16 @@ jobs:
           files: build/junit-unit.xml,build/junit-systest.xml,build/junit-docker.xml
 
   # Log level tests — verifies the NES_LOG_LEVEL compile-time flag works across all build types.
+  # Fanned out one job per build_type so each shard only loops the 6 log levels and fits the
+  # 40-min wall-clock cap even on fork PRs (which run without remote sccache credentials).
   build-and-test-log-level:
-    name: "Log Level Tests"
+    name: "Log Level Tests (${{ matrix.build_type }})"
     timeout-minutes: 40
     runs-on: ubuntu-24.04
+    strategy:
+      fail-fast: false
+      matrix:
+        build_type: [Debug, RelWithDebInfo, Release, Benchmark]
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


ls-1801 (changes requested):
This only easies the symptom, the fundamental issue that there is no duplication check when creating the id. And the space of potential ids is not particularly large so the test is bound to become flakey.

I don't mind the change itself, maybe we can append a smallish (4digits) random number at the end to decrease the odds of a collision, but I would remove the test.

The coordinator will eventually fix the issue, or we have to adapt the frontends to detect collisions and retry.

keyseven123:
> This only easies the symptom, the fundamental issue that there is no duplication check when creating the id. And the space of potential ids is not particularly large so the test is bound to become flakey.
> 
> I don't mind the change itself, maybe we can append a smallish (4digits) random number at the end to decrease the odds of a collision, but I would remove the test.
> 
> The coordinator will eventually fix the issue, or we have to adapt the frontends to detect collisions and retry.

I am fully aware that this is a quick fix and should be done once the coordinator is there. I have removed the test, though.
If @yschroeder97 is almost ready with a branch, or there is already a PR, I am more than happy to rebase this branch onto one that has a coordinator to implement the unique ID check. Please let me know.

ls-1801 (approved):

yschroeder97 (approved):

PR #1605, opened 2026-05-12
feat(plugins): add MQTT source and sink

ls-1801:
## Summary
Adds MQTT source and sink plugins backed by Eclipse Paho C++.

- **`MQTTSource`** — subscribes to a topic, pushes payloads into tuple
  buffers. Payloads that don't fit the current buffer are held in
  `PayloadStash`, a small consume-only scratchpad that retains the
  unconsumed tail across `fillTupleBuffer` calls and shrinks back to
  a default capacity after oversized payloads. A single large message
  can span multiple buffers without unbounded growth.
- **`MQTTSink`** — publishes formatted tuple buffers to a topic at the
  configured QoS. Outbound backpressure is delegated to
  `BackpressureHandler` (introduced in #1604): when
  `max_outstanding_messages` QoS≥1 deliveries are still unacked, the
  buffer is stashed and backpressure is acquired until acks drain.
  QoS 0 is fire-and-forget by spec so the knob is inert there.

## Config
| Param | Source | Sink | Default |
|---|---|---|---|
| `server_uri` | ✓ | ✓ | required |
| `client_id` | ✓ | ✓ | auto-generated |
| `topic` | ✓ | ✓ | required |
| `qos` | ✓ | ✓ | 1 (0/1/2) |
| `flush_interval_ms` | ✓ | — | 0 |
| `retained` | — | ✓ | false |
| `max_outstanding_messages` | — | ✓ | 128 |

## What changed
- `nes-plugins/Sources/MQTTSource/` — source plugin, `PayloadStash`,
  config parameters, bats suite, docker-compose generator, example
  fixtures with ASCII topology comments.
- `nes-plugins/Sinks/MQTTSink/` — sink plugin, config parameters,
  bats suite, docker-compose generator, example fixtures.
- `nes-plugins/CMakeLists.txt` — register both plugins.
- `vcpkg/vcpkg-registry/ports/paho-mqtt/` and
  `paho-mqttpp3/` — new ports with local patches (libcxx-19 fixes,
  missing `unsigned_char_traits`, ARM unresolved-symbol fix, libuuid
  ODR fix).
- `vcpkg/vcpkg.json` — pull in `paho-mqttpp3`.

## Test plan
- [ ] `ctest -L bats` runs the MQTT distributed suites against the
      mosquitto compose stack
- [ ] Source: single payload larger than the buffer is split across
      multiple buffers without loss; `PayloadStash` capacity returns
      to default after the burst
[...]

PR #1615, opened 2026-05-16
Add mqtt systest approach merged v3

alepping:
# External-system systests for plugins

## Goal

This PR adds support for testing plugins that depend on external systems while keeping the authoring and review experience as close as possible to ordinary NebulaStream systests.

## History

This feature started in #1354, which added MQTT support together with an initial “test container systest” approach. #1605 later reworked the MQTT plugin around dedicated distributed Bats tests. This PR keeps the useful infrastructure work from that line of development, but brings external-system testing back into the systest workflow and extends it so that these tests can be written, run, debugged, and reviewed like normal systests.
Note that the first 4 commits come from PR #1600, which currently lags behind main.

### 1. Keep one primary end-to-end test format

#1605 uses the existing distributed Bats path for this feature: Bats files, topology YAML, helper scripts, and ad-hoc result checks. That is a valid testing path in NebulaStream, but it further splits plugin end-to-end tests across two authoring models instead of concentrating this feature in the already dominant systest workflow. This PR keeps external-system coverage inside the systest infrastructure, so plugin authors and reviewers reuse the same DSL, runner, CI integration, and future systest improvements.

### 2. Keep the workload compact and reviewable

In a systest, input, query, and expected output live together in one file. In #1605, the same workload is split across shell code, YAML topology/configuration, and result files, which makes the actual test intent harder to read and review. The new approach keeps the external dependency setup separate, but the workload itself remains a normal, compact systest.

Screenshot from 2026-05-17 18-44-51

### 3. Give useful failure feedback

The Bats tests in #1605 mostly fail through shell assertions such as `grep ... results.csv` or line-count checks. When one of those fails, the signal is primarily “this shell predicate failed”; it does not naturally show how expected and actual results differ. Systests already provide explicit expected-vs-actual output, so failures are faster to understand and easier to review in CI.

Screenshot from 2026-05-17 18-46-21

### 4. Preserve the existing IDE workflow

Systests already have gutter actions for running tests directly from the file. This PR keeps that workflow for external-system tests instead of sending users to a separate Bats entry point. After the dispatcher rewrite, the same gutter flow also supports debugging, which matters because the quickest path after a failing plugin test is often to rerun it under the debugger immediately.

screencast doesn't really show it, but debug/run are triggered via gutter symbol:
[Screencast from 2026-05-17 18-49-22.webm](https://github.com/user-attachments/assets/9f00e7b8-b318-4b5f-871e-9916de655ec4)
[Screencast from 2026-05-17 18-47-16.webm](https://github.com/user-attachments/assets/e2963581-a07f-4d13-9a32-b8603893ee1f)

### 5. Keep logs and local diagnosis in the same place

External dependencies add enough complexity on their own; test execution should not make diagnosis harder. The new path keeps ordinary systest logging and adds targeted diagnostics for external-profile startup and reachability failures, so local runs and CI failures point at the real setup problem instead of only surfacing a generic downstream failure.

[Screencast from 2026-05-17 18-50-28.webm](https://github.com/user-attachments/assets/5bcdfe31-8350-433c-9d3f-ac6391747f9a)

### 6. Stay extensible beyond MQTT

The implementation is profile-based rather than MQTT-specific: a plugin provides a colocated external-system profile, while the systest infrastructure handles provisioning and execution. MQTT is the first consumer, but the intended proof point is that a second dependency shape, such as ODBC, can use the same flow without introducing another bespoke end-to-end test path.

TODO: ODBC poc impl

## Net result

After the external profile is set up, writing an external-system test is intentionally boring: create a normal `.test` file, add `# requires: <profile>`, and use the usual systest execution, logging, CI, and debugger workflow.


ls-1801:
> The Bats tests in https://github.com/nebulastream/nebulastream/pull/1605 mostly fail through shell assertions such as grep ... results.csv or line-count checks.

This is not a limitation of bats, that's just lazy by me, something that should be pointed out in a code review.

i could have just as well write something like:

```bash
    assert_equal_ignore_order output.csv <<EOF
111,000,11,23
100,2,1
42,8,9
EOF
```

```
Files differ (ignoring order):
  actual (output.csv)          | expected
100,2,1                          100,2,1
111,000,11,23                    111,000,11,23
999,5,7                       |  42,8,9
```

ls-1801 (changes requested):
re-request once ready

PR #1621, opened 2026-05-21
fix(docker): suggest matching images when requested combo is missing

keyseven123:

`scripts/install-local-docker-environment.sh` only checked whether the exact `${HASH}-${ARCH}-${STDLIB}-${SANITIZER}` tag existed; on miss it told the user to rebuild dependencies locally even when another image for the same hash was available. This PR queries Docker Hub for tags sharing the hash on the host arch and either lists them (`-y`) or shows an interactive menu (numeric pick reassigns `TAG`/`STDLIB`/`SANITIZER` and continues, `r` falls into the existing `-l` rebuild path, `q` aborts) — no need to re-invoke the script.

## Verifying this change

Run on arm64 where CI never builds `libstdcxx`:
- `bash scripts/install-local-docker-environment.sh -y --libstdcxx --thread --no-vcpkg-cache` — lists the four available `libcxx` combos and exits 1.
- Same command without `-y` — interactive menu; picking a numeric option proceeds with the new tag, `r` triggers the local rebuild, `q` aborts.
- Forcing a junk hash exercises the "no images for this hash" branch (interactive prompt to rebuild, or exit 1 under `-y`).

### DIFF
--- a/scripts/install-local-docker-environment.sh
+++ b/scripts/install-local-docker-environment.sh
@@ -56,6 +56,19 @@ usage() {
     exit 1
 }
 
+# Query Docker Hub for tags that share the given dependency hash on the given arch.
+# Prints one tag per line on stdout; prints nothing when no tags exist (or when curl is unavailable).
+list_hash_alternatives() {
+    local hash="$1"
+    local arch="$2"
+    command -v curl >/dev/null 2>&1 || return 0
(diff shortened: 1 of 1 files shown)


ls-1801 (approved):
works for me

greenBene (approved):
Nice!

PR #1629, opened 2026-05-27
feat(Functions): adds 'Contains' string function

alepping:
Adds 'Contains' function on VarSized/Strings.
Includes null handling and negative tests.
Does not include UTF8 tests, since we don't properly support UTF8 (or rather different string formats).

Lives in nes-plugins, may be moved into nes-logical/physical-operators if you consider it core functionality, though I'd like to keep it in nes-plugins as a showcase.

### DIFF
--- a/nes-plugins/Functions/ContainsFunction/CMakeLists.txt
+++ b/nes-plugins/Functions/ContainsFunction/CMakeLists.txt
@@ -1,0 +1,19 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#    https://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include(${PROJECT_SOURCE_DIR}/cmake/PluginRegistrationUtil.cmake)
+
+add_plugin_as_library(Contains LogicalFunction nes-logical-operators-registry contains_logical_function_plugin_library ContainsLogicalFunction.cpp)
+add_plugin_as_library(Contains PhysicalFunction nes-physical-operators-registry contains_physical_function_plugin_library ContainsPhysicalFunction.cpp)
+
+target_include_directories(contains_logical_function_plugin_library PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/)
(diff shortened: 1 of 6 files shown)


adrianmichalke (approved):

nes-plugins/Functions/ContainsFunction/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include(${PROJECT_SOURCE_DIR}/cmake/PluginRegistrationUtil.cmake)
+
+add_plugin_as_library(Contains LogicalFunction nes-logical-operators-registry contains_logical_function_plugin_library ContainsLogicalFunction.cpp)
+add_plugin_as_library(Contains PhysicalFunction nes-physical-operators-registry contains_physical_function_plugin_library ContainsPhysicalFunction.cpp)
+
```
Artraxon, line 17: Unreflection missing ^^

nes-plugins/Functions/ContainsFunction/ContainsLogicalFunction.cpp
```diff
@@ -0,0 +1,136 @@
...
+#include <fmt/format.h>
+#include <ErrorHandling.hpp>
+#include <LogicalFunctionRegistry.hpp>
+#include <SerializableVariantDescriptor.pb.h>
+
+namespace NES
+{
+
+ContainsLogicalFunction::ContainsLogicalFunction(LogicalFunction haystack, LogicalFunction needle)
+    : haystack(std::move(haystack)), needle(std::move(needle)), dataType(DataTypeProvider::provideDataType(DataType::Type::BOOLEAN))
```
Artraxon, line 42: I would prefer terms like searchIn, searchFor instead of something like an analogy, which might be difficult for some people with poor English to understand.

nes-plugins/Functions/ContainsFunction/ContainsLogicalFunction.hpp
```diff
@@ -0,0 +1,84 @@
...
+static_assert(LogicalFunctionConcept<ContainsLogicalFunction>);
+
+}
+
+namespace NES::detail
+{
+struct ReflectedContainsLogicalFunction
+{
+    std::optional<LogicalFunction> haystack;
+    std::optional<LogicalFunction> needle;
```
Artraxon, line 80: you shouldn't need the optional wrappers anymore

Artraxon (changes requested):
Looks fine in general, although of course I'd prefer to merge it in after schema inference.

But if you need it in mainline for anything else quickly, go ahead

PR #1639, opened 2026-05-29
feat(docs): add testing guidelines

keyseven123:
## Summary
- Add `docs/standards/testing_guidelines.md` covering when and how to write unit tests, system-level tests, connector tests, and distributed tests
- Includes contribution recipes by component type and a CI schedule table


docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
+NebulaStream uses a layered testing strategy.
+Each layer has a defined scope, and layers complement rather than replace each other.
+**Never rely on only one test type for a contribution.**
+A unit test verifying component logic and a systest verifying end-to-end correctness are not interchangeable — write both.
```
alepping: don't like 'write both'
imo we DEFINITELY should not force ppl into writing unittests
systests often suffice, only write a unit test if necessary, or obviously better suited

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+|-------------------|---------------------------------------------------|-----------------------------------------------------|-------------------------------------|
+| Unit test         | 1–3 translation units, no engine, no network      | Algorithmic components with many enumerable cases   | Every PR                            |
+| System-level test | Full query pipeline, single node                  | Operators, functions, formatters, source/sink types | Every PR                            |
+| Connector test    | Source/sink byte-correctness, engine not involved | New source or sink plugin                           | Subset per PR, full nightly/weekly  |
+| Distributed test  | Multi-node topology                               | Distributed routing or failure scenarios            | Subset per PR, chaos nightly/weekly |
+
+# Unit Tests
+
+Unit tests exercise a single component in isolation, typically spanning 1–3 `.cpp` files.
+No full queries, no network, no external services.
```
alepping: Sounds overly specific.
Why 1-3 .cpp files?

Why not full queries and no network?
We might want to mock a network connection to test e.g., the network sink/source.

The input-formatter/engine/pagedvector unit tests are all beyond 1-3 files for good reasons imho.

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Write a unit test when** the component has algorithmic logic with many enumerable deterministic cases — input combinations, edge cases, error paths.
+Optimizer rules, expression evaluation, data structure operations, and protocol parsers are typical candidates.
```
alepping: are they?

I really wonder whether we should be specific here, I mean our codebase is the documentation, being specific here just risks getting out-of-sync with the codebase

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+Unit tests let you verify these cases cheaply and precisely without spinning up a full engine.
```
alepping: I don't think spinning up the engine is a big problem though.
Without compiling, systests are extremely fast, I don't buy that argument fully.
Imho, the bigger issue is that we run systests using dozens of combinations, meaning we run all systests dozens of times.

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+
+- Place tests under `{module}/tests/UnitTests/` and register them in the module's `CMakeLists.txt`.
+- Shared utilities live in `nes-common/tests/Util/include/`. Link against `nes-test-util`.
+
+# System-Level Tests (Systests)
+
+Systests exercise the full query pipeline on a single embedded node: SQL parsing → logical plan → optimization → execution → result validation.
+They are the primary correctness layer for the query engine.
+
+**Always write a systest** when adding or changing an operator, SQL function, formatter, or source/sink type.
```
alepping: why when changing an operator? the existing systests may suffice (I might just did a minor refactor)

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+The systest proves the component is reachable via a real SQL query and produces correct output.
+
+Tests live in `nes-systests/` as declarative `.test` files.
```
alepping: might change, if we enable per plugin, which I personally think we should

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+| Differential (`====`)           | Two queries must produce identical output        |
+| Checksum sink                   | Large result sets compared against               |
+| Negative (error code)           | Query must fail, validating error handling paths |
+
+## Group assignment
+
+Tag each test in the header (`# groups: [...]`).
+Groups control CI scheduling — most new tests belong to `Fast`.
+Reserve `Large` and `benchmark` for tests that cannot run within a short CI window.
+
```
alepping, line 66: You could also mention in-test configurations, e.g., buffer sizes
imo very practical/useful (note that when setting multiple values, e.g., for buffer sizes, the tests are executed for each value

you might also want to talk about the 'bug' systests (.disabled)?

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+# Connector Tests
+
+Connector tests verify the byte-level correctness of a source or sink plugin **without the engine**.
```
alepping: It goes quite a bit beyond that.
It generally tests correct interaction, per connector,  between nebulastream and an external service, based on scenarios
one scenario focuses on expected vs actual correctness (is not byte-based for DBs (row based))
But also checks that reconnecting is possible, that buffers are lost/not-lost, given the right settings
It also checks valid/invalid configs
(and more)

The high-level way it works is sequence based (to avoid non-determinism).
It uses toxiproxy (shopify) to inject errors between nebulastream and external services.
There is a source and a sink test harness, and both essentially expose a 'open/fill/close' interface
The python testing framework can call these and sync with the framework. It can open (establishes connection), then cut the connection with toxi proxy, then call 'fill', and then check that it returns an expected error.
Scenarios can have multiple of these steps in different orders.

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+They test that the connector faithfully moves data between an external service and into the engine.
+
+- **Source side**: Python runner seeds data into the external service → source ingests it → emitted tuples are compared against the seed.
+- **Sink side**: buffers populated with known data → sink writes to the external service → Python runner reads back and compares against the original.
+
+These tests use a forthcoming Python-based framework that will replace BATS.
+The framework is scenario-based: connectors opt into the scenarios they support, and the harness iterates registered plugins and runs opted-in scenarios automatically.
```
alepping: thats correct

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+Groups control CI scheduling — most new tests belong to `Fast`.
+Reserve `Large` and `benchmark` for tests that cannot run within a short CI window.
+
+# Connector Tests
+
+Connector tests verify the byte-level correctness of a source or sink plugin **without the engine**.
+They test that the connector faithfully moves data between an external service and into the engine.
+
+- **Source side**: Python runner seeds data into the external service → source ingests it → emitted tuples are compared against the seed.
+- **Sink side**: buffers populated with known data → sink writes to the external service → Python runner reads back and compares against the original.
```
alepping: correct in the simple path, but more complex scenarios require more complex interaction

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+
+These tests use a forthcoming Python-based framework that will replace BATS.
+The framework is scenario-based: connectors opt into the scenarios they support, and the harness iterates registered plugins and runs opted-in scenarios automatically.
+
+## Scenarios
```
alepping: recently updated this significantly
but honestly, maybe just link to 'nes-conn-test/README.md'?
i put significant effort into curating it, might be better than needing to sync between the two (except if this really is just a throw-away document for our internal discussion)

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,132 @@
...
+Distributed tests also use the forthcoming Python framework for orchestration.
+
+# When Tests Run
+
+| Test type                            | Per PR   | Nightly | Weekly |
+|--------------------------------------|----------|---------|--------|
+| Unit tests                           | ✓ all    | ✓       | —      |
+| System-level tests (standard groups) | ✓ all    | ✓       | —      |
+| System-level tests (large groups)    | —        | ✓       | —      |
+| Connector tests                      | ✓ subset | ✓ full  | ✓ full |
```
alepping, line 91: not sure here yet (I'd say full for now (pretty fast) and maybe we change that  when it gets too slow?)

alepping (changes requested):

docs/standards/testing_guidelines.md
keyseven123: Changed it.

docs/standards/testing_guidelines.md
keyseven123: Removed the specific examples.

docs/standards/testing_guidelines.md
keyseven123: Removed 1-3 .cpp files and the specific examples

docs/standards/testing_guidelines.md
keyseven123: Got rid of the spinning up argument.

docs/standards/testing_guidelines.md
keyseven123: Changed it to always write a slt if one adds a new operator.

docs/standards/testing_guidelines.md
keyseven123: Just got rid of this line and reference now the systest.md file.

docs/standards/testing_guidelines.md
keyseven123: Added in-test configurations.

Not sure what I would write about the `bug` systests. What do you have in mind?

docs/standards/testing_guidelines.md
keyseven123: I will simply refer to an upcoming doc for this section that most probably lies under nes-conn-test/README.md I do not think that it makes sense to have this duplicated.

docs/standards/testing_guidelines.md
keyseven123: I will simply refer to an upcoming doc for this section that most probably lies under nes-conn-test/README.md I do not think that it makes sense to have this duplicated.

docs/standards/testing_guidelines.md
keyseven123: I will simply refer to an upcoming doc for this section that most probably lies under nes-conn-test/README.md I do not think that it makes sense to have this duplicated.

docs/standards/testing_guidelines.md
keyseven123: > maybe we change that when it gets too slow?)

Exactly what I had in mind. Then we would change this doc as well.

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,113 @@
...
+
+- Place tests under `{module}/tests/UnitTests/` and register them in the module's `CMakeLists.txt`.
+- Shared utilities live in `nes-common/tests/Util/include/`. Link against `nes-test-util`.
+
+# System-Level Tests (Systests)
+
+Systests exercise the full query pipeline on a single embedded node: SQL parsing → logical plan → optimization → execution → result validation.
+They are the primary correctness layer for the query engine.
+
+**Always write a systest** when adding an operator, SQL function, formatter, or source/sink type.
```
ls-1801: i thougt we are not testing sources and sinks in the systest?

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,113 @@
...
+|---------------------------------|--------------------------------------------------|
+| Inline expected output (`----`) | Result set is small and deterministic            |
+| Differential (`====`)           | Two queries must produce identical output        |
+| Checksum sink                   | Large result sets compared against               |
+| Negative (error code)           | Query must fail, validating error handling paths |
+
+## Group assignment
+
+Tag each test in the header (`# groups: [...]`).
+Groups control CI scheduling — most new tests belong to `Fast`.
```
ls-1801: does not Large implies Fast or do we have to be explicit?

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,113 @@
...
+Reserve `Large` for tests that cannot run within a short CI window.
+Additionally, please keep in mind to run a given `.test` file with in-test worker configuration.
+For example, adding `GlobalConfiguration worker.default_query_execution.operator_buffer_size: [4096, 8192]` automatically runs the queries with two different buffer sizes.
+
+# Connector Tests
+
+Connector tests verify the byte-level correctness of a source or sink plugin **without the engine**.
+On a quite high-level, they test that the connector faithfully moves data between an external service and into the engine.
+For more details, please see the upcoming documentation under `nes-conn-test/README.md`
```
ls-1801: probably should be a todo

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,113 @@
...
+# When Tests Run
+
+| Test type                            | Per PR   | Nightly | Weekly |
+|--------------------------------------|----------|---------|--------|
+| Unit tests                           | ✓ all    | ✓       | —      |
+| System-level tests (standard groups) | ✓ all    | ✓       | —      |
+| System-level tests (large groups)    | —        | ✓       | —      |
+| Connector tests                      | ✓ subset | ✓ full  | ✓ full |
+| Distributed tests (deterministic)    | ✓ subset | ✓       | —      |
+| Distributed tests (chaos)            | —        | ✓       | ✓      |
```
ls-1801, line 93: i would mention chaos testing as we dont have anything like that

ls-1801 (changes requested):

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,113 @@
...
+| Unit test         | 1–3 translation units, no engine, no network      | Algorithmic components with many enumerable cases   | Every PR                            |
+| System-level test | Full query pipeline, single node                  | Operators, functions, formatters, source/sink types | Every PR                            |
+| Connector test    | Source/sink byte-correctness, engine not involved | New source or sink plugin                           | Subset per PR, full nightly/weekly  |
+| Distributed test  | Multi-node topology                               | Distributed routing or failure scenarios            | Subset per PR, chaos nightly/weekly |
+
+# Unit Tests
+
+Unit tests exercise a single component in isolation.
+
+**Write a unit test when** the component has algorithmic logic with many enumerable deterministic cases — input combinations, edge cases, error paths.
```
ThanGeo: But we should be careful and not stretch this. Imo unit tests should be  smart-exhaustive, not trying to naively test every single edge case. For example we shouldn't be adding unit tests for basically every possible precondition/invariant violation in the new code, right? If so, maybe improve the phrasing?

greenBene: +1. Maybe use a framing such as "Create unit tests for the happy path and for the relevant edge cases. Be on point with the tests and keep the number of tests as small as sensible"?

docs/guide/how_to_add_a_sink.md
greenBene: `guide/how_to_add_a_source.md` does not contain the mentioned testing logic. I assume it will be part of Aljoscha's work?

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,113 @@
...
+TEST_F(MyComponentTest, handlesEdgeCase)
+{
+    /// ...
+}
+```
+
+- Place tests under `{module}/tests/UnitTests/` and register them in the module's `CMakeLists.txt`.
+- Shared utilities live in `nes-common/tests/Util/include/`. Link against `nes-test-util`.
+
+# System-Level Tests (Systests)
```
greenBene, line 52: I feel the whole section is redundant to `docs/development/systests.md`. I´d suggest focusing here on the reasons for choosing an SLT over a unit test, but keeping the documentation on how to write an SLT in its own doc. :)

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,113 @@
...
+Use judgment if a layer genuinely does not apply, but the default expectation is both a unit test and a systest.
+
+| Contribution             | Unit test                                      | Systest | Connector test                      |
+|--------------------------|------------------------------------------------|---------|-------------------------------------|
+| SQL function             | Yes — cover both interpreter and compiled path | Yes     | —                                   |
+| Operator                 | Yes                                            | Yes     | —                                   |
+| Optimizer rule           | Yes — enumerate each rule-triggering case      | Yes     | —                                   |
+| Input / output formatter | Yes                                            | Yes     | —                                   |
+| Source plugin            | Yes                                            | Yes     | Yes — opt into applicable scenarios |
+| Sink plugin              | Yes                                            | Yes     | Yes — opt into applicable scenarios |
```
greenBene: I'm not sure how much value the table gives if it basically just confirms that, yes, you should, where applicable, create Unit tests and SLT, but no strong obejction from my side if you want to keep it :)

greenBene (changes requested):
Just minor comments :) I really like the doc, and that we have a clearer understanding of what the test types are for.  Thank you for writing it!

docs/standards/testing_guidelines.md
Artraxon: IMO optimizer rules should generally speaking not be testable from systests, since optimization is hidden from the user.

Artraxon (changes requested):

docs/standards/testing_guidelines.md
greenBene: Fully agree!

alepping (approved):

docs/standards/testing_guidelines.md
keyseven123: Yes, sorry. Changed

docs/standards/testing_guidelines.md
keyseven123: Done

docs/standards/testing_guidelines.md
keyseven123: Done

docs/standards/testing_guidelines.md
keyseven123: TBH, I think we should not test only the happy path; if there is a precondition, it should be tested. Especially in the era of GenAI tools, I think tests are becoming even more useful. 
 I have added a sentence that advises being sensible with unit tests, though.

docs/guide/how_to_add_a_sink.md
keyseven123: Yes

docs/standards/testing_guidelines.md
keyseven123: Removed and referenced `docs/development/systests.md`

docs/standards/testing_guidelines.md
keyseven123: Done

greenBene (approved):

ls-1801 (approved):

docs/development/systests.md
```diff
@@ -33,6 +33,14 @@ The second line (starting with `# description:`) is a text that describes what t
 The third line (starting with `# groups:`) lists in brackets groups to which the given test suite belongs, such as "Sources", "Aggregation", or "Union".
 Groups are used to select all test cases from this group for testing, or exclude test cases from within this group when executing all other tests.
 
+### In-Test Worker Configuration
+A test file can override worker configuration via `GlobalConfiguration`.
+Listing multiple values runs every query of the file once per value:
+```
+GlobalConfiguration worker.default_query_execution.operator_buffer_size: [4096, 8192]
```
ThanGeo, line 40: I have used this so many times in the system tests files and the execution arguments and I always forget the syntax for both. 
Maybe it makes sense to add examples/documentation of all possible global configuration?
E.g. join strategy = [HASH_JOIN, NESTED_LOOP_JOIN] etc. and whatever else there is

docs/standards/testing_guidelines.md
```diff
@@ -0,0 +1,103 @@
...
+**Skip a unit test** for pure glue code — registration forwarding, trivial descriptor construction, wiring with no logic of its own.
+These are better verified at the systest level.
+
+All unit tests extend `BaseUnitTest` and use Google Test / Google Mock:
+
+```cpp
+#include <BaseUnitTest.hpp>
+#include <gtest/gtest.h>
+
+class MyComponentTest : public Testing::BaseUnitTest
```
ThanGeo, line 37: With PagedVector and ChainedHashMap redesigns, we introduced property testing. Shouldn't this be the new default at least for data structures or other data heavy testing? Maybe it makes sense to include it in the documentation. @ls-1801

ThanGeo (changes requested):

docs/development/systests.md
keyseven123: As these global configurations might change in the future and are a lot, I would rather like to refer to the worker and other configs

docs/standards/testing_guidelines.md
keyseven123: I have added a reference to property testing

docs/development/systests.md
ThanGeo: but can they be found somewhere?

ThanGeo (approved):

docs/development/systests.md
keyseven123: You can get them by runnign the executable with `--help`

adrianmichalke (approved):

Artraxon (approved):

yschroeder97 (approved):

PR #1642, opened 2026-05-29
feat(Input/Output-Formatting): Make Value (De)serializers registry-extensible

### LINKED ISSUE #1535: [FEAT] Add (de)serializer registry for different data types
### Description

Now that both InputFormatter and OutputFormatter, and thus string <-> c++ type serialization are properly configurable via a descriptor parameter map, we should add a (De)Serializer Registry for the currently supported data types. Value (de)serialization currently takes up a not insignificant amount of CPU time, both for input and output formatting. For the purpose of experimenting with different (de)serializers for data types, and enable the integration of (de)serialization of more complex, user-defined  data-types with possibly multiple possible input / output formats (like different timestamp standards), it would be helpful to be able to configure value (de)serializers in a plug & play fashion.

Unique, data type based (de)serializers will also help to handle the formatting demands for certain types of certain formats. For instance, JSON demands special chars to be delimited. This could be handled by creating a specific JSONCharValueSerializer instead of bloating the JSONOutputFormatter with switch cases over the data types.

### Acceptance Criteria

- [x] Value Deserializer Registry (textual to c++ type)
- [x] Value Serializer Registry (c++ type to textual)
- [x] Add (de)serializer parameter to parameter map of input and output formatter
- [x] Adjusting the input and output formatters to use the defined value (de)serializers
- [x] Adding the default options as first (de)serializer implementations
- [x] Create seperate nullhandling (de)serializer for both Input and Output Parsers
- [x] Adding a (de)serializer plugin as proof of concept

LeoRaasch:
Currently, the textual - binary and binary - textual value conversions for textual formats are hard-coded into the `InputFormatter` / `OutputFormatter` components (see `RawValueParser` and `OutputFormatterUtil` files). This PR aims to separate the value conversions from these components by introducing the registry-extensible `ValueSerializer` / `ValueDeserializer` components.

Each data type is assigned a specific (de)serializer implementation. The PR comes with the `Default<DataType>ValueSerializer` and `Default<DataType>ValueDeserializer` implementations respectively, which adapt the current methods to (de)serialize the values. Furthermore, the `JSONCHAR` and `JSONVARSIZED` ValueSerializers have been added as plugin to support the JSON-specfic character escape rules for textual values.

Users may use different serializers for specific data types, by adding new Value(De)serializers as plugin. While each input and output format specifies a default (de)serializer implementation for each data type, the used (de)serializer for each data type may be reconfigured by using the newly added `VALUE_DESERIALIZERS` and `VALUE_SERIALIZERS` config parameters of the `InputFormatterDescriptor` and `OutputFormatterDescriptor` respectively. These parameters expect a string containing a comma-separated lists consisting of `<DataType>:<Serializer-Key>` segments (documented in `query_api.md`).

Each specific deserializer type requires a separate `Nullable<Deserializer-Key>` implementation, which handles the deserialization of nullable values. The `provideValueDeserializer` function automatically fetches the nullable variant of the deserializer, if the field is nullable.

We did not add `ValueDeserializerDescriptor` or `ValueSerializerDescriptor` classes at the moment, as all currently relevant serializer related knobs are represented via `InputFormatterDescriptor` / `OutputFormatterDescriptor` config parameters. The ValueDeserializerRegistry takes the flags `quoted` and `hasTrailingSpaces` as arguments, while the `ValueSerializerRegistry` takes the `quoted` flag as argument.

This PR fulfills the following goals:

- Add means to configure specific (de)serializers for specific data types and add new (de)serializers via plugins.
- Keep `RawBufferIndex` phase as lean as possible -> the phase now simply accesses the raw field value based on the result of the `InputFormatIndexer` phase and calls the ValueDeserializer to deserialize the raw value.
- Keep `OutputFormatter` phase as lean as possible -> the phase is responsible to write format specific field and tuple delimiters as well as additional data like field names, while the ValueSerializer serializes the actual values.
- Remove all switch cases over the `DataType::Type` enum in the input formatter and output formatter components -> they are replaced with Value(De)serializer registry calls, which are more maintainable.
## Verifying this change
Added `ValueDeserializerOverrides.text` and ValueSerializerOverrides.test` systests to cover the `VALUE_DESERIALIZERS` and `VALUE_SERIALIZERS` parameters. All remaining tests still pass without adjustments taken.

- nes-input-formatters
- nes-output-formatters

## Documentation
Added `Value Deserializer` and `Value Serializer` sections to `query_api.md`

This PR closes #1535


nes-output-formatters/src/OutputParsers/DefaultOutputParsers.cpp
```diff
@@ -0,0 +1,367 @@
...
+#include <function.hpp>
+#include <val_arith.hpp>
+#include <val_bool.hpp>
+#include <val_ptr.hpp>
+
+namespace NES::DefaultOutputParser
+{
+
+template <typename T>
+T parseNumeric(
```
ls-1801: Why does this return `T`? It should return `uint64_t`, matching what `writeValueToBuffer` returns (bytes written to main memory).

nes-output-formatters/src/OutputParsers/DefaultOutputParsers.cpp
```diff
@@ -0,0 +1,367 @@
...
+                stringWithDoubledQuotes.append("\"\"");
+            }
+            else
+            {
+                stringWithDoubledQuotes += character;
+            }
+        }
+        stringFormattedValue = "\"" + stringWithDoubledQuotes + "\"";
+    }
+    return writeValueToBuffer(stringFormattedValue.c_str(), remainingSpace, tupleBuffer, bufferProvider, bufferStartingAddress);
```
ls-1801: `writeValueToBuffer` (`OutputFormatterUtil.hpp:48`) takes `const char*` and internally builds `std::string_view{value}`, which runs `strlen`. Passing `.c_str()` here silently truncates a VARSIZED payload at the first embedded `\0`. Same issue at line 54 (`parseChar` for a `\0` CHAR field writes zero bytes). Pre-existing on `main` — not introduced by this PR — but worth fixing during the reorg.

A few things worth pulling on here:

1. Perf: why does `parseVarsized` allocate a temporary `std::string` at all? For the non-quoted path we already have `(valueAddress, valueSize)` — no need to copy. Even the quoted path could stream the escape into the tuple buffer directly instead of building a temporary and then copying it in.

2. API: `writeValueToBuffer` should take `std::string_view` (or an explicit `const char* + size_t`) and callers pass `.data()` + `.size()`. That fixes the NUL truncation everywhere.

3. Broader (not this PR): Id slowly move away from treating VARSIZED as std::strings altogether. Their content is arbitrary bytes, user-provided, of unknown encoding — very different from system-internal strings. Having a type distinction between "system string" and "user byte payload" would let us stop reaching for `std::string` / `string_view` on the user-data side by accident.

nes-input-formatters/src/InputParsers/DefaultInputParsers.cpp
```diff
@@ -0,0 +1,643 @@
...
+    nautilus::val<bool> isNull = *getMemberWithOffset<bool>(parseResult, offsetof(DefaultInputParser::ParseResult<bool>, isNull));
+    return VarVal{nautilusValue, true, isNull};
+}
+
+VarVal DefaultCHARInputParser::parseToVarVal(
+    const nautilus::val<int8_t*>& fieldAddress, const nautilus::val<uint64_t>& fieldSize, const std::vector<std::string>& nullValues) const
+{
+    /// Cut off quotes if neccessary
+    const nautilus::val<int8_t*> trueFieldAddress = quoted ? fieldAddress + nautilus::val<uint32_t>(1) : fieldAddress;
+    const nautilus::val<uint64_t> trueFieldSize = quoted ? fieldSize - nautilus::val<uint32_t>(2) : fieldSize;
```
ls-1801: **`fieldSize - 2` underflows on unquoted / short / null fields.**

The quoted-cut runs unconditionally before any bounds check. If `quoted == true` and the field size is 0 or 1 (e.g. an empty JSON string `""` is exactly at the boundary, and a JSON `null` value comes back as `{nullptr, 0}` from `getRawValueFromIndex<true>`), then `fieldSize - 2` underflows the `uint64_t`. The garbage size is then passed to `checkIsNullProxy`, which constructs `std::string{ptr, ptr + huge}` → `std::length_error` → `std::terminate`.

Same pattern at lines **154, 171, 470, 507** (both nullable and non-nullable variants of CHAR and VARSIZED).

Reproducer, crashes on this branch:

```
CREATE LOGICAL SOURCE nullableCharStream(letter CHAR);
CREATE PHYSICAL SOURCE FOR nullableCharStream TYPE File SET(JSON AS INPUT_FORMATTER."TYPE");
ATTACH INLINE
{"LETTER": "a"}
{"LETTER": null}
{"LETTER": "b"}

CREATE SINK letterSink(letter CHAR) TYPE File;
SELECT * FROM nullableCharStream INTO letterSink;
```

Stack (excerpt):
```
Terminate called after throwing an instance of std::length_error: basic_string
#4 NES::DefaultInputParser::checkIsNullProxy(signed char const*, unsigned long, ...)
```

Fix options:
1. Do the null check first, then do the quoted-cut only when the field is non-null and `size >= 2`.
2. `PRECONDITION(fieldSize >= 2)` before the cut, and have the JSON path return `size = 2` (`""`) for null instead of `{nullptr, 0}`.
3. Move the quoted-cut into `parseFixedSized` after the null branch.

(1) is the least invasive.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
```diff
@@ -50,8 +50,102 @@
...
+    if (accessSIMDJsonFieldOrThrow(currentDoc, fieldName).is_null())
+    {
+        return true;
+    }
+    return false;
+}
+
+/// (Proxy) functions being called via nautilus::invoke() can not be member functions. Thus, we need to implement them outside of the class
+template <bool Nullable>
+RawJsonAccessResult* getRawValueFromIndex(const FieldIndex fieldIndex, RawBufferIndex* rawBufferIndex, const InputFormatIndexer* indexer)
```
ls-1801: **Non-nullable JSON fields silently mis-parse an explicit `null` value.**

The `Nullable` branch above does `checkIsNullJsonProxy` first; the non-nullable branch skips that entirely and just returns `raw_json()`, which for a `null` literal is 4 bytes `"null"`. That flows into whatever `Default*InputParser` is registered for the fields type. What the user gets depends on the type:

Reproducer (`nes-systests/formatter/NonNullFieldGetsJsonNull.test`):
```
CREATE LOGICAL SOURCE nonNullUint(u UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR nonNullUint TYPE File SET(JSON AS INPUT_FORMATTER."TYPE");
ATTACH INLINE
{"U": 1}
{"U": null}
{"U": 3}
```
(same shape for `CHAR NOT NULL` and `VARSIZED NOT NULL`)

Observed:

| Field type | Behavior on JSON `null` |
|---|---|
| `UINT64 NOT NULL` | throws `CannotFormatMalformedStringValue: Value null is not a valid value of type: unsigned long` |
| `CHAR NOT NULL` | throws `ul is not a supported char value` — the quoted-cut chopped `null` to `ul`, so the error references bytes that never appear in the input |
| `VARSIZED NOT NULL` | **silently emits `"ul"` as the field value.** No error, wrong data. |

The VARSIZED case is the worst — no exception, no error, just wrong data in the sink. The CHAR error message will also send anyone debugging on a wild goose chase looking for `"ul"` in the source.

Suggested fix: the non-nullable branch should also check for JSON `null` and throw a proper "field X received JSON null but schema declares NOT NULL" error identifying the field. That way all three types fail consistently and understandably.

nes-input-formatters/src/InputParsers/DefaultInputParsers.cpp
```diff
@@ -0,0 +1,643 @@
...
+{
+template <typename T>
+struct ParseResult
+{
+    T value;
+    bool isNull;
+};
+
+/// We expect a pointer and the size so that we can use this method from the nautilus runtime
+bool checkIsNullProxy(const int8_t* fieldAddress, const uint64_t fieldSize, const std::vector<std::string>* nullValues) noexcept
```
ls-1801: `checkIsNullProxy` is `noexcept`, but its body constructs `std::string{fieldAddress, fieldAddress + fieldSize}` which can throw `std::bad_alloc` / `std::length_error`. If it throws, `noexcept` calls `std::terminate` instead of unwinding.

Either drop the `noexcept`, or wrap the throwing call so the exception is handled inside the function.

nes-input-formatters/include/InputFormatters/InputFormatIndexer.hpp
```diff
@@ -42,13 +42,17 @@ class InputFormatIndexer
 
     [[nodiscard]] virtual std::string_view getTupleDelimitingBytes() const = 0;
     [[nodiscard]] virtual std::string_view getFieldDelimitingBytes() const = 0;
-    [[nodiscard]] virtual QuotationType getQuotationType() const = 0;
     [[nodiscard]] virtual const std::vector<std::string>& getNullValues() const = 0;
 
+    /// Get the parser type for a specific datatype
+    [[nodiscard]] std::string getParserType(const DataType::Type& dataType) const { return parserTypes.at(dataType); }
```
ls-1801: `.at()` throws `std::out_of_range`, not a project `Exception`. Use `find()` and throw a proper project exception. Also return `const std::string&` instead of by value to avoid the per-lookup copy.

nes-input-formatters/registry/include/ValueDeserializerRegistry.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+namespace NES
+{
+
+using InputParserRegistryReturnType = std::unique_ptr<InputParser>;
+
+struct InputParserRegistryArguments
+{
+    bool quotedText;
+    bool hasTrailingSpaces;
```
ls-1801, line 31: Naming inconsistency: `InputParserRegistryArguments::hasTrailingSpaces` is plural here, but the same concept is `InputParserConfig::hasTrailingSpace` (singular) and the `Default*Parser` member fields are also singular. The cross-boundary translation lives in `InputParserUtil.cpp` (`arguments{.hasTrailingSpaces = config.hasTrailingSpace}`). Pick one spelling and use it everywhere.

ls-1801 (changes requested):
Sorry for the late review, I have to put my veto against this — especially against the SQL API changes.

I don't know of any other system that has a knob like INPUT_PARSERS/OUTPUT_PARSERS per data type. I don't see how this could be superior to just using the default parser implementation and allowing an escape hatch that accepts a string and calls a function on it. A good optimizer could push that into the RawScan.

I don't mind reorganizing the parsing functions internally to avoid the duplications — that part is a genuine improvement. The inline comments cover the correctness bugs and internal-structure issues that survive that reorg. Everything tied to the user-facing INPUT_PARSERS/OUTPUT_PARSERS config surface I've deliberately left off, since I'd rather see that whole surface removed.

This PR would also be a greate place to get rid of this output praser naming psychosis and finally call the formatter or serializer or anything except parser...

having functions like this doing the actual formatting is super unintuative: 
```c++
T parseNumeric( 
    const T value,
    int8_t* bufferStartingAddress,
    const uint64_t remainingSpace,
    TupleBuffer* tupleBuffer,
    AbstractBufferProvider* bufferProvider)
```

nes-output-formatters/src/OutputParsers/DefaultOutputParsers.cpp
LeoRaasch: True, this is a bug. I fixed it

nes-output-formatters/src/OutputParsers/DefaultOutputParsers.cpp
LeoRaasch: Adopted your suggestions for 1 and 2. 3 sounds reasonable as well.

nes-input-formatters/src/InputParsers/DefaultInputParsers.cpp
LeoRaasch: Good catch for the NullableChar case. In the other cases, it should not be a problem: For not-nullable chars/varsized values, we always expect the quotes to be present, as they should not be null. NullableVarsized already performs the null-check before cutting off the quotes.

I chose option 3 for the fix, as I would like to avoid multiple invokes.
I added your test-case to the pr.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
LeoRaasch: I understand your concern here. At the same time, I see that the main branch currently omits the NULL-check if the field is not nullable. For SIMD-JSON, this is not an issue on main as SIMD-JSON will throw an error if we try to get a NULL value as numeric / bool / string. However, for CSV, this actually causes some interesting out-of-bounds errors regarding storing an empty string. Not exactly easy to debug either.

I was of the opinion that we trust that we do not receive NULL from the source in case NOT NULL is set for a field. If this should still be a concern, I suggest to extend your fix to the checkIsNullProxy usages of the ValueDeserializers.

LeoRaasch: The main reason why I want to split Nullable and Non-Nullable Value Deserializers into different classes is that I want to optimize the formatting process based on assumptions we can take about the data. For Non-Nullable Value Deserializers, this means omitting the null-checks. If the user is not sure, if the input data will contain NULL values, they can always choose to not declare fields as NOT NULL (which is the "Default" anyway)

nes-input-formatters/registry/include/ValueDeserializerRegistry.hpp
LeoRaasch: Fixed this

nes-input-formatters/include/InputFormatters/InputFormatIndexer.hpp
LeoRaasch: Fixed this

nes-input-formatters/src/InputParsers/DefaultInputParsers.cpp
LeoRaasch: Removed the noexcept

LeoRaasch:
> Sorry for the late review, I have to put my veto against this — especially against the SQL API changes.
> 
> I don't know of any other system that has a knob like INPUT_PARSERS/OUTPUT_PARSERS per data type. I don't see how this could be superior to just using the default parser implementation and allowing an escape hatch that accepts a string and calls a function on it. A good optimizer could push that into the RawScan.
> 
> I don't mind reorganizing the parsing functions internally to avoid the duplications — that part is a genuine improvement. The inline comments cover the correctness bugs and internal-structure issues that survive that reorg. Everything tied to the user-facing INPUT_PARSERS/OUTPUT_PARSERS config surface I've deliberately left off, since I'd rather see that whole surface removed.
> 
> This PR would also be a greate place to get rid of this output praser naming psychosis and finally call the formatter or serializer or anything except parser...
> 
> having functions like this doing the actual formatting is super unintuative:
> 
> ```c++
> T parseNumeric( 
>     const T value,
>     int8_t* bufferStartingAddress,
>     const uint64_t remainingSpace,
>     TupleBuffer* tupleBuffer,
>     AbstractBufferProvider* bufferProvider)
> ```

Regarding the "naming psychosis": I changed the name to ValueSerializers and ValueDeserializers now, which I hope captures the functionality of the component better.

Regarding the knob to control the (de)serializer implementations for certain types: I think there are multiple cases where this knob could be useful:

- A single DataType might have multiple input / output formats: What comes to mind here are for example timestamps (ISO, SQL etc. representations). The alternative here is to create one DataType per format, which would introduce a lot of redundancy, as the internal representation of the timestamp will be the same for all formats. With the knob, a user could simply specify in which format they would like their timestamps to be serialized in the sink / in which format the timestamps will be arriving.
- Some data sources may represent values of a DataType in a way that clashes with our current default (de)serializers (@alepping already encountered this). In this case, I personally think that presenting the option to create a new (de)serializer for this DataType as plugin and configure it as (de)serializer in the source/sink definition is more intuitive than the escape hatch you mentioned.
- Experimenting with alternative, external parsers / serializers, just to compare the performance to our defaults. In this case, the escape hatch would probably suffice, but it might be nice to have such (de)serializers as plugin. For my thesis and for @alepping's Piparse paper, we already used this approach to include plugins for [fast_float](https://github.com/fastfloat/fast_float) and [Zmij](https://github.com/vitaut/zmij) (see the benchmark-piparse... branches).

These use-cases are partially related to the extensible-datatype issue (I will implement a timestamp datatype for the DataType POC with several possible input/output formats, using the (de)serializer registry). Maybe we can merge the reorganization of the (de)serializer functions first, and withhold the knob to configure them individually until extensible-datatypes has been merged.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
ls-1801: Okay i think i'll trust your and @alepping opinion on that one. 

I personally wouldn't favor performance over correctness, even if it is the user who screwed up. But i can see the point, that if a user does not want to pay for the null check, then they should be able to opt out.

But than i can also see why sql defaults to nullable.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
```diff
@@ -50,8 +50,96 @@
...
+        if (checkIsNullJsonProxy(fieldIndex, simdJsonRawBufferIndex, simdIndexer))
+        {
+            return &result;
+        }
+    }
+
+    const auto fieldNameStr = simdIndexer->getFieldNameInJsonAt(fieldIndex).asCanonicalString();
+    const std::string_view fieldName = fieldNameStr;
+    auto currentDoc = *simdJsonRawBufferIndex->getDocStreamIterator();
+    std::string_view rawValue = accessSIMDJsonFieldOrThrow(currentDoc, fieldName).raw_json().value();
```
ls-1801: This is actually a regression in our JSON parsing capabilities. Previously, we would use `get_string` for string values, because simdjson would do the JSON string decoding; this systest was passing before the change:

```
# name: JSONEscapedString.test
# description: JSON input must decode escapes before producing string values
# groups: [Formatter, JSON]

CREATE LOGICAL SOURCE escapedJson(value VARSIZED NOT NULL, letter CHAR NOT NULL);
CREATE PHYSICAL SOURCE FOR escapedJson TYPE File SET('JSON' AS INPUT_FORMATTER."TYPE");
ATTACH INLINE
{"VALUE":"A\u0042C", "LETTER":"\u0078"}

SELECT value, letter FROM escapedJson INTO File();
----
ABC,x
```

The encoding is part of the JSON spec and comes in quite handy for representing new lines "\n" is decoded into a new line character

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
```diff
@@ -50,8 +50,96 @@
...
+    const auto* simdIndexer = static_cast<const SIMDJSONInputFormatIndexer*>(indexer);
+    PRECONDITION(
+        fieldIndex < simdIndexer->getNumberOfFields(),
+        "fieldIndex {} is out of bounds for schema keys of size: {}",
+        fieldIndex,
+        simdIndexer->getNumberOfFields());
+
+    /// We use the thread local to return multiple values.
+    /// C++ guarantees that the returned address is valid throughout the lifetime of this thread.
+    thread_local static RawJsonAccessResult result;
```
ls-1801: You should avoid abusing the thread_local for return values. We now have better tools for that. The thread_local prevents the future inlining changes.

```c++
nautilus::val<RawValueAccessResult> fieldAccessResult;
nautilus::invoke(
    getValueFromIndex<Nullable>,
    fieldIndex,
    rawBufferIndex,
    nautilus::val<const InputFormatIndexer*>(&indexer),
    &fieldAccessResult);
const nautilus::val<const int8_t*> address = fieldAccessResult.get(&RawValueAccessResult::ptrToValue);
const nautilus::val<uint64_t> size = fieldAccessResult.get(&RawValueAccessResult::sizeOfValue);

template <bool Nullable>
void getValueFromIndex<Nullable>(..., RawValueAccessResult* result) {
...
result.ptrToValue = ...;
result.sizeOfValue = ...;
}
```

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONRawBufferIndex.cpp
```diff
@@ -155,8 +175,34 @@ Record SIMDJSONRawBufferIndex::readSpanningRecord(
 
         auto fieldIndex = static_cast<nautilus::val<FieldIndex>>(i);
         const auto fieldDataType = bufferRef.getAllDataTypes().at(i);
-        writeValueToRecord(
-            fieldDataType, record, fieldName, fieldIndex, rawBufferIndex, nautilus::val<const InputFormatIndexer*>(&indexer));
+
+        /// Retrieve the address and size of the raw field value
+        /// Workaround to pass nullable into the template. Will be resolved during tracetime as nullable will always be known before compiling the query.
+        const auto fieldAccessResult = fieldDataType.nullable ? nautilus::invoke(
+                                                                    {.modRefInfo = nautilus::ModRefInfo::Ref},
```
ls-1801: You should not use this Attribute! The function writes to a thread_local variable; I guess we are lucky the optimizer didn't optimize away the access afterward, but it could have.

Even with the proposed change using the new nautilus::val<Struct>, we are writing to it, so you just cannot use this attribute.

nes-input-formatters/src/ValueDeserializerUtil.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+        const std::string typeName = overrides.substr(typeNameStart, typeNameEnd - typeNameStart);
+        const size_t deserializerTypeStart = typeNameEnd + 1;
+        const size_t deserializerTypeEnd = std::min(overrides.size(), overrides.find(',', deserializerTypeStart));
+        const std::string deserializerType = overrides.substr(deserializerTypeStart, deserializerTypeEnd - deserializerTypeStart);
+
+        if (std::optional<DataType::Type> dataType = magic_enum::enum_cast<DataType::Type>(typeName))
+        {
+            deserializersMap[dataType.value()] = deserializerType;
+        }
```
ls-1801: I don't think we should simply ignore those. Imagine someone writing UIN64:specialUint64Parser, only to realize that there was a typo and the parser was never applied.

ls-1801 (changes requested):
I’m trying to understand why serializer/deserializer selection is attached to a DataType::Type rather than to an
individual field.

The current map means that every field of the same datatype in a physical source must use the same deserializer. For
example, configuring:

UINT64:UnixTimestampDeserializer

applies that deserializer to every UINT64 field. It is impossible to select different parsing rules for two UINT64
fields because the field identity is not part of the lookup.

My mental model is closer to schema annotations: the underlying type describes the
in-memory representation, while field-level metadata describes how that particular field is represented externally.

For example, a schema could contain two timestamps that are both stored internally as epoch milliseconds:

```sql
CREATE
LOGICAL SOURCE
events(                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
event_time UINT64 NOT
NULL,                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
ingest_time UINT64 NOT
NULL                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
);
```

But the incoming values might use different representations:

```json
{
  "event_time": "2026-08-11T13:42:00Z",
  "ingest_time": 1786455720000
}
```

Conceptually, the field configuration would be:

```
event_time →
RFC3339ToEpochMilliseconds                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
ingest_time → UnixMilliseconds
```

Both fields produce the same internal UINT64 type, but require different deserialization. A simpler example would be one
timestamp arriving in Unix seconds and another in Unix milliseconds; one deserializer must scale the value while the
other must not.

This makes me expect an API along the lines of:

getDeserializerType(fieldName, dataType)

where the datatype provides a default and a field may override it.

Is there a use case or design constraint that requires the serializer to be selected exclusively by datatype? If the
intention is deliberately to configure one conversion policy per datatype for the entire physical source, I think that
limitation should be made explicit. Otherwise, field-level configuration seems more consistent with how serialization
metadata is normally associated with schemas.

LeoRaasch:
> I’m trying to understand why serializer/deserializer selection is attached to a DataType::Type rather than to an individual field.
> 
> The current map means that every field of the same datatype in a physical source must use the same deserializer. For example, configuring:
> 
> UINT64:UnixTimestampDeserializer
> 
> applies that deserializer to every UINT64 field. It is impossible to select different parsing rules for two UINT64 fields because the field identity is not part of the lookup.
> 
> My mental model is closer to schema annotations: the underlying type describes the in-memory representation, while field-level metadata describes how that particular field is represented externally.
> 
> For example, a schema could contain two timestamps that are both stored internally as epoch milliseconds:
> 
> ```sql
> CREATE
> LOGICAL SOURCE
> events(                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
> event_time UINT64 NOT
> NULL,                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
> ingest_time UINT64 NOT
> NULL                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
> );
> ```
> 
> But the incoming values might use different representations:
> 
> ```json
> {
>   "event_time": "2026-08-11T13:42:00Z",
>   "ingest_time": 1786455720000
> }
> ```
> 
> Conceptually, the field configuration would be:
> 
> ```
> event_time →
> RFC3339ToEpochMilliseconds                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     
> ingest_time → UnixMilliseconds
> ```
> 
> Both fields produce the same internal UINT64 type, but require different deserialization. A simpler example would be one timestamp arriving in Unix seconds and another in Unix milliseconds; one deserializer must scale the value while the other must not.
> 
> This makes me expect an API along the lines of:
> 
> getDeserializerType(fieldName, dataType)
> 
> where the datatype provides a default and a field may override it.
> 
> Is there a use case or design constraint that requires the serializer to be selected exclusively by datatype? If the intention is deliberately to configure one conversion policy per datatype for the entire physical source, I think that limitation should be made explicit. Otherwise, field-level configuration seems more consistent with how serialization metadata is normally associated with schemas.

I actually have field specific (de)serialization ultimately in mind. I guess the reason why I reduced it to datatype-specific (de)serialization for this PR was merely for simplicity reasons. But the longer I think about it, the more it seems like a rather silly decision, so I guess I will rework the PR to cover field-specific serialization like you described.

Perhaps we could change the SQL API so the specification of the desired input / output format (if diverging from the default) is set during the logical source / sink definition by appending the key of the (de)serializer to the field definition like:

```sql
CREATE
LOGICAL SOURCE
events(                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
event_time UINT64 NOT
NULL RFC3339,                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
ingest_time UINT64 NOT
NULL UnixMilliseconds                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
);
```

ls-1801 (changes requested):
Just one small thing I have noticed, after working with our new identifiers.

The comma/colon mini-language is not quote-aware. Legal quoted identifiers such as "sensor,value" and "event:time" cannot be configured because splitting occurs before identifier parsing. Please use a quote-aware parser or a structured field-to-key representation.

```sql
CREATE LOGICAL SOURCE measurements(
    "sensor,value" UINT64 NOT NULL
);

CREATE PHYSICAL SOURCE FOR measurements TYPE File SET(
    '"sensor,value":DefaultUINT64'
        AS "INPUT_FORMATTER".VALUE_DESERIALIZERS
);

ATTACH INLINE
42

SELECT "sensor,value" FROM measurements INTO File();
```

LeoRaasch:
> Just one small thing I have noticed, after working with our new identifiers.
> 
> The comma/colon mini-language is not quote-aware. Legal quoted identifiers such as "sensor,value" and "event:time" cannot be configured because splitting occurs before identifier parsing. Please use a quote-aware parser or a structured field-to-key representation.
> 
> ```sql
> CREATE LOGICAL SOURCE measurements(
>     "sensor,value" UINT64 NOT NULL
> );
> 
> CREATE PHYSICAL SOURCE FOR measurements TYPE File SET(
>     '"sensor,value":DefaultUINT64'
>         AS "INPUT_FORMATTER".VALUE_DESERIALIZERS
> );
> 
> ATTACH INLINE
> 42
> 
> SELECT "sensor,value" FROM measurements INTO File();
> ```

Interestingly, field names containing ',' cause issues within the SystestResultCheck, additionally to what you mentioned here: In `parseFieldNames` (SystestResultCheck.cpp, line 284), the header line of the result file, which is produced by `SchemaFormatter::getFormattedSchema()`, is split on ',' to obtain the fieldname-type pairs. The problem is that `getFormattedSchema` removes the quotes of quoted identifiers. Therefore, the parse in SystestResultCheck has currently no way to differentiate between ',' in identifiers and the delimiting ',' and triggers an invariant. 

A similar issue probably occurs when including ':' in field identifiers, which is also used to seperate the field identifier, field type, and is null in the output header. 

I would suggest keeping the quotes of quoted identifiers in the CSV headers. If this is fine, I will gladly open and fix this issue.

LeoRaasch:
> > Just one small thing I have noticed, after working with our new identifiers.
> > The comma/colon mini-language is not quote-aware. Legal quoted identifiers such as "sensor,value" and "event:time" cannot be configured because splitting occurs before identifier parsing. Please use a quote-aware parser or a structured field-to-key representation.
> > ```sql
> > CREATE LOGICAL SOURCE measurements(
> >     "sensor,value" UINT64 NOT NULL
> > );
> > 
> > CREATE PHYSICAL SOURCE FOR measurements TYPE File SET(
> >     '"sensor,value":DefaultUINT64'
> >         AS "INPUT_FORMATTER".VALUE_DESERIALIZERS
> > );
> > 
> > ATTACH INLINE
> > 42
> > 
> > SELECT "sensor,value" FROM measurements INTO File();
> > ```
> 
> Interestingly, field names containing ',' cause issues within the SystestResultCheck, additionally to what you mentioned here: In `parseFieldNames` (SystestResultCheck.cpp, line 284), the header line of the result file, which is produced by `SchemaFormatter::getFormattedSchema()`, is split on ',' to obtain the fieldname-type pairs. The problem is that `getFormattedSchema` writes the canonical identifiers, which are stripped of their quotes. Therefore, the parse in SystestResultCheck has currently no way to differentiate between ',' in identifiers and the delimiting ',' and triggers an invariant.
> 
> A similar issue probably occurs when including ':' in field identifiers, which is also used to seperate the field identifier, field type, and is null in the output header.
> 
> I would suggest keeping the quotes of quoted identifiers in the CSV headers. If this is fine, I will gladly open and fix this issue.

I opened a PR fixing this (#1983). Once this is merged, I can continue fixing your original issue here.

LeoRaasch:
> Just one small thing I have noticed, after working with our new identifiers.
> 
> The comma/colon mini-language is not quote-aware. Legal quoted identifiers such as "sensor,value" and "event:time" cannot be configured because splitting occurs before identifier parsing. Please use a quote-aware parser or a structured field-to-key representation.
> 
> ```sql
> CREATE LOGICAL SOURCE measurements(
>     "sensor,value" UINT64 NOT NULL
> );
> 
> CREATE PHYSICAL SOURCE FOR measurements TYPE File SET(
>     '"sensor,value":DefaultUINT64'
>         AS "INPUT_FORMATTER".VALUE_DESERIALIZERS
> );
> 
> ATTACH INLINE
> 42
> 
> SELECT "sensor,value" FROM measurements INTO File();
> ```

Fixed this for VALUE_DESERIALIZERS and VALUE_SERIALIZERS by parsing quote-aware. I added test-cases for this scenario to `ValueDeserializerOverrides.test` and `ValueSerializerOverrides.test`.

ls-1801 (approved):

alepping (approved):

PR #1643, opened 2026-05-29
feat(ci) build nightly created even if nightly test matrix fails.

greenBene:

We want to have an up-to-date and ready-to-use dockerized version of NebulaStream readily available. 
Thus, we want to release nightly builds. 

This was already enabled in the nightly GitHub workflows, but only triggered if the nightly testing matrix succeeds, which it often doesn't. Since all commits in main are tested before being pushed, and the nightly tests are a bit flaky, we don't require this dependency. 

Eventually, we want to have additionally stable releases that are heavily tested, but that is out of scope for this PR.

## Verifying this change
This change is tested by

* Validating that the images are built and available via docker-hub

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -140,13 +140,12 @@ jobs:
       run_reason: "Nightly Benchmark"
 
   create-nightly-image:
-    needs: [ build-and-test ]
-    if: ${{ !github.event.act && inputs.ref == 'main' }}
+    if: ${{ !github.event.act }}
     uses: ./.github/workflows/create_release_image.yml
     secrets: inherit
     with:
       head_sha: "main"
-      dockerhub-tag: "nightly"
+      dockerhub-tag: "latest"
 
   send-zulip-msg-on-failure:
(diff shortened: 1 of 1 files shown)


ls-1801 (approved):

.github/workflows/nightly.yml
```diff
@@ -17,152 +17,151 @@ on:
...
-        email: "pr-reminder-bot@nebulastream.zulipchat.com"
-        organization-url: "https://nebulastream.zulipchat.com"
-        to: "ci"
-        topic: "nightly ci status"
-        type: "stream"
-        content: |
-          @**ls-1801** @**Nils** [Nightly run ${{ github.run_id }}](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}) failed.
-          If there are novel faults, please open corresponding issues/topics and notify potential owners/investigators.
-          If applicable, please communicate workarounds and/or expected time-to-fix.
+#  send-zulip-msg-on-failure:
```
keyseven123: Can we then remove this dead code? Or are you already working on a branch that reintroduces it?

keyseven123 (changes requested):

keyseven123 (approved):

PR #1645, opened 2026-05-29
Feature/bloom filter

keyseven123:
  This pull request adds an optional BloomFilter to our chained hashmap. This helps not having to traverse potentially long chains.
Additionally, we provide a [nautilus fix of right shift](https://github.com/nebulastream/nautilus/pull/291). This fix was found by @LucaCode37

On my local machine, I saw speedups of up to 40x for the Hashjoin assuming a best case scenario for the bloom filter of one long chain.
  
  ## Verifying this change
  This change is tested by running our existing tests.
  
  - Dependencies — adds `rapidcheck` (vcpkg + nix flake)
  - ExecutionEngine — stream join `Build` / `Probe` operators, slice store
  - QueryCompiler — `LowerToPhysicalHashJoin` / `LowerToPhysicalNLJoin` + new `LowerToJoinUtil`
  - nes-nautilus (new `BloomFilterRef`)
  
  ## Documentation

### DIFF
--- a/nes-nautilus/include/Interface/Hash/BloomFilterRef.hpp
+++ b/nes-nautilus/include/Interface/Hash/BloomFilterRef.hpp
@@ -1,0 +1,83 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+#include <Interface/Hash/HashFunction.hpp>
+#include <val_bool.hpp>
+#include <val_ptr.hpp>
+
+namespace NES::Nautilus::Interface
+{
+
+/// BloomFilter sizing passed from the query compiler to the physical operators. Every instance is a valid,
+/// enabled sizing; the constructor is the single boundary enforcing that. "No filter" is an empty
+/// `std::optional` at the holder, so there is no disabled state here and no default constructor.
+class BloomFilterParams
+{
+    uint64_t bitCount;
+    uint64_t hashCount;
+
+public:
+    /// Bounds the sizing math (which runs in double) so the cast to uint64_t stays defined and the rounding
+    /// in `allocationByteCount()` cannot wrap. 2^40 bits is 128 GiB, far past what can be allocated.
+    static constexpr uint64_t maxBitCount = 1ULL << 40U;
+    /// A filter needs at least one whole word to index into.
+    static constexpr uint64_t minBitCount = 64;
+    /// `hashCount` unrolls the probe loop at trace time, so it has to stay small. No filter benefits from
+    /// more than a few dozen hash positions anyway.
+    static constexpr uint64_t minHashCount = 1;
+    static constexpr uint64_t maxHashCount = 64;
+
+    /// Optimal sizing from Bose et al., "On the false-positive rate of Bloom filters",
+    /// Information Processing Letters 108(4), 2008. https://doi.org/10.1016/j.ipl.2008.05.018
+    /// Requires a finite `fpRate` in (0, 1); the derived sizing is clamped into the bounds above.
+    [[nodiscard]] BloomFilterParams(std::size_t expectedEntries, double fpRate);
+
+    /// Bytes to allocate for the bit array, sizing the `ChainedHashMap`'s in-map bit area. Rounded up to
+    /// whole 64-bit words because `add`/`mightContain` index by word.
+    [[nodiscard]] uint64_t allocationByteCount() const { return ((bitCount + 63) / 64) * sizeof(uint64_t); }
+
+    [[nodiscard]] uint64_t getBitCount() const { return bitCount; }
+
+    [[nodiscard]] uint64_t getHashCount() const { return hashCount; }
+};
+
+/// Public-facing BloomFilter handle used by physical operators. Every instance is a real filter: "no filter"
+/// is expressed by the holder as an empty `std::optional<BloomFilterRef>`, not by an alternative in here.
+///
+/// The bit area is a runtime address, the sizing is trace-time data: hashCount unrolls the probe loop and
+/// bitCount folds into the modulo. Params therefore have to match the ones the bit area was allocated with.
+///
+/// Uses the Kirsch-Mitzenmacher double-hashing scheme (split hash into lower/upper 32 bits):
+/// A. Kirsch, M. Mitzenmacher, "Less Hashing, Same Performance: Building a Better Bloom Filter",
+/// Random Structures & Algorithms, 2008. https://doi.org/10.1002/rsa.20208
+class BloomFilterRef
+{
+public:
+    BloomFilterRef(const nautilus::val<uint64_t*>& memArea, BloomFilterParams params);
+
+    void add(const HashFunction::HashValue& hash) const;
+
+    /// Checks if a particular hash might be seen by the bloom filter
+    [[nodiscard]] nautilus::val<bool> mightContain(const HashFunction::HashValue& hash) const;
+
+private:
+    nautilus::val<uint64_t*> memArea;
+    BloomFilterParams params;
+};
+
+}
--- a/nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
+++ b/nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
@@ -26,7 +26,6 @@
 #include <Interface/HashMap/HashMap.hpp>
 #include <Runtime/AbstractBufferProvider.hpp>
 #include <Runtime/TupleBuffer.hpp>
-#include <Runtime/VariableSizedAccess.hpp>
 
 namespace NES
 {
@@ -106,8 +105,8 @@ public:
 
     [[nodiscard]] uint64_t getMask() const { return header().mask; }
 
-    [[nodiscard]] VariableSizedAccess::Index getStorageBufferIdx() const;
-    [[nodiscard]] VariableSizedAccess::Index getVarSizedBufferIdx() const;
+    [[nodiscard]] ChildBufferIndex getStorageBufferIdx() const;
+    [[nodiscard]] ChildBufferIndex getVarSizedBufferIdx() const;
     [[nodiscard]] ChainedHashMapEntry* getChain(uint64_t pos);
 
     /// @warning Be super careful with this. Sometimes you need a pointer to the TupleBuffer but you should never alter it outside of this
@@ -141,8 +140,8 @@ private:
         uint64_t entriesPerPage;
         uint64_t numRecords = 0;
         uint64_t mask;
-        VariableSizedAccess::Index storageSpaceIndex;
-        VariableSizedAccess::Index varSizedSpaceIndex;
+        ChildBufferIndex storageSpaceIndex;
+        ChildBufferIndex varSizedSpaceIndex;
 
         /// Chains array starts immediately after this header
         /// it is dynamically sized based on numChains, so nothing to store in here.
--- a/nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
+++ b/nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
@@ -18,10 +18,12 @@
 #include <cstdint>
 #include <functional>
 #include <memory>
+#include <optional>
 #include <span>
 #include <type_traits>
 #include <utility>
 #include <vector>
+#include <Interface/Hash/BloomFilterRef.hpp>
 #include <Interface/Hash/HashFunction.hpp>
 #include <Interface/HashMap/HashMap.hpp>
 #include <Runtime/AbstractBufferProvider.hpp>
@@ -31,6 +33,21 @@ namespace NES
 {
 /// Forward declaration of the ChainedHashMapRef, to avoid cyclic dependencies between ChainedHashMap and ChainedHashMapRef
 class ChainedHashMapRef;
+
+/// Sizing for a ChainedHashMap view, passed to init(). An engaged bloomFilter enables the in-map BloomFilter
+/// and reserves its bit area inline in the map's buffer; std::nullopt allocates nothing for it.
+struct ChainedHashMapConfig
+{
+    uint64_t entrySize;
+    uint64_t numberOfBuckets;
+    uint64_t pageSize;
+    std::optional<Nautilus::Interface::BloomFilterParams> bloomFilter = std::nullopt;
+
+    [[nodiscard]] uint64_t bloomFilterMemAreaSize() const { return bloomFilter ? bloomFilter->allocationByteCount() : 0; }
+
+    /// Bytes the TupleBuffer backing a map with this config must provide.
+    [[nodiscard]] uint64_t bufferSize() const;
+};
 
 /// Each entry contains a ptr to the next element, the hash of the current value and the keys and values.
 /// The physical layout of the storage space is the following
@@ -73,8 +90,9 @@ class ChainedHashMap final : public HashMap
 public:
     /// @brief Use init to initialize a ChainedHashMap view on a pre-allocated TupleBuffer
     /// Constructors are private
-    static void init(TupleBuffer& tupleBuffer, uint64_t entrySize, uint64_t numberOfBuckets, uint64_t pageSize);
-    static void init(TupleBuffer& tupleBuffer, uint64_t keySize, uint64_t valueSize, uint64_t numberOfBuckets, uint64_t pageSize);
+    /// The buffer must be at least config.bufferSize() bytes; entries built from a key/value split size their
+    /// entrySize as sizeof(ChainedHashMapEntry) + keySize + valueSize.
+    static void init(TupleBuffer& tupleBuffer, const ChainedHashMapConfig& config);
 
     /// @brief Loads a ChainedHashMap view from a pre-filled TupleBuffer
     static ChainedHashMap load(const TupleBuffer& tupleBuffer);
@@ -86,8 +104,12 @@ public:
 
     [[nodiscard]] TupleBuffer getPage(uint64_t pageIndex) const;
     [[nodiscard]] TupleBuffer getVarSizedPage(uint64_t pageIndex) const;
-    [[nodiscard]] static uint64_t calculateBufferSizeFromBuckets(uint64_t numberOfBuckets);
-    [[nodiscard]] static uint64_t calculateBufferSizeFromChains(uint64_t numberOfChains);
+
+    /// Size of the buffer a ChainedHashMap view needs: header, chains array and the in-map BloomFilter bit
+    /// area. The bloom size is a parameter rather than an afterthought so no caller can allocate a buffer
+    /// that init() then overruns.
+    [[nodiscard]] static uint64_t calculateBufferSizeFromBuckets(uint64_t numberOfBuckets, uint64_t bloomFilterMemAreaSize);
+    [[nodiscard]] static uint64_t calculateBufferSizeFromChains(uint64_t numberOfChains, uint64_t bloomFilterMemAreaSize);
     [[nodiscard]] uint64_t getNumberOfPages() const;
     [[nodiscard]] uint64_t getNumberOfVarSizedPages() const;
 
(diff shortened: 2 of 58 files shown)


nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
```diff
@@ -0,0 +1,484 @@
...
+    if (expectedEntries == 0)
+    {
+        return {.bitCount = 64, .hashCount = 1};
+    }
+    const auto n = static_cast<double>(expectedEntries);
+    const double ln2 = std::numbers::ln2;
+    const auto m = std::max(static_cast<uint64_t>(std::ceil(-n * std::log(fpRate) / (ln2 * ln2))), uint64_t{64});
+    const auto k = std::max(static_cast<uint64_t>(std::round(static_cast<double>(m) / n * ln2)), uint64_t{1});
+    return {.bitCount = m, .hashCount = k};
+}
```
ls-1801: why is this function here? it should be part of the bloom filter.

ls-1801 (changes requested):
I think that, in the current state, adding this level of complexity does not outweigh the benefits for me. Especially with long-awaited changes from @ThanGeo on the horizon, I would really like to postpone this.

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
keyseven123: Done

keyseven123:
> I think that, in the current state, adding this level of complexity does not outweigh the benefits for me. Especially with long-awaited changes from @ThanGeo on the horizon, I would really like to postpone this.

I have rebased and stacked this pr on top of the chained hash map refactoring pr

ls-1801 (changes requested):
re-request once relevant

keyseven123:
> re-request once relevant

I have rebased this pr/branch on the new chained hash map. Please let me know what you guys think

nes-nautilus/src/Interface/Hash/BloomFilterRef.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+    return BloomFilterRef{NoOp{}};
+}
+
+BloomFilterRef BloomFilterRef::createDisabled()
+{
+    return BloomFilterRef{NoOp{}};
+}
+
+BloomFilterRef::Impl::Impl(const BloomFilterParams params, std::unique_ptr<HashFunction> hashFunction)
```
ThanGeo: Is this Impl::Impl really the best way? Kinda reduces readability

nes-nautilus/include/Interface/Hash/BloomFilterRef.hpp
```diff
@@ -0,0 +1,125 @@
...
+    static BloomFilterRef createDisabled();
+
+    [[nodiscard]] const BloomFilterParams& params() const { return bloomFilterParams; }
+
+    void add(const nautilus::val<uint64_t*>& bitsData, const HashFunction::HashValue& hash) const
+    {
+        std::visit([&](const auto& bloomFilter) { bloomFilter.add(bitsData, hash); }, impl);
+    }
+
+    [[nodiscard]] nautilus::val<bool> mightContain(const nautilus::val<uint64_t*>& bitsData, const HashFunction::HashValue& hash) const
```
ThanGeo: what does "mightContain" mean?

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
```diff
@@ -0,0 +1,404 @@
...
+        return 0.0;
+    }
+    const auto exponent = -static_cast<double>(numHashes) * static_cast<double>(numEntries) / static_cast<double>(numBits);
+    return std::pow(1.0 - std::exp(exponent), static_cast<double>(numHashes));
+}
+
+/// Wraps the Nautilus-compiled add/mightContain pair. Engine setup and function registration happen
+/// once per instance so the hot loop is just function-call overhead. The `enabled` flag picks
+/// between the real BloomFilter variant and the no-op variant
+class TestableBloomFilter
```
ThanGeo: Following the changes to TestableChainedHashMap and TestablePagedVector being moved to their own class files, maybe consider doing the same here.

ThanGeo (changes requested):
Overall minor changes requested. One question, are you using a single hash function in the bloom filter? By definition usually $K$ hash functions are used even though with added overhead but better correctness

nes-nautilus/include/Interface/Hash/BloomFilterRef.hpp
keyseven123: If a hash/key might be seen by the bloom filter.

nes-nautilus/src/Interface/Hash/BloomFilterRef.cpp
keyseven123: Renamed it to `Active/Inactive`.

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
keyseven123: Done

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -346,4 +376,31 @@ TEST(ChainedHashMapIteratorTest, emptyMapIsAnEmptyRange)
...
+    putAndLookupKeysProperty(TestUtils::EngineMode::Interpreter, true);
+}
+
+RC_GTEST_PROP(ChainedHashMapPropertyTest, putAndGetAllWithBloomFilterCompiler, ())
+{
+    Logger::setupLogging("ChainedHashMapPropertyTest.log", LogLevel::LOG_DEBUG);
+    putAndGetAllProperty(TestUtils::EngineMode::Compiler, true);
+}
+
+RC_GTEST_PROP(ChainedHashMapPropertyTest, putAndGetAllWithBloomFilterInterpreter, ())
```
ThanGeo, line 504: Maybe add one more property test that check correctness: e.g. 2 chained hash maps, one with bloom filter and one with out, results should always match. Also maybe tests with very high/very low collision on the bloom filter? I',m trying to think of tests that will reveal instantly if any future changes break any edge cases or something

ThanGeo (changes requested):

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
keyseven123: Good idea. Added tests for it.

.github/workflows/build_and_test.yml
```diff
@@ -75,7 +75,7 @@ jobs:
             echo "=== sccache config before build ==="
             sccache --start-server
             sccache --show-stats
-            cmake -GNinja -B build -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} ${{ matrix.cmake_flags }}
+            cmake -GNinja -B build -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DNES_EXHAUSTIVE_TESTS=1 ${{ matrix.cmake_flags }}
```
ls-1801: The whole point was to not enable exhaustive tests by default; we can run them in the nightly.
But i would manage that via the matrix.yaml

nes-nautilus/include/Interface/Hash/BloomFilterRef.hpp
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Zeroes `dst`. Does nothing when the filter is Inactive.
+    void clearBits(const nautilus::val<std::byte*>& dst) const
+    {
+        std::visit([&](const auto& bloomFilter) { bloomFilter.clearBits(dst); }, filter);
+    }
+
+private:
+    /// Real BloomFilter implementation. Method bodies live in BloomFilterRef.cpp.
+    class Active
```
ls-1801: In general, I like the idea of being able to switch off the Bloom filter. But I think this should be a decision made by the Bloom filter owner. So the chained hash map can be configured to use a BF or not, and the implementation in the chained hash map should be conditional. 

I think it's kinda odd to have a Bloom filter that does nothing. I would be explicit and have a data structure that doesn't use a Bloom filter.

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
```diff
@@ -73,8 +73,16 @@ class ChainedHashMap final : public HashMap
...
+    /// @param bloomFilterMemAreaSize bytes for the optional in-map BloomFilter bit area, 0 disables it (see getBloomFilterMemArea)
+    static void
+    init(TupleBuffer& tupleBuffer, uint64_t entrySize, uint64_t numberOfBuckets, uint64_t pageSize, uint64_t bloomFilterMemAreaSize);
+    static void init(
+        TupleBuffer& tupleBuffer,
+        uint64_t keySize,
+        uint64_t valueSize,
+        uint64_t numberOfBuckets,
+        uint64_t pageSize,
+        uint64_t bloomFilterMemAreaSize);
```
ls-1801: At this point, I would use a ChainedHashMapConfig object, which can contain a std::optional<BloomFilterConfig>

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
```diff
@@ -105,10 +113,19 @@ class ChainedHashMap final : public HashMap
 
     [[nodiscard]] uint64_t getMask() const { return header().mask; }
 
+    /// Size in bytes of the optional in-map BloomFilter bit area (0 when disabled). Used to propagate the
+    /// filter sizing when a derived map is created from an existing one (e.g. the aggregation final map).
+    [[nodiscard]] uint64_t getBloomFilterMemAreaSize() const { return header().bloomFilterMemAreaSize; }
+
```
ls-1801: do we need to expose that?

nes-nautilus/src/Interface/Hash/BloomFilterRef.cpp
```diff
@@ -0,0 +1,127 @@
...
+    /// Each iteration computes a derived hash position via h_i(x) = (lower32 + i * upper32) mod bitCount.
+    /// The position is split into a word index (which uint64_t in the array) and a bit offset within that word.
+    /// Iterating hashCount times sets multiple independent bits, reducing the false-positive rate.
+    for (nautilus::val<uint64_t> i{0}; i < params.hashCount; i = i + nautilus::val<uint64_t>{1})
+    {
+        nautilus::val<uint64_t> x = (lower32 + i * upper32) % params.bitCount;
+        const nautilus::val<uint64_t> wordIndex = x / nautilus::val<uint64_t>{64ULL};
+        nautilus::val<uint64_t> bitOffset = x % nautilus::val<uint64_t>{64ULL};
+        const nautilus::val<uint64_t> mask = nautilus::val<uint64_t>{1ULL} << bitOffset;
+        *(bitsMemoryRef + wordIndex) = *(bitsMemoryRef + wordIndex) | mask;
```
ls-1801: I think the math for the setup is off. You divide by 8, which means for 65 bits we would use 9 bytes. However, here you would access 16 bytes. I think all your tests pad to 64 bits, so this wouldn't happen, but as soon as someone doesn't give a multiple of 64, this could crash.

nes-nautilus/src/Interface/Hash/BloomFilterRef.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+void BloomFilterRef::Active::add(const nautilus::val<uint64_t*>& bitsMemoryRef, const HashFunction::HashValue& hash) const
+{
+    nautilus::val<uint64_t> lower32 = hash & nautilus::val<uint64_t>{(1ULL << 32) - 1};
+    nautilus::val<uint64_t> upper32 = hash >> nautilus::val<uint64_t>{32ULL};
+
+    /// Each iteration computes a derived hash position via h_i(x) = (lower32 + i * upper32) mod bitCount.
+    /// The position is split into a word index (which uint64_t in the array) and a bit offset within that word.
+    /// Iterating hashCount times sets multiple independent bits, reducing the false-positive rate.
+    for (nautilus::val<uint64_t> i{0}; i < params.hashCount; i = i + nautilus::val<uint64_t>{1})
```
ls-1801: is there a reason this is not a static_val?

ThanGeo (approved):

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
```diff
@@ -0,0 +1,339 @@
...
+/// input - including numKeys = 0. Covers the old `edgeCaseZeroExpectedEntries` and `gettersReturnReasonableValues`.
+void sizingProperty()
+{
+    const auto numKeys = *rc::gen::inRange<uint64_t>(0, MAX_NUM_KEYS_SIZING + 1);
+    const auto fpRate = *rc::gen::elementOf(FP_RATES);
+    const auto params = BloomFilterParams::compute(numKeys, fpRate);
+    RC_ASSERT(params.bitCount >= 64);
+    RC_ASSERT(params.hashCount >= 1);
+    RC_ASSERT(params.allocationByteCount() >= 8);
+}
```
ls-1801, line 288: i think this is kinda pointless to test here. Especially since we don't really look at interesting large integer values where overflow could become an issue.

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
```diff
@@ -0,0 +1,339 @@
...
+namespace NES::Nautilus::Interface
+{
+namespace
+{
+
+using TestUtils::EngineMode;
+using TestUtils::TestableBloomFilter;
+
+/// Configured FP rates drawn per property.
+constexpr std::array<double, 6> FP_RATES = {0.0001, 0.001, 0.01, 0.05, 0.1, 0.2};
```
ls-1801: why limit our self to these? couldnt we just pass a range from 0.0001 to 0.2?

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
```diff
@@ -0,0 +1,339 @@
...
+/// reported present — and the measured false-positive rate stays within the theoretical bound for the
+/// actual saturation. Single-threaded: one filter, one bit array.
+void insertAndCheckProperty(EngineMode mode)
+{
+    const auto numKeysForSizing = *rc::gen::inRange<uint64_t>(MIN_NUM_KEYS_SIZING, MAX_NUM_KEYS_SIZING + 1);
+    const auto fpRate = *rc::gen::elementOf(FP_RATES);
+    const auto numUnique = *rc::gen::inRange<uint64_t>(MIN_NUM_UNIQUE, MAX_NUM_UNIQUE + 1);
+    const auto totalMultiplier = *rc::gen::inRange<uint64_t>(MIN_TOTAL_MULTIPLIER, MAX_TOTAL_MULTIPLIER + 1);
+    const auto numTotal = numUnique * totalMultiplier;
+    const auto seed = *rc::gen::arbitrary<uint64_t>();
```
ls-1801: You probably want to wrap this in a `rc::gen::noShrink()`

rc would attempt to find a minimal reproducible test case, however, shrinking the seed does not have any useful benefit

nes-physical-operators/include/HashMapOptions.hpp
```diff
@@ -42,7 +43,8 @@ struct HashMapOptions
         const uint64_t keySize,
         const uint64_t valueSize,
         const uint64_t pageSize,
-        const uint64_t numberOfBuckets)
+        const uint64_t numberOfBuckets,
+        Nautilus::Interface::BloomFilterRef bloomFilter)
```
ls-1801: Passing a bloom filter Ref in the HashMapOptions seems kinda odd. Probably saner to have the data structure own the BloomFilterRef and just construct it in the load function

ls-1801 (changes requested):

.github/workflows/build_and_test.yml
keyseven123: Yes, sorry. I forgot to drop the commit before re-requesting feedback

nes-nautilus/src/Interface/Hash/BloomFilterRef.cpp
keyseven123: Only query compilation time, but I think we can do this at a later stage, if the compilation time becomes a problem

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
keyseven123: Unfortuantely, it is not possible to pass a double to an `inRange()`. I have added a workaround. Let me know what you think about it.

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
keyseven123: Done

nes-nautilus/src/Interface/Hash/BloomFilterRef.cpp
keyseven123: Yes sir.

nes-nautilus/include/Interface/Hash/BloomFilterRef.hpp
keyseven123: ChainedHashMap has now an optional

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
keyseven123: Added a `ChainedHashMapConfig` object

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
keyseven123: Do not need this anymore

nes-physical-operators/include/HashMapOptions.hpp
keyseven123: It is now an optional BloomFilterParams.

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
keyseven123: Changed.

nes-nautilus/include/Interface/Hash/BloomFilterRef.hpp
```diff
@@ -0,0 +1,74 @@
...
+    /// size the `ChainedHashMap`'s in-map bit area so disabled maps allocate nothing.
+    /// Rounded up to whole 64-bit words: `add`/`mightContain` index by word (`bitCount - 1` may land in the
+    /// last, partially used word), so a byte-exact size would leave that word's tail out of bounds.
+    [[nodiscard]] uint64_t allocationByteCount() const
+    {
+        if (hashCount == 0)
+        {
+            return 0;
+        }
+        return ((bitCount + 63) / 64) * sizeof(uint64_t);
```
ls-1801: Could we validate the Bloom-filter input parameters at their construction boundaries? In practice, a `bitCount` above `UINT64_MAX - 63` is already invalid, but because `BloomFilterParams` is aggregate-constructible it can make this rounding overflow into an undersized allocation; similarly, an enabled filter with `bitCount == 0` later performs modulo by zero. `compute()` should also require a finite `fpRate` strictly between 0 and 1: negative/NaN values can reach a NaN-to-`uint64_t` conversion, while zero can derive an enormous `hashCount` for Nautilus to statically unroll. A precondition or checked factory would make these contracts explicit.

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
```diff
@@ -142,12 +173,23 @@ class ChainedHashMap final : public HashMap
         uint64_t mask;
         ChildBufferIndex storageSpaceIndex;
         ChildBufferIndex varSizedSpaceIndex;
+        /// Sizing of the optional in-map BloomFilter, nullopt when disabled. The bit area itself sits inline
+        /// behind the chains array, so no index into a child buffer is needed.
+        std::optional<Nautilus::Interface::BloomFilterParams> bloomFilter;
```
ls-1801, line 178: Maybe as a follow-up: Do we need to store the Bloom-filter configuration in both the hash-map header and the view? For example, `entrySize` is stored in the header even though it is also passed to `ChainedHashMapRef` when the view is created. The Bloom-filter configuration is now duplicated in the same way. Could the view be the source of these constant values, while the header stores only state that changes at runtime? This may also apply to our other data structures and does not need to block this PR.

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalHashJoin.cpp
```diff
@@ -255,7 +256,9 @@ HashMapOptions createHashMapOptions(
         keySize,
         valueSize,
         pageSize,
-        numberOfBuckets};
+        numberOfBuckets,
+        /// HashJoin does not yet enable the ChainedHashMap's optional filter.
+        std::nullopt};
```
ls-1801: Why is the Bloom filter not enabled for hash joins in general here? We probably want a configuration such as `worker.default_query_execution.bloom_filter_in_hash_join` to turn it off for debugging. To cover both modes without duplicating the complete system-test suite, we could add `worker.default_query_execution.bloom_filter_in_hash_join: [true, false]` to the configuration matrix in one of the existing hash-join system-test files.

ls-1801 (changes requested):

nes-nautilus/include/Interface/Hash/BloomFilterRef.hpp
keyseven123: Done

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalHashJoin.cpp
keyseven123: Because I suck...

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
keyseven123: I have created #1856 and #1858 for this.

nes-nautilus/tests/UnitTests/BloomFilterTest.cpp
```diff
@@ -0,0 +1,337 @@
...
+        RC_ASSERT(actualFpRate <= tolerance);
+    }
+    else
+    {
+        NES_INFO("Skipping FP rate check: only {} true negative samples (need {})", trueNegatives.size(), MIN_FP_SAMPLES_FOR_RATE_CHECK);
+    }
+}
+
+/// Property: zeroing the bit array reverts every previously-inserted key to mightContain=false.
+/// Single-threaded so the clear is uncontested; the multi-threaded race surface is covered by the
```
ls-1801: This says the main property covers the multi-threaded race surface, but `insertAndCheckProperty` is entirely single-threaded (and even documents that at line 159). Please remove this claim or add an actual concurrent test if concurrent access is part of the contract.

ls-1801 (approved):
just one comment fixup

PR #1648, opened 2026-06-01
Run systest in nix ci build

adrianmichalke:


.github/workflows/nix-build.yml
```diff
@@ -79,9 +79,15 @@ jobs:
         run: |
           "$SCCACHE_BIN" --show-stats
           ./.nix/nix-cmake.sh --build cmake-build-nix --target nes-single-node-worker -j"$(nproc)"
+          ./.nix/nix-cmake.sh --build cmake-build-nix --target systest -j"$(nproc)"
           "$SCCACHE_BIN" --show-stats
           "$SCCACHE_BIN" --stop-server
 
+      - name: Run system tests via devShell
```
keyseven123, line 90: TBH, I think that we can run the ctests as well. Compared to the SLT, they do not add that much overhead. But I am also fine solely testing the slts.

keyseven123 (approved):

.github/workflows/nix-build.yml
adrianmichalke: done

ls-1801 (approved):

PR #1651, opened 2026-06-03
feat(systest): allow skipping serialization validation

adrianmichalke:
Currently, systest always tests for correct serialization of queries. For large highly nested predicate queries de/serialization is very expensive and can take multiple seconds. We add the option to skip this test step with `--skip-query-plan-serialization-validation`.


Artraxon (approved):

ls-1801:
I addressed the underlying issue in #1732

PR #1652, opened 2026-06-03
ci(nightly): run SLT corpus

adrianmichalke:
Runs now the systest corpus of ported sqllite test files from https://github.com/nebulastream/Nightly-Systest-Corpus/tree/main during our nightly.
Currently they are ~200.000 test queries.

### DIFF
--- a/.github/workflows/nightly_slt_corpus.yml
+++ b/.github/workflows/nightly_slt_corpus.yml
@@ -1,0 +1,110 @@
+name: Nightly SLT Corpus
+
+on:
+  workflow_call:
+    inputs:
+      dev_image_tag:
+        required: true
+        type: string
+        description: "Docker image tag of the development image"
+      head_sha:
+        required: true
+        type: string
+        description: "head_sha of the branch"
+      corpus_ref:
+        required: false
+        type: string
+        default: main
+        description: "Nightly-Systest-Corpus ref to check out"
+      corpus_path:
+        required: false
+        type: string
+        default: slt
+        description: "Path to the generated systest corpus inside Nightly-Systest-Corpus"
+
+jobs:
+  run-slt-corpus:
+    name: "Run SLT corpus"
+    timeout-minutes: 720
+    runs-on: [ self-hosted, linux, Build, x64 ]
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.head_sha }}
+
+      - uses: actions/checkout@v5
+        with:
+          repository: nebulastream/Nightly-Systest-Corpus
+          ref: ${{ inputs.corpus_ref }}
+          path: nightly-systest-corpus
+
+      - uses: ./.github/steps/setup-sccache
+        with:
+          bucket: ${{ secrets.SCCACHE_CACHE_BUCKET }}
+          endpoint: ${{ secrets.SCCACHE_CACHE_ENDPOINT }}
+          access_key: ${{ secrets.SCCACHE_CACHE_ACCESS_KEY }}
+          secret_key: ${{ secrets.SCCACHE_CACHE_SECRET_KEY }}
+
+      - uses: ./.github/steps/prepare-github
+        name: Preparing Github Runners
+        with:
+          images: |
+            nebulastream/nes-ci:${{ inputs.dev_image_tag }}-libcxx-none
+          docker_username: ${{ secrets.DOCKER_USER_NAME }}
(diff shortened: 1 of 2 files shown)


keyseven123 (commented):
I am not sure what the idea behind this PR is? 
How are the queries generated / decided what queries to generate? 
I would require a little bit more information

ls-1801 (changes requested):
re-request once final, i.e., the change only contains the relevent ci changes. 

And i would really like to see if the systest tool can handle parsing all queries at once instead of running 1000 per iteration.

.github/workflows/nightly_slt_corpus.yml
```diff
@@ -0,0 +1,108 @@
...
+              echo "No .test files found in ${corpus_root}"
+              exit 1
+            fi
+            echo "Running ${test_count} generated SLT files from ${corpus_root} in a single systest invocation"
+            mkdir -p "${result_root}"
+            build/nes-systests/systest/systest \
+              -t "${corpus_root}" \
+              --workingDir "${result_root}" \
+              --clusterConfig nes-systests/configs/topologies/single-node.yaml \
+              -n 100 \
```
keyseven123: Is there a reason behind setting this to 100? Why not 1000 or even 10k?

.github/workflows/nightly_slt_corpus.yml
```diff
@@ -0,0 +1,108 @@
...
+              -- \
+              --worker.default_query_execution.execution_mode=COMPILER
```
keyseven123: I think we can also increase the number of cores to 8 or even 16. But this is not blocking, if you prefer here the default value

keyseven123 (changes requested):
Minor comments

.github/workflows/nightly_slt_corpus.yml
adrianmichalke: addressed, set it to 1000

.github/workflows/nightly_slt_corpus.yml
adrianmichalke: addressed

.github/workflows/nightly_slt_corpus.yml
keyseven123: Sorry to be so nit-picky, but I think there should be a comment specifing why this is set to now 1k. In couple of weeks or months, we do not remember if there was an actual reason behind setting it to 1k.

keyseven123 (changes requested):
Minor stuff.

.github/workflows/nightly_slt_corpus.yml
adrianmichalke: addressed

.github/workflows/nightly_slt_corpus.yml
```diff
@@ -0,0 +1,110 @@
...
+  run-slt-corpus:
+    name: "Run SLT corpus"
+    timeout-minutes: 720
+    runs-on: [ self-hosted, linux, Build, x64 ]
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.head_sha }}
+
+      - uses: actions/checkout@v4
```
ls-1801:
```suggestion
      - uses: actions/checkout@v5
```

.github/workflows/nightly_slt_corpus.yml
```diff
@@ -0,0 +1,110 @@
...
+              -t "${corpus_root}" \
+              --workingDir "${result_root}" \
+              --clusterConfig nes-systests/configs/topologies/single-node.yaml \
+              -n 1000 \
+              -- \
+              --worker.default_query_execution.execution_mode=COMPILER \
+              --worker.query_engine.number_of_worker_threads=16
+
+      - name: Upload SLT corpus logs on failure
+        uses: actions/upload-artifact@v4
```
ls-1801:
```suggestion
        uses: actions/upload-artifact@v5
```

ls-1801 (approved):
Update the actions to the newer versions. and please clean up the git history

.github/workflows/nightly_slt_corpus.yml
adrianmichalke: addressed

.github/workflows/nightly_slt_corpus.yml
adrianmichalke: addressed

keyseven123 (approved):

PR #1653, opened 2026-06-03
ci(nightly): only publish :latest image on main branch

ls-1801:
## Problem

The `NES Nightly` workflow can be triggered via `workflow_dispatch` from any branch (the `head_sha` input even defaults to `main` for testing). However, the `create-nightly-image` job unconditionally pushed the executable images to the `:latest` Docker tag, only excluding `act` runs.

As a result, **anyone triggering the nightly workflow from a non-`main` branch would unintentionally overwrite the published `:latest` release image.**

## Fix

Gate the `create-nightly-image` job on `github.ref == 'refs/heads/main'`. This evaluates to:

- `true` for the scheduled cron run (always runs on the default branch `main`)
- `true` for a manual dispatch from the `main` branch
- `false` for a manual dispatch from any other branch

So `:latest` is only published when the nightly actually runs on `main`. All other nightly jobs (tests, benchmarks, clang-tidy, etc.) are unaffected and still run on whichever branch is dispatched.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -141,7 +141,10 @@ jobs:
       run_reason: "Nightly Benchmark"
 
   create-nightly-image:
-    if: ${{ !github.event.act }}
+    # Only publish the `:latest` release image when the nightly job actually runs on `main`
+    # (scheduled run, or a manual dispatch from the `main` branch). This prevents anyone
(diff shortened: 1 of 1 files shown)


greenBene (approved):
Sorry for introducing this bug, thanks for fixing it!

keyseven123 (approved):

adrianmichalke (approved):

PR #1656, opened 2026-06-05
fix(data-types): make DataType::join symmetric for incompatible types

ls-1801:
## Summary
- `INT32.join(BOOLEAN)` returned `std::nullopt`, but `BOOLEAN.join(INT32)` returned a `DataType{UNDEFINED}` value. Same asymmetry for any numeric vs `CHAR`. Callers checking `has_value()` (e.g. `AddLogicalFunction::withInferredDataType`) silently accepted one direction and threw `DifferentFieldTypeExpected` for the other.
- `CHAR`/`BOOLEAN` branches now return `std::nullopt` when the other operand is a non-matching type, matching the existing numeric-vs-non-numeric branch (and emitting the same `NES_WARNING` for parity).
- Concat is unaffected: it already rejects any join result that is not `VARSIZED`.

## Test plan
- [x] New parameterized regression `JoinIncompatibleTypesTest.BothDirectionsReturnNullopt` in `nes-data-types/tests/UnitTests/API/NumericTypeConversionTest.cpp` covers all incompatible primitive pairs in both directions.
- [x] Existing `NumericTypeConversionTest.SimpleTest` (52 cases) still passes.
- [x] Full debug build passes (`cmake --build cmake-build-debug -j`).
- [x] `clang-tidy-diff` clean on the change.

### DIFF
--- a/nes-data-types/src/DataTypes/DataType.cpp
+++ b/nes-data-types/src/DataTypes/DataType.cpp
@@ -355,7 +355,8 @@ std::optional<DataType> DataType::join(const DataType& otherDataType) const
         {
             return {DataTypeProvider::provideDataType(Type::CHAR, isNullableResult)};
         }
-        return {DataTypeProvider::provideDataType(Type::UNDEFINED, isNullableResult)};
+        NES_WARNING("Cannot join {} and {}", *this, otherDataType);
+        return std::nullopt;
     }
     if (this->type == Type::BOOLEAN)
     {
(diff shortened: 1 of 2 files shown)


nes-data-types/src/DataTypes/DataType.cpp
```diff
@@ -355,15 +355,17 @@ std::optional<DataType> DataType::join(const DataType& otherDataType) const
         {
             return {DataTypeProvider::provideDataType(Type::CHAR, isNullableResult)};
         }
-        return {DataTypeProvider::provideDataType(Type::UNDEFINED, isNullableResult)};
```
Artraxon, line 358: IMO we should not ERROR/WARN Log in functions that have a optional/expected return type, because whether it is an error depends on the call side, and call side should produce an error.

IMO this should be NES_DEBUG.

Artraxon: nvm I just noticed that the function always prints to WARN. Should be changed in my opinion, but doesn't have to be in this PR.

Artraxon (approved):

keyseven123 (approved):

PR #1659, opened 2026-06-08
refactor(input-formatters): drop ad-hoc BufferManager from SequenceShredder

ls-1801:

The `SequenceShredder` seeded the first slot of its spanning-tuple ring buffer with a "dummy" `TupleBuffer` obtained from a **throwaway `BufferManager::create(1, 1)`**. That `BufferManager` stayed alive (via the buffer's recycler holding `shared_from_this`) until the dummy was consumed when the first tuple was read — so it was constructed and then destroyed at a seemingly random later point in time.

The visible symptom was **confusing debug logs**: for every source, a brand-new `BufferManager` would appear to be created and then, seemingly at random, destroyed once the first tuple was read (`BufferManager configuration bufferSize=1 numOfBuffers=1 ...` followed shortly by `Calling BufferManager::destroy()`). This made logs hard to read and suggested phantom allocators springing into and out of existence.

The dummy buffer's contents were in fact **never read**: it only needed to be a claimable, delimiter-bearing sentinel that anchors the first leading spanning tuple while contributing zero bytes. This PR replaces it with a **backing-free start-of-stream sentinel**, so no `BufferManager` is created at all:

- The first ring-buffer entry holds no `TupleBuffer` (`setStateOfFirstIndex()` only sets the atomic state).
- `tryClaimSpanningTuple()` hands back an empty `StagedBuffer` for the sentinel, detected via a **non-atomic null-pointer check** (`operator!`) on the trailing buffer reference (cheaper than reading the atomic reference counter).
- `getTrailingBytes()` returns empty for the (empty-view) sentinel before its delimiter-size bounds check.
- Drops the now-unused `sizeOfTupleDelimiterInBytes` constructor parameter of `SequenceShredder` / `SpanningTupleBuffer`.

No throwaway `BufferManager`, no confusing create/destroy logs.

## Verifying this change

- Input-formatter unit tests pass, including the multi-threaded **exhaustive concurrency test** (`ConcurrentSynchronizationTest`) and the start-of-buffer edge cases in `SpecificSequenceTest` (e.g. `simdJSONFirstObjectEndsAtBufferBoundary`).
- Correctness validated at scale: the **DEBS benchmark** (GB-scale CSV) passes all queries with matching checksums.
- Performance is unchanged: a head-to-head DEBS benchmark (20 runs each, `-b`, COMPILER, 16 threads) shows the new logic **within run-to-run noise** of the old (no per-query delta beyond combined stddev).

- InputFormatters (CSV/JSON spanning-tuple resolution via `SequenceShredder` / `SpanningTupleBuffer`)

## Documentation
- In-code documentation updated: the sentinel rationale is documented at `setStateOfFirstIndex`, `tryClaimSpanningTuple`, and `getTrailingBytes`.

N/A

### DIFF
--- a/nes-input-formatters/src/SpanningTupleBufferState.cpp
+++ b/nes-input-formatters/src/SpanningTupleBufferState.cpp
@@ -108,7 +108,13 @@ std::optional<StagedBuffer> SpanningTupleBufferEntry::tryClaimSpanningTuple(cons
 {
     if (this->atomicState.tryClaimSpanningTuple(abaItNumber))
     {
-        INVARIANT(this->trailingBufferRef.getReferenceCounter() != 0, "Tried to claim a trailing buffer with a nullptr");
+        /// The start-of-stream sentinel (very first entry, first iteration) owns no backing buffer (null data pointer). It
+        /// anchors the first leading spanning tuple but contributes no bytes, so we hand back an empty StagedBuffer for it.
+        if (!this->trailingBufferRef)
+        {
+            this->atomicState.setUsedTrailingBuffer();
+            return {StagedBuffer{}};
+        }
         const auto stagedBuffer
             = StagedBuffer(RawTupleBuffer{std::move(this->trailingBufferRef)}, firstDelimiterOffset, lastDelimiterOffset);
         this->atomicState.setUsedTrailingBuffer();
@@ -117,10 +123,10 @@ std::optional<StagedBuffer> SpanningTupleBufferEntry::tryClaimSpanningTuple(cons
     return std::nullopt;
 }
 
-void SpanningTupleBufferEntry::setStateOfFirstIndex(TupleBuffer dummyBuffer)
+void SpanningTupleBufferEntry::setStateOfFirstIndex()
 {
-    /// The first entry is a dummy that makes sure that we can resolve the first tuple in the first buffer
-    this->trailingBufferRef = std::move(dummyBuffer);
+    /// The first entry is a start-of-stream sentinel. It owns no buffer; it only carries the atomic state that lets the
+    /// SpanningTupleBuffer resolve the first tuple in the first buffer (acting as a virtual leading delimiter).
     this->atomicState.setStateOfFirstEntry();
     this->firstDelimiterOffset = 0;
     this->lastDelimiterOffset = 0;
--- a/nes-input-formatters/include/InputFormatters/InputFormatter.hpp
+++ b/nes-input-formatters/include/InputFormatters/InputFormatter.hpp
@@ -65,7 +65,7 @@ public:
         , inputFormatIndexer(std::move(inputFormatIndexer))
         , projections(memoryProvider->getAllFieldNames())
         , memoryProvider(std::move(memoryProvider))
-        , sequenceShredder(std::make_unique<SequenceShredder>(this->inputFormatIndexer->getTupleDelimitingBytes().size()))
+        , sequenceShredder(std::make_unique<SequenceShredder>())
     {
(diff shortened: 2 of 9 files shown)


nes-input-formatters/src/SpanningTupleBufferState.cpp
```diff
@@ -108,7 +108,13 @@ std::optional<StagedBuffer> SpanningTupleBufferEntry::tryClaimSpanningTuple(cons
...
     if (this->atomicState.tryClaimSpanningTuple(abaItNumber))
     {
-        INVARIANT(this->trailingBufferRef.getReferenceCounter() != 0, "Tried to claim a trailing buffer with a nullptr");
+        /// The start-of-stream sentinel (very first entry, first iteration) owns no backing buffer (null data pointer). It
+        /// anchors the first leading spanning tuple but contributes no bytes, so we hand back an empty StagedBuffer for it.
+        if (!this->trailingBufferRef)
+        {
+            this->atomicState.setUsedTrailingBuffer();
+            return {StagedBuffer{}};
+        }
```
alepping, line 117: I went from exactly this logic to the dummybuffer (I literally deleted equivalent lines in the PR back then).
While the dummybuffer is a bit ugly, concerning logs, its the way more elegant solution concerning the SequenceShredder.
The nullcheck here is just ugly. We do it for every every ST we try to claim, even though its a one-off for the first buffer.
The real problem is that the there is no way to construct a TupleBuffer with a control block except for using a throwaway buffermanager.

alepping: If it really is just the log noise that is the issue with the dummy buffer, I wonder if we can't go a separate route, instead re-adding logic that I purposefully deleted.

alepping (changes requested):

nes-input-formatters/src/SpanningTupleBufferState.cpp
alepping: Another solution that I contemplated was passing the dummy buffer from InputFormatter/Scan, but I didn't see a good way to do it back then

ls-1801: > The nullcheck here is just ugly. We do it for every every ST we try to claim, even though its a one-off for the first buffer.

I think it's perfectly reasonable to have that check. Why pay for a design decision that has been done within the sequence shredder, with something that is visible globally, and very likely to trip up everyone (who has no knowledge about this hack, so it's one person who has to deal with an "ugly" if condition, vs many people). Given future extension to the BufferManager (spilling, statistics), I would not be so sure that construction would be as simple as it is right now.

There was no measurable performance impact, as the branch is trivial to predict. 

@keyseven123 We need someone to decide

keyseven123: I agree with @ls-1801 here. During the paged vector refactoring, we saw more and more the need to refactor also the tuple buffers and buffer manager. Adding a `[[unlikely]]` to the if statement should have no measurable performance hit.
IMHO, the cleanest fix would be to have a dummy buffer that implements a tuple buffer interface so that this dummy buffer can be created in the constructor of the `SequenceShredder`. But this is not easily possible

keyseven123 (changes requested):
See my reply.

keyseven123 (commented):
We should have a short discussion about how much work the ' DummyBuffer ' would be. If it is too much work, then I think this PR should be approved

keyseven123 (approved):

alepping (approved):

PR #1665, opened 2026-06-08
Poison unused buffer/arena memory under AddressSanitizer

### LINKED ISSUE #1515: BufferManager: Add memory poisoning to catch out-of-bounds access in debug mode

ls-1801:
Closes #1515.

## What

Adds AddressSanitizer manual poisoning so that buffer memory which is *not* currently handed out is poisoned, catching use-before-allocate, use-after-recycle and out-of-bounds accesses that ASan otherwise misses (the memory lives inside our own pooled allocations, so ASan considers it addressable). All of this is a no-op without ASan — production memory layout and behaviour are unchanged.

## Commits (each builds on its own)

1. **`refactor(nes-memory): drop wasted control-block space in unpooled buffers`** — unpooled buffers heap-allocate their control block (wrapped `MemorySegment`), yet the chunk layout still reserved `sizeof(BufferControlBlock)` bytes in front of every data region. Drop that offset (saves 256 B per unpooled buffer); data region stays aligned.
2. **`feat(nes-common): add Util/Sanitizer.hpp`** — shared header exposing `NES_ASAN_ENABLED` and always-defined `ASAN_POISON_MEMORY_REGION` / `ASAN_UNPOISON_MEMORY_REGION` (no-ops without ASan).
3. **`feat(nes-memory): poison unused buffer memory under ASan`** — pooled slots become `[ redzone | control block | redzone | data region ]`. The arena is poisoned whole once at init; control block and data region then follow the buffer lifecycle (unpoisoned on hand-out in `prepare()`/`getBuffer*`, re-poisoned on recycle), with flips only at the two exclusive points and ordered across the `availableBuffers` queue, so concurrent `retain`/`release` are race-free. The redzones are never flipped and guard against over-/underflow into the control block. Unpooled chunks get the analogous treatment; their heap control blocks are already covered by ASan's heap redzones.
4. **`refactor(nes-nautilus): align arena allocations to 8 bytes`** — the arena bump allocator handed out unaligned, tightly-packed pointers; advance the cursor in multiples of 8 (safer for typed/atomic access and a prerequisite for precise poisoning).
5. **`feat(nes-nautilus): poison unused arena memory under ASan`** — poison a fresh arena buffer whole, unpoison exactly each allocation, insert a poisoned redzone between allocations (ASan-only, 0 in prod), and poison the unused tail of unpooled allocations.

## Design note

ASan's shadow memory is global and unsynchronized, so poison state is only flipped at exclusive points (hand-out / final release), never per-access — that's why control blocks use lifecycle poisoning plus permanent redzones rather than poisoning around every getter, which would race the concurrently-accessed reference counter.

## Testing

- Builds in both ASan (`address-libstdcxx` toolchain) and non-ASan configs.
- `nes-memory` suites (`unpooled-buffer-test`, `tuple-buffer-memory-access-tests`, `buffer-manager-destroy-test`, `child-buffer-tests`), plus `AndOrPhysicalFunctionTest` and `variable-sized-data-unit-tests` (arena paths) — all pass, no false positives.
- Targeted harnesses confirm detection of: data-after-recycle, overflow into a redzone, stray read of a recycled control block, arena overflow into a redzone, arena overflow into the next allocation, and unpooled overflow.

### DIFF
--- a/nes-common/include/Util/Sanitizer.hpp
+++ b/nes-common/include/Util/Sanitizer.hpp
@@ -1,0 +1,40 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+#pragma once
+
+/**
+ * @brief AddressSanitizer (ASan) detection and manual poisoning support.
+ *
+ * <sanitizer/asan_interface.h> ships with the compiler and already defines ASAN_POISON_MEMORY_REGION /
+ * ASAN_UNPOISON_MEMORY_REGION as no-ops when ASan is off, so call sites need no #ifdefs. We only add
+ * NES_ASAN_ENABLED for the places that must size a redzone at compile time.
+ * See https://github.com/google/sanitizers/wiki/AddressSanitizerManualPoisoning.
+ */
+#include <sanitizer/asan_interface.h> /// NOLINT(misc-include-cleaner)
+
+/// Nested, not `defined(__has_feature) && __has_feature(...)`: the preprocessor does not guarantee the
+/// second operand is left unexpanded on compilers without __has_feature.
+#if defined(__has_feature)
+    #if __has_feature(address_sanitizer)
+inline constexpr bool NES_ASAN_ENABLED = true;
+    #elif defined(__SANITIZE_ADDRESS__)
+inline constexpr bool NES_ASAN_ENABLED = true;
+    #else
+inline constexpr bool NES_ASAN_ENABLED = false;
+    #endif
+#elif defined(__SANITIZE_ADDRESS__)
+inline constexpr bool NES_ASAN_ENABLED = true;
+#else
+inline constexpr bool NES_ASAN_ENABLED = false;
+#endif
--- a/nes-memory/BufferManager.cpp
+++ b/nes-memory/BufferManager.cpp
@@ -27,6 +27,7 @@
 #include <memory_resource>
 #include <mutex>
 #include <optional>
+#include <span>
 #include <utility>
 #include <unistd.h>
 #include <Runtime/AbstractBufferProvider.hpp>
@@ -35,6 +36,7 @@
 #include <Runtime/TupleBuffer.hpp>
 #include <Util/Logger/Logger.hpp>
 #include <folly/MPMCQueue.h>
+#include <sanitizer/asan_interface.h>
 #include <ErrorHandling.hpp>
 #include <TupleBufferImpl.hpp>
 
@@ -94,6 +96,10 @@ void BufferManager::destroy()
     NES_DEBUG("Calling BufferManager::destroy()");
     if (isDestroyed.compare_exchange_strong(expected, true))
     {
+        /// All buffers should be idle now, so their control blocks (and data regions) are poisoned. Clear all
+        /// poison up front: the availability checks and the MemorySegment destructors in allBuffers.clear()
+        /// below read the control blocks, and the arena is about to be returned to the allocator anyway.
+        ASAN_UNPOISON_MEMORY_REGION(basePointer, allocatedAreaSize);
         bool success = true;
         if (allBuffers.size() != getNumberOfAvailableBuffers())
         {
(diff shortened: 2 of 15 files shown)


alepping (changes requested):
Cool change, thanks!
Many hours spent on debugging exactly the class of errors that this addresses.

The SIMDJSON issue is annoying and I think your solution is decent.
A potential solution that requires less source refactoring could be to reduce the index window of SIMDJSON at the cost of creating larger/more spanning tuples. I'd like to try that out first, before accepting/merging this.
If we can align pages, it would require less code to change, to revert to simdjson-adaption.

nes-common/include/Util/Sanitizer.hpp
```diff
@@ -0,0 +1,41 @@
...
+        #define NES_ASAN_ENABLED 1
+    #endif
+#endif
+#if !defined(NES_ASAN_ENABLED) && defined(__SANITIZE_ADDRESS__)
+    #define NES_ASAN_ENABLED 1
+#endif
+#if defined(NES_ASAN_ENABLED)
+    #include <sanitizer/asan_interface.h>
+#else
+    #define NES_ASAN_ENABLED 0
```
keyseven123: Why not? 
```suggestion
Index: nes-common/include/Util/Sanitizer.hpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/nes-common/include/Util/Sanitizer.hpp b/nes-common/include/Util/Sanitizer.hpp
--- a/nes-common/include/Util/Sanitizer.hpp	(revision 8013fd824d46f5ee9c4562923f698463a2689526)
+++ b/nes-common/include/Util/Sanitizer.hpp	(date 1784726954019)
@@ -16,26 +16,24 @@
 /**
  * @brief AddressSanitizer (ASan) detection and manual poisoning support.
  *
- * Defines NES_ASAN_ENABLED (1 when built with ASan, 0 otherwise) and guarantees that the manual
- * poisoning macros ASAN_POISON_MEMORY_REGION / ASAN_UNPOISON_MEMORY_REGION are always available:
- * they call into the ASan runtime when ASan is enabled and expand to no-ops otherwise. This lets
- * callers poison/unpoison memory regions unconditionally without sprinkling #ifdefs at each call
- * site. See https://github.com/google/sanitizers/wiki/AddressSanitizerManualPoisoning.
+ * <sanitizer/asan_interface.h> ships with the compiler and already defines ASAN_POISON_MEMORY_REGION /
+ * ASAN_UNPOISON_MEMORY_REGION as no-ops when ASan is off, so call sites need no #ifdefs. We only add
+ * NES_ASAN_ENABLED for the places that must size a redzone at compile time.
+ * See https://github.com/google/sanitizers/wiki/AddressSanitizerManualPoisoning.
  */
+#include <sanitizer/asan_interface.h>
+
+/// Nested, not `defined(__has_feature) && __has_feature(...)`: the preprocessor does not guarantee the
+/// second operand is left unexpanded on compilers without __has_feature.
 #if defined(__has_feature)
     #if __has_feature(address_sanitizer)
         #define NES_ASAN_ENABLED 1
     #endif
 #endif
-#if !defined(NES_ASAN_ENABLED) && defined(__SANITIZE_ADDRESS__)
-    #define NES_ASAN_ENABLED 1
-#endif
-#if defined(NES_ASAN_ENABLED)
-    #include <sanitizer/asan_interface.h>
-#else
-    #define NES_ASAN_ENABLED 0
-    #ifndef ASAN_POISON_MEMORY_REGION
-        #define ASAN_POISON_MEMORY_REGION(addr, size) (static_cast<void>(addr), static_cast<void>(size))
-        #define ASAN_UNPOISON_MEMORY_REGION(addr, size) (static_cast<void>(addr), static_cast<void>(size))
+#ifndef NES_ASAN_ENABLED
+    #ifdef __SANITIZE_ADDRESS__
+        #define NES_ASAN_ENABLED 1
+    #else
+        #define NES_ASAN_ENABLED 0
     #endif
 #endif
```

nes-memory/BufferManager.cpp
```diff
@@ -155,12 +167,12 @@ void BufferManager::initialize()
         controlBlockSize,
         alignof(detail::BufferControlBlock));
 
-    INVARIANT(basePointer, "memory allocation failed, because 'basePointer' was a nullptr");
     uint8_t* ptr = basePointer;
     for (size_t i = 0; i < numOfBuffers; ++i)
     {
-        uint8_t* controlBlock = ptr;
-        uint8_t* payload = ptr + controlBlockSize;
+        /// Slot layout: [ redzone | control block | redzone | data region ]. redzoneSize is 0 without ASan.
```
keyseven123: Do we need this comment here again?

keyseven123 (changes requested):

nes-memory/BufferManager.cpp
ls-1801: addressed

nes-common/include/Util/Sanitizer.hpp
ls-1801: addressed

ls-1801:
@alepping I was able to move that logic into the formatter. The indexer will not look at the last couple of bytes of a buffer, and they will be treated as a trailing tuple, which is copied into a dedicated memory area (which we can pad with however many bytes simdjson needs)

keyseven123 (approved):

alepping (approved):

PR #1666, opened 2026-06-09
Bump nautilus to latest main and port pipeline compilation to the module API

PhilippGrulich:
## What

Ports the nautilus dependency (vcpkg + Nix) from `1c6d94a8` to current nautilus `main` (`3e3da267`), and ports the nautilus-based operator implementation to the new **module API**: each pipeline now compiles into **exactly one module** containing multiple named functions.

## Dependency bump

- **vcpkg** (`vcpkg/vcpkg-registry/ports/nautilus/portfile.cmake`): update `REF`/`SHA512`; disable the newly-default-on asmjit backend and simd/specialization/inlining/gpu plugins (the asmjit backend breaks the installed CMake config by referencing a non-installed `asmjit::asmjit` target); keep the **std plugin ON** (NES includes `nautilus/std/{ostream,sstream,string,cstring}.h`).
- **Nix** (`.nix/nautilus/package.nix`): same `rev`/`hash` bump and flag alignment; regenerated `0003-disable-cond-branch-weights.patch` against the relocated `MLIRLoweringProvider::visitIf`.

## Module API port

New nautilus removed `engine::CallableFunction`; functions are now compiled together via `engine.createModule()` → N × `module.registerFunction(name, fn)` → one `module.compile()` → named `ModuleFunction` handles.

- **`CompilationContext`** (nes-runtime) is now the per-pipeline module builder. Operators register named functions during `setup()` and receive a **`PipelineFunction<Sig>`** handle that becomes invocable automatically after the stage compiles the module (invoking earlier trips an `INVARIANT`). Names are prefix + counter for uniqueness.
- **`CompiledExecutablePipelineStage::start()`** builds the single module per pipeline: operator `setup()` registers helpers, the main traced pipeline function joins the same module under the reserved name `"execute"`, then one `compile()` traces and compiles everything together. `CompiledModule` + the main `ModuleFunction` are stored as `std::optional` members; `execute()` INVARIANT-checks that `start()` ran.
- **Slice cleanup** (`NautilusCleanupExec` for aggregation/hash-join) is now `PipelineFunction<void(HashMap*)>`. `PipelineFunction` co-owns the compiled executable through nautilus's shared module state (`Module.hpp`: `ModuleFunction` holds `shared_ptr<ModuleState>`, which owns the `Executable`), so cleanup handles stored in `HashMapSlice`s stay valid even if a slice outlives the pipeline stage. Downstream `shared_ptr` plumbing unchanged.
- **Test utils** stay on the still-supported single-function engine API; only the named result type changes to `CompiledFunction<R(Args...)>`.

Interpreted mode (`engine.Compilation=false`) flows through the same path — `ModuleFunction` dispatches to the registered `std::function` directly.

## Verification

- Built nautilus locally in interpreter-only mode and ran a harness driving the **actual** `CompilationContext.hpp` end-to-end: deferred-handle resolution, both `registerFunction` overloads, raw-signature deduction, single-compile flow, pre-compilation invocation invariant, and handle validity after `NautilusModule`/`CompiledModule` destruction — all pass.
- Hashes validated by reproducing the old known-good values with the same tooling before computing the new ones; all referenced port patches apply cleanly.
- `grep CallableFunction` over the repo is empty; full compile/test validation via CI (smoke tests cover COMPILER mode; aggregation/join systests exercise the module-compiled cleanup path).

https://claude.ai/code/session_01KmKRyv6JurTp8gGg89kwyt


nes-runtime/include/CompilationContext.hpp
```diff
@@ -13,32 +13,102 @@
...
+    uint64_t functionNameCounter = 0;
 
 public:
-    explicit CompilationContext(const nautilus::engine::NautilusEngine& engine) : engine(engine) { }
+    explicit CompilationContext(nautilus::engine::NautilusModule& module) : module(module) { }
 
     template <typename R, typename... FunctionArguments>
-    auto registerFunction(R (*fnptr)(nautilus::val<FunctionArguments>...)) const
+    auto
+    registerFunction(std::function<R(nautilus::val<FunctionArguments>...)> func, const std::string_view namePrefix = "operatorFunction")
```
ls-1801: NIT: `registerFunction` has no precondition guarding against being called after `resolveAfterCompilation`. If that happens the resolver is appended after the vector has already been iterated, so the returned `PipelineFunction` stays permanently unresolved and only trips its `INVARIANT` on first use — far from the bug site. Suggest tracking a `compiled` bool alongside `pendingResolvers` and `PRECONDITION(!compiled, ...)` here (and in the `fnptr` overload).

nes-physical-operators/src/Functions/ArithmeticalFunctions/AbsolutePhysicalFunction.cpp
```diff
@@ -35,25 +35,28 @@ VarVal AbsolutePhysicalFunction::execute(const Record& record, ArenaRef& arena)
...
     if (not inputType.isSignedInteger() and not inputType.isFloat())
     {
-        return value;
+        return value.castToType(outputType.type);
     }
 
     /// We need to built a zero and negativeOne via castToType, as we can not make any assumptions on the input type.
     const auto zero = VarVal{0}.castToType(inputType.type);
     const auto negativeOne = VarVal{-1}.castToType(inputType.type);
     if (value < zero)
```
ls-1801: `value` here is a `VarVal`, and `VarVal` has a static `select` (see `nes-nautilus/include/DataTypes/VarVal.hpp:62`). Prefer `return VarVal::select(value < zero, value * negativeOne, value).castToType(outputType.type);`

nes-runtime/src/Pipelines/CompiledExecutablePipelineStage.cpp
```diff
@@ -40,63 +42,52 @@ CompiledExecutablePipelineStage::CompiledExecutablePipelineStage(
...
-nautilus::engine::CallableFunction<void, PipelineExecutionContext*, const TupleBuffer*, const Arena*>
-CompiledExecutablePipelineStage::compilePipeline() const
+void CompiledExecutablePipelineStage::registerPipelineFunction(nautilus::engine::NautilusModule& module) const
 {
-    CPPTRACE_TRY
+    /// We must capture the operatorPipeline by value to ensure it is not destroyed before the function is called
+    /// Additionally, we can NOT use const or const references for the parameters of the lambda function
+    /// NOLINTBEGIN(performance-unnecessary-value-param)
+    const std::function<void(nautilus::val<PipelineExecutionContext*>, nautilus::val<const TupleBuffer*>, nautilus::val<const Arena*>)>
+        compiledFunction = [this](
```
ls-1801, line 70: NIT: Comment says the pipeline is captured by value to keep it alive, but the capture is `[this]` — a raw pointer to the enclosing stage. The lambda reaches `pipeline` via `this->pipeline`, so if `CompiledExecutablePipelineStage` is destroyed before the compiled function executes the lambda dangles. To actually match the stated intent, capture the `shared_ptr` by value: `[pipeline = pipeline](...) { ... }` (and drop the `this->` indirection inside).

I know this is unrelated to the actual change, feel free to ignore it

ls-1801 (approved):
Please address the clang tidy warnings (except the default param if it's a hassle, or add a second overload).

PhilippGrulich:
Re: the Codecov patch-coverage report — leaving the 3 uncovered lines as-is. They're all defensive/diagnostic paths in the new code that aren't worth contrived tests:

- `CompiledExecutablePipelineStage.cpp` — the `throw wrapExternalException(...)` inside `CPPTRACE_CATCH` (only reached when nautilus compilation fails) and the `if (compiledModule->getStatistics())` debug-logging branch (only reached with debug logging + compiled mode).
- `CompilationContext.hpp` — the `INVARIANT` guard that fires only if a `PipelineFunction` is invoked before its module has compiled.

Codecov reports "All tests successful" with 97.6% patch coverage; the ✗ is only the sub-100% patch threshold, not a failed test. Exercising these would require building a real engine + a pipeline that deliberately fails to compile, which is disproportionate to the value of covering error guards.

---
_Generated by [Claude Code](https://claude.ai/code)_

nes-runtime/include/CompilationContext.hpp
PhilippGrulich: Done — `CompilationContext` now tracks a `compiled` bool that is set in `resolveAfterCompilation()`, and both `registerFunction` overloads start with `PRECONDITION(!compiled, ...)`, so registering after compilation now fails at the call site instead of leaving a permanently-unresolved handle.

nes-physical-operators/src/Functions/ArithmeticalFunctions/AbsolutePhysicalFunction.cpp
PhilippGrulich: Done — `execute()` now uses `VarVal::select(value < zero, value * negativeOne, value)`, casting each branch to the output type to keep the branch types aligned after integer promotion.

nes-runtime/src/Pipelines/CompiledExecutablePipelineStage.cpp
PhilippGrulich: Kept the `[this]` capture intentionally. Capturing the `pipeline` shared_ptr by value makes the compiled module co-own the pipeline, and since cached slices keep that module alive via their cleanup handle, the pipeline (and its slice buffers) is retained past teardown — which leaks buffers in the sliceCache systests. The stage always outlives every invocation of the compiled function (it is only called through `execute()`/`start()`/`stop()`), so `this` cannot dangle. Added a comment block above the lambda documenting this trade-off.

keyseven123 (changes requested):
Please squash some commits and resolve the conflicts, otherwise LGTM.

keyseven123 (approved):

zeuchste:
@PhilippGrulich could you rebuild/republish the dev + CI dependency images? This PR added the `nautilus::nautilus-std` link (`nes-nautilus/CMakeLists.txt:46`), but the published dependency images at the current hash (`0504de94a8…`) don't ship that target yet, so `cmake` configure fails before any build:

```
CMake Error at nes-nautilus/CMakeLists.txt:45 (target_link_libraries):
  Target "nes-nautilus" links to: nautilus::nautilus-std  — but the target was not found.
```

Verified it's the image, not a local setup issue: configure fails identically **inside the official `nebulastream/nes-ci:0504de…-libcxx-none` (x64) image** as well as the arm64 variant — neither's vcpkg nautilus defines `nautilus-std` (only `nautilus::nautilus` + `nautilus::InliningPass`). `get-dev-images` is currently *skipping* (the hash matches the stale image), so a fresh build/publish of the dependency images is needed to unblock local builds (and likely arm64 CI). Thanks!

ls-1801:
@zeuchste docker images are build automatically.

```bash
./scripts/install-local-docker-environment.sh --libcxx --no-sanitizer -y
```

PR #1667, opened 2026-06-10
Chained Hash Map redesign

ThanGeo:
This pull requests refactors the ChainedHashMap so that it uses TupleBuffer in a more structured way. 

## Requirements
This branch builds on top of the paged_vector_redesign branch. When that branch is merged, then this one should follow. 

## Documentation
The ChainedHashMap did not use our self-contained, single-TupleBuffer-based data structure model. This redesign aims to fix this, enabling easier seralization, migration, representation and memory optimization for the ChainedHashMap. 

The HashMapSlice was also slightly redesigned, since it contains hash maps based on the number of workers and input streams. 

image


nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -12,180 +12,1045 @@
...
+            {
+                availableMemoryArea = DIRTY_FILL_PATTERN;
+            }
+        }
+        return buffer;
     }
 
-    static void TearDownTestSuite() { NES_INFO("Tear down ChainedHashMapTest class."); }
+    std::shared_ptr<BufferManager> bm;
+};
```
ls-1801: We dont need this anymore, this has been integrated into the default BufferManager. you can just use the normal buffer manager and delete this code. same would apply for the paged vector test.

nes-physical-operators/src/Aggregation/AggregationBuildPhysicalOperator.cpp
```diff
@@ -90,10 +54,14 @@ void AggregationBuildPhysicalOperator::execute(ExecutionContext& ctx, Record& re
 
     /// Getting the corresponding slice so that we can update the aggregation states
     const auto timestamp = timeFunction->getTs(ctx, record);
-    const auto hashMapPtr
+    const auto hashMapBuffer
         = sliceStoreRef->getDataStructureRef(timestamp, ctx.workerThreadId, operatorHandler, ctx.pipelineMemoryProvider.bufferProvider);
     ChainedHashMapRef hashMap(
```
keyseven123: use {} instead of () for constructor calls

nes-physical-operators/src/Aggregation/AggregationProbePhysicalOperator.cpp
```diff
@@ -66,20 +64,67 @@ void AggregationProbePhysicalOperator::open(ExecutionContext& executionCtx, Reco
...
+            /// get a buffer and for the final hash map with the same config
+            auto neededFinalBufferSize = ChainedHashMap::calculateBufferSizeFromChains(chm.getNumberOfChains());
+            std::optional<TupleBuffer> finalHashMapTupleBuffer = bufferProvider->getUnpooledBuffer(neededFinalBufferSize);
+            if (not finalHashMapTupleBuffer.has_value())
+            {
+                throw CannotAllocateBuffer("{}B for the hash join window trigger were requested", neededFinalBufferSize);
+            }
+            /// initialize the final hash map tuple buffer
+            *finalHashMapBuffer = finalHashMapTupleBuffer.value();
+            std::ignore = ChainedHashMap::init(*finalHashMapBuffer, chm.getEntrySize(), chm.getNumberOfBuckets(), chm.getPageSize());
```
keyseven123: Same as with the paged vector. I think we should remove the return type of the `ChainedHashMap::init()`

nes-physical-operators/src/Aggregation/AggregationProbePhysicalOperator.cpp
```diff
@@ -66,20 +64,67 @@ void AggregationProbePhysicalOperator::open(ExecutionContext& executionCtx, Reco
...
+    /// create final hash map and pin it
+    OwnedNautilusBuffer finalHashMapNautilusBuffer;
+    nautilus::invoke(
+        +[](const TupleBuffer* parent, AbstractBufferProvider* bufferProvider, TupleBuffer* finalHashMapBuffer)
+        {
+            INVARIANT(parent != nullptr, "Parent Tuplebuffer MUST NOT be null at this point");
+            /// load the first hash map
+            const VariableSizedAccess::Index bufferIndex{0};
+            auto buffer = parent->loadChildBuffer(bufferIndex);
+            const ChainedHashMap chm = ChainedHashMap::load(buffer);
```
keyseven123: use auto whenever is possible. applies to multiple occurrences

nes-physical-operators/src/Aggregation/AggregationProbePhysicalOperator.cpp
```diff
@@ -66,20 +64,67 @@ void AggregationProbePhysicalOperator::open(ExecutionContext& executionCtx, Reco
...
+            /// get a buffer and for the final hash map with the same config
+            auto neededFinalBufferSize = ChainedHashMap::calculateBufferSizeFromChains(chm.getNumberOfChains());
```
keyseven123, line 79: I think this should be passed via the `EmittedAggregationWindow`. Then we have everything in one place

nes-physical-operators/src/Join/HashJoin/HJBuildPhysicalOperator.cpp
```diff
@@ -52,47 +53,6 @@ namespace NES
 void HJBuildPhysicalOperator::setup(ExecutionContext& executionCtx, CompilationContext& compilationContext) const
```
keyseven123: I think this method can be removed as we simply call the parent method

nes-physical-operators/src/Join/HashJoin/HJOperatorHandler.cpp
```diff
@@ -115,33 +100,30 @@ void HJOperatorHandler::emitSlicesToProbe(
...
-    const auto leftHashMaps = getHashMapsForSlice(sliceLeft, JoinBuildSideType::Left);
-    const auto rightHashMaps = getHashMapsForSlice(sliceRight, JoinBuildSideType::Right);
+    const auto leftTupleBuffers = getHashMapsForSlice(sliceLeft, JoinBuildSideType::Left);
+    const auto rightTupleBuffers = getHashMapsForSlice(sliceRight, JoinBuildSideType::Right);
 
     /// We need a buffer that is large enough to store:
     /// - size of EmittedHJWindowTrigger
-    /// - all pointers to (left + right) hashmaps of the window to be triggered
-    const auto neededBufferSize = sizeof(EmittedHJWindowTrigger) + ((leftHashMaps.size() + rightHashMaps.size()) * sizeof(HashMap*));
+    /// Hash map buffers are stored as child buffers, not inline in the main buffer.
```
keyseven123, line 189: Nice comment!

nes-physical-operators/src/Join/HashJoin/HJProbePhysicalOperator.cpp
```diff
@@ -85,35 +86,71 @@ void HJProbePhysicalOperator::open(ExecutionContext& executionCtx, RecordBuffer&
...
-            const nautilus::val<HashMap*> rightHashMapPtr = rightHashMapRefs[rightHashMapIndex];
+            /// Pin the current right hashmap buffer
+            /// Right buffers start after all left buffers
+            OwnedNautilusBuffer rightHashMapBufferRef;
+            nautilus::invoke(
+                +[](TupleBuffer* parent, uint32_t leftNumberOfHashMaps, uint32_t rightHashMapIndex, TupleBuffer* rightHashMapBuffer)
+                {
+                    INVARIANT(parent != nullptr, "Parent Tuplebuffer MUST NOT be null at this point");
+                    /// Right buffers start after all left buffers
+                    const VariableSizedAccess::Index bufferIndex{leftNumberOfHashMaps + rightHashMapIndex};
```
keyseven123: One could think about calculating this number in the `triggerSlices` of the operator handler so that one does not have to think about adding this value here.

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
```diff
@@ -106,156 +118,302 @@ ChainedHashMap::ChainedHashMap(const uint64_t keySize, const uint64_t valueSize,
...
+    auto chainsArray = chm.chains();
+    std::ranges::fill(chainsArray, nullptr);
+    /// NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)
+    chainsArray[numberOfChains] = reinterpret_cast<ChainedHashMapEntry*>(&chainsArray[numberOfChains]);
+
+    return chm;
 }
 
-ChainedHashMapEntry* ChainedHashMap::findChain(const HashFunction::HashValue::raw_type hash) const
+ChainedHashMap ChainedHashMap::load(const TupleBuffer& tupleBuffer)
```
keyseven123, line 137: Not sure if this will lead to a problem but we are copying a tuplebuffer in the constructor. This passed TB object will be returned as a mutable view.

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
```diff
@@ -106,156 +118,302 @@ ChainedHashMap::ChainedHashMap(const uint64_t keySize, const uint64_t valueSize,
...
-}
+    /// Create object
+    ChainedHashMap chm(tupleBuffer);
 
-void ChainedHashMap::setDestructorCallback(const std::function<void(ChainedHashMapEntry*)>& callback)
-{
-    destructorCallBack = callback;
-}
+    /// Initialize its header
+    new (tupleBuffer.getAvailableMemoryArea<Header>().data())
```
keyseven123, line 127: I think we should check if the tupleBuffer is large enough as a precondition

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
```diff
@@ -106,156 +118,302 @@ ChainedHashMap::ChainedHashMap(const uint64_t keySize, const uint64_t valueSize,
...
+    return page;
 }
 
-ChainedHashMapEntry* ChainedHashMap::getStartOfChain(const uint64_t entryIdx) const
+uint64_t ChainedHashMap::calculateBufferSizeFromBuckets(uint64_t numberOfBuckets)
 {
-    PRECONDITION(entryIdx <= numberOfChains, "Entry index {} is greater than the capacity {}", entryIdx, numberOfChains);
-    return entries[entryIdx];
+    const uint64_t numberOfChains = calcCapacity(numberOfBuckets, assumedLoadFactor);
+    return sizeof(Header) + ((numberOfChains + 1) * sizeof(uint64_t));
```
keyseven123:
```suggestion
    return calculateBufferSizeFromChains(numberOfChains);
```

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMapRef.cpp
```diff
@@ -269,57 +268,89 @@ ChainedHashMapRef::EntryIterator ChainedHashMapRef::begin() const
...
-    const nautilus::val<ChainedHashMapEntry*> chainStart = entries[entryStartPos];
+    const auto chainStart = invoke(
+        +[](TupleBuffer* tupleBuffer, size_t entryStartPos)
+        {
+            ChainedHashMap chm = ChainedHashMap::load(*tupleBuffer);
+            return chm.getChain(entryStartPos);
+        },
+        tupleBuffer,
+        entryStartPos);
     return chainStart;
```
keyseven123:
```suggestion
return nautilus::invoke(
        +[](const TupleBuffer* tupleBuffer, const HashFunction::HashValue::raw_type hashValue) -> ChainedHashMapEntry*
        {
            ChainedHashMap chm = ChainedHashMap::load(*tupleBuffer);
            const auto numberOfTuples = chm.getTotalNumberOfRecords();
            if (numberOfTuples == 0)
            {
                return nullptr;
            }
            const auto entryPos = hashValue & chm.getMask();
            return chm.getChain(entryPos);
        },
        tupleBuffer,
        hash);
```

I would have also expected that we can do all of this directly in nautilus and would not need to go to the c++ runtime. In my mind, we can calculate all of these metadata fields as trace-time-constants.

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMapRef.cpp
```diff
@@ -269,57 +268,89 @@ ChainedHashMapRef::EntryIterator ChainedHashMapRef::begin() const
...
-    return {hashMapRef, currentEntry, entrySize, tupleIndex, indexOnPage, numberOfTuplesInCurrentPage, pageIndex, numberOfPages};
+    return {tupleBuffer, currentEntry, entrySize, tupleIndex, indexOnPage, numberOfTuplesInCurrentPage, pageIndex, numberOfPages};
 }
 
 ChainedHashMapRef::EntryIterator ChainedHashMapRef::end() const
 {
     /// The iterator pointing to the end() should NEVER be advanced. Therefore, we do not need to set a lot of its members
-    const auto numberOfTuples = readValueFromMemRef<uint64_t>(getMemberRef(hashMapRef, &ChainedHashMap::numberOfTuples));
-    return {hashMapRef, nullptr, entrySize, numberOfTuples, -1, -1, -1, -1};
+    const auto numberOfTuples = invoke(
```
keyseven123, line 302: I think that we should be able to read this directly in nautilus and calculate the memory address to read from via a trace-time-constant

keyseven123 (changes requested):
First round of feedback, as discussed. I like this pr. 
On a high level. I think we can reduce the number of invokes() and thus increase the performance. We should be able to read the metadata, no. tuples etc, directly in nautilus.

One small thing: I think that `getSliceCleanupFunction()` is not necessary anymore and can be removed.

ls-1801 (changes requested):
Please rebase and remove the paged vector commit.

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
ThanGeo: Then the constructor is doing std::move(buffer). Should i move the std::move in the load() method instead?

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMapRef.cpp
ThanGeo: I think I understand what you mean. Instead of invoking and loading, we can read the value from memory directly as we know the offset. Not sure how to implement. Also, wouldn't this go a bit against the whole idea with the API over tuplebuffer, as nautilus will be directly reading from its memory instead?

PR #1669, opened 2026-06-10
426 logical plan debug vis

### LINKED ISSUE #426: [TASK] Custom query plan type renderer for debugger
### Description

Currently, we don't have a ergonomic way of looking at query plans in the debugger.

Nodes have a toString method and QueryConsoleDumpHandler assembles the string representations of individual nodes to a multi line plan string. 
This string representation is only used for logging.

CLion unfortunately does not provide custom type renderers like in IntelliJ.
We should investigate if there are other debugging tools (that can be used easily with the existing toolchain) that can display some custom type rendering.

If that is not possible, we should at least some convenience functions that can be used as watches in the debugger.

### Acceptance Criteria

One of:
 - A type renderer integrated into the debugger that can display a configurable string representation of plans or nodes (we might not want to display all the information that is available at hand for better readability, but this needs to be tested out)
 - Convenience functions that can be easily called in the debugger

tim-nick:
### Purpose of the Change and Brief Change Log

 This is about how Logical plans are shown in the debugger. 

I have implemented two options for rendering the plan better. They are two separate commits so you can keep one, both, or drop either:

- Debug::view() — a small C++ helper. You type `NES::Debug::view(plan)` in the evaluate box and get a clean little tree with short labels like SINK(name), SELECTION(field), SOURCE(name). 
- GDB python pretty-printer — auto-adds a one-line label to every plan/operator variable, e.g. SinkLogicalOperator (id: 7, children: [Selection]), so you don't have to call anything. Loads automatically in the docker container; for native debugging you add two lines to your ~/.gdbinit.

### Verifying this change
With LogicalPlanTest
Just set a breakpoint on an EXPECT_EQ line in one of the DebugView_* tests and look at `NES::Debug::view(plan)`.

What components does this pull request potentially affect?

- nes-logical-operators — the new view() helper + tests (doesn't touch any real plan logic)
- Development.dockerfile / DevelopmentLocal.dockerfile — GDB setup so the printer loads
- .gdbinit — new file at repo root, loads the printer script for native debugging

This PR closes #426

### DIFF
--- a/.gdbinit
+++ b/.gdbinit
@@ -1,0 +1,34 @@
+# NES project-level GDB init file
+# Lives at the repo root: nebulastream-public/.gdbinit
+#
+# GDB loads this automatically when you debug from the nebulastream-public
+# root directory — which is what CLion does.
+#
+# ONE-TIME SETUP per developer (only for native debugging on the host):
+#   Add this line to your ~/.gdbinit (create it if it doesn't exist):
+#
+#       add-auto-load-safe-path /path/to/nebulastream-public
+#
+#   This tells GDB it is safe to auto-load this project's .gdbinit.
+#   Without it GDB silently ignores this file for security reasons.
+#
+#   Inside the dev container this is already set up (the image trusts the
+#   mounted repo path), so no ~/.gdbinit change is needed there.
+
+# Load the NES pretty printer. This file only auto-loads from the repo root, so
+# the cwd is that root and the scripts/ path can be derived from it.
+python
+import os
+import sys
+sys.path.insert(0, os.path.join(os.getcwd(), 'scripts'))
+import pretty_print_logic_plans
+end
+
+# Make STL containers readable (vectors, strings etc.)
+set print pretty on
+
+# Show all characters in strings (plans can be long)
+set print elements 0
+
+# Follow shared_ptr and virtual dispatch automatically
+set print object on
--- a/docker/dependency/Development.dockerfile
+++ b/docker/dependency/Development.dockerfile
@@ -36,13 +36,25 @@ RUN git clone https://github.com/aras-p/ClangBuildAnalyzer.git \
 
 # Install GDB Libc++ Pretty Printer
 RUN wget -P /usr/share/libcxx/  https://raw.githubusercontent.com/llvm/llvm-project/refs/tags/llvmorg-19.1.7/libcxx/utils/gdb/libcxx/printers.py && \
-    cat <<EOF > /etc/gdb/gdbinit
-    python
-    import sys
-    sys.path.insert(0, '/usr/share/libcxx')
-    import printers
-    printers.register_libcxx_printer_loader()
-    end
+    cat << 'EOF' > /etc/gdb/gdbinit
+python
+import sys
+sys.path.insert(0, '/usr/share/libcxx')
+import printers
+printers.register_libcxx_printer_loader()
+end
+set auto-load safe-path /
+EOF
+
+# Same libc++ pretty printer for udb (UndoDB time-travel debugger)
+RUN cat << 'EOF' > /root/.udbinit
+python
+import sys
+sys.path.insert(0, '/usr/share/libcxx')
+import printers
+printers.register_libcxx_printer_loader()
+end
+set auto-load safe-path /
 EOF
 
 # Installing the stable and nightly rust toolchain
--- a/docker/dependency/DevelopmentLocal.dockerfile
+++ b/docker/dependency/DevelopmentLocal.dockerfile
@@ -21,4 +21,10 @@ RUN (${ROOTLESS} || (echo "uid: ${UID} gid ${GID} username ${USERNAME}" && \
 RUN mkdir -p /run/containerd && \
     chown -R ${UID}:${GID} /run/containerd
 
+# Mirror the root-level gdbinit trust setting for the non-root dev user: GDB reads
+# per-user config from ~/.config/gdb/gdbinit, and without this it refuses to
+# auto-load the repo-root .gdbinit (which loads the pretty-printer) when running
+# as ${USERNAME}.
+RUN mkdir -p /home/${USERNAME}/.config/gdb && \
+    echo "set auto-load safe-path /" > /home/${USERNAME}/.config/gdb/gdbinit
 USER ${USERNAME}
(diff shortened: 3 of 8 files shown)


nes-logical-operators/tests/LogicalPlanTest.cpp
```diff
@@ -237,3 +260,158 @@ TEST_F(LogicalPlanTest, OutputOperator)
...
+
+/// ── Breakpoint-target tests ──────────────────────────────────────────────────
+/// Build a plan, pause at the marked line, and inspect `plan` or
+/// `NES::Debug::view(plan)` in the debugger Variables panel.
+
+TEST_F(LogicalPlanTest, DebugView_DeepFilterPipeline)
+{
+    /// Source → WatermarkAssigner → Selection(temperature > limit) → Selection(id) → Sink
+    const LogicalOperator source{SourceNameLogicalOperator{"sensors"}};
+    const LogicalOperator watermark{IngestionTimeWatermarkAssignerLogicalOperator{}.withChildren({source})};
```
Artraxon: 2 things regarding constructing operators:

1. Always use the constructor with directly passing the children. It directly runs local schema inference and insures the correct number of children via the ctor signature. Consider the ctor without children deprecated, it will be removed in the near future.
2. Always use the emplace constructor with argument forwarding, i.e. TypedLogicalOperator<SourceNameLogicalOperator>{"sensors"}, when you rebase on main you will see that you are not going to be able to create the operators directly anymore since you'd need the weak self-reference to the shared ptr of the operator, which you won't (and should not) have.

nes-logical-operators/tests/LogicalPlanTest.cpp
```diff
@@ -237,3 +260,158 @@ TEST_F(LogicalPlanTest, OutputOperator)
     ss << plan;
     EXPECT_FALSE(ss.str().empty());
 }
+
+/// ── Breakpoint-target tests ──────────────────────────────────────────────────
+/// Build a plan, pause at the marked line, and inspect `plan` or
+/// `NES::Debug::view(plan)` in the debugger Variables panel.
+
+TEST_F(LogicalPlanTest, DebugView_DeepFilterPipeline)
```
Artraxon: In general, I don't quite see why we'd need these tests, since they are not asserting anything.

I understand that you want to give some example to point to, but tbh we already have tests for other optimization stages (like field ordering), or you could just put a breakpoint in RuleBasedOptimizer::optimize or something like that.

So these are just deadweight that we need to carry through without them testing regressions, I think they should be removed.

scripts/pretty_print_logic_plans.py
```diff
@@ -0,0 +1,311 @@
+# NES Query Plan Pretty Printers for GDB
```
Artraxon, line 13: don't forget the license preamble

.gdbinit
```diff
@@ -0,0 +1,32 @@
+# NES project-level GDB init file
+# Lives at the repo root: nebulastream-public/.gdbinit
+#
+# GDB loads this automatically when you debug from the nebulastream-public
+# root directory — which is what CLion does.
+#
+# ONE-TIME SETUP per developer:
+#   Add these two lines to your ~/.gdbinit (create it if it doesn't exist):
+#
+#       set auto-load safe-path /
```
Artraxon: is adding the root really necessary? That has nothing to do with this project, does it?

Also, one sentence regarding paths in a docker deployment would be good.

docker/dependency/Development.dockerfile
```diff
@@ -34,15 +28,17 @@ RUN git clone https://github.com/aras-p/ClangBuildAnalyzer.git \
...
-    printers.register_libcxx_printer_loader()
-    end
+    cat << 'EOF' > /etc/gdb/gdbinit
+python
+import sys
+sys.path.insert(0, '/usr/share/libcxx')
+import printers
+printers.register_libcxx_printer_loader()
+end
+set auto-load safe-path /
```
Artraxon, line 46: Ideally add it for udb as well
```bash
RUN cat << 'EOF' > /root/.udbinit
python
import sys
sys.path.insert(0, '/usr/share/libcxx')
import printers
printers.register_libcxx_printer_loader()
end
set auto-load safe-path /
EOF
```

Artraxon (changes requested):
Very nice that you are tackling this.
I think its overall a good first version and we can extend on it later as we see fit, but its a good foundation.

docker/dependency/DevelopmentLocal.dockerfile
```diff
@@ -21,4 +21,6 @@ RUN (${ROOTLESS} || (echo "uid: ${UID} gid ${GID} username ${USERNAME}" && \
 RUN mkdir -p /run/containerd && \
     chown -R ${UID}:${GID} /run/containerd
 
+RUN mkdir -p /home/${USERNAME}/.config/gdb && \
```
keyseven123, line 28: Add a comment why we need this line

nes-logical-operators/include/Debug/DebugHelpers.hpp
```diff
@@ -0,0 +1,55 @@
...
+/// Debugger-only helpers. Not used by production code paths.
+///
+/// Intended usage: while paused at a breakpoint in GDB/CLion, evaluate
+/// `NES::Debug::view(plan)` in the "Threads & Variables" expression box or the
+/// GDB console to get a shallow, expandable operator tree.
+///
+/// The *View structs are deliberately tiny (one label string + children) so
+/// the debugger renders them as a compact tree: operator -> children -> ...,
+/// without the self/get()/impl indirection of the real operator objects.
+
```
keyseven123: Remove line

nes-logical-operators/src/Debug/DebugHelpers.cpp
```diff
@@ -0,0 +1,19 @@
...
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <Debug/DebugHelpers.hpp>
+
+/// Definitions intentionally live in src/Plans/LogicalPlan.cpp so the linker
+/// always keeps the symbols (LogicalPlan.cpp is referenced by production code,
+/// this TU is not).
```
keyseven123: Remove this file

nes-logical-operators/tests/LogicalPlanTest.cpp
```diff
@@ -237,3 +239,44 @@ TEST_F(LogicalPlanTest, OutputOperator)
     ss << plan;
     EXPECT_FALSE(ss.str().empty());
 }
+
```
keyseven123: Remove the newly added test cases

keyseven123 (changes requested):

keyseven123 (approved):

Artraxon (approved):

PR #1675, opened 2026-06-11
  refactor(worker): queries are now directly started instead of requiring prior registration

greenBene:

Previously, submitting a query required two separate gRPC calls: registerQuery (compile and store), followed by startQuery (move to execution). Since the two were always called in immediate succession, this PR collapses them into one step.

`QueryStatus::Registered` stays the initial state of a query and is updated when the query successfully starts or fails to start. 

**IMPORTANT** We had internal discussions about this PR. The consensus agreed to add it because it simplifies a distinction that is nowhere actually used. Nevertheless, if there are major objections about this, please feel free to mention and argue them in the comments of the PR.  

## Verifying this change

All existing tests continue to run successfully

- Worker (gRPC interface)
- NodeEngine
- Frontend (QueryManager, StatementHandler)

## Documentation
- Inline documentation was updated

### DIFF
--- a/grpc/SingleNodeWorkerRPCService.proto
+++ b/grpc/SingleNodeWorkerRPCService.proto
@@ -18,24 +18,18 @@ import "SerializableQueryPlan.proto";
 import "google/protobuf/empty.proto";
 
 service WorkerRPCService {
-  rpc RegisterQuery (RegisterQueryRequest) returns (RegisterQueryReply) {}
-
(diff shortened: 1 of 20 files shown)


ls-1801 (approved):
I am in favor of moving functionality out of the worker... If we ever need this behaivor i think the registration logic can be implemented on the coordinator. 

If we need functionality to verify that a query is runnable on a worker, we could add a concrete endpoint for that. Currently i think the model of registration first is just causing confusion.

yschroeder97 (approved):

PR #1677, opened 2026-06-14
Implement Median as Nautilus Function

PhilippGrulich:

The compilation of physical aggregation functions is expensive because each aggregation's `lift`/`combine`/`lower` body is traced and **inlined** into the build/probe pipeline functions. For median this is the worst case: its `lower` is a doubly-nested loop, and a query with many `MEDIAN` aggregations inlines that loop once per aggregation, blowing up the traced pipeline and slowing compilation.

This PR extracts the median bodies into dedicated **nautilus functions** (`NautilusFunction`, i.e. cross-calls compiled once and called from the pipeline) so they are no longer inlined. Since nautilus deduplicates functions by name, naming them purely by data type means all medians of the same type in a pipeline share a single compiled instantiation.

Change log:
- Added a global, statically-allocated templated table `MedianFunctionTable<InputT, ResultT, Nullable>` holding the four extracted nautilus functions (`lift`, `combine`, `lower`, `reset`) with clear, type-derived names (e.g. `Median_lower_INT64_FLOAT64_X`). `combine`/`reset` depend only on nullability and are named/shared accordingly.
- The bodies are free, type-parametric function templates (no per-instance capture) so they can be global statics shared across all medians of the same type.
- The virtual `lift`/`combine`/`lower`/`reset` are now thin wrappers that keep the column-specific work in the main pipeline trace (compute the input value via `inputFunction` into a scratch slot; read the lowered result from a scratch slot before writing the result field) and forward to the selected functions via a `MedianFunctions` pointer struct.
- The paged vector now stores **only the single aggregated value** per row (a one-column buffer derived from the input type) instead of the whole input record, so `lower` reads that value directly rather than re-applying `inputFunction`. `getSizeOfStateInBytes()` is unchanged, so the hash-map layout and the lowering rule are untouched.
- Scope is **median only**; the same pattern can be applied to Sum/Min/Max/Count/Avg later. The change is self-contained in `MedianAggregationPhysicalFunction.hpp/.cpp`.

## Verifying this change

This change should be covered by the existing aggregation systests (no behavior change intended):
- `nes-systests/operator/aggregation/WindowAggregationNull.test`, `WindowAggregationDifferentDataTypes*.test`, `WindowAggregationStackedAggregations.test`, `ExpressionInsideAggregation.test`, `ColumnNames.test`, `AggregationOfUnion.test`
- `nes-systests/differential/AggregationEquiValence.test`, `nes-systests/guide/guide8.test`

The intended win can be confirmed by compiling a query with many `MEDIAN`s and checking that the probe module emits a single `Median_lower_*` per distinct type instead of one inlined copy per aggregation.

- ExecutionEngine (physical aggregation functions / nautilus code generation)
- QueryCompiler (only indirectly; the windowed-aggregation lowering rule is unchanged)

## Documentation

- The change is reflected in in-code documentation (class and helper comments in `MedianAggregationPhysicalFunction`).

### DIFF
--- a/nes-physical-operators/include/Aggregation/Function/MedianAggregationPhysicalFunction.hpp
+++ b/nes-physical-operators/include/Aggregation/Function/MedianAggregationPhysicalFunction.hpp
@@ -15,6 +15,8 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdint>
+#include <functional>
 #include <memory>
 #include <Aggregation/Function/AggregationPhysicalFunction.hpp>
 #include <DataTypes/DataType.hpp>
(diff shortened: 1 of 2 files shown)


keyseven123 (changes requested):
I really like this PR as it helps in a longer standing problem of spending lot of time in the compilation. I have taken the liberty to rebase it onto main and also implemented it in a more general way so that we can use/register nautilus function across all execution operators and not only in the median. See the branch `nls/agg_functions`. I can either open a pr, you can take a look at the changes or just let me know how you would like to proceed. 
On the branch, I have also removed `CompilationIntensive`labels from our systests @alepping . With this fix here, we go from 40s to 12s in the total query runtime of `WindowAggregationDifferentDataTypes`for example.

keyseven123:
After discussion with @PhilippGrulich , we close this pr and have opened #1805 instead.

PR #1683, opened 2026-06-16
Feature: Liveness reserve in BufferManager to avoid buffer-exhaustion deadlock

zeuchste:
## Problem

The global buffer pool is fixed-size. When it is fully drained into in-flight buffers and queues, a producer blocks in `getBufferBlocking()` while the only path that frees buffers (the drain/emit path) also needs a buffer it cannot get — a **buffer-exhaustion deadlock**.

## Approach (Milestone 0 — liveness, no disk)

Introduce an opt-in **liveness reserve** in `BufferManager`: a configurable number of buffers held back in a separate queue and only handed out as a last resort.

- **`AbstractBufferProvider`**: defaulted `tryGetReservedBuffer()` / `getReservedBufferWithTimeout()` that delegate to the normal API (`BufferManager` is the only override).
- **`BufferManager`**: dedicated `reservedBuffers` queue; the normal acquisition path can never consume the reserve; recycled buffers top up the reserve first; both queues accounted in `getNumberOfAvailableBuffers()`/`destroy()`.
- **`DefaultPEC::allocateTupleBuffer()`** (the per-pipeline output allocation): tries the normal pool first, then falls back to the reserve only when exhausted — so an in-flight task can always run to completion and free its input buffer, guaranteeing forward progress.
- **Config**: `number_of_reserved_buffers_in_global_buffer_manager` (default **64** — comfortably exceeds the default worker-thread count, negligible against the 32768 pool; `0` disables).

With reserve `0`, behaviour is identical to the previous `getBufferBlocking()`.

## Tests

- New `BufferManagerReserveTest` (6 cases): reserve seeding, normal-can't-drain-reserve, reserved-can-use-reserve, recycle-tops-up-reserve-first, zero-reserve-unchanged, clamp-to-pool-size.
- `query-engine-test`: 29/29 pass.
- End-to-end systests with the reserve active: arithmetic, multi-stream join, keyed windowed aggregation — all pass.
- clang-format-19 + clang-tidy clean on changed lines.

## Scope / follow-ups

This is the liveness half. Disk-backed **slice-level state spilling** (capacity / larger-than-memory state) is the planned Milestone 1 follow-up.

### DIFF
--- a/nes-memory/BufferManager.cpp
+++ b/nes-memory/BufferManager.cpp
@@ -44,8 +44,13 @@ BufferManager::BufferManager(
     const uint32_t bufferSize,
     const uint32_t numOfBuffers,
     std::shared_ptr<std::pmr::memory_resource> memoryResource,
-    const uint32_t withAlignment)
+    const uint32_t withAlignment,
+    const uint32_t numOfReservedBuffers)
     : availableBuffers(numOfBuffers)
+    /// Both queues are sized to the full pool so a recycled segment can always be routed to either one
+    /// (writeIfNotFull never fails). The reserve target is clamped to the pool size.
+    , reservedBuffers(numOfBuffers)
+    , numOfReservedBuffers(std::min<size_t>(numOfReservedBuffers, numOfBuffers))
     , unpooledChunksManager(std::make_shared<UnpooledChunksManager>(memoryResource))
     , bufferSize(bufferSize)
     , numOfBuffers(numOfBuffers)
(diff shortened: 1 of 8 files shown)


keyseven123 (commented):
I don't think this actually breaks the deadlock and simply delays the deadlock with adding complexity. Two things:

  1. Fan-out still drains the reserve. A window/join emit grabs many output buffers but only frees one input, and those outputs only come back once a downstream task picks them. If all workers are stuck emitting, nothing refills the reserve, and we just run into the 1s timeout → BufferAllocationFailure. So we simply delay the deadlock.
  2. The sink/network paths from the description don't even use the reserve. Only DefaultPEC::allocateTupleBuffer calls the reserved API. Sinks still go through
  OutputFormatterUtil.hpp:68/86 → getBufferBlocking() on the normal pool, same for NLJOperatorHandler.cpp:75. So the path you're describing isn't actually covered.

zeuchste:
Closing as superseded by #1700, which replaces the liveness-reserve approach with buffer-exhaustion load-shedding (terminating a victim query on pool exhaustion). See #1700 for the replacement; the underlying problem and discussion continue there.

PR #1686, opened 2026-06-16
Task(#1580): vendor Corrosion into cmake/

### LINKED ISSUE #1580: [TASK] Vendor Corrosion into cmake/ folder
### Description

We currently fetch Corrosion from a fork via `FetchContent` (see `cmake/EnableRust.cmake`). To support offline builds, the dev Docker image pre-clones the repo to `$CORROSION_SRC`, and `EnableRust.cmake` probes GitHub at configure time to decide between the remote and the local copy.

We should drop this and vendor Corrosion directly into `cmake/` (e.g., `cmake/corrosion/`). The fork only carries a small patch on top of upstream Corrosion, so a vendored snapshot is easy to keep up to date and removes a moving network dependency from every fresh configure.

### Approach

1. Snapshot `/corrosion@fix/avoid-always-dirty-targets` into `cmake/corrosion/` (preserving its license + a short note on origin and the patch we carry)
2. Replace the `FetchContent_Declare(Corrosion ...)` block in `cmake/EnableRust.cmake` with `add_subdirectory(${CMAKE_SOURCE_DIR}/cmake/corrosion)` (or `FetchContent_Declare(... SOURCE_DIR ...)` if we want to keep the FetchContent surface)
3. Remove the GitHub reachability probe and the conditional `.cargo/config.toml` writing — `net.offline` handling for crates is independent of Corrosion and stays in the Docker image
4. Drop `CORROSION_SRC` plumbing from `docker/dependency/Development.dockerfile` and `hash_dependencies.sh`
5. Document how to bump the vendored copy (single script or short README in `cmake/corrosion/`)

### Relevant files

- `cmake/EnableRust.cmake` — fetch/fallback logic (lines 43–86)
- `docker/dependency/Development.dockerfile` — pre-clones Corrosion, sets `CORROSION_SRC`
- `docker/dependency/hash_dependencies.sh` — hashes the Corrosion ref for cache invalidation

### Acceptance Criteria

- [ ] Corrosion sources live in `cmake/corrosion/` with origin + patch info
- [ ] `cmake/EnableRust.cmake` no longer calls `FetchContent` against GitHub for Corrosion
- [ ] No GitHub reachability probe and no online/offline branch in `EnableRust.cmake`
- [ ] `CORROSION_SRC` plumbing removed from the dev Dockerfile and `hash_dependencies.sh`
- [ ] Documented procedure for updating the vendored Corrosion
- [ ] Configure + build + tests pass on a machine with no network access

zeuchste:
Closes #1580

Vendors Corrosion in-tree instead of fetching it from a fork via FetchContent, removing a moving network dependency (GitHub reachability) from every fresh configure.

## Changes
- **`cmake/corrosion/`** — vendored Corrosion snapshot (`.git` stripped), upstream `LICENSE` preserved. Added `cmake/corrosion/VENDORING.md` documenting the upstream repo, the fork URL + ref, the always-dirty-targets patch the fork carries, and a "how to bump" procedure.
- **`cmake/EnableRust.cmake`** — replaced the `FetchContent_Declare(Corrosion ...)` block, the GitHub reachability probe, and the online/offline conditional `.cargo/config.toml` writing with a single `add_subdirectory(${CMAKE_SOURCE_DIR}/cmake/corrosion ...)`. Cargo offline/crate handling (independent of Corrosion fetching) is left untouched.
- **`docker/dependency/Development.dockerfile`** — removed the Corrosion pre-clone and the `CORROSION_GIT_REPO`/`CORROSION_VERSION`/`CORROSION_SRC` env. Rust toolchain + cargo crate prefetch (incl. `cxxbridge-cmd`) are unchanged.
- **`docker/dependency/hash_dependencies.sh`** — no change required: the Corrosion ref was only hashed indirectly via the Dockerfile under `docker/dependency`, which no longer references a moving ref. Vendored sources are now covered by normal in-tree source hashing. (Noted as the minimal change per the issue.)

## Vendored ref
- Fork: `https://github.com/nebulastream/corrosion.git`
- Ref: `v0.6.1-always-dirty-fix` (branch `fix/avoid-always-dirty-targets`)
- Commit: `5ff20ed3b025d2170617d9844073cc7221439877`

Opened as draft — not built locally; needs CI (esp. the offline build) to verify.


ls-1801 (changes requested):
We only need the relevant files, last time I checked this boiled down to just 2 cmake files

zeuchste:
@ls-1801 trimmed as requested: the vendored tree is now just the modules our `add_subdirectory()` use needs — `Corrosion.cmake`, `CorrosionGenerator.cmake` (unconditionally `include()`d by Corrosion.cmake), `FindRust.cmake`, a 12-line trimmed `CMakeLists.txt` entry point, plus `LICENSE` and `VENDORING.md`. Upstream's tests/docs/CI/presets/install rules are gone (−259 files / −6.5k lines). It ended up as 3 cmake modules rather than 2 because `Corrosion.cmake` unconditionally includes the generator module and `find_package(Rust)` needs `FindRust.cmake`. VENDORING.md now documents exactly what is vendored and the minimal-copy bump procedure.

docker/dependency/Development.dockerfile
```diff
@@ -114,16 +114,8 @@ RUN python3 -m venv /opt/iree && \
...
-# first and only uses CORROSION_SRC when the probe fails. Tracks the nebulastream fork that
-# avoids always-dirty Rust target rebuilds.
-ENV CORROSION_GIT_REPO=https://github.com/nebulastream/corrosion.git \
-    CORROSION_VERSION=v0.6.1-always-dirty-fix \
-    CORROSION_SRC=/opt/corrosion
-RUN git clone --depth 1 --branch ${CORROSION_VERSION} \
-        ${CORROSION_GIT_REPO} ${CORROSION_SRC} && \
-    chmod -R a+rX ${CORROSION_SRC}
+# Corrosion is vendored in-tree at cmake/corrosion (see cmake/corrosion/VENDORING.md), so there is
+# no Corrosion pre-clone / CORROSION_SRC fallback to bake into the image anymore.
```
ls-1801: remove this

scripts/check_preamble.py
```diff
@@ -108,7 +108,7 @@
 
         suffix = filename.split(".")[-1]
 
-        if filename in VENDORED_FILES:
+        if filename in VENDORED_FILES or filename.startswith("cmake/corrosion/"):
```
ls-1801: why cant we add the files to vendored files?

cmake/EnableRust.cmake
```diff
@@ -40,49 +40,12 @@ if (${USING_LIBCXX})
...
-            GIT_TAG ${_NES_CORROSION_GIT_TAG}
-    )
-    file(REMOVE "${CMAKE_BINARY_DIR}/.cargo/config.toml")
-endif ()
-
-FetchContent_MakeAvailable(Corrosion)
+### Corrosion is vendored in-tree at cmake/corrosion (see cmake/corrosion/VENDORING.md) instead of
+### being fetched over the network, so fresh configures (including fully offline builds in the dev
+### Docker image) do not depend on GitHub reachability. Adding it as a subdirectory defines the
+### corrosion_import_crate() function used below.
```
ls-1801: We don't need these; instead of xyz comments. The git history already documents the old code, and commits explain why changes were made. 

It's hard to expect a developer to read these "walls" of text.

Just remove the comment

.github/workflows/clang_tidy_diff.yml
```diff
@@ -80,13 +80,13 @@ jobs:
         with:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |
-            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml  -checks='-*,readability-duplicate-include' -j $(nproc)
+            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' ':!cmake/corrosion/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml  -checks='-*,readability-duplicate-include' -j $(nproc)
```
ls-1801: This should also be adjusted in the clang-tidy-diff CMake target?

ls-1801 (changes requested):

docker/dependency/Development.dockerfile
zeuchste: Removed.

scripts/check_preamble.py
zeuchste: Done — the four corrosion cmake files are now entries in VENDORED_FILES (the path special-case is gone); verified the check passes locally.

cmake/EnableRust.cmake
zeuchste: Trimmed to a single pointer at VENDORING.md.

.github/workflows/clang_tidy_diff.yml
zeuchste: Done — the exclude now lives in scripts/tidy-diff.sh's EXCLUDES (shared by the tidy-diff target and the CI pre-check) and in the remaining inline full pass in the workflow.

ls-1801 (approved):

keyseven123 (approved):

PR #1688, opened 2026-06-16
Task(#1660): align ABS() with SQL semantics

### LINKED ISSUE #1660: [TASK] Align ABS() function with SQL semantics for signed-integer overflow and result types
### Description

The `ABS()` arithmetical function should follow well-defined SQL semantics with respect to result data types and overflow. The defining property is that the absolute value of a number must always be representable in the result type. Because the absolute value of the most-negative value of a signed integer type is **not** representable in that same type (e.g. `ABS(-128)` = `128` does not fit in `INT8`, `ABS(INT64_MIN)` does not fit in `INT64`), `ABS()` must not simply preserve the input type for signed integers.

The expected type mapping and behavior for `ABS(x)` is:

- **Floating-point input** → result keeps the same floating-point type; the magnitude (sign flipped if negative) is returned.
- **Unsigned integer input** → result keeps the same unsigned type; the value is returned unchanged (no-op).
- **Signed integer input** → the result type is widened by one rank so the absolute value of the type's most-negative value is always representable:
  - `INT8` → `INT16`
  - `INT16` → `INT32`
  - `INT32` → `INT64`
  - `INT64` → `INT64` (cannot be widened further)
  - Since `INT64` has no wider type, `ABS(INT64_MIN)` overflows and must raise a **runtime error** rather than silently wrapping.
- **Nullability** propagates from input to output (a nullable input yields a nullable output; `NULL` → `NULL`).
- Any non-arithmetic input type is rejected during schema inference.

This is the behavior described by the type mapping in `AbsoluteLogicalFunction::withInferredDataType`, and the physical execution must agree with the inferred output type.

### Acceptance Criteria

- [ ] `ABS()` schema inference produces the result types described above (signed integers widened by one rank; floats and unsigned integers keep their type; nullability propagated).
- [ ] `ABS()` of the most-negative value of a signed integer type returns the correct positive value in the widened result type (e.g. `ABS(INT8 -128)` = `INT16 128`).
- [ ] `ABS(INT64_MIN)` raises a runtime error instead of silently overflowing/wrapping.
- [ ] `ABS()` on unsigned integers is a no-op and `ABS()` on floats returns the magnitude.
- [ ] Schema inference rejects non-arithmetic input types.
- [ ] Tests cover each type-mapping case, the boundary (most-negative) values, the `INT64` overflow error, and nullability propagation.

zeuchste:
Closes #1660

Makes `ABS()` follow well-defined SQL semantics for result types and signed-integer overflow.

## Type mapping implemented (`AbsoluteLogicalFunction::withInferredDataType`)
- Signed integers are widened by one rank so the absolute value of the type's most-negative value is always representable:
  - `INT8` -> `INT16`
  - `INT16` -> `INT32`
  - `INT32` -> `INT64`
  - `INT64` -> `INT64` (cannot widen further)
- Unsigned integers and floats keep their type.
- Nullability is propagated from the input (nullable in -> nullable out; `NULL` -> `NULL`).
- Non-arithmetic input is rejected during schema inference with `DifferentFieldTypeExpected`.

## Runtime overflow
The physical function (`AbsolutePhysicalFunction`) now receives the inferred (widened) output type and casts the value into it **before** negating, so e.g. `ABS(INT8 -128)` produces `INT16 128` without overflow. Because `INT64` cannot be widened, a non-null `ABS(INT64_MIN)` raises a runtime `ArithmeticalError` (error code 3013) via `nautilus::invoke`, mirroring the existing divide-by-zero / modulo-by-zero guards in `VarVal`. The guard is skipped for NULL values (whose raw bits are arbitrary).

## Tests (`nes-systests/function/arithmetical/FunctionAbsolute.test`)
- Updated the sink schema to the widened result types and adjusted expected output.
- Boundary case: most-negative values (`ABS(INT8 -128)=128`, `ABS(INT16 -32768)=32768`, `ABS(INT32 -2147483648)=2147483648`).
- `ABS(INT64_MIN)` -> `ERROR 3013`.
- Unsigned no-op and float magnitude.
- Nullability propagation including `NULL` -> `NULL`.

## Judgment calls for review
- **No new widening helper in the DataType system.** `DataType` already has `join()` for binary widening, but no unary "next wider rank" helper. I added a small local `inferAbsoluteResultType()` in `AbsoluteLogicalFunction.cpp` rather than introducing a parallel public helper. If a shared rank/widen helper is preferred, this could be promoted to `DataType`.
- **Reused the existing `ArithmeticalError` (3013) exception** for the overflow rather than adding a new exception type.
- **Physical function now takes `outputType`** (in addition to the existing `inputType`), consistent with `Ceil`/`Floor`/`Round`/`Pow`, sourced from `logicalFunction.getDataType()` in `FunctionProvider`.
- **NULL-input test semantics**: empty inline CSV fields are treated as NULL input and expected NULL output is written as the literal `NULL`, matching `operator/projection/ProjectionNull.test`. Worth confirming the File source parses empty fields as NULL in CI.

Opened as draft — not built/tested locally; needs CI to verify. (Both changed translation units were syntax- and object-compiled in isolation against the existing build's compile commands, but the full build and systests were not run.)


ls-1801 (changes requested):
This PR needs rebasing, ideally on top of #1666 as that also changes the abs implementation and uses the prefered `select` to remove the branch

keyseven123 (changes requested):
Re-request once rebased onto current main and the branch got removed via `nautilus::select()`

zeuchste:
Rebased onto current main (which now carries #1666's `select`-based ABS). The PR is now a single commit on top of main with no control-flow branch — the implementation uses `VarVal::select` exactly as requested.

The remaining delta over main is the actual SQL-semantics fix: result-type rank-widening in `AbsoluteLogicalFunction` (INT8→INT16, INT16→INT32, INT32→INT64; INT64 stays INT64) and an INT64_MIN runtime overflow guard in the physical function (since `select` evaluates both branches and negating INT64_MIN is UB). `FunctionAbsolute.test` covers widening, the most-negative boundary values, the INT64_MIN overflow (`ERROR 3013`), and nullability propagation.

Verified locally against current main: `FunctionAbsolute.test` 4/4 and `ChainedFunctionsInWhere.test` 4/4 pass, clang-format clean. Re-requesting review.

nes-physical-operators/src/Functions/ArithmeticalFunctions/AbsolutePhysicalFunction.cpp
```diff
@@ -39,15 +43,32 @@ VarVal AbsolutePhysicalFunction::execute(const Record& record, ArenaRef& arena)
...
+    /// (e.g. ABS(INT8 -128) = INT16 128). INT64 cannot be widened, so ABS(INT64_MIN) is not representable and must raise a runtime error.
+    /// The check has to happen before the select below, because select evaluates both branches and negating INT64_MIN overflows (UB).
+    if (inputType.isType(DataType::Type::INT64))
+    {
+        const auto rawValue = value.getRawValueAs<nautilus::val<int64_t>>();
+        const auto isMostNegative = rawValue == nautilus::val<int64_t>{std::numeric_limits<int64_t>::min()};
+        /// A NULL value carries arbitrary raw bits, so only raise the overflow error for a non-null most-negative value.
+        if (isMostNegative and not value.isNull())
+        {
+            nautilus::invoke(+[] { throw ArithmeticalError("ABS() of the most-negative INT64 value overflows"); });
```
ls-1801: Sounds weird, not sure if most—negative is really a thing. Not sure what the correct term would be so up to you

ls-1801 (approved):

nes-physical-operators/src/Functions/ArithmeticalFunctions/AbsolutePhysicalFunction.cpp
zeuchste: Good point — reworded to the standard term: the comments now say "minimum value" and the error message is "ABS() of INT64_MIN overflows" (also renamed `isMostNegative` -> `isMinValue`, and aligned the same wording in AbsoluteLogicalFunction.cpp + the systest). No behavioural change.

keyseven123 (approved):

PR #1571, opened 2026-04-27
Paged Vector redesign

ThanGeo:
## Purpose of the Change and Brief Change Log
This pull request refactors the PagedVector internals and how it is used. The change list is as follows:
- PagedVector now uses a single main TupleBuffer with attached children to represent its structure instead of vector<TupleBuffer>. TupleBuffer are organized for fixed-size and varsized data using strictly a single tree of TupleBuffers starting from the main buffer. Each buffer contains a header for its metadata and access/edit methods.
- Instead of using TupleBufferRef, a new TupleLayout class is introduced to more flexibly store tuples in general purpose tuple buffers. 
- Added a property testing unit test framework for testing paged vector's new logic.
- Added a `getChildRef()` method to the TupleBuffer that materializes a child buffer and attaches it for persistence to the parent TupleBuffer and cross C++/Nautilus references.

## What components does this pull request potentially affect?
- It adds a rapidscheck dependency
- Data Structures, TupleBuffer
- Some compilation stuff

## Documentation
- I will add some figures in the discussion

## Issue Closed by this pull request:

This PR closes #1448
and partly #1263

PR #1699, opened 2026-06-17
Feature: Disk-backed state spilling for windowed aggregation and joins

zeuchste:
## Problem

Stateful streaming operators (windowed aggregation, hash join, nested-loop join) accumulate slice state in memory. With enough keys/windows this state exceeds RAM, and there is no way to run such queries. This PR lets the engine spill cold operator state to disk and reload it on access, so queries can run with state larger than a configured memory budget.

This is the **capacity** half of the spilling effort and is **independent of the liveness reserve (the deadlock fix), which ships separately**. Default-off, so it is behaviour-neutral until enabled.

## Design

A slice is the unit of spilling. When spilling is enabled, each slice allocates all of its state from its own **arena-backed `BufferManager`** (`ArenaMemoryResource`: an anonymous mmap region; evict = `pwrite` + `madvise(DONTNEED)`, reload = `pread` to the *same* virtual address, so live pointers into the state survive). A process-wide **`SpillManager`** governor tracks resident bytes and, above a high-watermark, evicts the coldest unpinned slices down to a low-watermark.

Eviction-vs-access safety: spilling requires the no-op slice cache (`enable_slice_cache=false`), so every slice access goes through the C++ cache-miss proxy, which reload+pins the slice a worker is about to touch (one pinned slice per worker; eviction skips pinned slices). Triggered slices are pinned until the probe consumes them. A setup-time precondition enforces the cache requirement, so it cannot be misconfigured into corruption.

## What's included

- `ArenaMemoryResource` + `SpillManager` (in `nes-memory`), with unit tests (15 cases): evict/reload pointer & data stability, byte accounting, coldest-first eviction, pinned-not-evicted.
- `ChainedHashMap` / `PagedVector` gain an optional pinned buffer provider (null ⇒ unchanged).
- Slices made spillable: `AggregationSlice`/`HJSlice` (via `HashMapSlice`) and `NLJSlice`.
- Governor threaded `NodeEngine` → `QueryEngine` → `PipelineExecutionContext` → operator handlers → slices.
- New worker config: `enable_state_spilling`, `state_memory_budget_in_bytes`, `spill_directory` (all default off/empty).

## Validation

With `enable_state_spilling=true` and a deliberately tiny 16 KB budget forcing continuous evict→disk→reload, these produce **byte-identical results to the in-memory baseline**:
- keyed / stacked / expression windowed aggregation
- multi-stream join under `HASH_JOIN` and `NESTED_LOOP_JOIN`

Spilling-off regressions and the spill unit tests pass. Example configs: `nes-systests/configs/spill-enabled.yaml`, `spill-enabled-nlj.yaml`.

## Notes / follow-ups

- When spilling is on, the no-op-cache requirement costs one JIT→C++ call per record (the documented tradeoff); a residency-token cache fast-path could remove it later.
- clang-tidy: clean on changed lines (flagged checks are disabled in the repo `.clang-tidy`).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,6 @@ scripts/__pycache__
 .claude/
 .codex/
 [Cc][Ll][Aa][Uu][Dd][Ee].md
+
+# build directories
+/build*
--- a/nes-common/include/ExceptionDefinitions.inc
+++ b/nes-common/include/ExceptionDefinitions.inc
@@ -83,6 +83,7 @@ EXCEPTION(CannotAllocateBuffer, 3003, "cannot allocate buffer")
 EXCEPTION(SkippingDelayedTaskDuringShutdown, 3004, "skipping delayed task during shutdown")
 EXCEPTION(TuplesTooLargeForPipelineBufferSize, 3005, "tuples too large for pipeline buffer size")
 EXCEPTION(InferenceRuntimeFailure, 3006, "inference runtime failure")
+EXCEPTION(SpillIoFailure, 3007, "spill I/O failure")
 EXCEPTION(ArithmeticalError, 3013, "arithmetical error")
 EXCEPTION(WindowingError, 3014, "windowing error")
 
--- a/nes-executable/include/PipelineExecutionContext.hpp
+++ b/nes-executable/include/PipelineExecutionContext.hpp
@@ -25,6 +25,8 @@
 
 namespace NES
 {
+class SpillManager;
+
 class PipelineExecutionContext
 {
 public:
@@ -57,6 +59,11 @@ public:
     [[nodiscard]] virtual WorkerThreadId getWorkerThreadId() const = 0;
     [[nodiscard]] virtual uint64_t getNumberOfWorkerThreads() const = 0;
     [[nodiscard]] virtual std::shared_ptr<AbstractBufferProvider> getBufferManager() const = 0;
+
+    /// The process-wide state-spilling governor, or nullptr if state spilling is not configured. Window/aggregation
+    /// operator handlers obtain it here (in start()) to make their slices spillable.
+    [[nodiscard]] virtual std::shared_ptr<SpillManager> getSpillManager() const { return nullptr; }
+
     [[nodiscard]] virtual PipelineId getPipelineId() const = 0;
 
     /// TODO #30 Remove OperatorHandler from the pipeline execution context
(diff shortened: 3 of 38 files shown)


zeuchste:
## Re-based onto the redesigned (TupleBuffer-native) PagedVector

`main` merged #1571 ("Paged Vector redesign"), which rewrote `PagedVector` into a **transient, single-`TupleBuffer`** structure (created via `PagedVector::init`/`load`; pages are child buffers). The old spilling seam — `PagedVector::setOwnBufferProvider` on a *persistent* PagedVector — no longer fits, since there is no persistent PagedVector object anymore. I re-based the NLJ spilling integration onto the new design:

- **PagedVector**: taken from #1571 unchanged. `setOwnBufferProvider`/`moveAllPages` removed from the PV/NLJ path (main's `movePagesFrom` already replaces `moveAllPages`). The `setOwnBufferProvider` API still exists for the **HashMap** path, which is untouched (ChainedHashMap isn't redesigned on main yet).
- **NLJSlice**: now allocates **all** its PagedVector main buffers from its own per-slice arena-backed `BufferManager` (`effectiveProvider`) when spilling is enabled. Keeps main's `vector<TupleBuffer>` model + `movePagesFrom`. Implements `SpillableState` (arena `evict`/`reload`/`liveBytes`), reload-before-release in the destructor.
- **Page allocation seam**: because pages are allocated during inserts via Nautilus (`PagedVectorRef::pushBack(record, provider)`), `NLJBuildPhysicalOperator` now routes the **slice's arena provider** (via the already-pinned build slice, exposed by the new `WindowBasedOperatorHandler::getPinnedBuildSlice`) into `pushBack` — so main buffers **and** pages live in the same arena, i.e. the whole slice is a single spill unit. Behaviour is unchanged when spilling is off (proxy returns the pipeline provider).
- HashMap / HJ / Aggregation spilling, the `ArenaMemoryResource`/`SpillManager` infra, config, and the `pin`/`unpin`/`pinSliceForBuild` protocol merged cleanly (no conflicts) — reused as-is.

⚠️ **Not built locally** (dev image lacks the Rust/vcpkg nautilus dependency), so this needs CI + a review of the NLJ build-operator provider routing and the arena reload-before-release ordering in `~NLJSlice`.

zeuchste:
We will do this after the memory rework is done

PR #1700, opened 2026-06-18
Feature: Terminate a victim query on buffer-pool exhaustion

zeuchste:
## Problem

All queries on a node share one fixed-size `BufferManager` pool. When it is exhausted, a worker blocks in `getBufferBlocking()` and the engine stalls/deadlocks (a producer waits for a buffer while the only thing that frees buffers also needs one).

## Approach — load-shedding instead of a reserve

When the pool is exhausted, the engine **terminates a query to free buffers** so the rest of the system recovers, and the offender fails *loudly* with a clear `QueryBufferExhausted` error (rather than every query stalling silently). This **supersedes the liveness-reserve approach in #1683** (which wastes buffers and lets an over-budget query stall forever). It is independent of the state-spilling work (#1699).

The policy lives in the query engine (which knows queries); the buffer layer only reports availability.

## Mechanism

- **`BufferExhaustionArbiter`** in `DefaultPEC::allocateTupleBuffer`: hands out a buffer while more than a **recovery margin** of buffers remain free (the "don't take the last buffer" cushion, default = worker-thread count, so all workers can detect exhaustion at once). On exhaustion it **proactively selects the offending query, terminates it, waits for its buffers to drain, then retries** — escalating to the next victim only if still exhausted, so the minimum number of queries are shed. If the caller's own query is the victim, it self-terminates (guaranteeing liveness either way). Under a deterministic policy all workers agree on one victim, so the idempotent `QueryCatalog::failQuery` terminates it exactly once.
- **Configurable policy** (`query_engine.buffer_exhaustion_policy`, default `TERMINATE_LARGEST`): `LARGEST` (most buffers held — proxied by summed `RunningQueryPlanNode::pendingTasks`), `OVER_FAIR_SHARE`, `YOUNGEST`, or `SELF`. Also `query_engine.buffer_recovery_margin` (0 ⇒ worker count).
- `QueryCatalog::selectVictim` enumerates running/starting queries (reading `RunningQueryPlan::sumPendingTasks` via an identity `AtomicState` transition); `QueryCatalog::failQuery` terminates by id (mirrors the existing `onFailure` path; teardown is allocation-free, so it nets buffers).
- `AbstractBufferProvider::getNumberOfAvailableBuffers()` added (unbounded default; `BufferManager` overrides).

## Validation

- **Regression:** `query-engine-test` 29/29 and normal-pool systests unchanged — the arbiter is behaviour-neutral when the pool is ample.
- **No hang:** a multi-stream join on a deliberately tiny pool (which previously deadlocked) completes in ~3s.
- **Clean load-shedding:** offending queries fail with `QueryBufferExhausted` ("terminated to relieve buffer-pool exhaustion") while the **majority survive (3/5)**; verified under `TERMINATE_LARGEST` and `TERMINATE_SELF`.
- Config fixture: `nes-systests/configs/tiny-pool.yaml`. format clean; clang-tidy clean on changed lines.

## Notes / follow-ups

- The included systest has ~equal queries, so `LARGEST` ≈ "kill whichever is largest at the exhaustion instant"; a lopsided workload (one big query + small ones) best showcases policy differences — can add a purpose-built test if desired.
- clang-tidy diff is the CI gate.


codecov.yaml
ls-1801, line 1: We don't do this for any component in the system, and I wouldn't start here. I would try to activly provoke this in a system test.

nes-systests/configs/tiny-pool.yaml
ls-1801, line 1: I don't think this works? You would need to create a custom topology and assign these parameters to a worker than use the ——clusterConfig I think it would be worth to add a ctest that runs benchmark systest with limit configs and has a timeout to verify that the systest does not hang and not crash. Some queries should fail with out of buffers

ls-1801 (changes requested):
1. I think we need this! But I want to do some exploring first myself.
2. we should make sure this does not effect the hot path. But for now this is fine. Especially the out of buffer handling should be treated as something very rare and unlikely.
3. I would love to move some of the logic outside the query engine (at least in a dedicated source file).

ls-1801 (changes requested):
1. Testing this is probably the trickiest part. I am just worried that we introduce race conditions, which we cannot test reliably. Maybe adding a systest flag that allows queries to fail with Error Code <OUT_OF_BUFFER>, could help here and we reduce the number of buffers available to provoke this behavior. 
2. Currently, only `allocateTupleBuffer` uses the arbiter; the execution uses an AbstractBufferProvider, which we could control. That way, we can control all buffer allocations during the pipeline execution.
3. If we choose the arbiters fail current query, we should consider taking a lock if the available buffers are empty (to prevent killing multiple queries at the same time); the lock is released after the failed query is done and has released buffers. I think this can avoid expensive locks in the happy case and avoid overkilling queries.

keyseven123 (commented):
We went through the arbiter and the new test in detail. In summary, we found several paths on which we terminate a healthy query instead of the offender, and the new test cannot distinguish correct shedding from terminating everything. Details inline.

nes-query-engine/QueryEngine.cpp
```diff
@@ -184,28 +191,119 @@ class QueryCatalog
...
+            {
+                if (auto buffer = bufferProvider->getBufferNoBlocking())
+                {
+                    return std::move(buffer.value());
+                }
+            }
+
+            /// Exhausted: select the offending query. Under a deterministic policy (e.g. LARGEST) all exhausted workers
+            /// agree on the same victim, so failQuery (which is idempotent + thread-safe) terminates it exactly once.
+            const auto victim = catalog->selectVictim(policy, currentQuery);
```
keyseven123: **Workers do not agree on a victim.** The comment states that all exhausted workers select the same victim under a deterministic policy. However, `selectLargest` ranks on `sumPendingTasks`, which changes continuously, so two workers snapshotting microseconds apart can rank A and B differently and kill both. As a result, we shed two queries where freeing one would have sufficed. This matches the overkill concern raised in the last review round. To address this, we should serialize the kill under a lock held until the victim has released its buffers.

nes-query-engine/QueryEngine.cpp
```diff
@@ -184,28 +191,119 @@ class QueryCatalog
...
+            if (!victim.has_value() || *victim == currentQuery || std::chrono::steady_clock::now() >= deadline)
+            {
+                /// We are the victim (or no other victim exists, or the backstop fired): terminate this query.
+                throw QueryBufferExhausted("query {} terminated to relieve buffer-pool exhaustion", currentQuery);
+            }
+            /// Terminate the offender, then wait for its buffers to drain before considering another kill. If the pool
+            /// recovers we take a buffer on the next outer iteration; if not, we escalate to the new largest victim.
+            catalog->failQuery(
```
keyseven123, line 472: **Teardown runs inline on the worker thread.** `failQuery` disposes the victim plan, and `~RunningQueryPlan` clears `sources`, which joins the victim's source threads (`RunningQueryPlan.cpp:314-341`). However, those source threads can themselves be blocked allocating from the exhausted pool, or blocked in `addAdmissionTaskBlocking` on a queue that only worker threads drain. As a result, with a small worker pool every worker can sit in `allocate() -> failQuery -> join`, and no buffer is ever released. The 5s deadline does not help here, since we only check it in the outer loop and not inside `failQuery`. To address this, we should hand the teardown to a separate thread.

nes-query-engine/QueryEngine.cpp
```diff
@@ -242,7 +344,7 @@ struct DefaultPEC final : PipelineExecutionContext
     TupleBuffer allocateTupleBuffer() override
```
keyseven123, line 568: **The arbiter covers one allocation site.** Only `DefaultPEC::allocateTupleBuffer` routes through the arbiter. However, `nes-nautilus/src/Arena.cpp:49,58`, `PagedVectorRef.cpp:146`, `OutputFormatterUtil.hpp:68,86`, and `SequenceShredder.cpp:52` still call `getBufferBlocking()` directly, and these are exactly the hoarding paths behind joins and windows. As a result, a windowing query that exhausts the pool through a paged vector blocks forever as before, while we kill an unrelated query that merely called `allocateTupleBuffer`. This confirms the point from the last review round, i.e. the enforcement belongs in the `AbstractBufferProvider` and not in one PEC method.

nes-query-engine/QueryEngine.cpp
```diff
@@ -242,7 +344,7 @@ struct DefaultPEC final : PipelineExecutionContext
...
     {
         PRECONDITION(!wasRepeated, "A task should terminate after repeating");
-        return bm->getBufferBlocking();
+        return arbiter->allocate(queryId);
```
keyseven123: **Lost backpressure.** The removed `bm->getBufferBlocking()` waited up to `GET_BUFFER_TIMEOUT` (1000ms, `nes-memory/TupleBufferImpl.hpp:49`) for a recycled buffer, so transient pressure resolved itself. In contrast, `arbiter->allocate()` kills on the first observation of exhaustion, since the 200ms `perVictimRecovery` wait only applies after a kill. For example, a single windowed aggregation that briefly occupies the pool is now terminated, whereas it previously continued. To address this, we should wait once for a recycled buffer before selecting any victim.

nes-query-engine/QueryEngine.cpp
```diff
@@ -991,4 +1108,126 @@ void QueryCatalog::stopQuery(QueryId id)
...
+    if (toDispose)
+    {
+        std::visit([]<typename T>(T&& plan) { T::element_type::dispose(std::forward<T>(plan)); }, std::move(toDispose).value());
+    }
+    if (didTransition)
+    {
+        const auto timestamp = std::chrono::system_clock::now();
+        exception.what() += fmt::format(" in Query {}.", id);
+        ENGINE_LOG_ERROR("Query Failed: {}", exception.what());
+        listener->logQueryFailure(id, std::move(exception), timestamp);
```
keyseven123: **Duplicate failure reporting.** `failQuery` reports the failure itself, but the victim's in-flight tasks still run `RealQueryLifeTimeListener::onFailure`, whose transition list accepts `Terminated` and whose logging at lines 999-1002 is not guarded by `didTransition`. As a result, we report the same query as failed twice, with two different messages. In particular, this breaks `expectQueryStatusEvents`, which registers `logQueryFailure(id, _, _).Times(1)`, so the new test can fail intermittently with "called more times than expected". To address this, we should guard the log and the listener call with `didTransition`.

nes-query-engine/QueryEngine.cpp
```diff
@@ -991,4 +1108,126 @@ void QueryCatalog::stopQuery(QueryId id)
...
+                auto candidate = std::move(running);
+                pending = candidate.plan->sumPendingTasks();
+                return candidate;
+            });
+        if (!eligible)
+        {
+            eligible = state->transition(
+                [&pending](Starting&& starting)
+                {
+                    auto candidate = std::move(starting);
```
keyseven123: **Stopping queries are never victims.** `gatherVictimCandidates` only accepts `Running` and `Starting`. However, `StopPipelineTask` now allocates through the arbiter with its own query id, and a stopping query can hold many buffers while flushing a large window. As a result, `selectVictim` skips the actual consumer and terminates a healthy `Running` query instead. In other words, shutting down one query kills an unrelated one. To address this, we should treat `Stopping` as a candidate, at least for the caller itself.

nes-query-engine/QueryEngine.cpp
```diff
@@ -991,4 +1108,126 @@ void QueryCatalog::stopQuery(QueryId id)
...
+    return candidates;
+}
+
+QueryId QueryCatalog::selectLargest(const std::vector<VictimCandidate>& candidates)
+{
+    QueryId bestQuery = INVALID_QUERY_ID;
+    uint64_t bestPending = 0;
+    for (const auto& candidate : candidates)
+    {
+        if (bestQuery == INVALID_QUERY_ID || candidate.pending > bestPending)
```
keyseven123: **Victim ranking.** `selectLargest` ranks candidates by `sumPendingTasks`, which counts queued tasks and not buffers held. However, a query can hold the entire pool with zero pending tasks, e.g. through operator handler state, window slices, or `allocateAndHoldPerInvocation` in the new test. In that case the comparison at line 1200 never triggers and we return the first entry of an `unordered_map`, i.e. an arbitrary query. As a result, the actual offender survives and we kill innocent queries one by one. To address this, we should track buffers held per query and rank on that.

nes-query-engine/tests/BufferExhaustionTest.cpp
```diff
@@ -0,0 +1,102 @@
...
+        ASSERT_TRUE(test.waitForQepRunning(test.queryId(i), DEFAULT_LONG_AWAIT_TIMEOUT));
+    }
+
+    /// Continuously feed both sources so the pipelines keep allocating until each query is terminated.
+    DataGenerator<> dataGenerator;
+    dataGenerator.start({test.sourceControls[sourceIds.at(0)], test.sourceControls[sourceIds.at(1)]});
+
+    for (size_t i = 0; i < numberOfQueries; ++i)
+    {
+        EXPECT_TRUE(test.waitForQepTermination(test.queryId(i), DEFAULT_LONG_AWAIT_TIMEOUT));
```
keyseven123, line 96: **The test cannot fail.** Both queries are configured as buffer hogs (line 73) and both are expected to fail (line 74), so the only assertion is that everything dies. As a result, an arbiter that unconditionally throws `QueryBufferExhausted` for every allocation passes this test unchanged. In particular, recovery, `TERMINATE_SELF`, the recovery margin, and the 5s safety deadline are never exercised. To address this, we should add an asymmetric case, i.e. one hoarding query and one well-behaved query, and assert that the well-behaved query keeps processing and stops gracefully. A second case with one large and one small holder would pin the ranking. Additionally, we should add a systest flag `--tolerate_error_code`, which allows a query to either pass or throw the given error code. Thus, we can provoke exhaustion in a systest with a reduced buffer count and still assert a deterministic outcome.

keyseven123 (changes requested):
Sorry, misclicked

codecov.yaml
zeuchste: Done — the codecov patch override is out of this PR (main's generic codecov.yaml applies); exhaustion is being provoked in tests instead, per the sibling threads.

nes-query-engine/QueryEngine.cpp
zeuchste: Fixed in 99fc60bd62: victim selection and termination are now serialized under an arbiter-level `victimMutex`. The killing worker holds it from selection until the victim's buffers drained (or a bounded per-victim recovery window lapsed), and every other exhausted worker that acquires the mutex re-checks pool availability first and retries the normal allocation path before it may select another victim. So two workers snapshotting concurrently can no longer shed two queries where one sufficed.

nes-query-engine/QueryEngine.cpp
zeuchste: Fixed in 99fc60bd62: `failQuery` now only initiates the kill (atomic state transition + one-shot failure report) and hands the blocking disposal to a dedicated `QueryPlanReaper` thread, which joins the victim's source threads off the worker path. The reaper registers itself like the terminator thread (it emits the pipeline-stop tasks that disposal produces) and is drained and joined in ~QueryEngine while the ThreadPool is still alive; after shutdown, disposal falls back inline on the (worker) caller. The allocating worker returns to its allocation loop immediately after enqueueing.

nes-query-engine/QueryEngine.cpp
zeuchste: Fixed in 99fc60bd62 (plus the small BufferManager API in c981da0f33): pipelines now get a per-query `QueryBufferProvider` through `PipelineExecutionContext::getBufferManager()`, and its `getBufferBlocking()` routes through the arbiter — so Arena, PagedVectorRef, the output formatters, SequenceShredder and every other PEC-provider call site participate in exhaustion handling, not just `allocateTupleBuffer`. Sources still draw from the global pool directly, but they poll with a bounded 25 ms `getBufferWithTimeout` retry loop and honor stop tokens, so they cannot deadlock-throw; covering them would require rewiring the NodeEngine's SourceProvider, which I kept out of scope here.

nes-query-engine/QueryEngine.cpp
zeuchste: Fixed in 99fc60bd62: before any victim is considered, `allocate()` now waits one bounded grace period (1 s, mirroring the removed `getBufferBlocking` GET_BUFFER_TIMEOUT) for a recycled buffer, so transient pressure self-resolves exactly as before. Only if the pool is still exhausted after the grace period does the (serialized) victim selection run, and after each kill the worker again waits up to 1 s for the victim's buffers before escalating.

nes-query-engine/QueryEngine.cpp
zeuchste: Fixed in 99fc60bd62: `RealQueryLifeTimeListener::onFailure` no longer performs the Terminated->Terminated transition, and the log line, `logQueryFailure` listener call and QueryFail statistic are all guarded by `didTransition`. A victim shed via `failQuery` is therefore reported exactly once; in-flight tasks that fail afterwards find the query already Terminated and stay silent, which makes `expectQueryStatusEvents`'s `logQueryFailure...Times(1)` deterministic.

nes-query-engine/QueryEngine.cpp
zeuchste: Fixed in 99fc60bd62 with the broader variant: `gatherVictimCandidates` now accepts Running, Starting and Stopping queries. I went beyond caller-only because the state machine already supports it safely — both `failQuery` and `onFailure` carry a Stopping -> Terminated::Failed transition, and disposal of a `StoppingQueryPlan` goes through the same reaper path. A stopping query flushing a large window can now be shed (including when it is the caller itself, which then self-terminates with QueryBufferExhausted) instead of killing a healthy query.

nes-query-engine/QueryEngine.cpp
zeuchste: Fixed in 99fc60bd62 + c981da0f33: each per-query `QueryBufferProvider` counts outstanding pooled buffers — incremented on hand-out, decremented in the recycle callback, which flows through the provider because buffers are prepared with it as their `BufferRecycler` (new `getBufferNoBlockingFor`/`getBufferWithTimeoutFor` on BufferManager). `selectLargest` ranks candidates on a snapshot of these counts instead of `sumPendingTasks`, which is removed. A query holding the whole pool with zero pending tasks is now correctly ranked first; buffers held by sources are the one thing still uncounted (sources sit outside the query engine's provider wiring).

nes-query-engine/tests/BufferExhaustionTest.cpp
zeuchste: Fixed in 1a5d760b96: added (a) the asymmetric case — one hoarding query plus one well-behaved query; the hog must fail with QueryBufferExhausted while the well-behaved query demonstrably keeps delivering buffers to its sink and then stops gracefully on EoS — and (b) the ranking case — a 56-buffer holder vs. a 16-buffer holder with single-buffer inputs for deterministic counts; the LARGE holder must be the victim and the small one must still be alive afterwards and stop gracefully. The original total-exhaustion test is kept; an arbiter that kills everything now fails both new tests.

nes-systests/configs/tiny-pool.yaml
zeuchste: Reworked in b742269b0f: tiny-pool.yaml is deleted — the hook is the systest binary's `-- --worker.*` override arguments, as you suggested. I added the `--tolerate_error_code <code>` flag (a query passes iff it succeeds or fails with exactly that code; cascading secondary errors are tolerated alongside it) and a `systest_buffer_exhaustion` ctest entry that runs the operator/join suite with `--worker.total_memory_in_bytes=${SYSTEST_JOIN_TOTAL_MEMORY_BYTES} --worker.unpooled_memory_fraction=0.999` (~196 pooled 4 KiB buffers, unpooled share kept generous so only the pooled pool is under pressure), `--tolerate_error_code 3007` and a TIMEOUT property as the hang detector.

PR #1702, opened 2026-06-18
Feature(#1701): Bound unpooled buffer memory to prevent worker OOM

### LINKED ISSUE #1701: Unpooled buffers are unbounded and can OOM-kill the worker
> **Epic:** #1714 · **Subsumed by:** #1712

## Problem

The pooled buffer pool is fixed-size and validated against physical RAM at startup, but **unpooled** buffers (`BufferManager::getUnpooledBuffer` → `UnpooledChunksManager`) are allocated on-the-fly from the heap with **no global cap**.

Operator state that uses unpooled buffers — `ChainedHashMap` (entry/storage/var-sized pages), `PagedVector` pages, `Arena` var-sized data, network deserialization child buffers — grows with cardinality / window count. A large group-by or join can therefore exhaust physical RAM and the OS **OOM-kills the entire worker process** (all queries die at once, with no error).

There is also a latent crash: when the underlying allocation returns `nullptr`, `getUnpooledBuffer` builds a `TupleBuffer` over a null payload (segfault/UB) instead of failing cleanly.

## Proposed fix

Bound total unpooled memory to a budget (auto: a fraction of physical RAM minus the pooled-pool footprint; overridable via config). On breach, `getUnpooledBuffer` returns `std::nullopt` so the requesting query fails cleanly with `CannotAllocateBuffer`/`BufferAllocationFailure` (existing caller handling) instead of OOM-killing the worker.

zeuchste:
## Context

Closes #1701.

The pooled buffer pool is fixed-size and validated against physical RAM at startup, but **unpooled** buffers (`BufferManager::getUnpooledBuffer` → `UnpooledChunksManager`) were allocated from the heap with **no global cap**. Large operator state (`ChainedHashMap`/`PagedVector` pages, `Arena` var-sized data, network child buffers) could therefore exhaust physical RAM and **OOM-kill the entire worker**. A null underlying allocation also built a `TupleBuffer` over a null payload (segfault/UB).

## Change

- **`UnpooledChunksManager`**: tracks total unpooled bytes via a shared atomic counter; `allocateSpace` optimistically reserves and rolls back if it would breach the budget (or if the underlying allocate returns null). `getUnpooledBuffer` now returns `std::optional<TupleBuffer>`, yielding `nullopt` on breach instead of crashing. Counter is decremented when a chunk is freed.
- The clean-failure path already existed but was dead code: callers (`ChainedHashMap`, `PagedVector`, `Arena`, network `TupleBufferBuilder`, aggregation trigger) already turn a `nullopt` into `CannotAllocateBuffer`/`BufferAllocationFailure` → a clean query failure. This wires it up.
- **`BufferManager`**: resolves the budget — explicit override, else **70% of physical RAM minus the pooled-pool footprint** (16 MiB floor) — and passes it to `UnpooledChunksManager`. `DEFAULT_ALIGNMENT` exposed; new trailing `unpooledMemoryLimitInBytes` param on ctor/`create()`.
- **Config**: new `unpooled_memory_limit_in_bytes` (`0` = auto) in `WorkerConfiguration`, wired through `NodeEngineBuilder`.

## Testing

- New unit test `UnpooledBufferTests.UnpooledMemoryBudgetIsEnforcedAndReleased`: budget is enforced (allocation refused), then succeeds again once held buffers are released. The two GiB-scale stress tests opt out via an unbounded budget.
- New systest `aggregation/UnpooledMemoryBudget.test`: a 1 KiB cap makes a keyed aggregation fail cleanly with `BufferAllocationFailure (3000)` instead of OOM.
- Regression: all **125** Aggregation + Join systest queries pass under the auto budget (no spurious failures).
- clang-format clean; clang-tidy shows only pre-existing findings.

### DIFF
--- a/.github/workflows/large_test.yml
+++ b/.github/workflows/large_test.yml
@@ -85,7 +85,8 @@ jobs:
               --worker.default_query_execution.operator_buffer_size=${{ inputs.buffer_size }} \
               --worker.default_query_execution.page_size=${{ inputs.page_size }} \
               --worker.default_query_execution.number_of_partitions=${{ inputs.number_of_partitions }} \
-              --worker.number_of_buffers_in_global_buffer_manager=200000 \
+              --worker.total_memory_in_bytes=76911476736 \
+              --worker.unpooled_memory_fraction=0.8935 \
               --worker.default_query_execution.slice_cache.enable_slice_cache=${{ inputs.enable_slice_cache }}
 
       - name: Upload Test Logs on Failure
@@ -120,7 +121,7 @@ jobs:
         with:
           ref: ${{ inputs.head_sha }}
       - name: Configure NebulaStream
-        run: cmake -GNinja -B build -DUSE_SANITIZER=${{ matrix.sanitizer }} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_LARGE_TESTS=ON -DExternalData_OBJECT_STORES=/test-file-cache
+        run: cmake -GNinja -B build -DUSE_SANITIZER=${{ matrix.sanitizer }} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_LARGE_TESTS=ON -DExternalData_OBJECT_STORES=/test-file-cache -DLARGE_TEST_TOTAL_MEMORY_BYTES=76911476736
       - name: Build NebulaStream
         run: cmake --build build --target systest nes-single-node-worker -j
       - name: Run systest remote test
--- a/nes-memory/BufferManager.cpp
+++ b/nes-memory/BufferManager.cpp
@@ -41,20 +41,40 @@ namespace NES
 {
 
 BufferManager::BufferManager(
-    Private, const uint32_t bufferSize, const uint32_t numOfBuffers, std::shared_ptr<std::pmr::memory_resource> memoryResource)
+    Private,
+    const uint32_t bufferSize,
+    const uint32_t numOfBuffers,
+    std::shared_ptr<std::pmr::memory_resource> memoryResource,
+    const size_t unpooledMemoryLimitInBytes,
+    const uint32_t alignment)
     : availableBuffers(numOfBuffers)
-    , unpooledChunksManager(std::make_shared<UnpooledChunksManager>(memoryResource))
+    , unpooledChunksManager(std::make_shared<UnpooledChunksManager>(memoryResource, unpooledMemoryLimitInBytes))
     , bufferSize(bufferSize)
     , numOfBuffers(numOfBuffers)
+    , alignment(alignment)
     , memoryResource(std::move(memoryResource))
 {
+    PRECONDITION(numOfBuffers > 0, "BufferManager requires at least one pooled buffer, but the configured budget yields {}", numOfBuffers);
     initialize();
 }
 
 std::shared_ptr<BufferManager>
-BufferManager::create(uint32_t bufferSize, uint32_t numOfBuffers, const std::shared_ptr<std::pmr::memory_resource>& memoryResource)
+BufferManager::create(const size_t totalMemoryInBytes,
+    const double unpooledMemoryFraction,
+    const uint32_t alignment,
+    const uint32_t bufferSize,
+    const std::shared_ptr<std::pmr::memory_resource>& memoryResource)
 {
-    return std::make_shared<BufferManager>(Private{}, bufferSize, numOfBuffers, memoryResource);
+    PRECONDITION(
+        unpooledMemoryFraction >= 0.0 and unpooledMemoryFraction <= 1.0,
+        "unpooledMemoryFraction must be in [0.0, 1.0] but is {}",
+        unpooledMemoryFraction);
+    PRECONDITION(bufferSize > 0, "bufferSize must be greater than zero");
+
+    const auto unpooledMemoryLimitInBytes = static_cast<size_t>(static_cast<double>(totalMemoryInBytes) * unpooledMemoryFraction);
+    const size_t pooledMemoryInBytes = totalMemoryInBytes - unpooledMemoryLimitInBytes;
+    const auto numOfBuffers = static_cast<uint32_t>(pooledMemoryInBytes / bufferSize);
+    return std::make_shared<BufferManager>(Private{}, bufferSize, numOfBuffers, memoryResource, unpooledMemoryLimitInBytes, alignment);
 }
 
 BufferManager::~BufferManager()
@@ -95,7 +115,7 @@ void BufferManager::destroy()
 
         availableBuffers = decltype(availableBuffers)();
         NES_DEBUG("Shutting down Buffer Manager completed");
-        memoryResource->deallocate(basePointer, allocatedAreaSize, DEFAULT_ALIGNMENT);
+        memoryResource->deallocate(basePointer, allocatedAreaSize, alignment);
         allocatedAreaSize = 0;
 
         /// Destroying the unpooled chunks
@@ -105,8 +125,8 @@ void BufferManager::destroy()
 
 void BufferManager::initialize()
 {
-    /// Buffers are always aligned to a cache line; the alignment is no longer configurable.
-    constexpr uint32_t withAlignment = DEFAULT_ALIGNMENT;
+    /// Buffer alignment is configured at construction
+    const uint32_t withAlignment = alignment;
     const size_t pages = sysconf(_SC_PHYS_PAGES);
     size_t page_size = sysconf(_SC_PAGE_SIZE);
     auto memorySizeInBytes = pages * page_size;
@@ -215,7 +235,7 @@ std::optional<TupleBuffer> BufferManager::getBufferWithTimeout(const std::chrono
 
 std::optional<TupleBuffer> BufferManager::getUnpooledBuffer(const size_t bufferSize)
 {
-    return unpooledChunksManager->getUnpooledBuffer(bufferSize, DEFAULT_ALIGNMENT, shared_from_this());
+    return unpooledChunksManager->getUnpooledBuffer(bufferSize, alignment, shared_from_this());
 }
 
 void BufferManager::recyclePooledBuffer(detail::MemorySegment* segment)
--- a/nes-memory/BufferManager.cpp
+++ b/nes-memory/BufferManager.cpp
@@ -22,6 +22,7 @@
 #include <cstdio>
 #include <cstring>
 #include <deque>
+#include <limits>
 #include <memory>
 #include <memory_resource>
 #include <mutex>
@@ -58,10 +59,10 @@ BufferManager::BufferManager(
     initialize();
 }
 
-std::shared_ptr<BufferManager>
-BufferManager::create(const size_t totalMemoryInBytes,
+std::shared_ptr<BufferManager> BufferManager::create(
+    const size_t totalMemoryInBytes,
     const double unpooledMemoryFraction,
-    const uint32_t alignment,
+    const BufferAlignment alignment,
     const uint32_t bufferSize,
     const std::shared_ptr<std::pmr::memory_resource>& memoryResource)
 {
@@ -73,8 +74,13 @@ BufferManager::create(const size_t totalMemoryInBytes,
 
     const auto unpooledMemoryLimitInBytes = static_cast<size_t>(static_cast<double>(totalMemoryInBytes) * unpooledMemoryFraction);
     const size_t pooledMemoryInBytes = totalMemoryInBytes - unpooledMemoryLimitInBytes;
+    PRECONDITION(
+        pooledMemoryInBytes / bufferSize <= std::numeric_limits<uint32_t>::max(),
+        "pooled buffer count {} exceeds uint32_t; lower total_memory_in_bytes or raise buffer_size",
+        pooledMemoryInBytes / bufferSize);
     const auto numOfBuffers = static_cast<uint32_t>(pooledMemoryInBytes / bufferSize);
-    return std::make_shared<BufferManager>(Private{}, bufferSize, numOfBuffers, memoryResource, unpooledMemoryLimitInBytes, alignment);
+    return std::make_shared<BufferManager>(
+        Private{}, bufferSize, numOfBuffers, memoryResource, unpooledMemoryLimitInBytes, alignment.getRawValue());
 }
 
 BufferManager::~BufferManager()
--- a/nes-memory/UnpooledChunksManager.cpp
+++ b/nes-memory/UnpooledChunksManager.cpp
@@ -16,11 +16,13 @@
 #include <Runtime/UnpooledChunksManager.hpp>
 
 #include <algorithm>
+#include <atomic>
 #include <cstddef>
 #include <cstdint>
 #include <memory>
 #include <memory_resource>
 #include <numeric>
+#include <optional>
 #include <ranges>
 #include <thread>
 #include <utility>
@@ -34,8 +36,9 @@
 
 namespace NES
 {
-UnpooledChunksManager::UnpooledChunksManager(std::shared_ptr<std::pmr::memory_resource> memoryResource)
-    : memoryResource(std::move(memoryResource))
+UnpooledChunksManager::UnpooledChunksManager(
+    std::shared_ptr<std::pmr::memory_resource> memoryResource, const size_t unpooledMemoryBudgetInBytes)
+    : memoryResource(std::move(memoryResource)), unpooledMemoryBudgetInBytes(unpooledMemoryBudgetInBytes)
 {
 }
 
@@ -114,9 +117,28 @@ UnpooledChunksManager::allocateSpace(const std::thread::id threadId, const size_
     /// For now, we allocate multiple localLastAllocateChunkKeyrolling averages. If this is too small for the current bufferSize, we allocate at least the bufferSize
     const auto newAllocationSizeExact = std::max(neededSize, newRollingAverage * NUM_PRE_ALLOCATED_CHUNKS);
     const auto newAllocationSize = (newAllocationSizeExact + 4095U) & ~4095U; /// Round to the nearest multiple of 4KB (page size)
+
+    /// Enforce the global unpooled-memory budget before touching the allocator: optimistically reserve the bytes, and
+    /// roll the reservation back if we would breach the budget. This bounds unbounded operator state (hash maps, paged
+    /// vectors, var-sized data) so a runaway query fails cleanly instead of OOM-killing the whole worker process.
+    const auto bytesInUseBefore = currentlyAllocatedUnpooledBytes->fetch_add(newAllocationSize, std::memory_order_relaxed);
+    /// Underflow-safe budget check: bytesInUseBefore + newAllocationSize can wrap size_t (and SIZE_MAX is the
+    /// unbounded sentinel), and concurrent fetch_adds can transiently push bytesInUseBefore past the budget.
+    if (bytesInUseBefore >= unpooledMemoryBudgetInBytes || newAllocationSize > unpooledMemoryBudgetInBytes - bytesInUseBefore)
+    {
+        currentlyAllocatedUnpooledBytes->fetch_sub(newAllocationSize, std::memory_order_relaxed);
+        NES_WARNING(
+            "Unpooled memory budget of {}B would be exceeded by a {}B chunk ({}B already in use); refusing allocation.",
+            unpooledMemoryBudgetInBytes,
+            newAllocationSize,
+            bytesInUseBefore);
+        return {};
+    }
+
     auto* const newlyAllocatedMemory = static_cast<uint8_t*>(memoryResource->allocate(newAllocationSize, alignment));
     if (newlyAllocatedMemory == nullptr)
     {
+        currentlyAllocatedUnpooledBytes->fetch_sub(newAllocationSize, std::memory_order_relaxed);
         NES_WARNING("Could not allocate {} bytes for unpooled chunk!", newAllocationSize);
         return {};
     }
@@ -134,7 +156,7 @@ UnpooledChunksManager::allocateSpace(const std::thread::id threadId, const size_
     return {localKeyForUnpooledBufferChunk, localMemoryForNewTupleBuffer};
 }
 
-TupleBuffer
+std::optional<TupleBuffer>
 UnpooledChunksManager::getUnpooledBuffer(const size_t neededSize, size_t alignment, const std::shared_ptr<BufferRecycler>& bufferRecycler)
 {
     const auto threadId = std::this_thread::get_id();
@@ -146,6 +168,14 @@ UnpooledChunksManager::getUnpooledBuffer(const size_t neededSize, size_t alignme
     const auto& [localKeyForUnpooledBufferChunk, localMemoryForNewTupleBuffer]
         = this->allocateSpace(threadId, alignedBufferSizePlusControlBlock, alignment);
 
+    /// allocateSpace returns a null pair when the budget is exhausted or the underlying allocation failed. Signal this
+    /// to the caller as an empty optional (callers turn it into CannotAllocateBuffer/BufferAllocationFailure) instead of
+    /// constructing a TupleBuffer over a null payload.
+    if (localMemoryForNewTupleBuffer == nullptr)
+    {
+        return std::nullopt;
+    }
+
     /// Creating a new memory segment, and adding it to the unpooledMemorySegments
     const auto alignedBufferSize = alignBufferSize(neededSize, alignment);
     const auto controlBlockSize = alignBufferSize(sizeof(detail::BufferControlBlock), alignment);
@@ -156,7 +186,8 @@ UnpooledChunksManager::getUnpooledBuffer(const size_t neededSize, size_t alignme
         [copyOfMemoryResource = this->memoryResource,
          copyOLastChunkPtr = localKeyForUnpooledBufferChunk,
          copyOfChunk = chunk,
-         copyOfAlignment = alignment](detail::MemorySegment* memorySegment, BufferRecycler*)
+         copyOfAlignment = alignment,
+         copyOfAllocatedBytes = currentlyAllocatedUnpooledBytes](detail::MemorySegment* memorySegment, BufferRecycler*)
         {
             auto lockedLocalUnpooledBufferData = copyOfChunk->wlock();
             auto& curUnpooledChunk = lockedLocalUnpooledBufferData->chunks[copyOLastChunkPtr];
@@ -175,6 +206,7 @@ UnpooledChunksManager::getUnpooledBuffer(const size_t neededSize, size_t alignme
                 lockedLocalUnpooledBufferData.unlock();
                 copyOfMemoryResource->deallocate(
                     extractedChunkControlBlock.startOfChunk, extractedChunkControlBlock.totalSize, copyOfAlignment);
+                copyOfAllocatedBytes->fetch_sub(extractedChunkControlBlock.totalSize, std::memory_order_relaxed);
             }
         });
 
--- a/nes-memory/include/Runtime/BufferManager.hpp
+++ b/nes-memory/include/Runtime/BufferManager.hpp
@@ -24,6 +24,7 @@
 #include <mutex>
 #include <optional>
 #include <vector>
+#include <Identifiers/NESStrongType.hpp>
 #include <Runtime/AbstractBufferProvider.hpp>
 #include <Runtime/BufferRecycler.hpp>
 #include <Runtime/UnpooledChunksManager.hpp>
@@ -31,6 +32,10 @@
 
 namespace NES
 {
+
+/// Strong type for create()'s alignment argument to reduce potential swapped as both are u32 and also
+/// readability-suspicious-call-argument on every call site
+using BufferAlignment = NESStrongType<uint32_t, struct BufferAlignment_, 0, 0>;
 
 /**
  * @brief The BufferManager is responsible for:
@@ -89,7 +94,7 @@ public:
     static std::shared_ptr<BufferManager> create(
         size_t totalMemoryInBytes,
         double unpooledMemoryFraction,
-        uint32_t alignment,
+        BufferAlignment alignment,
         uint32_t bufferSize,
         const std::shared_ptr<std::pmr::memory_resource>& memoryResource);
 
--- a/nes-memory/include/Runtime/BufferManager.hpp
+++ b/nes-memory/include/Runtime/BufferManager.hpp
@@ -16,6 +16,8 @@
 
 #include <atomic>
 #include <condition_variable>
+#include <cstddef>
+#include <cstdint>
 #include <deque>
 #include <memory>
 #include <memory_resource>
@@ -23,7 +25,6 @@
 #include <optional>
 #include <vector>
 #include <Runtime/AbstractBufferProvider.hpp>
-#include <Runtime/Allocator/NesDefaultMemoryAllocator.hpp>
 #include <Runtime/BufferRecycler.hpp>
 #include <Runtime/UnpooledChunksManager.hpp>
 #include <folly/MPMCQueue.h>
@@ -68,21 +69,29 @@ class BufferManager final : public std::enable_shared_from_this<BufferManager>, 
         explicit Private() = default;
     };
 
-    static constexpr auto DEFAULT_BUFFER_SIZE = 8 * 1024;
-    static constexpr auto DEFAULT_NUMBER_OF_BUFFERS = 1024;
-    static constexpr auto DEFAULT_ALIGNMENT = 64;
-
 public:
-    explicit BufferManager(Private, uint32_t bufferSize, uint32_t numOfBuffers, std::shared_ptr<std::pmr::memory_resource> memoryResource);
+    explicit BufferManager(
+        Private,
+        uint32_t bufferSize,
+        uint32_t numOfBuffers,
+        std::shared_ptr<std::pmr::memory_resource> memoryResource,
+        size_t unpooledMemoryLimitInBytes,
+        uint32_t alignment);
 
-    /// Creates a new global buffer manager. Buffers are always aligned to DEFAULT_ALIGNMENT (a cache line).
-    /// @param bufferSize the size of each buffer in bytes
-    /// @param numOfBuffers the total number of buffers in the pool
+    /// Creates a new global buffer manager from a total memory budget. The pooled buffer count and the unpooled
+    /// memory limit are derived: unpooledLimit = totalMemoryInBytes * unpooledMemoryFraction, the remaining
+    /// pooled bytes are split into floor(pooledBytes / bufferSize) buffers.
+    /// @param totalMemoryInBytes total memory budget shared by the pooled and unpooled pools
+    /// @param unpooledMemoryFraction share of the total budget reserved for unpooled buffers, in [0.0, 1.0]
+    /// @param alignment byte alignment of every buffer; must be a power of two <= page size (a cache line is 64 bytes)
+    /// @param bufferSize the size of each pooled buffer in bytes
     /// @param memoryResource resource for allocating and deallocating memory
     static std::shared_ptr<BufferManager> create(
-        uint32_t bufferSize = DEFAULT_BUFFER_SIZE,
-        uint32_t numOfBuffers = DEFAULT_NUMBER_OF_BUFFERS,
-        const std::shared_ptr<std::pmr::memory_resource>& memoryResource = std::make_shared<NesDefaultMemoryAllocator>());
+        size_t totalMemoryInBytes,
+        double unpooledMemoryFraction,
+        uint32_t alignment,
+        uint32_t bufferSize,
+        const std::shared_ptr<std::pmr::memory_resource>& memoryResource);
 
     BufferManager(const BufferManager&) = delete;
     BufferManager& operator=(const BufferManager&) = delete;
@@ -92,8 +101,8 @@ public:
 
 private:
     /**
-     * @brief Configure the BufferManager to use numOfBuffers buffers of size bufferSize bytes, aligned to
-     * DEFAULT_ALIGNMENT. This is a one shot call. A second invocation of this call will fail
+     * @brief Configure the BufferManager to use numOfBuffers buffers of size bufferSize bytes.
+     * This is a one shot call. A second invocation of this call will fail
      */
     void initialize();
 
@@ -146,6 +155,7 @@ private:
 
     size_t bufferSize;
     size_t numOfBuffers;
+    uint32_t alignment;
 
     uint8_t* basePointer{nullptr};
     size_t allocatedAreaSize;
--- a/nes-memory/include/Runtime/UnpooledChunksManager.hpp
+++ b/nes-memory/include/Runtime/UnpooledChunksManager.hpp
@@ -14,11 +14,13 @@
 
 #pragma once
 
+#include <atomic>
 #include <cstddef>
 #include <cstdint>
 #include <iosfwd>
 #include <memory>
 #include <memory_resource>
+#include <optional>
 #include <thread>
 #include <unordered_map>
 #include <utility>
@@ -42,6 +44,16 @@ class UnpooledChunksManager
 
     /// Needed for allocating and deallocating memory
     std::shared_ptr<std::pmr::memory_resource> memoryResource;
+
+    /// Hard cap on the total bytes that may be allocated for unpooled chunks. Once exceeded, getUnpooledBuffer
+    /// returns std::nullopt instead of allocating, so the requesting query fails cleanly (via the callers'
+    /// CannotAllocateBuffer/BufferAllocationFailure paths) rather than the worker running out of physical memory.
+    /// std::numeric_limits<size_t>::max() means "unbounded" (the historic behaviour, e.g. for tests).
+    size_t unpooledMemoryBudgetInBytes;
+
+    /// Total bytes currently allocated across all unpooled chunks. Held via shared_ptr so the per-segment recycle
+    /// callback can decrement it even if this manager has been destroyed in the meantime.
+    std::shared_ptr<std::atomic<size_t>> currentlyAllocatedUnpooledBytes = std::make_shared<std::atomic<size_t>>(0);
 
     /// Helper struct that stores necessary information for accessing unpooled chunks
     /// Instead of allocating the exact needed space, we allocate a chunk of a space calculated by a rolling average of the last n sizes.
@@ -85,9 +97,12 @@ class UnpooledChunksManager
     std::shared_ptr<folly::Synchronized<UnpooledChunk>> getChunk(std::thread::id threadId);
 
 public:
-    explicit UnpooledChunksManager(std::shared_ptr<std::pmr::memory_resource> memoryResource);
+    explicit UnpooledChunksManager(std::shared_ptr<std::pmr::memory_resource> memoryResource, size_t unpooledMemoryBudgetInBytes);
     size_t getNumberOfUnpooledBuffers() const;
-    TupleBuffer getUnpooledBuffer(size_t neededSize, size_t alignment, const std::shared_ptr<BufferRecycler>& bufferRecycler);
+
+    /// Returns std::nullopt if the unpooled memory budget would be exceeded or the underlying allocation fails.
+    std::optional<TupleBuffer>
+    getUnpooledBuffer(size_t neededSize, size_t alignment, const std::shared_ptr<BufferRecycler>& bufferRecycler);
 };
 
 }
--- a/nes-memory/tests/UnpooledBufferTests.cpp
+++ b/nes-memory/tests/UnpooledBufferTests.cpp
@@ -13,12 +13,15 @@
 */
 
 #include <cstddef>
+#include <cstdint>
 #include <ctime>
 #include <memory>
 #include <optional>
 #include <random>
 #include <thread>
+#include <utility>
 #include <vector>
+#include <Runtime/Allocator/NesDefaultMemoryAllocator.hpp>
 #include <Runtime/BufferManager.hpp>
 #include <Runtime/TupleBuffer.hpp>
 #include <gtest/gtest.h>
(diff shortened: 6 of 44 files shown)


nes-memory/include/Runtime/UnpooledChunksManager.hpp
```diff
@@ -43,6 +46,16 @@ class UnpooledChunksManager
...
 
+    /// Hard cap on the total bytes that may be allocated for unpooled chunks. Once exceeded, getUnpooledBuffer
+    /// returns std::nullopt instead of allocating, so the requesting query fails cleanly (via the callers'
+    /// CannotAllocateBuffer/BufferAllocationFailure paths) rather than the worker running out of physical memory.
+    /// std::numeric_limits<size_t>::max() means "unbounded" (the historic behaviour, e.g. for tests).
+    size_t unpooledMemoryBudgetInBytes;
+
+    /// Total bytes currently allocated across all unpooled chunks. Held via shared_ptr so the per-segment recycle
+    /// callback can decrement it even if this manager has been destroyed in the meantime.
+    std::shared_ptr<std::atomic<size_t>> currentlyAllocatedUnpooledBytes = std::make_shared<std::atomic<size_t>>(0);
```
ls-1801, line 56: Since the TulpeBuffer keeps the buffer manager alive: We can get away with not using a shared_ptr but a reference

ls-1801 (approved):
Just a performance improvment suggestion, but i dont mind keeping the shared_ptr until we find it to be a bottleneck

nes-memory/BufferManager.cpp
```diff
@@ -40,14 +40,40 @@
 namespace NES
 {
 
+namespace
+{
+/// Resolve the unpooled-memory budget. An explicit non-zero limit is used verbatim. Otherwise we derive a budget from
```
keyseven123:
```suggestion
/// Resolve the unpooled-memory budget. An explicit non-zero limit is used verbatim. Otherwise, we derive a budget from

```

nes-memory/BufferManager.cpp
```diff
@@ -40,14 +40,40 @@
...
+/// physical RAM: a fraction of total RAM minus the pooled-pool footprint (which is already validated to fit). This caps
+/// unbounded operator state so a runaway query fails cleanly instead of the OS OOM-killing the whole worker.
+size_t resolveUnpooledMemoryLimit(const size_t pooledFootprintBytes, const size_t explicitLimitInBytes)
+{
+    if (explicitLimitInBytes > 0)
+    {
+        return explicitLimitInBytes;
+    }
+    constexpr double unpooledBudgetFractionOfRam = 0.7;
+    const size_t physicalMemoryInBytes = static_cast<size_t>(sysconf(_SC_PHYS_PAGES)) * static_cast<size_t>(sysconf(_SC_PAGE_SIZE));
```
keyseven123: In a docker container, this would read the host memory resulting in failures. Add a comment for this so that people do not spend too much time chasing down this bug.

nes-memory/BufferManager.cpp
```diff
@@ -40,14 +40,40 @@
...
+/// Resolve the unpooled-memory budget. An explicit non-zero limit is used verbatim. Otherwise we derive a budget from
+/// physical RAM: a fraction of total RAM minus the pooled-pool footprint (which is already validated to fit). This caps
+/// unbounded operator state so a runaway query fails cleanly instead of the OS OOM-killing the whole worker.
+size_t resolveUnpooledMemoryLimit(const size_t pooledFootprintBytes, const size_t explicitLimitInBytes)
+{
+    if (explicitLimitInBytes > 0)
+    {
+        return explicitLimitInBytes;
+    }
+    constexpr double unpooledBudgetFractionOfRam = 0.7;
```
keyseven123: I think we should have this fraction be part of the buffer manager constructor. There is no clamp and a user has now a lot of different configs. See also my high-level comment for this.

keyseven123 (changes requested):
Not part of your code changes, but could you also get rid of either the exception `BufferAllocationFailure` or `CannotAllocateBuffer`? We currently have exceptions that are the same.

Also I think we should have onyl three configs:
1. Total Size for NebulaStream
2. Fraction for unpooled
3. Fixed buffer size

I have implemented it on this branch as a separate commit. Please feel free to change the commit to your liking.

zeuchste:
@keyseven123 addressed your review in 1e08089:
- **Duplicate exception**: removed `CannotAllocateBuffer` (3003); all call sites now throw `BufferAllocationFailure` (3000).
- **3 configs**: the branch now uses `total_memory_in_bytes` + `unpooled_memory_fraction` + fixed buffer size. I kept budget resolution in `NodeEngineBuilder::resolveMemoryBudgets` rather than moving it into the `BufferManager` ctor (moving it would force the BM to also own pooled-count derivation + cgroup/host-memory detection — worse layering), but added the **clamp of the fraction to [0,1]** you noted was missing.
- **Docker memory**: added a warning comment on `detectAvailableMemoryBytes()` that `sysconf(_SC_PHYS_PAGES)` reports host (not container) memory.
- Comment reword applied (it moved to `resolveMemoryBudgets` after your refactor).

Two flagged for your eyes: (1) the budget resolution staying in NodeEngineBuilder vs the BM ctor; (2) `BufferManager::create`'s trailing default args couldn't be removed (C++ ordering) so they're `NOLINT`ed with rationale. Verified locally: nes-memory + downstream build clean, UnpooledBufferTests 5/5.

nes-memory/include/Runtime/UnpooledChunksManager.hpp
zeuchste: done

zeuchste: done — `memoryResource` is now a `std::pmr::memory_resource&` reference.

nes-memory/BufferManager.cpp
zeuchste: done

nes-memory/BufferManager.cpp
zeuchste: done — added a warning that sysconf reports host (not container) memory.

nes-memory/BufferManager.cpp
zeuchste: done — switched to `unpooled_memory_fraction` with a [0,1] clamp; kept resolution in NodeEngineBuilder (layering) rather than the BM ctor — flagged in my summary comment.

zeuchste:
Fixed the RelWithDebInfo smoke-test OOM (exit 137).

**Root cause:** `total_memory_in_bytes` defaulted to `0` = auto-detect (cgroup/host RAM). Each worker sizes its pooled pool by pre-allocating `total × (1 − unpooled_memory_fraction)` up front. Under `ctest -j N`, all N concurrent test processes each sized their pools against the *full machine*, collectively overcommitting → runner OOM-killed.

**Fix:** default `total_memory_in_bytes` to a fixed `447392427` bytes, which reproduces main's legacy 32768-buffer pooled pool (`32768 × DEFAULT_OPERATOR_BUFFER_SIZE / (1 − 0.7)`) and so matches main's per-process footprint. Auto-detect is still available by explicitly setting `0`; the doc comment now warns it's unsafe when many workers share one host/container. Also rebased onto current main.

zeuchste:
Reworked the memory model so the unpooled cap is **opt-in**, after the bound (a fraction of a fixed total) proved strictly tighter than main's unbounded unpooled pool and broke the heavier tests.

**Failures it resolves** (all green on main because unpooled was unbounded there):
- Large-scale Nexmark exhausted the cap → `No unpooled TupleBuffer available`.
- The slice-cache aggregation systest thrashed under the cap → ASan timeout (`Run build and test / Done` gate).

**Change:**
- `unpooled_memory_fraction == 0` (new default) ⇒ unpooled pool is **unbounded** (limit 0, which `BufferManager` already maps to `SIZE_MAX`); the full `total_memory_in_bytes` sizes the pre-allocated pooled pool.
- `unpooled_memory_fraction > 0` ⇒ hard cap at `total × fraction`, remainder sizes the pooled pool (the OOM-bounding feature, now opt-in).
- Default `total_memory_in_bytes = 134217728` = the 32768-buffer pool, so a default worker matches main exactly (fixed small pooled + unbounded on-demand unpooled), and `ctest -j64` no longer pre-allocates a host-RAM-sized pool.
- Large-scale configs use `fraction=0` + a 200000-buffer pool (mirroring main's old `number_of_buffers=200000`).

Auto-detect (`total=0`) is retained but is no longer the default, which also addresses @keyseven123's concern that it reads host memory inside a container.

Verified locally: `UnpooledMemoryBudget.test` still bounds and returns `ERROR 3000` (opt-in path), and default-config aggregation tests pass with unbounded unpooled.

nes-memory/include/Runtime/UnpooledChunksManager.hpp
zeuchste: Done — `memoryResource` is now a `std::pmr::memory_resource&` rather than a `shared_ptr`, relying on the TupleBuffer keeping the BufferManager (and thus the resource) alive.

nes-memory/BufferManager.cpp
zeuchste: Budget resolution moved out of `BufferManager` into `NodeEngineBuilder::resolveMemoryBudgets` and was reworked so the unpooled cap is opt-in (`unpooled_memory_fraction == 0` ⇒ unbounded). The doc comment is updated there to match, so this suggestion is now obsolete.

nes-memory/BufferManager.cpp
zeuchste: Addressed — auto-detect (`total_memory_in_bytes=0`) is no longer the default, so a default worker never reads host/cgroup memory. The caveat (and the warning that it's unsafe when many workers share one host) is now documented on the `total_memory_in_bytes` option in `WorkerConfiguration.hpp`.

nes-memory/BufferManager.cpp
zeuchste: The fraction is now clamped to [0,1] in `resolveMemoryBudgets`. Rather than passing the fraction into the BufferManager constructor, `resolveMemoryBudgets` (in `NodeEngineBuilder`) resolves it to a concrete pooled-buffer count + unpooled byte limit and passes those, keeping the BM constructor free of policy knobs. The opt-in default (unbounded, matching main) also reduces the surprising-config surface.

zeuchste:
Reworked so the unpooled bound is **on by default** (closes #1701), after the opt-in default left a default worker unbounded — the exact OOM scenario the issue describes.

**Root cause:** the `total + fraction` model couples the *pre-allocated* pooled pool with the *on-demand* unpooled cap. Bounding unpooled by inflating `total` also inflates the pooled pre-allocation, which is what OOM-killed `ctest -j64`.

**Fix — decouple the pools in the default (`total_memory_in_bytes == 0`) path:**
- Pooled = fixed **32768 buffers** (pre-allocated) → N workers sharing a host/container don't OOM at startup.
- Unpooled = capped at the detected **cgroup limit (else physical RAM)** minus the pooled pool. On demand, so it costs nothing until used, but it bounds runaway operator state → **a default worker can no longer OOM-kill itself (closes #1701)**.
- An explicit `total_memory_in_bytes > 0` keeps the fraction split unchanged, so every existing config — including `UnpooledMemoryBudget.test` — behaves exactly as before.

Verified locally: `UnpooledMemoryBudget.test` still returns `ERROR 3000` (the opt-in bound), and default-config aggregations pass with the bounded cap. clang-format clean.

nes-runtime/interface/Configuration/WorkerConfiguration.hpp
```diff
@@ -42,13 +43,29 @@ class WorkerConfiguration final : public BaseConfiguration
...
+        = {"total_memory_in_bytes",
+           "0",
+           "Total worker buffer memory in bytes (0 = auto: fixed pooled pool + unpooled bounded by cgroup/physical memory).",
            {std::make_shared<NumberValidation>()}};
 
+    /// Only used when totalMemoryInBytes > 0: share of it reserved as a hard cap on unpooled (variable-sized) operator
+    /// state (hash maps, paged vectors, var-sized data); the remainder sizes the pre-allocated pooled pool. A positive
+    /// fraction bounds the unpooled pool, so on breach the requesting query fails cleanly (BufferAllocationFailure)
+    /// instead of the worker OOM-ing; 0 leaves the unpooled pool unbounded (e.g. for large-scale stress configs).
+    FloatOption unpooledMemoryFraction
```
ls-1801: In general it's better to have these comments in the description if the content is for the user of the system, which in this case they are

ls-1801 (changes requested):
I have to say the whole fraction of available memory is kinda brittle at this point. I would suggest using a fixed budget which we have to adjust if test run oom. I think this is also a good way to indicate that something is using more memory than expected.

.github/workflows/large_test.yml
```diff
@@ -85,7 +85,8 @@ jobs:
               --worker.default_query_execution.operator_buffer_size=${{ inputs.buffer_size }} \
               --worker.default_query_execution.page_size=${{ inputs.page_size }} \
               --worker.default_query_execution.number_of_partitions=${{ inputs.number_of_partitions }} \
-              --worker.number_of_buffers_in_global_buffer_manager=200000 \
+              --worker.total_memory_in_bytes=$((200000 * ${{ inputs.buffer_size }})) \
+              --worker.unpooled_memory_fraction=0 \
```
keyseven123: I do not like that we use 0 for unlimited. We can now not set the unpooled fraction to 0 anymore. I think we should use in `inf` instead

nes-common/include/ExceptionDefinitions.inc
```diff
@@ -79,7 +79,7 @@ EXCEPTION(DistributedFailure, 2303, "distributed failure")
 EXCEPTION(BufferAllocationFailure, 3000, "buffer allocation failure")
 EXCEPTION(QueryNotRegistered, 3001, "query not registered")
 EXCEPTION(CannotFormatMalformedStringValue, 3002, "cannot format malformed string value")
-EXCEPTION(CannotAllocateBuffer, 3003, "cannot allocate buffer")
+/// 3003 was CannotAllocateBuffer; merged into BufferAllocationFailure (3000). Code retired to avoid renumbering.
```
keyseven123: Please do not keep these comments into the code. LLMs like to write these types of comments but they do not have any benefit

nes-memory/include/Runtime/BufferManager.hpp
```diff
@@ -70,26 +70,34 @@ class BufferManager final : public std::enable_shared_from_this<BufferManager>,
 
     static constexpr auto DEFAULT_BUFFER_SIZE = 8 * 1024;
     static constexpr auto DEFAULT_NUMBER_OF_BUFFERS = 1024;
-    static constexpr auto DEFAULT_ALIGNMENT = 64;
 
 public:
+    static constexpr auto DEFAULT_ALIGNMENT = 64;
```
keyseven123: I think this member and also the other two default members can be removed. We already have a default value for the buffer size and number of buffers in the worker configuration.

nes-memory/tests/UnpooledBufferTests.cpp
```diff
@@ -65,8 +66,11 @@ void runAllocations(
     const size_t maxAllocationSize,
     const size_t numberOfThreads)
 {
-    /// Creating component under test. We do not care about the bufferSize and the number of buffers, as we test getUnpooledBuffer()
-    const auto bufferManager = BufferManager::create(1, 1);
+    /// Creating component under test. We do not care about the bufferSize and the number of buffers, as we test getUnpooledBuffer().
+    /// These stress tests deliberately allocate tens of GiB of (mostly untouched) unpooled memory to exercise chunk
+    /// management, so we opt out of the unpooled budget here (the budget itself is covered by a dedicated test below).
```
keyseven123: Opt out or is not better to state that we set the unpooled buffer fraction here to infinity?

nes-memory/BufferManager.cpp
```diff
@@ -45,9 +46,14 @@ BufferManager::BufferManager(
...
-    const uint32_t withAlignment)
+    const uint32_t withAlignment,
+    const size_t unpooledMemoryLimitInBytes)
     : availableBuffers(numOfBuffers)
-    , unpooledChunksManager(std::make_shared<UnpooledChunksManager>(memoryResource))
+    /// 0 means "unbounded" at this layer; the worker derives a concrete budget in NodeEngineBuilder and passes it here.
+    /// The UnpooledChunksManager holds the memory_resource by reference; this BufferManager owns the shared_ptr (member
+    /// `memoryResource` below) and outlives the manager, so the reference stays valid for the manager's whole lifetime.
+    , unpooledChunksManager(std::make_shared<UnpooledChunksManager>(
+          *memoryResource, unpooledMemoryLimitInBytes == 0 ? std::numeric_limits<size_t>::max() : unpooledMemoryLimitInBytes))
```
keyseven123: I do not like using 0 for unlimited. I think we should always pass a unpooled memory limits

nes-memory/BufferManager.cpp
```diff
@@ -57,9 +63,13 @@ BufferManager::BufferManager(
 }
 
 std::shared_ptr<BufferManager> BufferManager::create(
-    uint32_t bufferSize, uint32_t numOfBuffers, const std::shared_ptr<std::pmr::memory_resource>& memoryResource, uint32_t withAlignment)
+    uint32_t bufferSize,
```
keyseven123: I might be nitpicking, but should we not simply pass the configs here, i.e., totalMemoryInBytes, fraction unpooled, and buffer size, and then let the conversion into number of buffers happen inside of the create or the constructor? I think this is cleaner

nes-runtime/interface/Configuration/WorkerConfiguration.hpp
keyseven123: I agree

nes-runtime/src/Runtime/NodeEngineBuilder.cpp
```diff
@@ -14,19 +14,114 @@
...
 #include <QueryEngine.hpp>
 
 namespace NES
 {
 
+namespace
+{
+/// Reads a single non-negative integer from a cgroup memory file. Returns nullopt when the file is absent (not
+/// containerized) or holds the v2 sentinel "max"/an unparseable value, i.e. "no concrete limit".
+std::optional<size_t> readMemoryLimitFromFile(const char* path)
```
keyseven123:
```suggestion
/// Reads a non-negative integer from a cgroup memory file. nullopt = file absent (not containerized) or holds
/// "max"/garbage, i.e. no concrete limit.
```

nes-runtime/src/Runtime/NodeEngineBuilder.cpp
```diff
@@ -14,19 +14,114 @@
...
+}
+
+/// Best-effort detection of the memory the OOM-killer will actually enforce: the cgroup limit when containerized
+/// (v2 first, then v1), otherwise physical RAM. A cgroup limit is only honoured when it is below physical RAM, since
+/// an unset limit is reported as a huge sentinel rather than "max" on some kernels.
+/// WARNING: in a Docker (or otherwise containerized) deployment sysconf(_SC_PHYS_PAGES) reports the *host* physical
+/// memory, not the container's limit. If the cgroup files are unreadable/absent we therefore fall back to the host
+/// size and may size the pools far larger than the container allows -- which then OOM-kills the worker. If you are
+/// chasing a phantom OOM in a container, set total_memory_in_bytes explicitly rather than trusting this auto-detection.
+size_t detectAvailableMemoryBytes()
```
keyseven123:
```suggestion
/// Memory the OOM-killer enforces: cgroup limit if containerized (v2 then v1), else physical RAM. The cgroup limit
/// is honoured only when below physical RAM, since an unset limit reads as a huge sentinel, not "max".
/// WARNING: under containerization sysconf(_SC_PHYS_PAGES) reports HOST RAM, not the container limit; if the cgroup
/// files are unreadable we oversize the pools and get OOM-killed. Chasing a phantom container OOM? Set
/// total_memory_in_bytes explicitly instead of trusting this.
```

nes-runtime/src/Runtime/NodeEngineBuilder.cpp
```diff
@@ -14,19 +14,114 @@
...
+
+struct MemoryBudgets
+{
+    uint32_t numberOfBuffers;
+    size_t unpooledLimitInBytes;
+};
+
+/// The default pre-allocated pooled pool when totalMemoryInBytes is auto (0): the legacy buffer count, kept fixed so
+/// that N workers sharing one host/container do not collectively OOM at startup.
+constexpr uint32_t DEFAULT_POOLED_BUFFERS = 32768;
```
keyseven123:
```suggestion
/// Auto-mode (totalMemoryInBytes==0) pooled pool: fixed legacy count so N workers on one host don't OOM at startup.
```

nes-runtime/src/Runtime/NodeEngineBuilder.cpp
```diff
@@ -14,19 +14,114 @@
...
+/// Resolve the pooled (pre-allocated) and unpooled (on-demand, capped) memory budgets.
+///  - totalMemoryInBytes == 0 (auto, the default): decouple the pools. The pooled pool is a fixed DEFAULT_POOLED_BUFFERS
+///    buffers, and the unpooled pool is capped at the detected memory (cgroup limit if containerized, else physical RAM)
+///    minus that pooled pool. The fixed pooled pool avoids a host-sized pre-allocation under parallelism, while the cap
+///    still bounds runaway unpooled state so it fails cleanly instead of OOM-killing the worker. unpooledFraction is
+///    ignored here.
+///  - totalMemoryInBytes > 0 (explicit): split it by unpooledFraction. unpooledFraction > 0 caps the unpooled pool at
+///    totalMemoryInBytes * unpooledFraction with the remainder sizing the pooled pool; unpooledFraction == 0 leaves the
+///    unpooled pool unbounded (0 is BufferManager's unbounded sentinel). The fraction is clamped into [0, 1].
+MemoryBudgets resolveMemoryBudgets(const size_t totalMemoryInBytes, const uint32_t bufferSize, double unpooledFraction)
```
keyseven123:
```suggestion
/// Splits memory into the pre-allocated pooled pool and the on-demand-capped unpooled pool. The two modes are
/// documented in full on WorkerConfiguration::totalMemoryInBytes: ==0 auto (fixed pooled pool, unpooled capped to
/// detected RAM minus it, fraction ignored); >0 explicit (split by unpooledFraction, 0 = unbounded unpooled).
```

keyseven123 (changes requested):
On a high-level. Some of the comments are quite verbose. I have added a suggestion

nes-runtime/src/Runtime/NodeEngineBuilder.cpp
zeuchste: Done — applied your suggestion.

nes-runtime/src/Runtime/NodeEngineBuilder.cpp
zeuchste: Done — applied your suggestion.

nes-runtime/src/Runtime/NodeEngineBuilder.cpp
zeuchste: Done — applied your suggestion.

nes-runtime/src/Runtime/NodeEngineBuilder.cpp
zeuchste: Done — applied your suggestion.

nes-common/include/ExceptionDefinitions.inc
zeuchste: Done — removed the retired-code comment.

nes-memory/tests/UnpooledBufferTests.cpp
zeuchste: Done — reworded to state we set the unpooled limit to infinity (size_t max).

nes-runtime/interface/Configuration/WorkerConfiguration.hpp
zeuchste: Done — trimmed the verbose user-facing comments; the semantics now live in the option descriptions and the PR description.

nes-memory/include/Runtime/BufferManager.hpp
zeuchste: Flagging scope: these three defaults are the default args of `BufferManager::create()`, which has 41 call sites — 20 pass only bufferSize, 23 pass bufferSize+count — all relying on the remaining defaults. Removing them means touching 40+ (mostly test) call sites. Happy to do it as a dedicated follow-up cleanup, but it's wider than this PR; OK to defer?

nes-memory/BufferManager.cpp
zeuchste: Agreed 0-for-unlimited is magic. The test path already passes `std::numeric_limits<size_t>::max()` (infinity). I held off switching the internal sentinel from 0→inf in this push because the :102 comment suggestion I applied still documents the 0 sentinel — didn't want contradictory changes in one commit. Want the inf switch in this PR or a follow-up?

.github/workflows/large_test.yml
zeuchste: Same as the BufferManager thread — happy to switch `fraction=0` to an explicit `inf` unpooled limit; flagging it as a small follow-up so it lands together with the sentinel change.

nes-memory/BufferManager.cpp
zeuchste: Reasonable — passing total/fraction/bufferSize into `create()` and resolving inside is cleaner. `resolveMemoryBudgets` currently lives in NodeEngineBuilder with its own unit test, so I'd move it as a focused follow-up rather than reshuffle it mid-PR. OK to defer?

zeuchste:
Addressed the review feedback (commit `1a391d5`):

**Done & resolved**
- keyseven123's four comment suggestions in `NodeEngineBuilder` (terser cgroup/budget docs) — applied verbatim.
- Dropped the retired-3003 tombstone comment in `ExceptionDefinitions.inc`.
- Reworded the `UnpooledBufferTests` comment to say the unpooled limit is set to infinity.
- Trimmed the verbose user-facing config comments (ls-1801) — semantics now live in the option descriptions.
- Exception dedup (`CannotAllocateBuffer`→`BufferAllocationFailure`) and the 3-config model are already in (earlier commits + Nils's refactor).

**Flagged as follow-ups** (replied inline) — happy to do these, just want to confirm scope before touching widely:
- Remove the three `BufferManager` default members → impacts 40+ `create()` call sites (mostly tests).
- Switch the `0`-for-unlimited sentinel to explicit `inf` (the test path already uses `size_t` max).
- Move budget resolution into `create()` per keyseven123's nitpick.

@ls-1801 on the *fraction-of-available-memory is brittle* point: the auto path (`total=0`) bounds the unpooled cap to the **cgroup limit** (not host RAM) precisely so it can't exceed the container, and the pooled pool is fixed — but I take your point that auto-detection is a footgun in containers. I'm fine switching the default to a **fixed budget** we bump when tests OOM; want me to make that the default here, or keep `0=auto` available and just document it as opt-in?

nes-memory/BufferManager.cpp
keyseven123: As we are changing the conf in this PR, I think we should do it here.

nes-memory/include/Runtime/BufferManager.hpp
keyseven123: I think this should be done in this PR. We are already touching the files and adding/changing the configurations. Also, this change should not be quite complex.

.github/workflows/large_test.yml
keyseven123: See other comments. Please do it in this PR.

keyseven123 (changes requested):

nes-memory/include/Runtime/BufferManager.hpp
zeuchste: Done — removed DEFAULT_BUFFER_SIZE and DEFAULT_NUMBER_OF_BUFFERS; create() now requires bufferSize + numOfBuffers, and the 16 test call sites pass explicit values. The worker path no longer touches these at all (it goes through createFromMemoryBudget).

nes-memory/BufferManager.cpp
zeuchste: Done — the unbounded sentinel is now std::numeric_limits<size_t>::max(); BufferManager passes the limit through verbatim (no 0-magic).

.github/workflows/large_test.yml
zeuchste: Done — there is no 0-for-unlimited in code anymore: createFromMemoryBudget converts an unpooled fraction of 0 into the explicit inf sentinel (size_t max). The large-scale configs keep fraction=0 only as the knob that selects 'unbounded'; happy to add a dedicated unpooled-limit option if you'd rather set inf directly in the yaml.

nes-memory/BufferManager.cpp
zeuchste: Done — moved the pooled/unpooled split into BufferManager::createFromMemoryBudget(bufferSize, total, fraction); NodeEngineBuilder::build just calls it now.

zeuchste:
Implemented the remaining requests from both of you (commit `120980a`):

- **Fixed budget, no auto-detection** (@ls-1801): ripped out `detectAvailableMemoryBytes`/cgroup reading. `total_memory_in_bytes` is a fixed budget (default 1 GiB at fraction 0.875 = the legacy 32768-buffer pool), documented to be bumped per-test when a workload OOMs.
- **`inf` not `0`** (@keyseven123): unbounded = `numeric_limits<size_t>::max()`, passed through verbatim.
- **Conversion in BufferManager** (@keyseven123): new `createFromMemoryBudget(bufferSize, total, fraction)` does the split; `NodeEngineBuilder::build` calls it.
- **Removed the default members** (@keyseven123): `DEFAULT_BUFFER_SIZE`/`DEFAULT_NUMBER_OF_BUFFERS` gone, 16 test call sites updated.

Verified locally: `UnpooledMemoryBudget.test` → `ERROR 3000`, default-config aggregations pass, full project compiles, clang-format clean. Re-requesting review.

nes-memory/include/Runtime/BufferManager.hpp
```diff
@@ -68,26 +69,41 @@ class BufferManager final : public std::enable_shared_from_this<BufferManager>,
         explicit Private() = default;
     };
 
-    static constexpr auto DEFAULT_BUFFER_SIZE = 8 * 1024;
-    static constexpr auto DEFAULT_NUMBER_OF_BUFFERS = 1024;
+public:
     static constexpr auto DEFAULT_ALIGNMENT = 64;
```
keyseven123: I think we should remove all three defaults.

nes-runtime/interface/Configuration/WorkerConfiguration.hpp
```diff
@@ -42,13 +43,23 @@ class WorkerConfiguration final : public BaseConfiguration
...
+    /// when a workload OOMs -- a fixed budget makes "this query uses more memory than expected" visible rather than
+    /// silently sizing to the machine. Default reproduces the legacy 32768-buffer pooled pool at fraction 0.875.
+    UIntOption totalMemoryInBytes
+        = {"total_memory_in_bytes",
+           "1073741824",
+           "Total worker buffer memory in bytes; split into pooled and unpooled pools by unpooled_memory_fraction.",
            {std::make_shared<NumberValidation>()}};
 
+    /// Fraction of total_memory_in_bytes that caps unpooled operator state; the rest sizes the pre-allocated pooled
+    /// pool. 0 leaves the unpooled pool unbounded. On breach the query fails cleanly (BufferAllocationFailure).
```
keyseven123: I thought that we use inf instead of 0 for unbounded

keyseven123 (changes requested):

nes-runtime/interface/Configuration/WorkerConfiguration.hpp
```diff
@@ -49,6 +49,15 @@ class WorkerConfiguration final : public BaseConfiguration
            "Number buffers in global buffer pool.",
            {std::make_shared<NumberValidation>()}};
 
+    /// Hard cap on total unpooled (variable-sized) buffer memory in bytes, used by operator state (hash maps, paged
+    /// vectors, var-sized data). 0 = auto: a fraction of physical RAM minus the pooled pool. On breach, the requesting
```
keyseven123: I think something went wrong during rebase. At least, I think we decided on using `inf` instead of `0` for unlimited.

nes-memory/include/Runtime/BufferManager.hpp
```diff
@@ -70,26 +70,31 @@ class BufferManager final : public std::enable_shared_from_this<BufferManager>,
 
     static constexpr auto DEFAULT_BUFFER_SIZE = 8 * 1024;
     static constexpr auto DEFAULT_NUMBER_OF_BUFFERS = 1024;
-    static constexpr auto DEFAULT_ALIGNMENT = 64;
 
 public:
+    static constexpr auto DEFAULT_ALIGNMENT = 64;
```
keyseven123: Why is this not removed?

keyseven123 (changes requested):

nes-memory/include/Runtime/BufferManager.hpp
zeuchste: Done — all three default constexprs (DEFAULT_BUFFER_SIZE, DEFAULT_NUMBER_OF_BUFFERS, DEFAULT_ALIGNMENT) are removed from the public header. `create()` now requires all four arguments (no default arguments), and the cache-line alignment is an implementation-detail constant in BufferManager.cpp.

zeuchste:
@ls-1801 the large-scale systests now run with an explicit unpooled-memory cap so the #1701/#1702 bound is exercised under load (commit `f9eef6e3df`).

`nes-systests/systest/CMakeLists.txt`:
- Added `set(LARGE_TEST_UNPOOLED_MEMORY_LIMIT_BYTES 17179869184) # 16 GiB` and passed `--worker.unpooled_memory_limit_in_bytes=${LARGE_TEST_UNPOOLED_MEMORY_LIMIT_BYTES}` to both `systest_large_HASH_JOIN` and `systest_large_benchmark_HASH_JOIN` (next to `number_of_buffers_in_global_buffer_manager=2000000`).

On the value: 16 GiB is a **starting** cap chosen to let the current large HASH_JOIN workload pass while still bounding runaway operator state — I can't pin down the exact "gerade groß genug" number locally (the large test needs the runner + ~2h), so I sized it generously. Once the large-test CI runs we can tune it down to just above the actual unpooled footprint (or raise it if a query legitimately needs more). It's a single named CMake variable so adjusting is a one-line change.

keyseven123 (changes requested):
I have left my ideas as commits on the branch.

keyseven123 (approved):

nes-executable/src/BackpressureChannel.cpp
```diff
@@ -77,22 +78,23 @@ bool BackpressureController::releasePressure()
...
+    };
+
+    /// Bounded poll instead of an unbounded wait: a missed condition_variable_any/stop_token wakeup (a known
+    /// hazard on some libc++/libstdc++ versions) must not park a source forever and stall query teardown. We
+    /// re-arm on a short timeout and re-check the stop token, but still only return once the channel is OPEN/
+    /// DESTROYED or a stop was requested -- so backpressure semantics are preserved (the source does not resume
+    /// producing while the channel is CLOSED).
+    while (!channelChanged() && !stopToken.stop_requested())
+    {
+        channel->change.wait_for(state.as_lock(), stopToken, std::chrono::milliseconds{100}, channelChanged);
```
ls-1801: @nils-schubert this looks out of scope for #1702 — please revert it and open a separate issue/PR.

I am not aware of any missed-wakeup bug in the libstdc++/libc++ versions we ship against. The class of `condition_variable_any` + `stop_token` bugs (LWG 3567, GCC 104928, libc++ pre-18) is real, but fixed in the toolchains we actually use. Do you have a reproducer, a stack trace of the parked source, or a linked bug? Or is this speculation?

I do not see how a `stop_request` on a `stop_token` can be missed here — the CV wakeup on stop is part of the standard-mandated contract for the `stop_token` overload. If the concern is really a spec-level missed wakeup, note that `wait_for` uses the same overload, so only the 100 ms poll actually guards against it — the `wait` → `wait_for` swap by itself does nothing. That should be called out in the comment (or better: dropped entirely along with the change).

nes-memory/UnpooledChunksManager.cpp
```diff
@@ -114,9 +117,26 @@ UnpooledChunksManager::allocateSpace(const std::thread::id threadId, const size_
     /// For now, we allocate multiple localLastAllocateChunkKeyrolling averages. If this is too small for the current bufferSize, we allocate at least the bufferSize
     const auto newAllocationSizeExact = std::max(neededSize, newRollingAverage * NUM_PRE_ALLOCATED_CHUNKS);
     const auto newAllocationSize = (newAllocationSizeExact + 4095U) & ~4095U; /// Round to the nearest multiple of 4KB (page size)
+
+    /// Enforce the global unpooled-memory budget before touching the allocator: optimistically reserve the bytes, and
+    /// roll the reservation back if we would breach the budget. This bounds unbounded operator state (hash maps, paged
+    /// vectors, var-sized data) so a runaway query fails cleanly instead of OOM-killing the whole worker process.
+    const auto bytesInUseBefore = currentlyAllocatedUnpooledBytes->fetch_add(newAllocationSize, std::memory_order_relaxed);
+    if (bytesInUseBefore + newAllocationSize > unpooledMemoryBudgetInBytes)
```
ls-1801: `bytesInUseBefore + newAllocationSize` can wrap `size_t`. Not realistic in practice, but since `SIZE_MAX` is also the unbounded sentinel, an overflowed sum would silently pass the check and then get charged to the counter. Cheap swap: check `newAllocationSize > unpooledMemoryBudgetInBytes - bytesInUseBefore` (underflow-safe form), or skip the counter entirely when `budget == SIZE_MAX`.

nes-systests/configs/topologies/two-node-more-capacity.yaml
```diff
@@ -11,7 +11,10 @@ workers:
     max_operators: 10000
     config:
       worker:
-        number_of_buffers_in_global_buffer_manager: 2000000
+        # Fallback budget for standalone bats runs; CMake/distributed.bats overrides this
+        # (WORKER_TOTAL_MEMORY_IN_BYTES) with 32 GiB when ENABLE_LARGE_TESTS=ON, 16 GiB otherwise.
```
ls-1801: This comment is stale. `CMakeLists.txt:191` sets `REMOTE_TEST_TOTAL_MEMORY_BYTES = 76911476736` (~72 GiB, 8 GiB pooled + 64 GiB unpooled) when `ENABLE_LARGE_TESTS=ON`, not 32 GiB. The 32 GiB figure is the *unpooled* share only. Please update to "~72 GiB (8 GiB pooled + 64 GiB unpooled), 16 GiB otherwise" — same on line 29.

nes-systests/configs/topologies/two-node-more-capacity.yaml
```diff
@@ -22,7 +25,10 @@ workers:
       - "sink-node:8080"
     config:
       worker:
-        number_of_buffers_in_global_buffer_manager: 2000000
+        # Fallback budget for standalone bats runs; CMake/distributed.bats overrides this
+        # (WORKER_TOTAL_MEMORY_IN_BYTES) with 32 GiB when ENABLE_LARGE_TESTS=ON, 16 GiB otherwise.
```
ls-1801: This comment is stale. `CMakeLists.txt:191` sets `REMOTE_TEST_TOTAL_MEMORY_BYTES = 76911476736` (~72 GiB, 8 GiB pooled + 64 GiB unpooled) when `ENABLE_LARGE_TESTS=ON`, not 32 GiB. The 32 GiB figure is the *unpooled* share only. Please update to "~72 GiB (8 GiB pooled + 64 GiB unpooled), 16 GiB otherwise" — same on line 29.

nes-systests/systest/CMakeLists.txt
```diff
@@ -177,11 +184,30 @@ add_dependencies(nes-systest-lib test-data)
...
+if (ENABLE_LARGE_TESTS)
+    set(REMOTE_TEST_TOTAL_MEMORY_BYTES 76911476736) # 8 GiB pooled (2,000,000 * 4096) + 64 GiB unpooled
+else ()
+    set(REMOTE_TEST_TOTAL_MEMORY_BYTES 17179869184) # 16 GiB
+endif ()
+
+# The remote (distributed) systest spins up worker containers that each request the budget above. Skip it on hosts
+# that do not have enough physical RAM to back that budget, so a dev laptop does not OOM (or thrash) running it.
+cmake_host_system_information(RESULT _host_physical_memory_mib QUERY TOTAL_PHYSICAL_MEMORY)
+math(EXPR _remote_test_required_mib "${REMOTE_TEST_TOTAL_MEMORY_BYTES} / 1048576")
```
ls-1801: Two things about this silent-skip block:

**1. Silent skip is inconsistent.** We silently drop `systest-remote-test` when the host has less RAM than the required budget, but we do not do the same for the large-scale in-process tests (`large_scale_${joinStrategy}` above), which also request `LARGE_TEST_TOTAL_MEMORY_BYTES` (~40 GiB). Either both should be memory-gated, or (preferred) `cmake` should `FATAL_ERROR` when the machine has less than the required RAM and require the user to explicitly disable `ENABLE_LARGE_TESTS` (or the remote test set). Silent skips vanish tests from CI runs (`ctest -R systest-remote-test` exits 0 with no matches, no `--no-tests=error`) and vanish coverage on any runner reprovisioning without any signal.

**2. Why two different budgets?** `REMOTE_TEST_TOTAL_MEMORY_BYTES=76911476736` (~72 GiB) vs `LARGE_TEST_TOTAL_MEMORY_BYTES=42551738368` (~40 GiB). Both configurations spin up multiple worker processes, each with its own BufferManager and its own `total_memory_in_bytes`. If the per-worker budget is what matters, having two different per-worker limits for what looks like the same class of workload seems arbitrary. Could you justify the split (or unify)?

nes-single-node-worker/tests/offline.bats
```diff
@@ -70,7 +70,7 @@ teardown() {
 
   # Randomly pick words from the expected output
   echo "$output" | grep -q "worker"
-  echo "$output" | grep -q "number_of_buffers_in_global_buffer_manager"
+  echo "$output" | grep -q "total_memory_in_bytes"
```
ls-1801, line 73: Related to the new `total_memory_in_bytes` / `unpooled_memory_fraction` / `buffer_alignment_in_bytes` options in `WorkerConfiguration.hpp`: none of them have offline coverage. Degenerate values silently abort via `PRECONDITION` today (e.g. `total_memory_in_bytes < buffer_size` ⇒ `numOfBuffers > 0` PRECONDITION trips, process aborts).

Please add bats coverage. Sketch:

```bash
@test "worker rejects total_memory_in_bytes too small for one buffer" {
  # buffer_size default is 4096; 0 (and anything under buffer_size) yields 0 pooled buffers.
  run timeout 5 $NES_WORKER --worker.total_memory_in_bytes=0
  [ "$status" -ne 0 ] && [ "$status" -ne 124 ]
  grep -q "numOfBuffers" singleNodeWorker.log || grep -q "PRECONDITION" singleNodeWorker.log
}

@test "worker rejects unpooled_memory_fraction out of range" {
  run timeout 5 $NES_WORKER --worker.unpooled_memory_fraction=1.5
  [ "$status" -ne 0 ] && [ "$status" -ne 124 ]
  grep -qE "unpooled_memory_fraction|fraction" singleNodeWorker.log

  run timeout 5 $NES_WORKER --worker.unpooled_memory_fraction=-0.1
  [ "$status" -ne 0 ] && [ "$status" -ne 124 ]
}

@test "worker rejects non-power-of-two buffer_alignment_in_bytes" {
  run timeout 5 $NES_WORKER --worker.buffer_alignment_in_bytes=48
  [ "$status" -ne 0 ] && [ "$status" -ne 124 ]
  grep -qE "power of two|alignment" singleNodeWorker.log
}

@test "worker rejects buffer_alignment_in_bytes larger than page size" {
  run timeout 5 $NES_WORKER --worker.buffer_alignment_in_bytes=8192
  [ "$status" -ne 0 ] && [ "$status" -ne 124 ]
  grep -qE "page size|alignment" singleNodeWorker.log
}

@test "worker accepts unpooled_memory_fraction at boundary values" {
  run timeout 5 $NES_WORKER --worker.unpooled_memory_fraction=0.0
  [ "$status" -eq 124 ]
  run timeout 5 $NES_WORKER --worker.unpooled_memory_fraction=1.0
  [ "$status" -eq 124 ]
}
```

These fail-fast tests would (a) lock the config contract, (b) catch regressions where PRECONDITION strings mutate, (c) surface the "0 as sentinel vs 0 as invalid" ambiguity between `total_memory_in_bytes` and `unpooled_memory_limit_in_bytes`. The FloatValidation currently only checks that the string parses as a float — it does not clamp to [0,1]. Same for alignment: `NumberValidation` accepts non-power-of-two silently and the PRECONDITION on alignment aborts later.

Ideally, if you want any of these to abort gracefully at config-parse time (instead of PRECONDITION-abort during BufferManager construction), tighten the validators too.

nes-memory/BufferManager.cpp
```diff
@@ -45,21 +45,37 @@ BufferManager::BufferManager(
...
-    return std::make_shared<BufferManager>(Private{}, bufferSize, numOfBuffers, memoryResource, withAlignment);
+    PRECONDITION(
+        unpooledMemoryFraction >= 0.0 and unpooledMemoryFraction <= 1.0,
+        "unpooledMemoryFraction must be in [0.0, 1.0] but is {}",
+        unpooledMemoryFraction);
+    PRECONDITION(bufferSize > 0, "bufferSize must be greater than zero");
+
+    const auto unpooledMemoryLimitInBytes = static_cast<size_t>(static_cast<double>(totalMemoryInBytes) * unpooledMemoryFraction);
+    const size_t pooledMemoryInBytes = totalMemoryInBytes - unpooledMemoryLimitInBytes;
+    const auto numOfBuffers = static_cast<uint32_t>(pooledMemoryInBytes / bufferSize);
```
ls-1801, line 81: `numOfBuffers = static_cast<uint32_t>(pooledMemoryInBytes / bufferSize)` silently truncates. For a multi-TB `totalMemoryInBytes` with a 4 KiB `bufferSize` the quotient exceeds `UINT32_MAX` and the truncated value can wrap to 0 — which then trips the `numOfBuffers > 0` PRECONDITION downstream with an unrelated-looking message. Either widen the field to `size_t` or add an explicit `PRECONDITION(pooledMemoryInBytes / bufferSize <= std::numeric_limits<uint32_t>::max(), ...)` before the cast.

ls-1801 (changes requested):

nes-executable/src/BackpressureChannel.cpp
keyseven123: Done

nes-memory/UnpooledChunksManager.cpp
keyseven123: Done

nes-systests/configs/topologies/two-node-more-capacity.yaml
keyseven123: Done

nes-systests/configs/topologies/two-node-more-capacity.yaml
keyseven123: Done

nes-systests/systest/CMakeLists.txt
keyseven123: 1. It is now a fatal error
2. Have the same budget now

nes-single-node-worker/tests/offline.bats
keyseven123: Done and tightened the `FloatValidation`

nes-memory/BufferManager.cpp
keyseven123: Done

ls-1801 (approved):

Issue #1714, opened 2026-06-20
[EPIC] Dynamic, variable-sized memory management


zeuchste:
## Vision
NES manages execution memory with rigid mechanisms: a single fixed pooled buffer size, an unbounded/slow **unpooled** path for everything else, the buffer size baked into code generation and the network wire format, and statically-capped per-source buffering. This wastes memory (a 3-tuple result still pins a full buffer), risks OOM (#1701), fragments, and mis-distributes memory across heterogeneous sources.

This epic tracks the move to a **unified, dynamic, variable-sized memory subsystem**: every execution buffer comes from fast lock-free **size-class pools**, sized to the actual data; code generation, the network protocol, and operator state are all size-aware; emit is **lazy / right-sized**; and per-source provisioning **adapts to load** under a bounded global budget. This matters for the cloud-edge-sensor target — thousands of heterogeneous queries/sources under tight, variable memory.

## Workstreams
(Rough order; items marked ▸ can proceed in parallel once the foundation lands.)

- [ ] **#1706 — Size-class buffer manager (foundation)** _(in review)_ — variable-sized pooled buffers via power-of-two size classes + pluggable provisioning (TotalBudgetSplit / EagerPerClass / LazyElastic). Backward compatible; reroutes Arena / network-child / `MemoryUtils` off unpooled.
- [ ] ▸ **#1703 — Dynamic columnar capacity in Nautilus codegen** — drop the baked `capacity = bufferSize / tupleSize` so columnar buffers can vary in size.
- [ ] ▸ **#1704 — Carry buffer size in the network wire protocol** — so a variable-sized buffer survives the hop.
- [ ] ▸ **#1705 — PagedVector per-size-class page selection** — operator state uses right-sized pooled pages.
- [ ] **#1711 — Lazy / right-sized emit** — emit gathers output cardinality, then allocates a buffer sized to the actual tuple count _(depends on #1706, #1703)_.
- [ ] **#1712 — Retire the unpooled path** — route all sizes through size classes (large/huge LazyElastic classes), budget-bounded; remove `UnpooledChunksManager`; subsumes #1701 _(depends on #1706, #1711, #1705, #1704)_.
- [ ] ▸ **#1713 — Dynamic, adaptive source buffer-pool provisioning** — replace the static per-source inflight cap with elastic, fair provisioning from the global budget _(depends on #1706)_.

## Dependency graph
```
[...]

Issue #1703, opened 2026-06-19
Variable-size buffers: dynamic columnar capacity in Nautilus codegen


zeuchste:
> **Epic:** #1714 · **Depends on:** #1706 · **Blocks:** #1711

Follow-up to the size-class buffer manager (variable-sized pooled buffers).

The size-class allocator now hands out differently-sized pooled buffers, but **columnar** operator buffers still cannot vary in size because `LowerSchemaProvider::lowerSchema` constant-folds the per-buffer capacity into the generated Nautilus code:

- `nes-nautilus/src/Interface/BufferRef/LowerSchemaProvider.cpp:106` — `const uint64_t capacity = bufferSize / tupleSize;`
- `:114` — column offsets are computed as `field_size * capacity`, baking a fixed capacity (and thus a fixed buffer size) into the column layout.

A buffer handed to columnar-generated code must therefore have exactly the `bufferSize` the code was compiled for. To let columnar pipelines use variable-sized buffers we need the generated code to read the capacity from `buffer.getBufferSize()` at runtime instead of a compile-time constant.

**Scope**
- Make `ColumnTupleBufferRef` compute column offsets from the runtime buffer size.
- Verify generated row vs columnar paths; benchmark the runtime-divide cost vs constant-fold.
- Allow operator emit to request a non-default size class for columnar output.

Deferred from the allocator-first PR (size classes).

🤖 Filed with [Claude Code](https://claude.com/claude-code)

zeuchste:
Part of #1714 — **[EPIC] Dynamic, variable-sized memory management**.

PR #1716, opened 2026-06-22
Feature(#1703): runtime buffer capacity for row-layout codegen

zeuchste:
Part of EPIC #1714 (dynamic memory). Makes buffer capacity a runtime value in row-layout codegen (was a baked compile-time constant), so one generated pipeline can process buffers of any size — a prerequisite for variable-sized and size-class buffers. Independent; based on `main`.


ls-1801 (changes requested):
I think this should require a design decision, whether buffers produced by a pipeline can be dynamically resized, or if we keep the buffer size per pipeline a static query compilation parameter, which would require us to adapt the pipeline and recompile.

zeuchste:
@ls-1801 thanks for the comment. I do only see space savings when we make the buffer size adjustable and I think we do not need to re-compile as the code can use the getBufferSize Call

Issue #1713, opened 2026-06-20
Dynamic, adaptive source buffer-pool provisioning


zeuchste:
> **Epic:** #1714 · **Depends on:** #1706

Part of the variable-sized memory management work (builds on #1706).

## Problem
Source buffer provisioning is **static**. There are no real per-source pools — `LocalBufferPool`/`FixedSizeBufferPool` are forward-declared only (`nes-memory/TupleBufferImpl.hpp`), and the `{GLOBAL, LOCAL, FIXED}` `BufferManagerType` enum is aspirational (only GLOBAL exists). Every source draws from the single shared global `BufferManager` (`NodeEngineBuilder.cpp`), and each source is capped by a **static** inflight limit — a `std::counting_semaphore` sized once at `maxInflightBuffers` (default 64) in `RunningSource.cpp`s emit function (`SourceRuntimeConfiguration::inflightBufferLimit`).

In a cloud-edge-sensor deployment (thousands of heterogeneous sources):
- a **slow/idle source** is permitted 64 inflight buffers it never uses;
- a **fast source** is capped at 64 even when memory is free -> poor distribution and head-of-line throughput limits.

## Proposed change
Replace the static cap with **adaptive, elastic provisioning** drawn from the global (size-class) budget:
- **Start small, grow on demand**: each source begins with a few inflight permits; when it consistently saturates its quota and global memory is available, grant more.
- **"Slower and slower" backoff / fair share**: grant additional permits on a diminishing curve as a source consumes more of the shared budget (AIMD-style), with weighted max-min fairness across sources — a fast source can burst but cannot monopolize.
- **Decay on idle**: reclaim permits from sources that stop using them and return them to the global budget.
- **Signals**: buffer-acquisition stalls (`SourceThread.cpp` retry loop), backpressure (`BackpressureChannel`), and global memory pressure.

## Hook points
- `nes-query-engine/RunningSource.cpp` — the `counting_semaphore` inflight limit -> resizable/adaptive throttle.
- `nes-sources/include/Sources/SourceHandle.hpp` `SourceRuntimeConfiguration::inflightBufferLimit` -> dynamic (min/max/elasticity).
- `nes-sources/src/SourceProvider.cpp::lower` -> initial small quota + elasticity policy.
- New `WorkerConfiguration` options (enable dynamic provisioning, min/max inflight, source memory budget, elasticity mode) and `SourceDescriptor` hints.
[...]

PR #1717, opened 2026-06-22
Feature(#1713): adaptive in-flight source provisioning (InflightThrottle + AIMD)

zeuchste:
Part of EPIC #1714. Replaces the static per-source in-flight `counting_semaphore` with a resizable `InflightThrottle` + AIMD `InflightPolicy` (grow under load, decay when idle, bounded by the old cap), gated by `enable_adaptive_inflight_buffers`. Uses `condition_variable_any` to compose with `stop_token`. Unit + differential test included. Independent; based on `main`.


ls-1801 (changes requested):
I am not sure what I am looking at. This is a limit that increases if the source needs more slots upto a fixed limit?

Can you explain what problem this is solving?

I think there is room for improvement in our sources, when it comes to a limits, but that's a larger design decision and we probably need a component that can manage the limit of individual sources in relation to other sources (and sources of the same query).

But I would say we need a concrete goal first, the current solution is the minimal version of preventing that a single source fills the queue, your version achieves the same with more complexity.

PR #1719, opened 2026-06-22
Feature(#1705): route operator-state pages through pooled size classes

zeuchste:
Part of EPIC #1714. Routes operator-state pages (PagedVector, ChainedHashMap) through `getPagedBuffer` + `servesSizeClasses()` so state is served from fitting size classes under the unified budget instead of the unpooled heap. Stacked on #1706 (base: `feature/variable-size-buffer-size-classes`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1706, opened 2026-06-19
[nes-memory] Variable-sized pooled buffers via power-of-two size classes

zeuchste:
## Purpose of the Change and Brief Change Log

The global `BufferManager` only handed out a **single fixed buffer size**; any other size fell through to the slower per-thread **unpooled** path (heap-allocated control block, RW-locked chunk maps, chunk free deferred until its last buffer is released). That unpooled path exists *only because* pooling was single-size.

This PR generalizes the pooled path into **segregated power-of-two size classes**, so a variable-sized request is rounded up to the smallest fitting class and served O(1) from a lock-free pool.

### Design: same idea as DuckDB / ClickHouse / Velox / Umbra — and how this differs

The shared, proven idea is a **segregated free list by size class**: instead of one buffer size, keep N pools, one per power-of-two class, and round requests up.
- **Umbra** (CIDR'20): variable-size pages in power-of-two classes from 64 KiB, each class a contiguous VM segment, size class encoded in the swizzled pointer.
- **Velox**: `MmapAllocator` with ~9 power-of-two `SizeClass` objects; sub-page allocs fall back to `malloc`.
- **ClickHouse**: `S·2^N` segregated blocks (2 MB chunks for big structures) → "no external fragmentation, fast alloc/dealloc", over jemalloc thread-arenas.
- **DuckDB**: deliberately keeps almost everything on *one* fixed block size and only registers arbitrary-size buffers (≥ block) for evictable temp memory — it *avoids* mixing sizes to fight fragmentation.

**How this implementation distinguishes itself:**
1. **Reuses NES's own fixed-pool machinery per class.** Each class is the existing pooled path — lock-free `folly::MPMCQueue` of preallocated `MemorySegment`s with placement-new'd intrusive control blocks — instantiated once per class. No new allocator core, so the per-class hot path keeps NES's exact zero-GC, lock-free reference-counting/recycling semantics. (Umbra/Velox build a bespoke page allocator; we generalize ours.)
2. **Streaming, not paging.** Unlike the Umbra/DuckDB buffer managers (disk paging, eviction under a memory limit, pin/unpin + pointer swizzling), NES buffers are in-memory, reference-counted, never evicted. Size classes here are purely an allocation-latency/fragmentation optimization — no swizzling/pinning.
[...]

Issue #1705, opened 2026-06-19
Variable-size buffers: PagedVector per-size-class page selection


zeuchste:
> **Epic:** #1714 · **Depends on:** #1706 · **Blocks:** #1712

Follow-up to the size-class buffer manager (variable-sized pooled buffers).

`PagedVector` allocates its pages at the provider's default buffer size:

- `nes-nautilus/src/Interface/PagedVector/PagedVector.cpp` — new pages use `bufferProvider->getBufferSize()`.
- `nes-physical-operators/.../HJBuildPhysicalOperator.cpp:144` and `NLJSlice.cpp:42` init it with the default size.
- `PagedVector::Header` (`PagedVector.hpp`) precondition requires `bufferSize > headerSize + tupleSize`, so a tuple larger than the default buffer cannot be stored.

Now that the buffer manager can serve larger pooled classes, a PagedVector should be able to pick a fitting size class at init (one class for the whole vector keeps the per-page capacity uniform, which the indexing relies on) so large-tuple joins/aggregations get bigger pooled pages instead of hitting the precondition or wasting space.

**Scope**
- Choose a size class per PagedVector at init based on tuple size.
- Allocate all pages of that vector from the chosen class.
- Thread the choice through HashJoin build / NLJ slice.

Deferred from the allocator-first PR (size classes).

🤖 Filed with [Claude Code](https://claude.com/claude-code)

zeuchste:
Part of #1714 — **[EPIC] Dynamic, variable-sized memory management**.

PR #1721, opened 2026-06-22
Feature(#1705): route operator-state pages through pooled size classes

zeuchste:
Part of EPIC #1714. Routes operator-state pages (PagedVector, ChainedHashMap) through `getPagedBuffer` + `servesSizeClasses()` so state is served from fitting size classes under the unified budget. Rebased onto upstream #1706 (`feature/variable-size-buffer-size-classes`) — scoped to 5 files. (Replaces closed #1719.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ scripts/__pycache__
 .claude/
 .codex/
 [Cc][Ll][Aa][Uu][Dd][Ee].md
+/build-bench*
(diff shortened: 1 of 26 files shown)


ThanGeo:
I suggest you stack this over #1667 as the design for the Chained Hash Map is significantly changed and there are plenty of more getBuffer calls there

PR #1723, opened 2026-06-23
Chore: drop the dead withAlignment parameter from BufferManager

zeuchste:
## What

`BufferManager`'s constructor took a `withAlignment` argument but immediately discarded it (`((void)withAlignment)`) and always aligned to `DEFAULT_ALIGNMENT` (64 bytes = one cache line). So the parameter was dead — whatever a caller passed was ignored.

This removes the vestigial parameter from `create()` and the private constructor, and lets `initialize()` use `DEFAULT_ALIGNMENT` directly.

## What does *not* change
The alignment behaviour: buffers are still cache-line aligned via `DEFAULT_ALIGNMENT`. Only the ignored knob is gone. No call site passed `withAlignment` explicitly, so this is a pure no-op cleanup.

Verified: `nes-memory` builds, `child-buffer-tests` (3/3) and `tuple-buffer-memory-access-tests` (8/8) pass, clang-format clean.

### DIFF
--- a/nes-memory/BufferManager.cpp
+++ b/nes-memory/BufferManager.cpp
@@ -41,25 +41,20 @@ namespace NES
 {
 
 BufferManager::BufferManager(
-    Private,
-    const uint32_t bufferSize,
-    const uint32_t numOfBuffers,
-    std::shared_ptr<std::pmr::memory_resource> memoryResource,
-    const uint32_t withAlignment)
+    Private, const uint32_t bufferSize, const uint32_t numOfBuffers, std::shared_ptr<std::pmr::memory_resource> memoryResource)
     : availableBuffers(numOfBuffers)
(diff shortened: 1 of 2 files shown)


ls-1801 (approved):

nes-memory/BufferManager.cpp
```diff
@@ -108,8 +103,10 @@ void BufferManager::destroy()
     }
 }
 
-void BufferManager::initialize(uint32_t withAlignment)
+void BufferManager::initialize()
 {
+    /// Buffers are always aligned to a cache line; the alignment is no longer configurable.
```
keyseven123, line 108: I think we should make this configurable if a user wants to. Nothing guarantees that the cache line stays for ever at 64B. But I think this should be done in PR #1702 , as #1702 deletes the other default constexpr.

keyseven123 (approved):

PR #1728, opened 2026-07-02
chore(schema): Move Schema to nes-common

Artraxon:
Move the schema class (and its fwd header) to nes-common.
Does NOT move the field types, they stay in nes-datatypes.

IMO I should have done this in the schema inference PR directly, since the schema class itself does not have any dependency on datatypes, but rather is templated over any type that has a getFullyQualifiedName method.

This is a prerequisite to leveraging the schema class to also resolve config options in the SQL frontend, since any catalog object must be configurable through a set of named options, where collisions and possible, safe, shorthands must be resolved through the logic implemented in the schema (whether its unambiguous to use just PATH instead of SOURCE.PATH).

Also, I put the schema header into its own include folder so that future refactors are easier.


greenBene (approved):
I like this! :) It makes sense and makes the code base cleaner! Thank you!

keyseven123 (changes requested):
In general, I like this PR. I am not happy with the destination/new location of the schema. I can remember us having a discussion that we want to have a logical schema and a physical schema (or however we will call it). By adding the Schema now into `nes-common ', it signals that everywhere in the project one can and should include Schema even though this is not the case for the physical schema.

Artraxon:
@keyseven123 added an extra target

keyseven123 (approved):
As discussed, I have created #1855 as a follow-up for this.

PR #1729, opened 2026-07-03
feat(sql) Improved EXPLAIN statement output, and make EXPLAIN statements testable via systests 

### LINKED ISSUE #1679: [TASK] Improve `EXPLAIN`  handling
### Description

The current `EXPLAIN [query]` output gives only limited information and is not well-suited for testing. 

The goal of this task is to add options to the `EXPLAIN` call, allowing users to define whether they want a graphical visualization, a simple text-based output, or a verbose text-based output. 

### Acceptance Criteria

* Add options to the `EXPLAIN` statement to request a visual, simple-text, or verbose-text representation of the query
* Extend SLT to support EXPLAIN statements
* Tests

greenBene:

This pull request improves the output of EXPLAIN statements and adds options to control what is shown. The
change list is as follows:
- Extended the SQL grammar (AntlrSQL.g4) so EXPLAIN accepts optional parenthesized options: a stage
(LOGICAL, OPTIMIZED, DISTRIBUTED, ALL) and a FORMAT (VISUAL, TEXT, VERBOSE), e.g. EXPLAIN (LOGICAL, FORMAT
TEXT) SELECT ....
- Added bindExplainStatement to the StatementBinder, which parses the new options, rejects duplicate
stage/format options, and produces an ExplainQueryStatement carrying the requested ExplainStage and
ExplainFormat.
- Updated the StatementHandler and QueryConsoleDumpHandler to render the selected stage in the requested
format with clearer, stage-labeled output (e.g. == Initial Logical Plan ==, == Optimized Global Plan ==).
- Improved the textual representation of plans and operators (LogicalPlan, PhysicalPlan,
SinkLogicalOperator, FieldAccessLogicalFunction, SourceDescriptor).
- Extended the systest framework to parse, run, and validate EXPLAIN statements against expected output.

## Verifying this change

This change is tested by:
- Added nes-systests/explain/EXPLAIN.test, a system test that validates the EXPLAIN output for logical and
optimized plans.
- Added unit tests in StatementBinderTest covering binding of the new EXPLAIN stage/format options.
- Added SystestParserTests covering parsing and validation of EXPLAIN statements in systests.

- QueryCompiler / SQL parser (grammar and statement binding)
- Frontend statement handling and plan dumping
- Logical, physical, and source operator string representations
- Systest framework (parser, binder, result checking, runner)

This PR closes #1679

### DIFF
--- a/nes-logical-operators/src/Operators/Sinks/SinkLogicalOperator.cpp
+++ b/nes-logical-operators/src/Operators/Sinks/SinkLogicalOperator.cpp
@@ -115,6 +115,10 @@ std::string SinkLogicalOperator::explain(ExplainVerbosity verbosity, OperatorId 
         }
         return fmt::format("SINK(opId: {}, sinkName: {})", id, sinkName);
     }
+    if (sinkDescriptor.has_value() && sinkDescriptor->isInline())
+    {
+        return fmt::format("SINK({})", sinkDescriptor->getSinkType());
+    }
     return fmt::format("SINK({})", sinkName);
 }
 
--- a/nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
+++ b/nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
@@ -16,6 +16,7 @@
 
 #include <algorithm>
 #include <array>
+#include <ranges>
 #include <set>
 #include <string_view>
 #include <typeindex>
@@ -69,6 +70,17 @@ std::unordered_set<Field> getAccessedFields(const LogicalFunction& logicalFuncti
     return result;
 }
 
+std::vector<Field> sortFields(const std::unordered_set<Field>& fields)
+{
+    auto sortedFields = std::ranges::to<std::vector>(fields);
+    std::ranges::sort(
+        sortedFields,
+        [](const Field& left, const Field& right)
+        { return left.getLastName().asCanonicalString() < right.getLastName().asCanonicalString(); });
+
+    return sortedFields;
+}
+
 LogicalOperator pushBeyondSink(const TypedLogicalOperator<SinkLogicalOperator>& op)
 {
     /// Use identifiers from output schema as starting point for required fields.
@@ -94,7 +106,9 @@ LogicalOperator pushBeyondSource(TypedLogicalOperator<SourceDescriptorLogicalOpe
     {
         std::vector<ProjectionLogicalOperator::UnboundProjection> projections;
         projections.reserve(required.size());
-        for (const auto& field : required)
+
+        /// Set a fix order in which the projections are applied to ensure deterministic behaviour of rule.
+        for (const auto& field : sortFields(required))
         {
             projections.emplace_back(field.getLastName(), UnboundFieldAccessLogicalFunction{field.getLastName()});
         }
@@ -134,8 +148,8 @@ LogicalOperator pushBeyondProjection(const TypedLogicalOperator<ProjectionLogica
     {
         projectionMap.emplace(field, function);
     }
-
-    for (const auto& field : required)
+    /// Set a fix order in which the projections are applied to ensure deterministic behaviour of rule.
+    for (const auto& field : sortFields(required))
     {
         if (auto iter = projectionMap.find(field); iter != projectionMap.end())
         {
--- a/nes-sources/include/Sources/SourceDescriptor.hpp
+++ b/nes-sources/include/Sources/SourceDescriptor.hpp
@@ -65,6 +65,8 @@ public:
     [[nodiscard]] Host getHost() const;
     [[nodiscard]] PhysicalSourceId getPhysicalSourceId() const;
 
+    [[nodiscard]] bool isInlineSource() const;
+
     [[nodiscard]] std::string explain(ExplainVerbosity verbosity) const;
 
 private:
@@ -78,6 +80,7 @@ private:
     std::string sourceType;
     Host host;
     InputFormatterDescriptor inputFormatterDescriptor;
+    bool isInline = false;
 
 
     /// Used by Sources to create a valid SourceDescriptor.
@@ -87,7 +90,8 @@ private:
         std::string_view sourceType,
         Host host,
         DescriptorConfig::Config config,
-        const InputFormatterDescriptor& inputFormatterDescriptor);
+        const InputFormatterDescriptor& inputFormatterDescriptor,
+        bool isInline);
 
 public:
     /// Per default, we set an 'invalid' number of max inflight buffers. We choose zero as an invalid number as giving zero buffers to a source would make it unusable.
@@ -137,6 +141,7 @@ struct ReflectedSourceDescriptor
     std::string type;
     Host host;
     InputFormatterDescriptor inputFormatterDescriptor;
+    bool isInline;
     Reflected config;
 };
 }
--- a/nes-sql-parser/AntlrSQL.g4
+++ b/nes-sql-parser/AntlrSQL.g4
@@ -60,7 +60,11 @@ terminatedStatement: statement ';';
 multipleStatements: (statement (';' statement)* ';'?)? EOF;
 statement: queryWithOptions | createStatement | dropStatement | showStatement | explainStatement;
 
-explainStatement: EXPLAIN query;
+explainStatement: EXPLAIN ('(' explainStages ')')? (FORMAT explainFormat)? query;
+explainStages: explainStage (',' explainStage)*;
+explainStage: identifier | LOGICAL;
+explainFormat: identifier | TEXT;
+
 createStatement: CREATE createDefinition;
 createDefinition: createLogicalSourceDefinition | createPhysicalSourceDefinition | createSinkDefinition | createWorkerDefinition | createModelDefinition;
 createLogicalSourceDefinition: LOGICAL SOURCE sourceName=identifier schemaDefinition fromQuery?;
@@ -619,10 +623,10 @@ UNSIGNED_TYPE_QUALIFIER: 'UNSIGNED ';
 
 
 SHOW : 'SHOW';
-FORMAT : 'FORMAT';
+FORMAT : 'FORMAT' | 'format';
 CREATE : 'CREATE';
 SOURCE : 'SOURCE';
-LOGICAL: 'LOGICAL';
+LOGICAL: 'LOGICAL' | 'logical';
 PHYSICAL: 'PHYSICAL';
 WORKER: 'WORKER';
 SINK : 'SINK';
--- a/nes-sql-parser/include/SQLQueryParser/StatementBinder.hpp
+++ b/nes-sql-parser/include/SQLQueryParser/StatementBinder.hpp
@@ -14,6 +14,7 @@
 
 
 #pragma once
+
 #include <cstddef>
 #include <cstdint>
 #include <expected>
@@ -24,6 +25,7 @@
 #include <string>
 #include <string_view>
 #include <unordered_map>
+#include <unordered_set>
 #include <variant>
 #include <vector>
 #include <AntlrSQLParser.h>
@@ -125,9 +127,25 @@ struct QueryStatement
     std::optional<DistributedQueryId> id;
 };
 
+enum class ExplainFormat : uint8_t
+{
+    Visual,
+    Text,
+    Verbose
+};
+
+enum class ExplainStage : uint8_t
+{
+    Logical,
+    Optimized,
+    Distributed
+};
+
 struct ExplainQueryStatement
 {
     LogicalPlan plan;
+    ExplainFormat explainFormat = ExplainFormat::Visual;
+    std::unordered_set<ExplainStage> explainStages = {ExplainStage::Logical, ExplainStage::Optimized, ExplainStage::Distributed};
 };
 
 struct ShowQueriesStatement
--- a/nes-sql-parser/src/StatementBinder.cpp
+++ b/nes-sql-parser/src/StatementBinder.cpp
@@ -30,6 +30,7 @@
 #include <Util/Strings.hpp>
 
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <variant>
 #include <vector>
@@ -531,6 +532,95 @@ public:
         throw InvalidStatement("Unrecognized DROP statement");
     }
 
+    std::unordered_set<ExplainStage> bindExplainStages(AntlrSQLParser::ExplainStagesContext* stagesContext) const
+    {
+        std::unordered_set<ExplainStage> explainStages;
+
+        for (auto* const stageContext : stagesContext->explainStage())
+        {
+            if (stageContext->LOGICAL() != nullptr)
+            {
+                if (!explainStages.insert(ExplainStage::Logical).second)
+                {
+                    throw InvalidStatement("Duplicate EXPLAIN stage: LOGICAL");
+                }
+            }
+            else
+            {
+                const auto identifier = bindIdentifier(stageContext->identifier());
+
+                if (identifier == Identifier::parse("logical"))
+                {
+                    if (!explainStages.insert(ExplainStage::Logical).second)
+                    {
+                        throw InvalidStatement("Duplicate EXPLAIN stage: LOGICAL");
+                    }
+                }
+                else if (identifier == Identifier::parse("optimized"))
+                {
+                    if (!explainStages.insert(ExplainStage::Optimized).second)
+                    {
+                        throw InvalidStatement("Duplicate EXPLAIN stage: OPTIMIZED");
+                    }
+                }
+                else if (identifier == Identifier::parse("distributed"))
+                {
+                    if (!explainStages.insert(ExplainStage::Distributed).second)
+                    {
+                        throw InvalidStatement("Duplicate EXPLAIN stage: DISTRIBUTED");
+                    }
+                }
+                else
+                {
+                    throw InvalidStatement("Unrecognized EXPLAIN stage: {}", identifier);
+                }
+            }
+        }
+
+        return explainStages;
+    }
+
+    ExplainFormat bindExplainFormat(AntlrSQLParser::ExplainFormatContext* formatAST) const
+    {
+        if (formatAST->TEXT() != nullptr)
+        {
+            return ExplainFormat::Text;
+        }
+
+        auto parsedFormatIdentifier = bindIdentifier(formatAST->identifier());
+
+        if (parsedFormatIdentifier == Identifier::parse("text"))
+        {
+            return ExplainFormat::Text;
+        }
+        if (parsedFormatIdentifier == Identifier::parse("verbose"))
+        {
+            return ExplainFormat::Verbose;
+        }
+        if (parsedFormatIdentifier == Identifier::parse("visual"))
+        {
+            return ExplainFormat::Visual;
+        }
+        throw InvalidStatement("Unrecognized EXPLAIN format: {}", parsedFormatIdentifier);
+    }
+
+    Statement bindExplainStatement(AntlrSQLParser::ExplainStatementContext* explainAst) const
+    {
+        INVARIANT(explainAst->query() != nullptr, "Should be enforced by antlr");
+        ExplainFormat format = ExplainFormat::Visual;
+        std::unordered_set<ExplainStage> stages = {ExplainStage::Logical, ExplainStage::Optimized, ExplainStage::Distributed};
+        if (auto* const stagesAST = explainAst->explainStages())
+        {
+            stages = bindExplainStages(stagesAST);
+        }
+        if (auto* const formatAST = explainAst->explainFormat())
+        {
+            format = bindExplainFormat(formatAST);
+        }
+
+        return ExplainQueryStatement{.plan = queryBinder(explainAst->query()), .explainFormat = format, .explainStages = stages};
+    }
+
     std::expected<Statement, Exception> bind(AntlrSQLParser::StatementContext* statementAST) const
     {
         try
@@ -549,8 +639,7 @@ public:
             }
             if (auto* const explainStatementAST = statementAST->explainStatement())
             {
-                INVARIANT(explainStatementAST->query() != nullptr, "Should be enforced by antlr");
-                return ExplainQueryStatement{queryBinder(explainStatementAST->query())};
+                return bindExplainStatement(explainStatementAST);
             }
             if (auto* const queryAst = statementAST->queryWithOptions(); queryAst != nullptr)
             {
--- a/nes-systests/explain/EXPLAIN.test
+++ b/nes-systests/explain/EXPLAIN.test
@@ -1,0 +1,122 @@
+# name: explain/EXPLAIN.test
+# description: Validates the output of EXPLAIN statements.
+# groups: [explain, inference]
+
+CREATE LOGICAL SOURCE stream(id UINT64 NOT NULL, value UINT64 NOT NULL, timestamp UINT64 NOT NULL);
+CREATE PHYSICAL SOURCE FOR stream TYPE File;
+ATTACH INLINE
+1,1,1000
+2,3,2000
+
+CREATE LOGICAL SOURCE stream2(id2 UINT64 NOT NULL, value2 UINT64 NOT NULL, timestamp2 UINT64 NOT NULL);
+CREATE PHYSICAL SOURCE FOR stream2 TYPE File;
+ATTACH INLINE
+1,2,1000
+2,4,2000
+
+CREATE SINK sinkFilter(id UINT64 NOT NULL, value UINT64 NOT NULL) TYPE File;
+CREATE SINK sinkJoin(start UINT64 NOT NULL, end UINT64 NOT NULL, id UINT64 NOT NULL, value UINT64 NOT NULL, timestamp UINT64 NOT NULL, id2 UINT64 NOT NULL, value2 UINT64 NOT NULL, timestamp2 UINT64 NOT NULL) TYPE File;
+
+CREATE MODEL iris ('model/iris.onnx')
+INPUT (p1 FLOAT32, p2 FLOAT32, p3 FLOAT32, p4 FLOAT32)
+OUTPUT (setosa FLOAT32, versicolor FLOAT32, virginica FLOAT32);
+
+CREATE LOGICAL SOURCE stream_inference(p1 FLOAT32 NOT NULL, p2 FLOAT32 NOT NULL, p3 FLOAT32 NOT NULL, p4 FLOAT32 NOT NULL);
+CREATE PHYSICAL SOURCE FOR stream_inference TYPE File;
+ATTACH INLINE
+6.1,3.5,1.4,0.2
+4.9,3.0,1.4,0.2
+4.7,3.2,1.3,0.2
+
+
+
+EXPLAIN (LOGICAL) FORMAT TEXT SELECT id, value FROM stream WHERE value > UINT64(2) INTO sinkFilter;
+----
+== Initial Logical Plan ==
+SINK(SINKFILTER)
+  PROJECTION(fields: [ID, VALUE])
+    SELECTION(VALUE > 2)
+      SOURCE(STREAM)
+==END==
+
+
+EXPLAIN (OPTIMIZED) FORMAT TEXT SELECT id, value FROM stream WHERE value > UINT64(2) INTO sinkFilter;
+----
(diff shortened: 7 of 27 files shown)


nes-logical-operators/src/Operators/Sinks/SinkLogicalOperator.cpp
```diff
@@ -115,6 +115,10 @@ std::string SinkLogicalOperator::explain(ExplainVerbosity verbosity, OperatorId
         }
         return fmt::format("SINK(opId: {}, sinkName: {})", id, sinkName);
     }
+    if (sinkDescriptor->isInline())
+    {
+        return fmt::format("SINK({})", sinkDescriptor->getSinkType());
```
ls-1801, line 120: This is unchecked and crashes

greenBene: Thanks! Fixed it.

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -861,6 +903,51 @@ struct SystestBinder::Impl
...
+            const auto parseResult = managedParser->parseSingle();
+            if (not parseResult.has_value())
+            {
+                throw InvalidQuerySyntax("failed to parse the statement \"{}\"", replaceAll(statementString, "\n", " "));
+            }
+
+            auto binding = binder.bind(parseResult.value().get());
+            if (not binding.has_value())
+            {
+                throw InvalidQuerySyntax("failed to bind the statement \"{}\": ", statementString, binding.error());
```
ls-1801: The format string has only one `{}` placeholder but two arguments — the actual bind error from `binding.error()` is dropped and the trailing colon dangles. Fix to `"failed to bind the statement \"{}\": {}"`.

nes-sql-parser/AntlrSQL.g4
```diff
@@ -621,6 +625,10 @@ LOGICAL: 'LOGICAL';
 PHYSICAL: 'PHYSICAL';
 WORKER: 'WORKER';
 SINK : 'SINK';
+OPTIMIZED: 'OPTIMIZED' | 'optimized';
```
ls-1801: The new EXPLAIN option keywords have inconsistent case-sensitivity. `OPTIMIZED`, `DISTRIBUTED`, `VISUAL`, `VERBOSE`, and `ALL` accept both uppercase and lowercase (`OPTIMIZED | optimized`), but `LOGICAL`, `TEXT`, `FORMAT` are uppercase-only. So `EXPLAIN (optimized) ...` works while `EXPLAIN (logical) ...` fails. Please make this uniform (SQL keywords are typically case-insensitive throughout the grammar).

Suggested regression test in `StatementBinderTest.cpp`:

```cpp
TEST_F(StatementBinderTest, ExplainStatementCaseInsensitive)
{
    const auto bind = [&](std::string_view sql) { return binder->parseAndBindSingle(sql); };
    // These currently fail while their uppercase counterparts succeed.
    EXPECT_TRUE(bind("EXPLAIN (logical) SELECT * FROM testSource INTO testSink").has_value());
    EXPECT_TRUE(bind("EXPLAIN (format text) SELECT * FROM testSource INTO testSink").has_value());
    EXPECT_TRUE(bind("EXPLAIN (logical, format verbose) SELECT * FROM testSource INTO testSink").has_value());
}
```

nes-systests/systest/src/SystestParser.cpp
```diff
@@ -296,6 +328,7 @@ void SystestParser::parse()
                 break;
             }
             case TokenType::DIFFERENTIAL: {
+                INVARIANT(readResultVerbatim == false, "Differential tests are not supported expected verbatim results ");
```
ls-1801: This INVARIANT message is ungrammatical: `"Differential tests are not supported expected verbatim results "` — two clauses jammed together with no conjunction/punctuation, plus a trailing space. Suggest: `"DIFFERENTIAL block cannot follow an EXPLAIN statement (verbatim result expected)"`.

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -69,6 +69,17 @@ std::unordered_set<Field> getAccessedFields(const LogicalFunction& logicalFuncti
     return result;
 }
 
+std::vector<Field> sortFields(const std::unordered_set<Field>& fields)
+{
+    auto sortedFields = std::ranges::to<std::vector>(fields);
+    std::ranges::sort(
```
ls-1801, line 76: The comparator only uses `getLastName().asCanonicalString()`, so two fields sharing a last name (e.g. `t1.a` vs `t2.a` in a join / self-join) tie and their relative order is unspecified — the "deterministic behaviour" comment is not strictly true today. Is this OK for now until proper qualifier support lands, or should we tie-break on the fully-qualified name here?

nes-sources/include/Sources/SourceDescriptor.hpp
```diff
@@ -78,6 +80,7 @@ class SourceDescriptor final : public Descriptor
     std::string sourceType;
     Host host;
     InputFormatterDescriptor inputFormatterDescriptor;
+    bool isInline;
```
ls-1801: New `isInline` member has no in-class default. All current construction paths pass it explicitly, but adding `bool isInline = false;` here would guard against a future deserialization / default-init path silently reading indeterminate output.

nes-systests/explain/EXPLAIN.test
```diff
@@ -0,0 +1,53 @@
...
+)
+INTO sinkJoin;
+----
+== Optimized Global Plan ==
+SINK(SINKJOIN3)
+  Join(INNER_JOIN, ID = ID2)
+    WATERMARK_ASSIGNER(Event time)
+      SOURCE(STREAM)
+    WATERMARK_ASSIGNER(Event time)
+      SOURCE(STREAM2)
```
ls-1801, line 66: The OPTIMIZED plan output here is missing information a reader would expect from the query — e.g. the TUMBLING window descriptor (size, timestamp fields) is gone, and there is no explicit projection for the eight selected columns. Some of this is probably a limitation of the current `explain()` implementations for windowed/join operators.

Two suggestions:

1. Open follow-up issues to enrich the `explain()` output for operators that currently drop information (window descriptor, join predicates beyond the equi-key, projection field lists). This test then becomes a regression check for those improvements.

2. Rather than curating "the right" expected plan, add a broad set of test queries covering every major operator (selection, projection, join, window, watermark, aggregation, sink types, sources) and freeze whatever the current `explain()` returns. Any future change to a stringification then forces the author to update the expected output — automatic coverage of the explain functions with no manual auditing.

ls-1801, line 66: DISTRIBUTED EXPLAIN would be much easier to nail down in the repl test suite than in a systest: the embedded repl already binds a topology purely via `CREATE WORKER`.

Adding a two-node test there gives us regression coverage for the exact placement wording (which stage/format produces what, network sinks/sources, per-worker plan grouping) without depending on the systest cluster config. Suggestion:

```diff
diff --git a/nes-frontend/apps/repl/tests/embedded.bats b/nes-frontend/apps/repl/tests/embedded.bats
index 5152d82dcc..3df2b4d5d8 100644
--- a/nes-frontend/apps/repl/tests/embedded.bats
+++ b/nes-frontend/apps/repl/tests/embedded.bats
@@ -78,3 +78,137 @@ setup()         { nes_offline_setup; }
   [ "$status" -ne 0 ]
   grep "invalid config parameter; Enum for INVALID was not found." nes-repl.log
 }
+
+@test "EXPLAIN over 2-node topology places sources upstream and join downstream" {
+  run $NES_REPL -f JSON <tests/sql-file-tests/good/explain_distributed.sql
+  [ "$status" -eq 0 ]
+
+  # The six EXPLAIN statements sit at the tail of the JSON output stream.
+  local n=${#lines[@]}
+  local i_logical_text=$((n - 6))
+  local i_optimized_text=$((n - 5))
+  local i_distributed_text=$((n - 4))
+  local i_logical_visual=$((n - 3))
+  local i_optimized_visual=$((n - 2))
+  local i_distributed_visual=$((n - 1))
+
+  # `sed` inside extract_explain right-trims each line so VISUAL padding does not need to
+  # live as trailing whitespace in the .bats source.
+  extract_explain() {
+    echo "$1" | jq -j '.[0].explain' | sed 's/[[:space:]]*$//'
+  }
+
+  assert_equal "$(extract_explain "${lines[$i_logical_text]}")" "$(cat <<'EOF'
+== Initial Logical Plan ==
+SINK(SINKJOIN)
+  PROJECTION(fields: [START, END, ID, VALUE, TIMESTAMP, ID2, VALUE2, TIMESTAMP2])
+    PROJECTION(fields: [*])
+      Join(INNER_JOIN, ID = ID2)
+        WATERMARK_ASSIGNER(Event time)
+          PROJECTION(fields: [*])
+            SOURCE(STREAM)
+        WATERMARK_ASSIGNER(Event time)
+          PROJECTION(fields: [*])
+            SOURCE(STREAM2)
+EOF
+)"
+
+  assert_equal "$(extract_explain "${lines[$i_optimized_text]}")" "$(cat <<'EOF'
+== Optimized Global Plan ==
+SINK(SINKJOIN)
+  Join(INNER_JOIN, ID = ID2)
+    WATERMARK_ASSIGNER(Event time)
+      SOURCE(STREAM)
+    WATERMARK_ASSIGNER(Event time)
+      SOURCE(STREAM2)
+EOF
+)"
+
+  assert_equal "$(extract_explain "${lines[$i_distributed_text]}")" "$(cat <<'EOF'
+== Decomposed Plans ==
+-- 1 plan(s) on downstream-node:8080 --
+0:
+SINK(SINKJOIN)
+  Join(INNER_JOIN, ID = ID2)
+    SOURCE(NETWORK)
+    SOURCE(NETWORK)
+
+
+-- 2 plan(s) on upstream-node:8080 --
+0:
+SINK(NETWORK)
+  WATERMARK_ASSIGNER(Event time)
+    SOURCE(STREAM)
+
+
+1:
+SINK(NETWORK)
+  WATERMARK_ASSIGNER(Event time)
+    SOURCE(STREAM2)
+EOF
+)"
+
+  assert_equal "$(extract_explain "${lines[$i_logical_visual]}")" "$(cat <<'EOF'
+== Initial Logical Plan ==
+
+                       SINK(SINKJOIN)
+                              │
+PROJECTION(fields: [START, END, ID, VALUE, TIMESTAMP, ID2...
+                              │
+                   PROJECTION(fields: [*])
+                              │
+                 Join(INNER_JOIN, ID = ID2)
+               ┌──────────────┴───────────────┐
+WATERMARK_ASSIGNER(Event time) WATERMARK_ASSIGNER(Event time)
+               │                              │
+   PROJECTION(fields: [*])        PROJECTION(fields: [*])
+              ┌┘                             ┌┘
+       SOURCE(STREAM)                SOURCE(STREAM2)
+EOF
+)"
+
+  assert_equal "$(extract_explain "${lines[$i_optimized_visual]}")" "$(cat <<'EOF'
+== Optimized Global Plan ==
+
+                       SINK(SINKJOIN)
+                              │
+                 Join(INNER_JOIN, ID = ID2)
+               ┌──────────────┴───────────────┐
+WATERMARK_ASSIGNER(Event time) WATERMARK_ASSIGNER(Event time)
+              ┌┘                             ┌┘
+       SOURCE(STREAM)                SOURCE(STREAM2)
+EOF
+)"
+
+  assert_equal "$(extract_explain "${lines[$i_distributed_visual]}")" "$(cat <<'EOF'
+== Decomposed Plans ==
+-- 1 plan(s) on downstream-node:8080 --
+0:
+
+        SINK(SINKJOIN)
+               │
+  Join(INNER_JOIN, ID = ID2)
+       ┌───────┴───────┐
+SOURCE(NETWORK) SOURCE(NETWORK)
+
+
+-- 2 plan(s) on upstream-node:8080 --
+0:
+
+        SINK(NETWORK)
+               │
+WATERMARK_ASSIGNER(Event time)
+               │
+        SOURCE(STREAM)
+
+
+1:
+
+        SINK(NETWORK)
+               │
+WATERMARK_ASSIGNER(Event time)
+               │
+       SOURCE(STREAM2)
+EOF
+)"
+}
```

nes-sql-parser/include/SQLQueryParser/StatementBinder.hpp
```diff
@@ -125,9 +125,26 @@ struct QueryStatement
...
+    Optimized,
+    Distributed,
+    All
+};
+
 struct ExplainQueryStatement
 {
     LogicalPlan plan;
+    ExplainFormat explainFormat = ExplainFormat::Visual;
+    ExplainStage explainStage = ExplainStage::All;
```
ls-1801: Follow-up issue: `nes-cli dump` at `nes-frontend/apps/cli/CLIStarter.cpp:798` constructs `ExplainQueryStatement(plan)` with defaults, so after this PR it always emits `stage=All, format=Visual` — three stage-labeled sections concatenated. Please open a follow-up to make `dump` accept the explain stage (`LOGICAL|OPTIMIZED|DISTRIBUTED|ALL`) and the format (`VISUAL|TEXT|VERBOSE`) as CLI flags, and forward them into the `ExplainQueryStatement` here. Same knobs the SQL grammar already exposes.

ls-1801 (changes requested):
i would like to extend the EXPLAIN test to cover all major operators, this should reveal that some of them will require improvements.

the distributed explain requires support for a systest to specify a specific topology, for now you could write a repl test. I provided a diff.

nes-systests/explain/EXPLAIN.test
adrianmichalke, line 1: Did you get the differential query check support running? If not we should make sure produce meaningful error messages and track it in an issue

nes-sources/include/Sources/SourceDescriptor.hpp
```diff
@@ -65,6 +65,8 @@ class SourceDescriptor final : public Descriptor
     [[nodiscard]] Host getHost() const;
     [[nodiscard]] PhysicalSourceId getPhysicalSourceId() const;
 
+    [[nodiscard]] bool isInlineSource() const;
```
adrianmichalke, line 68: As inline sources are for testsing purposes only I would prefer to not change the descriptor

nes-sql-parser/src/StatementBinder.cpp
```diff
@@ -531,6 +531,78 @@ class StatementBinder::Impl
...
+        }
+        if (stageAst->DISTRIBUTED() != nullptr)
+        {
+            return ExplainStage::Distributed;
+        }
+        if (stageAst->ALL() != nullptr)
+        {
+            return ExplainStage::All;
+        }
+        throw InvalidStatement("Unrecognized EXPLAIN stage");
```
adrianmichalke: Wouldnt that be a bug? In this case use PRECONDTION(false, "...

nes-sql-parser/src/StatementBinder.cpp
```diff
@@ -531,6 +531,78 @@ class StatementBinder::Impl
...
+        }
+        if (formatAst->VERBOSE() != nullptr)
+        {
+            return ExplainFormat::Verbose;
+        }
+        if (formatAst->VISUAL() != nullptr)
+        {
+            return ExplainFormat::Visual;
+        }
+        throw InvalidStatement("Unrecognized EXPLAIN format");
```
adrianmichalke: Same here

nes-sql-parser/src/StatementBinder.cpp
```diff
@@ -531,6 +531,78 @@ class StatementBinder::Impl
...
+        {
+            bool stageSet = false;
+            bool formatSet = false;
+            for (auto* const optionAST : optionsAST->explainOption())
+            {
+                if (auto* const stageAST = optionAST->explainStage())
+                {
+                    if (stageSet)
+                    {
+                        throw InvalidQuerySyntax("EXPLAIN: at most one stage option (LOGICAL, OPTIMIZED, DISTRIBUTED, ALL) is allowed");
```
adrianmichalke:
```suggestion
                        throw InvalidQuerySyntax("EXPLAIN: at most one stage option is allowed");
```
Or, add it for the FORMAT option message as well. I would prefer to rm it bc it can get easily out of sync with the code

nes-systests/systest/src/SystestParser.cpp
```diff
@@ -259,17 +267,41 @@ void SystestParser::parse()
                 }
                 break;
             }
+            case TokenType::EXPLAIN: {
+                auto statement = expectQuery(DefaultQueryStopTokens);
+                readResultVerbatim = true;
```
adrianmichalke: nit: this can be a snum for better readability

nes-systests/systest/src/SystestParser.cpp
```diff
@@ -259,17 +267,41 @@ void SystestParser::parse()
...
+                /// EXPLAIN statements cannot be part of a differential block
```
adrianmichalke, line 273: Do we enforce this somewhere and throw meaningful error messages?

nes-systests/explain/EXPLAIN.test
```diff
@@ -0,0 +1,53 @@
...
+2,4,2000
+
+CREATE SINK sinkFilter(id UINT64 NOT NULL, value UINT64 NOT NULL) TYPE File;
+CREATE SINK sinkJoin(start UINT64 NOT NULL, end UINT64 NOT NULL, id UINT64 NOT NULL, value UINT64 NOT NULL, timestamp UINT64 NOT NULL, id2 UINT64 NOT NULL, value2 UINT64 NOT NULL, timestamp2 UINT64 NOT NULL) TYPE File;
+
+
+EXPLAIN (LOGICAL, FORMAT TEXT) SELECT id, value FROM stream WHERE value > UINT64(2) INTO sinkFilter;
+----
+== Initial Logical Plan ==
+SINK(SINKFILTER1)
```
adrianmichalke: Do you think we can somehow get rid of the indexing in the sink names?

adrianmichalke (changes requested):
Cool addition! 

I have some comments.

Btw, looking what duckdb i found some test cases in which they allow to specify regexes as results. That might be a more efficient way to write EXPLAIN test cases in the future.

See: https://github.com/duckdb/duckdb/blob/main/test/sql/explain/test_explain_table_scan.test
https://github.com/duckdb/duckdb/blob/main/test/sql/explain/test_explain_limit.test
https://github.com/duckdb/duckdb/blob/main/test/sql/explain/test_explain.test

nes-systests/systest/src/SystestBinder.cpp
greenBene: Good catch! Thanks!

nes-sql-parser/AntlrSQL.g4
greenBene: I fixed it for the mentioned keywords. @Artraxon is currently working on the binder/grammar to make it more consistent.

nes-systests/systest/src/SystestParser.cpp
greenBene: Good catch!

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: In my understanding, fields return `QualifiedIdentifierBase<1>`  for `getFullyQualifiedName()`  which is basically the same as `getLastName()` but with extra steps. Thus, I'd leave it as is right now. For actual computing, the order doesn't matter anyway, just for consistency in the output. :)

nes-sources/include/Sources/SourceDescriptor.hpp
greenBene: Good point. I included your suggestion.

nes-systests/explain/EXPLAIN.test
greenBene: Added tests ensuring each operator is at least explained once, and created backlog ticket to improve `::explain` output (#1733 )

greenBene: Good idea! Added the test.

nes-sql-parser/include/SQLQueryParser/StatementBinder.hpp
greenBene: Created follow up issue #1735

nes-systests/explain/EXPLAIN.test
greenBene: Differential support is not yet included. I throw an error message via an invariant when parsing a DIFFERENTIAL token right after an explain query was read. 


Created backlog issue #1739, because I personally think this feature is not critical, but nice to have. If you think it should be part of this PR, let me know! :)

nes-sources/include/Sources/SourceDescriptor.hpp
greenBene: Network sources also use them in multi-worker deployments. 
Calling it "InlineSource" might be misleading, but that is the terminology we're using. I suggest leaving it as is and renaming it if necessary in the future.

nes-sql-parser/src/StatementBinder.cpp
greenBene: In the grammar, as it was when you reviewed the code, yes! But after discussing the grammar changes with Leonhard, we switched to identifiers and thus must catch this case with a meaningful error message.

nes-sql-parser/src/StatementBinder.cpp
greenBene: Same as above

nes-sql-parser/src/StatementBinder.cpp
greenBene: Good point, but the code was removed because of the grammar change

nes-systests/systest/src/SystestParser.cpp
greenBene: Changed it to use an enum.

nes-systests/systest/src/SystestParser.cpp
greenBene: Yes, via INVARIANT (see comment above)

nes-systests/explain/EXPLAIN.test
greenBene: Great point! :) I found a way by creating, in addition to the numbered sinks, also the sinks with the default names.

nes-systests/explain/EXPLAIN.test
```diff
@@ -0,0 +1,119 @@
+# name: explain/EXPLAIN.test
+# description: Validates the output of EXPLAIN statements.
+# groups: [explain]
+
+CREATE LOGICAL SOURCE stream(id UINT64 NOT NULL, value UINT64 NOT NULL, timestamp UINT64 NOT NULL);
+CREATE PHYSICAL SOURCE FOR stream TYPE File;
+ATTACH INLINE
+1,1,1000
+2,3,2000
```
adrianmichalke, line 9: I guess there is no easy way to not having to specify input tuples?
We are not using them here at all

adrianmichalke (commented):
LGTM

adrianmichalke (approved):

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -95,6 +95,31 @@ class SLTSinkFactory
         const Schema<UnqualifiedUnboundField, Ordered>& schema,
         const std::unordered_map<Identifier, std::string>& /*config*/)
     {
+        std::unordered_map<Identifier, std::string> config{};
+        std::unordered_map<Identifier, std::string> formatConfig{};
+        if (sinkType == Identifier::parse("File"))
+        {
+            config[Identifier::parse("file_path")] = "/temp/none.txt";
```
ls-1801: Typo: `/temp/none.txt` should be `/tmp/none.txt` (also on line 107 for the CHECKSUM branch).

nes-sql-parser/src/StatementBinder.cpp
```diff
@@ -531,6 +532,83 @@ class StatementBinder::Impl
...
     }
 
+    std::unordered_set<ExplainStage> bindExplainStages(AntlrSQLParser::ExplainStagesContext* stagesContext) const
+    {
+        std::unordered_set<ExplainStage> explainStages;
+
+        for (auto* const stageContext : stagesContext->explainStage())
+        {
+            if (stageContext->LOGICAL() != nullptr)
+            {
```
ls-1801, line 542: `bindExplainStages` uses `unordered_set::insert`, which silently deduplicates. The PR body says duplicate stages should be rejected — as written, `EXPLAIN (LOGICAL, LOGICAL) ...` is accepted. Track a `std::vector<ExplainStage>` (or check `insert().second`) and `throw InvalidStatement` on a duplicate.

nes-systests/systest/src/SystestParser.cpp
```diff
@@ -472,6 +507,23 @@ std::vector<std::string> SystestParser::expectTuples(const bool ignoreFirst)
...
+std::vector<std::string> SystestParser::expectVerbatimResultLines()
+{
+    INVARIANT(currentLine < lines.size(), "current line to parse should exist: {}", currentLine);
+    std::vector<std::string> resultLines;
+    /// skip the result line `----`
+    if (toLowerCase(lines[currentLine]) == toLowerCase(ResultDelimiter))
+    {
+        currentLine++;
+    }
+    while (currentLine < lines.size() && !lines[currentLine].empty())
```
ls-1801: `expectVerbatimResultLines` terminates on the first blank line, so any EXPLAIN whose output contains a blank line cannot be tested here. In particular, `EXPLAIN (DISTRIBUTED) ...` (and `EXPLAIN (ALL) ...`) will not work as a systest fixture because the plan text separates per-worker sections and stage headers with blank lines — the expected block would silently truncate at the first one. Not blocking for this PR (current tests only use single-section LOGICAL/OPTIMIZED), but worth documenting or switching to an explicit end marker (`----`, `==END==`) before adding multi-section coverage.

nes-sql-parser/src/StatementBinder.cpp
```diff
@@ -531,6 +532,83 @@ class StatementBinder::Impl
...
+        }
+        if (parsedFormatIdentifier == Identifier::parse("verbose"))
+        {
+            return ExplainFormat::Verbose;
+        }
+        if (parsedFormatIdentifier == Identifier::parse("visual"))
+        {
+            return ExplainFormat::Visual;
+        }
+        throw InvalidStatement("Unrecognized EXPLAIN format");
```
ls-1801: Include the offending identifier so users can spot typos (e.g. `FORMAT TXET`):

```suggestion
        throw InvalidStatement("Unrecognized EXPLAIN format: {}", parsedFormatIdentifier);
```

Matches the pattern used above for `bindExplainStages` (`"Unrecognized EXPLAIN stage: ", identifier`).

nes-sql-parser/src/StatementBinder.cpp
```diff
@@ -531,6 +532,83 @@ class StatementBinder::Impl
...
+                {
+                    explainStages.insert(ExplainStage::Optimized);
+                }
+                else if (identifier == Identifier::parse("distributed"))
+                {
+                    explainStages.insert(ExplainStage::Distributed);
+                }
+                else
+                {
+                    throw InvalidStatement("Unrecognized EXPLAIN stage: ", identifier);
```
ls-1801: missing `{}` format placeholder

ls-1801 (changes requested):
Will approve on re-request

nes-systests/systest/src/SystestBinder.cpp
greenBene: Fixed

nes-sql-parser/src/StatementBinder.cpp
greenBene: Good catch! I applied your suggested change and added a unit test for this

nes-systests/systest/src/SystestParser.cpp
greenBene: Good point, I changed the logic to 
```cpp
while (currentLine < lines.size() && !lines[currentLine].starts_with("==END=="))
```

nes-sql-parser/src/StatementBinder.cpp
greenBene: Applied your fix.

nes-sql-parser/src/StatementBinder.cpp
greenBene: Good catch! :)

ls-1801 (approved):

PR #1734, opened 2026-07-06
fix(inference): share IREE instance + device to remove TSan races

ls-1801:
## Summary
- Share one IREE runtime instance + HAL device process-wide via a Meyers singleton (`sharedRuntime()`) in `nes-inference/runtime/src/InferenceRuntime.cpp`. IREE registers the VM builtin type registry per instance without internal synchronization (`iree_vm_instance_register_type`), so creating an instance per worker thread raced; with one shared instance those globals register exactly once at singleton construction.
- Serialize each setup's `iree_runtime_session_create_with_device` **and** `iree_runtime_session_append_bytecode_module_from_memory` under a single `std::mutex`. Session creation writes the shared `device->topology_info` (`iree_hal_sync_device_assign_topology_info`, `sync_device.c:249`), and bytecode append triggers executable creation, which — via `iree_hal_executable_environment_initialize` — **unconditionally re-runs `iree_cpu_initialize` on every executable creation** against the unguarded `iree_cpu_data` global (not a once-only lazy init, so warming it up at singleton construction is not sufficient). Neither write is synchronized by IREE, so both paths must be serialized. Setup runs once per query registration (off the hot path); inference itself stays fully lock-free.
- Drop the two `race:iree_*` entries from `.github/.env/tsan_suppressions.txt` and their justification block — the underlying races are gone rather than suppressed.

Motivating symptom: CI's `x64 libstdcxx RelWithDebInfo thread (unit+systest)` job occasionally hit silent SIGABRT during the `InferModel` systests under `-n 20` + `slice_cache=true`. TSan was suppressing the two known IREE races so nothing surfaced in the ctest log; the abort tracked back to concurrent first-touch of the same IREE process globals.

## Verification
- `-DUSE_SANITIZER=thread`, RelWithDebInfo, `.github/.env/test-thread.env` sourced but with the two IREE suppressions removed.
- Full CI-flavor sweep matching the failing job (COMPILER, `-n 20`, slice cache enabled) → 519/519 systest queries pass, exit 0, no TSan warnings on any IREE path. INTERPRETER + HASH_JOIN also clean.
- Serializing only session creation was **not** enough: the Inference systest group still raced on the unguarded `iree_cpu_initialize` during concurrent bytecode append (repro: 34/200 iterations aborted with exit 66). Widening the lock to also cover the append eliminated it.

## Test plan
- [ ] `USE_SANITIZER=thread` CI job (`x64 libstdcxx RelWithDebInfo thread`) is green.
- [ ] Reviewer sanity check on the sharing model: IREE HAL devices are documented as safe for concurrent submission, and `iree_runtime_session_create_with_device` accepts a shared device pointer; the only non-synchronized writes on the shared device/globals happen during setup, which is fully serialized.

### DIFF
--- a/.github/.env/tsan_suppressions.txt
+++ b/.github/.env/tsan_suppressions.txt
@@ -19,8 +19,3 @@ race:__shared_weak_count
 race:__assoc_state
 race:__assoc_sub_state
 
-# IREE registers global type tables and detects CPU features lazily on first instance/session/module load, and that init is not
-# thread-safe. InferenceRuntime warms up a shared instance+session to do most of it once, but each bytecode module's executable/CPU
-# init still registers lazily, racing across worker threads. The writes are idempotent (same values), so they are benign.
-race:iree_vm_instance_register_type
-race:iree_cpu_initialize
--- a/nes-inference/runtime/src/InferenceRuntime.cpp
+++ b/nes-inference/runtime/src/InferenceRuntime.cpp
@@ -17,6 +17,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <memory>
+#include <mutex>
 #include <string>
 #include <utility>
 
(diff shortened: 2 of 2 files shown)


Artraxon (approved):
I don't mind, but we should get input from  @gaturchenko as well

gaturchenko (approved):
LGTM @ls-1801 @Artraxon

PR #1736, opened 2026-07-07
Chained Hash Map redesign

ThanGeo:
This pull requests refactors the ChainedHashMap so that it uses TupleBuffer in a more structured way. 

## Requirements
This branch builds on top of the paged_vector_redesign branch. When that branch is merged, then this one should follow. 

## Documentation
The ChainedHashMap did not use our self-contained, single-TupleBuffer-based data structure model. This redesign aims to fix this, enabling easier seralization, migration, representation and memory optimization for the ChainedHashMap. 

The HashMapSlice was also slightly redesigned, since it contains hash maps based on the number of workers and input streams. 

image

Previous PR: [https://github.com/nebulastream/nebulastream/pull/1667](url)


nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ls-1801, line 1: we can probably pull this code into a shared test utility for other upcoming datastructures.

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -12,180 +12,1001 @@
...
+                    [&](const nautilus::val<AbstractHashMapEntry*>& newEntry)
+                    {
+                        const auto chainedEntry = static_cast<nautilus::val<ChainedHashMapEntry*>>(newEntry);
+                        const ChainedHashMapRef::ChainedEntryRef newEntryRef(chainedEntry, chainedHashMapBuffer, fieldKeys, fieldValues);
+                        newEntryRef.copyValuesToEntry(record, bm);
+                    },
+                    bm);
+            })));
+
+        readAtFn.emplace(engine->registerFunction(std::function(
```
ls-1801: what is read at? why is the index numerical? The testable interface should look like this:
```
testableMap.at(AnyVec key) -> std::optional<AnyVec> value
testableMap.put(AnyVec key, AnyVec value)
```

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ls-1801, line 1: In general the file is to much copy paste of the pagedvector.
Effectivly we want to compare map<AnyVec, AnyVec> with our chained hashmap and map<AnyVec,vector<AnyVec>> with our chm.

So you need a TestableHashMap that has at(key), put(key, value), and maybe a getAll to compare iterators

ls-1801 (changes requested):
mostly looked at the tests so far

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ThanGeo: Done, but i'm thinking maybe that eventually for more property testing on data structures we should find some better way to abstract/interface this anyvec code for correctness more easily.

ThanGeo: done

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ThanGeo: you're right. Fixed it as per your suggestion

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
```diff
@@ -64,53 +66,114 @@ class ChainedHashMapEntry final : public AbstractHashMapEntry
...
+
+    [[nodiscard]] const Header& header() const { return *buffer.getAvailableMemoryArea<const Header>().data(); }
+
+    [[nodiscard]] std::span<ChainedHashMapEntry*> chains()
+    {
+        auto* data = buffer.getAvailableMemoryArea<uint8_t>().data();
+
+        /// NOLINTBEGIN(cppcoreguidelines-pro-type-reinterpret-cast, cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        auto* entries = reinterpret_cast<ChainedHashMapEntry**>(data + sizeof(Header));
+        return {entries, getNumberOfChains() + 1};
```
ls-1801, line 178: All current callers allocate this buffer through `calculateBufferSizeFromBuckets/Chains`, so the layout should fit. Since `chains()` relies on that implicit pairing, could we add a `PRECONDITION` or `INVARIANT` in `init` checking that the buffer contains `Header` plus `numChains + 1` pointers? That would encode the layout contract and protect future serialization/migration call sites.

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
```diff
@@ -56,16 +61,10 @@ uint64_t calcCapacity(const uint64_t numberOfKeys, const double loadFactor)
...
-    , pageSize(pageSize)
-    , entrySize(entrySize)
-    , entriesPerPage(pageSize / entrySize)
-    , numberOfChains(calcCapacity(numberOfBuckets, assumedLoadFactor))
-    , entries(nullptr)
-    , mask(numberOfChains - 1)
-    , destructorCallBack(nullptr)
+void ChainedHashMap::init(TupleBuffer& tupleBuffer, const uint64_t entrySize, const uint64_t numberOfBuckets, const uint64_t pageSize)
 {
+    const uint64_t entriesPerPage = pageSize / entrySize;
```
ls-1801, line 67: `entrySize` is used as a divisor before the `entrySize > 0` precondition runs, so `init(..., entrySize = 0, ...)` hits division by zero instead of the intended contract failure. Could we move the precondition ahead of the `entriesPerPage` calculation?

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMapRef.cpp
```diff
@@ -269,57 +268,75 @@ ChainedHashMapRef::EntryIterator ChainedHashMapRef::begin() const
     const nautilus::val<uint64_t> indexOnPage = 0;
     const nautilus::val<uint64_t> pageIndex = 0;
     const auto currentEntry = nautilus::invoke(
-        +[](const HashMap* hashMap, const uint64_t pageIndexVal, const uint64_t indexOnPageVal)
+        +[](TupleBuffer* tupleBuffer, const uint64_t pageIndexVal, const uint64_t indexOnPageVal)
         {
-            const auto& page = dynamic_cast<const ChainedHashMap*>(hashMap)->getPage(pageIndexVal);
+            const auto chm = ChainedHashMap::load(*tupleBuffer);
+            const auto& page = chm.getPage(pageIndexVal);
```
ls-1801, line 283: An initialized empty map has no storage page yet, but `begin()` unconditionally loads page 0. The following regression test builds on this branch and currently aborts with exit code 134 (`terminate called without an active exception`):

```cpp
TEST(ChainedHashMapIteratorTest, emptyMapIsAnEmptyRange)
{
    constexpr uint64_t bufferSize = 4096;
    constexpr uint64_t numberOfBuckets = 1;
    constexpr uint64_t entrySize = sizeof(ChainedHashMapEntry);
    constexpr uint64_t entriesPerPage = 1;

    auto bufferManager = BufferManager::create(bufferSize, 16);
    auto hashMapBuffer = bufferManager->getUnpooledBuffer(
        ChainedHashMap::calculateBufferSizeFromBuckets(numberOfBuckets)).value();
    ChainedHashMap::init(hashMapBuffer, entrySize, numberOfBuckets, entrySize * entriesPerPage);

    auto engine = makeEngine(EngineMode::Interpreter);
    auto iterate = engine.registerFunction(std::function(
        [](nautilus::val<TupleBuffer*> buffer)
        {
            const ChainedHashMapRef ref(
                buffer, {}, {}, nautilus::val<uint64_t>{entriesPerPage}, nautilus::val<uint64_t>{entrySize});
            for (const auto entry : ref)
            {
                std::ignore = entry;
            }
        }));

    EXPECT_NO_THROW(iterate(&hashMapBuffer));
}
```

Could `begin()` return an iterator equal to `end()` when the map has zero records, without loading a page? This would also let the test remove its existing caller-side empty-map guard.

ls-1801 (changes requested):

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
ThanGeo: added a guard in init() that takes the chains into account for the size check

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
ThanGeo: moved the precondition before using the entrySize value

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMapRef.cpp
ThanGeo: merged all invoked by introducing a new struct that wraps the two unsigned long values together. Added a check in there about ```numPages = 0``` and returns ```end()``` if true.

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -12,180 +12,558 @@
...
+                }
+            }
+            return true;
+        });
+    return it == reference.end() ? nullptr : &it->second;
 }
 
-TEST_P(ChainedHashMapTest, update)
+/// Builds a reference association list tracking only the first-seen value per unique key,
+/// mirroring CHM's first-write-wins deduplication, and put()s every generated pair into the map under test.
```
ls-1801: Could you please document all behaviors in which our chained hash map differs from std::unordered_map?

This first-write-wins is new to me.

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ls-1801: You mean code like this? Yeah, we could move them into the util

```c++
                Record record;
                /// static_val ensures each field iteration gets a distinct trace tag.
                for (nautilus::static_val<uint64_t> i = 0; i < dataTypes.size(); ++i)
                {
                    record.write(projections[i], buildVarVal(rec, i, dataTypes[i]));
                }
```

or the reverse:

```c++
for (nautilus::static_val<uint64_t> i = 0; i < fieldValues.size(); ++i)
                    {
                        const auto& fld = fieldValues[i];
                        storeVarValToAnyVec(out, static_cast<uint64_t>(i), valueRecord.read(fld.fieldIdentifier), fld.type);
                    }
```

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -12,180 +12,558 @@
...
+        const auto valueProjections = std::vector<Record::RecordFieldIdentifier>(splitPoint, projections.end());
+        auto [fk, fv]
+            = ChainedEntryMemoryProvider::createFieldOffsets(createSchemaFromDataTypes(fieldTypes), keyProjections, valueProjections);
+        return {.keys = std::move(fk), .values = std::move(fv)};
     }
+};
 
-    /// Now we are searching for the entries and checking if the values are correct.
-    checkIfValuesAreCorrectViaFindEntry(hashMap, exactMap);
+using KeyValueReference = std::vector<std::pair<AnyVec, AnyVec>>;
```
ls-1801: What is the reason we are comparing against a std::vector?

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -12,180 +12,558 @@
...
+        numEntriesPerPage,
+        fmt::join(fieldTypes, ", "));
+
+    auto bufferManager = BufferManager::create(bufferSize, pooledBufferCountFor(bufferSize));
+    TestableChainedHashMap chainedHashMap(fieldTypes, *bufferManager, mode, numberOfBuckets, numKeyFields, numEntriesPerPage);
+    const auto reference = buildReferenceAndPopulate(chainedHashMap, fieldTypes, numberOfItems);
+
+    NES_INFO("putAndLookupKeys: CHM has {} entries, {} unique keys", chainedHashMap.size(), reference.size());
+    RC_ASSERT(chainedHashMap.size() == reference.size());
+    verifyLookups(chainedHashMap, reference, fieldTypes);
```
ls-1801: You could run this in a loop.

HashMap m
for(num_iterations) {
m.put_some_value();
m.verify_lookups();
}

That way we also test interleaving, and it's not just write-then-read.

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -12,180 +12,558 @@
...
-        return ss.str();
-    });
+    }
+}
+
+/// Verify put()/at() round-trip: every reference key is found with its first-seen value, and independently
+/// generated keys that don't collide with the reference correctly report absent.
+void putAndLookupKeysProperty(EngineMode mode)
+{
+    const auto fieldTypes = *genDataTypeSchema(ALL_VALUE_TYPES, 1, MAX_SCHEMA_FIELDS);
```
ls-1801: Not quite sure what currently happens if we don't have any fields. I.e, just using the HashMap as a HashSet

nes-physical-operators/src/HashMapSlice.cpp
```diff
@@ -21,61 +21,74 @@
...
+                createNewHashMapSliceArgs.valueSize,
+                createNewHashMapSliceArgs.numberOfBuckets,
+                createNewHashMapSliceArgs.pageSize);
+            hashMapBuffers.emplace_back(childBuffer.value());
+        }
+        else
+        {
+            throw BufferAllocationFailure("No unpooled TupleBuffer available for chained hash map child buffer!");
+        }
     }
```
ls-1801: We decided to keep the vector in the slice for now! Is there a reason we do the allocation upfront now?
For a large number of worker threads and tiny windows, this can be quite slow and wasteful. I think we can get away with reserving the number of slots in the vector, but have each thread construct its buffer on demand. Maybe we switch to a std::vector<std::optional<TupleBuffer>> instead, or we keep the buffer default-constructed (empty).

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
```diff
@@ -64,53 +66,114 @@ class ChainedHashMapEntry final : public AbstractHashMapEntry
...
+            , storageSpaceIndex(TupleBuffer::INVALID_CHILD_BUFFER_INDEX_VALUE)
+            , varSizedSpaceIndex(TupleBuffer::INVALID_CHILD_BUFFER_INDEX_VALUE)
+        {
+        }
+    };
+
+    static_assert(std::is_trivially_destructible_v<Header>, "Header must be trivially destructible");
+
+    /// Helper util methods for safe access
+    [[nodiscard]] Header& header() { return *buffer.getAvailableMemoryArea<Header>().data(); }
```
ls-1801, line 168: You should probably assert the `VALID_CHM` somewhere; might as well put it here.

ls-1801 (changes requested):

zeuchste:
We benchmark hash-join storage/processing variants on this codebase (edge-join design-space study) and ran this PR's head (d85467f6b5) through our workload battery — 23 points (window sizes 10ms–100s, duplicates 0.1–32/key, 2–32 cores, 0–8 co-running queries), keyed default join config, 2 reps each vs. base 609d016fcb, same nodes, checksum-verified. Sharing the results since they may be useful input:

**Correctness:** every run reproduced the exact golden checksum. 👍

**Throughput:** consistent **−10–16% steady-state time in the mid-range** (1s–100s windows, low-to-mid duplication) — the redesign measurably reduces the per-key allocation cost that dominates the current map on many-key workloads. Three regressions at the extremes:
- 10ms windows (2.5M windows): **+26%** — per-window map setup appears heavier;
- 2-core pinned: +9%, unpinned 64-CPU: +7%;
- benefit vanishes at heavy duplication (32/key: −2%).

**Tail latency under concurrency:** with ≥4 co-running queries, trigger→sink latency explodes: p50 34ms → 7.3s (n=4) / 15.2s (n=8), p99 up to 66s, while throughput still improves. Caveat: we could not pin memory config equal (see below), so the new default memory bound may be a confound — but worth a look before merge.

Two harness observations from testing (possibly main, not this PR):
1. Passing `--worker.total_memory_in_bytes=<any value>` through systest's `--` passthrough exits with code 232 and no error message.
2. When the disable-config filters out all selected tests, systest prints its banner and exits with success-like output — an A/B run of ours silently executed zero queries.

Happy to share the full per-point table and repro scripts.

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ThanGeo: Added some elaboration on how the HashMap works overall

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ThanGeo: Switched to comparing against a std::unordered_map

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ThanGeo: Done

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
ThanGeo: HashSet was already supported internally, i added a property test for it too

nes-physical-operators/src/HashMapSlice.cpp
ThanGeo: switched back to lazy initialization of buffers. I associated a new vector of flags along with the vector of tuplebuffers to keep track of what buffers have been initialized. I know it's not the prettiest solution, but If i had to switch to storing children buffers instead in order to use the INVALID_CHILD_BUFFER_INDEX placeholder here it would re-introduce the complexity that we got rid of by reverting back to the vector again.

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
ThanGeo: I assert it in the load() method of the chained hash map. The status flag is always the first field in the header(), so if you load a trash/unrelated buffer then you can catch this in the load() method. The header() is only used internally, after a load

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
```diff
@@ -80,18 +79,34 @@ ChainedHashMap::ChainedHashMap(uint64_t entrySize, const uint64_t numberOfBucket
...
+    PRECONDITION(
+        tupleBuffer.getBufferSize() >= sizeof(Header) + (numberOfChains + 1) * sizeof(ChainedHashMapEntry*),
+        "Buffer of size {} is not big enough to hold the header ({} bytes) plus {} chain pointers ({} bytes each)",
+        tupleBuffer.getBufferSize(),
+        sizeof(Header),
+        numberOfChains + 1,
+        sizeof(ChainedHashMapEntry*));
+
+    /// Create object
+    ChainedHashMap chm(tupleBuffer);
```
keyseven123: Use `{}` for constructors

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
```diff
@@ -106,156 +121,277 @@ ChainedHashMap::ChainedHashMap(const uint64_t keySize, const uint64_t valueSize,
...
+        const uint64_t lastPageSize = lastPage.getBufferSize();
+        const uint64_t lastPageNumTuples = lastPage.getNumberOfTuples();
+        needNewPage = (lastPageNumTuples + neededSize > lastPageSize);
+    }
+
+    if (needNewPage)
+    {
+        /// Currently if a single varsized entry can not fit in one page we throw an exception
+        PRECONDITION(
+            neededSize <= bufferProvider->getBufferSize(),
```
keyseven123: But should we not rather use the unpooled buffers instead of having this hard limit of varsized data in hash maps?

nes-nautilus/include/Interface/HashMap/HashMapRef.hpp
```diff
@@ -59,7 +61,7 @@ class HashMapRef
     virtual nautilus::val<AbstractHashMapEntry*> findEntry(const nautilus::val<AbstractHashMapEntry*>& otherEntry) = 0;
 
 protected:
-    nautilus::val<HashMap*> hashMapRef;
+    nautilus::val<TupleBuffer*> tupleBuffer;
```
keyseven123: please rename this member. We also discussed that we want to go away from `tupleBuffer`. What about simply `buffer` or `hashmap`?

nes-nautilus/tests/TestUtils/src/DataStructureTestUtils.cpp
```diff
@@ -0,0 +1,498 @@
...
+#include <static.hpp>
+
+namespace NES::TestUtils
+{
+
+std::shared_ptr<BufferManager> createBufferManager(const uint64_t bufferSize, const uint64_t numberOfBuffers)
+{
+    constexpr double unpooledMemoryFraction = 0.9;
+    constexpr BufferAlignment alignment{64};
+    const size_t totalMemoryInBytes = 10 * numberOfBuffers * bufferSize;
```
keyseven123, line 56: Is there a reason for this `10`? If so then add a comment. Otherwise, people are tempted to remove it

nes-nautilus/tests/TestUtils/src/DataStructureTestUtils.cpp
```diff
@@ -0,0 +1,498 @@
...
+}
+
+Schema<QualifiedUnboundField, Ordered> createSchemaFromDataTypes(const std::vector<DataType>& dataTypes)
+{
+    return views::enumerate(dataTypes)
+        | std::views::transform(
+               [](const auto& pair)
+               {
+                   const auto& [typeIdx, dataType] = pair;
+                   const auto name = Record::RecordFieldIdentifier(Identifier::parse("field" + std::to_string(typeIdx)));
```
keyseven123:
```suggestion c++
const Record::RecordFieldIdentifier name{Identifier::parse("field" + std::to_string(typeIdx))};
```

nes-nautilus/tests/TestUtils/src/DataStructureTestUtils.cpp
```diff
@@ -0,0 +1,498 @@
...
+                   return QualifiedUnboundField{name, dataType};
+               })
+        | std::ranges::to<Schema<QualifiedUnboundField, Ordered>>();
+}
+
+rc::Gen<std::vector<DataType>> genDataTypeSchema(std::span<const DataType::Type> pool, size_t minFields, size_t maxFields)
```
keyseven123: not a fan of naming it pool. What about `possibleDataTypes`?

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -12,180 +12,671 @@
...
+    types.reserve(fields.size());
+    for (const auto& field : fields)
+    {
+        names.push_back(field.fieldIdentifier);
+        types.push_back(field.type);
+    }
+    return {std::move(names), std::move(types)};
+}
+
+class TestableChainedHashMap
```
keyseven123: I think we should move this into its own class as it is quite large. We already have a custom `DataStructureTestUtils` so another `TestableChainedHashmap` is fine. Then this file here can take care of only different testing cases  / scenarios.

nes-nautilus/tests/UnitTests/ChainedHashMapTest.cpp
```diff
@@ -12,180 +12,671 @@
...
+        if (not chainedHashMapBufferOpt.has_value())
+        {
+            throw BufferAllocationFailure(
+                "No unpooled TupleBuffer of size {} available!", ChainedHashMap::calculateBufferSizeFromBuckets(numberOfBuckets));
+        }
+        chainedHashMapBuffer = chainedHashMapBufferOpt.value();
+        ChainedHashMap::init(chainedHashMapBuffer, entrySize, numberOfBuckets, pageSize);
+
+        const auto numKeyProjections = keyDataTypes.size();
+        const auto keyProjections = std::vector<Record::RecordFieldIdentifier>(
```
keyseven123: So we assume that the projections for the key are always the first in the projections? Not sure if this is the best we can do... Can we at least have a precondition that checks if the projections is large enough?

nes-physical-operators/src/HashMapSlice.cpp
ls-1801: sounds like what a std::vector<std::optional<TupleBuffer>> would also do, without a second allocation

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
```diff
@@ -59,27 +60,12 @@
 #include <fmt/ranges.h>
 #include <rapidcheck/gtest.h>
 
-/// NOLINTBEGIN(misc-include-cleaner, bugprone-unchecked-optional-access)
+/// NOLINTBEGIN(misc-include-cleaner, bugprone-unchecked-optional-access, google-build-using-namespace)
```
keyseven123: I am not a fan of this, as these are good clang-tidy warnings. Is there a bigger reason why you disabled them?

nes-physical-operators/include/Join/NestedLoopJoin/NLJSlice.hpp
```diff
@@ -68,6 +68,7 @@ class NLJSlice final : public Slice
     void combinePagedVectors();
 
 private:
+    /// This does not really follow our compact-buffer data structure logic
```
keyseven123: So should we not change it then?

ls-1801 (approved):

nes-physical-operators/src/Aggregation/AggregationOperatorHandler.cpp
```diff
@@ -104,13 +105,18 @@ void AggregationOperatorHandler::triggerSlices(
         /// - all pointers to all hashmaps of the window to be triggered
         /// - a new hashmap for the probe operator, so that we are not overwriting the thread local hashmaps
         /// - size of EmittedAggregationWindow
```
keyseven123: Please update this comment

nes-physical-operators/src/Join/HashJoin/HJBuildPhysicalOperator.cpp
```diff
@@ -49,53 +49,6 @@
...
-    /// ReSharper disable once CppPassValueParameterByConstReference
-    /// NOLINTBEGIN(performance-unnecessary-value-param)
-    auto* const operatorHandler = dynamic_cast<HJOperatorHandler*>(
-        nautilus::details::RawValueResolver<OperatorHandler*>::getRawValue(executionCtx.getGlobalOperatorHandler(operatorHandlerId)));
-    if (operatorHandler->wasSetupCalled(joinBuildSide))
-    {
-        return;
-    }
-
-    const auto cleanupStateNautilusFunction
```
keyseven123, line 69: so nice to see this finally gone :)

nes-physical-operators/src/Join/HashJoin/HJBuildPhysicalOperator.cpp
```diff
@@ -133,32 +90,40 @@ void HJBuildPhysicalOperator::execute(ExecutionContext& ctx, Record& record) con
...
+                    +[](TupleBuffer* hashMapBuf, uint32_t* valueMemArea, AbstractBufferProvider* bufferProvider, uint64_t tupleSize) -> void
                     {
                         if (auto pagedVectorBuffer = bufferProvider->getUnpooledBuffer(PagedVector::getMainBufferSize()))
                         {
-                            /// initialize paged vector buffer
                             PagedVector::init(pagedVectorBuffer.value(), bufferProvider->getBufferSize(), tupleSize);
-                            /// @warning: this will be refactored again during the ChainedHashMap refactor
-                            new (pagedVectorBufferMemArea) TupleBuffer(pagedVectorBuffer.value());
+                            auto childIndex = hashMapBuf->storeChildBuffer(pagedVectorBuffer.value());
+                            *valueMemArea = childIndex.getRawIndex();
```
keyseven123, line 103: Please add here why we are storing the raw index at it.

keyseven123 (changes requested):
Generally, I am really happy that we finally get rid of the cleanup methods.

nes-physical-operators/src/HashMapSlice.cpp
ThanGeo: Should I switch to that?

nes-nautilus/tests/TestUtils/src/DataStructureTestUtils.cpp
ThanGeo: it is the exact same thing being used in a lot of other tests as well, e.g. EmitPhysicalOperatorTest.cpp:72 (commited by you).

nes-physical-operators/include/Join/NestedLoopJoin/NLJSlice.hpp
ThanGeo: Not really because the NLJSlice is not a data structure. It added significant complexity and overhead trying to model it that way with no real benefits

nes-physical-operators/src/Join/HashJoin/HJBuildPhysicalOperator.cpp
ThanGeo: This gets overwritten by the ChildBufferIndex PR. But the idea is you only care about the number of the child buffer and write it in a nautilus-wrappable type

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMap.cpp
ThanGeo: You're right, fixed it in the same way as in the PagedVector (#1814)

nes-nautilus/tests/TestUtils/src/DataStructureTestUtils.cpp
ThanGeo: kept the pool but made it more explicit

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
ThanGeo: fixed it

keyseven123 (approved):

PR #1743, opened 2026-07-09
Allow raw numeric literals in SQL

adrianmichalke:
With this change we simplify our SQL syntax to not require type literals for numerics.

Before type constructors where always required:

SELECT speed * **FLOAT32(3.6)** AS speed_m_sec FROM s INTO sink;

Now also allowed:

SELECT speed * **3.6** AS speed_m_sec FROM s INTO sink;

### Inference Rules

Inference rules follow [Clickhouse](https://clickhouse.com/docs/sql-reference/syntax#literals).

NebulaStream infers a raw numeric literal's data type from its value:
- Integer literals without a leading minus sign use the smallest unsigned integer type that can represent the value: `UINT8`, `UINT16`, `UINT32`, or `UINT64`.
- Negative integer literals use the smallest signed integer type that can represent the value: `INT8`, `INT16`, `INT32`, or `INT64`.
- Floating-point literals, including fractional and exponent notation such as `0.1`, `42.0`, `.5`, and `1E3`, use `FLOAT64`.

Use an explicit type constructor when a query depends on an exact literal type:

```sql
SELECT UINT64(1) AS id, FLOAT32(3.6) AS scale FROM s INTO sink
```


nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -174,6 +176,102 @@ static LogicalFunction createLogicalBinaryFunction(LogicalFunction leftFunction,
...
+    else
+    {
+        constantValue.insert(0, 1, '-');
+    }
+
+    const auto dataType = constantFunction.getDataType().isFloat() ? DataTypeProvider::provideDataType(DataType::Type::FLOAT64)
+                                                                   : inferIntegerLiteralDataType(constantValue);
+    return ConstantValueLogicalFunction(dataType, std::move(constantValue));
+}
+
```
ls-1801: Is there a reason we have to infer the data type from the parser and not from the LogicalConstantFunction? 

I think passing the DataType via the constructor is just an artifact from an earlier version. The DataType should be derived during the inferSchema check, which can analyze the string.

nes-physical-operators/src/Functions/FunctionProvider.cpp
```diff
@@ -91,9 +103,9 @@ PhysicalFunction FunctionProvider::lowerConstantFunction(const ConstantValueLogi
...
     {
         case DataType::Type::UINT8:
-            return ConstantUInt8ValueFunction(parseConstantValue<int8_t>(stringValue));
+            return ConstantUInt8ValueFunction(parseSmallUnsignedConstantValue<uint8_t>(stringValue));
         case DataType::Type::UINT16:
-            return ConstantUInt16ValueFunction(parseConstantValue<int16_t>(stringValue));
+            return ConstantUInt16ValueFunction(parseSmallUnsignedConstantValue<uint16_t>(stringValue));
         case DataType::Type::UINT32:
             return ConstantUInt32ValueFunction(parseConstantValue<uint32_t>(stringValue));
         case DataType::Type::UINT64:
```
ls-1801, line 99: This looks odd. What's the reason for that? Why do we handle uint8 and uint16 differently from uint32?

We have a from_chars<T> in our util that would also reject out-of-bounds values; you should use a wrapper around that to create the exception string.

ls-1801 (changes requested):

nes-physical-operators/src/Functions/FunctionProvider.cpp
adrianmichalke: Removed the special parseSmallUnsignedConstantValue helper again

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
adrianmichalke: Moved the main logic now into the data type

ls-1801 (approved):

nes-logical-operators/src/Functions/ConstantValueLogicalFunction.cpp
```diff
@@ -14,26 +14,92 @@
...
+        }
+        return inferUnsignedIntegerLiteralDataType(static_cast<uint64_t>(value));
+    }
+    return inferUnsignedIntegerLiteralDataType(from_chars_with_exception<uint64_t>(literal));
+}
+
+DataType inferConstantValueDataType(const std::string_view literal)
+{
+    if (literal.find('.') != std::string_view::npos or literal.find('E') != std::string_view::npos
+        or literal.find('e') != std::string_view::npos)
```
alepping: nit: We could also 'try' to cast to integer first and try double if that failed.
I'd prefer it to this rather arbitrary logic.

alepping (approved):
one minor comment that you can address at will

PR #1744, opened 2026-07-09
fix(optimizer): support constant-only projections

adrianmichalke:
Until now queries consisting entirely of constants fail.

E.g.:

```
SELECT UINT8(1) AS x FROM input INTO out;
----
1
```

fails with bc of an empty schema:
`
Precondition violated: (!std::ranges::empty(schema)): We can not lower an empty schema!
`

This PR adds a mitigation and a regression systest.

### DIFF
--- a/nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
+++ b/nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
@@ -100,7 +100,7 @@ LogicalOperator pushBeyondSource(TypedLogicalOperator<SourceDescriptorLogicalOpe
     /// Recursion stop. Add projection if required is a strict subset of output schema.
 
     const auto outputSchema = op.getOutputSchema();
-    if (required.empty() && outputSchema.begin() != outputSchema.end())
+    if (required.empty() && not(outputSchema.empty()))
     {
         required.insert(*outputSchema.begin());
     }
--- a/nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
+++ b/nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
@@ -99,8 +99,14 @@ LogicalOperator pushBeyondSource(TypedLogicalOperator<SourceDescriptorLogicalOpe
 {
     /// Recursion stop. Add projection if required is a strict subset of output schema.
 
+    const auto outputSchema = op.getOutputSchema();
+    if (required.empty() && outputSchema.begin() != outputSchema.end())
+    {
+        required.insert(*outputSchema.begin());
+    }
+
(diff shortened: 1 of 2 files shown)


nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -87,8 +87,14 @@ LogicalOperator pushBeyondSource(TypedLogicalOperator<SourceDescriptorLogicalOpe
 {
     /// Recursion stop. Add projection if required is a strict subset of output schema.
 
+    const auto outputSchema = op.getOutputSchema();
+    if (required.empty() && outputSchema.begin() != outputSchema.end())
```
alepping:
```suggestion
    if (required.empty() && not(outputSchema.empty()))
```

alepping: assuming outputSchema is a well behaved container

alepping (approved):
one minor comment

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -99,6 +99,12 @@ LogicalOperator pushBeyondSource(TypedLogicalOperator<SourceDescriptorLogicalOpe
 {
     /// Recursion stop. Add projection if required is a strict subset of output schema.
 
+    /// Preserve the input schema for constant-only projections.
```
greenBene, line 102:
```suggestion
    /// If no field is required, the source schema is preserved, while it is assumed that the 
    /// parent operator ignores the fields. We choose the full schema instead of an arbitrary subset
    /// to avoid confusion. Ideally, we would send tuples without fields, but our system does not support this.
```

I'd suggest a more elaborate comment, similar to the suggested one. Let's double-check with @ls-1801 and @keyseven123 if the assumption is true that we do not support tuples without fields.

greenBene (changes requested):

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
adrianmichalke: I would prefer to keep this comment shorter to avoid documenting details that may go stale. The important point here is that this source-side pushdown must not create an empty projected schema when no fields are required. At least in this lowering path, field-less tuples are not supported: `LowerSchemaProvider::lowerSchema` has a non-empty-schema precondition.

greenBene: Fair point :)

greenBene (approved):
Thank you for fixing this bug! 💪

PR #1747, opened 2026-07-10
chore(logger): drop periodic flusher thread

ls-1801:
## Summary

Removes the spdlog `periodic_worker` flusher owned by `Logger`. Motivation is reducing thread count to improve the debugging and performance-analysis experience; the existing shutdown path already flushes the sink.

## Test plan
- [x] `format` target clean
- [ ] CI build + tidy


ls-1801:
Closing: dropping the periodic flusher causes async log tail loss; distributed-cli-test greps worker log for a Backpressure line that no longer reaches disk before the check.

PR #1749, opened 2026-07-10
chore(systest): run EmbeddedWorker on its own thread for logging

ls-1801:
## Summary

Runs the `SingleNodeWorker` owned by `EmbeddedWorkerQuerySubmissionBackend` on a dedicated `jthread` initialized with the worker's host as `WorkerNodeId`. Communication with the backend goes through request/reply queues (`folly::UnboundedQueue`) and a request variant so the backend API stays synchronous.

Motivation: query submission calls (`start`, `stop`, `status`, `workerStatus`) previously ran on the systest driver thread, so the `%G` host tag in log lines resolved to the driver's `WorkerNodeId` rather than the target worker's — making multi-worker systest runs indistinguishable in the log.

Captured from a live `YearOfExtraction.test` run:

Before:
```
[I] [systest] [main] [queryId:...] [SingleNodeWorker.cpp:116] [registerQuery] Registering query ...
[I] [systest] [main] [queryId:...] [SingleNodeWorker.cpp:116] [registerQuery] Registering query ...
```

After:
```
[I] [source-node:8080] [main] [queryId:...] [SingleNodeWorker.cpp:116] [registerQuery] Registering query ...
[I] [sink-node:8080]   [main] [queryId:...] [SingleNodeWorker.cpp:116] [registerQuery] Registering query ...
```

## Test plan
- [x] Debug build passes locally
- [x] `format` target clean
- [ ] CI build + tidy

### DIFF
--- a/nes-frontend/src/QueryManager/EmbeddedWorkerQuerySubmissionBackend.cpp
+++ b/nes-frontend/src/QueryManager/EmbeddedWorkerQuerySubmissionBackend.cpp
@@ -16,60 +16,208 @@
 
 #include <chrono>
 #include <memory>
+#include <mutex>
+#include <stop_token>
 #include <utility>
-#include <Identifiers/Identifiers.hpp>
+#include <variant>
 #include <Listeners/QueryLog.hpp>
 #include <Plans/LogicalPlan.hpp>
 #include <QueryManager/QueryManager.hpp>
-
-#include <Util/Logger/impl/NesLogger.hpp>
+#include <Util/Overloaded.hpp>
+#include <folly/concurrency/UnboundedQueue.h>
 #include <ErrorHandling.hpp>
 #include <QueryId.hpp>
 #include <QueryStatus.hpp>
+#include <SingleNodeWorker.hpp>
 #include <SingleNodeWorkerConfiguration.hpp>
+#include <Thread.hpp>
+#include <WorkerConfig.hpp>
 #include <WorkerStatus.hpp>
+
+namespace NES::detail
+{
+namespace
+{
+struct Stop
+{
+};
+
+struct StartQuery
+{
+    LogicalPlan plan;
+};
+
+struct StopQuery
+{
+    QueryId id;
+};
+
+struct QueryStatusQuery
+{
+    QueryId id;
+};
+
+struct WorkerStatusQuery
+{
+    std::chrono::system_clock::time_point after;
+};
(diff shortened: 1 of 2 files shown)


nes-frontend/src/QueryManager/EmbeddedWorkerQuerySubmissionBackend.cpp
```diff
@@ -16,60 +16,203 @@
...
+#include <WorkerConfig.hpp>
 #include <WorkerStatus.hpp>
 
+namespace NES::detail
+{
+namespace
+{
+struct Stop
+{
+};
```
greenBene, line 43: Why do we need the `Stop` struct if it only leads to an `std::unreachable`?

nes-frontend/src/QueryManager/EmbeddedWorkerQuerySubmissionBackend.cpp
```diff
@@ -16,60 +16,203 @@
...
+public:
+    Channel(WorkerConfig cfg, const SingleNodeWorkerConfiguration& workerConfiguration)
+        : config(std::move(cfg))
+        , thread(
+              "main",
+              this->config.host,
+              [&requests = this->requests, &replies = this->replies, config = this->config, workerConfiguration](
+                  const std::stop_token& stopToken)
+              {
+                  /// Start with the per-worker topology config, then overlay only
```
greenBene: I find this nesting a bit difficult to read. Could we move the function definition to a static/free function?

greenBene (changes requested):
I like this separation of concerns :) Just minor feedback

nes-frontend/src/QueryManager/EmbeddedWorkerQuerySubmissionBackend.cpp
ls-1801: Stop is handled earlier and would break the loop, this is why it's unreachable here

nes-frontend/src/QueryManager/EmbeddedWorkerQuerySubmissionBackend.cpp
ls-1801: i moved the code into a dedicated function

greenBene (approved):

keyseven123 (approved):

PR #1757, opened 2026-07-14
Enable raw string & boolean literals

adrianmichalke:
### Varsize & boolean literals

With this change we simplify our SQL syntax to not require type literals for varsize and booleans.

Before type constructors where always required:

SELECT **VARSIUZED('text')**, **BOOLEAN(true)** FROM s INTO sink;

Now also allowed:

SELECT **'text'**, **true** FROM s INTO sink;

### Identifiers syntax

Aligned with standard SQL we reserve double quotes for identiers, i.e.:

SELECT 'text', true FROM **"source name with whitespace"** INTO sink;

This PR add documetation for this differenciation.

### DIFF
--- a/docs/guide/query_api.md
+++ b/docs/guide/query_api.md
@@ -370,6 +370,21 @@ Use an explicit type constructor when a query depends on an exact literal type:
 SELECT UINT64(1) AS id, FLOAT32(3.6) AS scale FROM s INTO sink
 ```
 
+### String and boolean literals
+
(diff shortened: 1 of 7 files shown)


ls-1801 (approved):

keyseven123 (approved):
I think that @Artraxon should also take a look, as I am not an expert/codeowner in this part of the code. But up to you @adrianmichalke

PR #1761, opened 2026-07-15
feat(Clang-Tidy): Add origin/main tidy-diff targets

keyseven123:

`tidy-diff` only covered uncommitted changes, so there was no one-click way to run clang-tidy over a whole branch — that meant hand-writing a `git diff origin/main | clang-tidy-diff-19.py ...` invocation. This pull request adds branch-wide tidy-diff targets. The change list is as follows:

- Added `tidy-diff-to-main` and `tidy-diff-to-main-fix` to `cmake/ClangTidyDiff.cmake`, pinning the diff base to `origin/main` via `cmake -E env` so the base is set in the child process even when the Docker toolchain does not forward `NES_TIDY_DIFF_BASE`. They show up in the CLion target dropdown with no environment setup.
- Changed the documented default base of `tidy-diff` / `tidy-diff-fix` from `git diff --cached` to `git diff HEAD`, so they cover all uncommitted changes rather than only staged ones.
- Documented all four targets in `fix_clang_tidy_warnings.md` and `development.md`, and demoted the hand-written `clang-tidy-diff.py` command to a fallback for custom bases and non-CMake setups.

## Verifying this change

Build tooling only — no library, engine, or test code is touched, so no automated tests were added. Verified manually in the `nebulastream/nes-development:local` container:

- All four targets (`tidy-diff`, `tidy-diff-fix`, `tidy-diff-to-main`, `tidy-diff-to-main-fix`) register in the generated build.
- `cmake --build <build> --target tidy-diff-to-main` runs end-to-end against a branch with local commits: it resolves `origin/main` as the base, restricts findings to the branch's own lines, and writes `clang-tidy-diff-report.txt`.
- The pinned base takes effect inside the container without any `-e NES_TIDY_DIFF_BASE=...` plumbing, which is the point of the `cmake -E env` wrapper.

### DIFF
--- a/cmake/ClangTidyDiff.cmake
+++ b/cmake/ClangTidyDiff.cmake
@@ -63,7 +63,7 @@ function(project_enable_tidy_diff)
     message(STATUS "Enabling tidy-diff targets using ${CLANG_TIDY_DIFF_EXECUTABLE} (binary: ${CLANG_TIDY_EXECUTABLE})")
 
     # tidy-diff: check mode. The driver discovers the diff base from the
-    # NES_TIDY_DIFF_BASE env var (default 'git diff --cached'), prints a summary,
+    # NES_TIDY_DIFF_BASE env var (default 'git diff HEAD'), prints a summary,
     # then runs clang-tidy-diff, teeing colored output to stdout and a de-colored
     # copy to the report file. USES_TERMINAL preserves the colored stdout.
     add_custom_target(tidy-diff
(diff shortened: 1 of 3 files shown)


greenBene (approved):
One clarifying question: In the manual command, I use " -j 24" to ensure parallelism. Can the introduced targets also use all my cores?

keyseven123:
> One clarifying question: In the manual command, I use " -j 24" to ensure parallelism. Can the introduced targets also use all my cores?

It does per default, but you can overwrite it.

zeuchste (approved):

PR #1764, opened 2026-07-15
Interval join

keyseven123:
This PR adds an [interval join](https://nightlies.apache.org/flink/flink-docs-stable/docs/dev/datastream/operators/joining/#interval-join) to our code base. Similar to Flink, we do not support outer joins for interval joins.

## Verifying this change
This change is tested by newly added tests.

- ExecutionEngine
- Parser


nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -1032,28 +1058,62 @@ void AntlrSQLQueryPlanCreator::exitJoinRelation(AntlrSQLParser::JoinRelationCont
...
-    }
 
-    const auto currentWindowTimestampOpt = helpers.top().windowTimestamp;
-    if (!currentWindowTimestampOpt.has_value()
-        || !std::holds_alternative<std::array<Windowing::UnboundTimeCharacteristic, 2>>(currentWindowTimestampOpt.value()))
+    const LogicalPlan queryPlan = [&]
     {
-        throw InvalidQuerySyntax("join requires two timestamps, but got {}", currentWindowTimestampOpt.has_value() ? "only one" : "none");
-    }
+        if (helpers.top().intervalLowerBound.has_value() && helpers.top().intervalUpperBound.has_value())
```
ls-1801, line 1162: The interval bounds remain set after the first join is lowered, so every following join in the same relation enters this interval-join branch—even when its SQL uses `WINDOW`. For example, this ordinary left tumbling join following an interval join is rejected as an interval outer join:

```sql
SELECT id1 FROM (SELECT * FROM leftInput)
INNER JOIN (SELECT * FROM intervalRight) ON (id1 = id2)
INTERVAL (timestampLeft, timestampRight, lower 0 ms, upper 3 ms)
LEFT JOIN (SELECT * FROM windowRight) ON (id1 = id3)
WINDOW TUMBLING (timestampLeft, timestampWindow, size 1 sec)
INTO discard;
```

The reproducer fails with `Interval joins only support INNER JOIN`; the same left join passes when the first join is a normal tumbling join. Clear the interval bounds after constructing the interval join (or keep the bounds scoped to the individual `JoinRelation`).

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinOperatorHandler.cpp
```diff
@@ -0,0 +1,343 @@
...
+        }
+        anchor->combinePagedVectors();
+
+        EmittedIntervalJoinWindowTrigger trigger{};
+        trigger.anchorSliceEnd = anchor->getSliceEnd();
+        trigger.windowInfo = WindowInfo{anchor->getSliceStart().getRawValue(), anchor->getSliceEnd().getRawValue()};
+        trigger.partnerCount = 0;
+        auto minPartnerStart = anchor->getSliceStart();
+
+        const auto anchorStartRaw = static_cast<std::int64_t>(anchor->getSliceStart().getRawValue());
```
ls-1801: `Timestamp`/slice boundaries are unsigned, but this narrows the anchor timestamp to `int64_t`. For timestamps at or above `INT64_MAX + 1`, it becomes negative; all computed partner ends are then non-positive and skipped, so valid matches disappear.

Reproducer (the normal `1000` pair is emitted, while the high-timestamp pair is missing):

```sql
CREATE LOGICAL SOURCE highLeft(id1 UINT64 NOT NULL, timestampLeft UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR highLeft TYPE File;
ATTACH INLINE
1,1000
1,9223372036854775808

CREATE LOGICAL SOURCE highRight(id2 UINT64 NOT NULL, timestampRight UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR highRight TYPE File;
ATTACH INLINE
1,1000
1,9223372036854775808

SELECT * FROM (SELECT * FROM highLeft) INNER JOIN (SELECT * FROM highRight) ON (id1 = id2)
INTERVAL (timestampLeft, timestampRight, lower 0 ms, upper 3 ms)
INTO highSink;
```

Please keep `anchorStartRaw` and the bucket index unsigned where possible. The interval offset loop must support negative bounds, so apply negative offsets with an underflow check rather than converting the timestamp to a signed value.

ls-1801 (changes requested):
Wrong button, see review above

nes-logical-operators/include/Operators/Windows/IntervalJoinLogicalOperator.hpp
```diff
@@ -0,0 +1,152 @@
...
+    [[nodiscard]] const UnqualifiedUnboundField& getStartField() const;
+    [[nodiscard]] const UnqualifiedUnboundField& getEndField() const;
+
+    [[nodiscard]] bool operator==(const IntervalJoinLogicalOperator& rhs) const;
+
+    [[nodiscard]] IntervalJoinLogicalOperator withTraitSet(TraitSet traitSet) const;
+    [[nodiscard]] TraitSet getTraitSet() const;
+
+    [[nodiscard]] Schema<Field, Unordered> getOutputSchema() const;
+    [[nodiscard]] IntervalJoinLogicalOperator withChildrenUnsafe(std::vector<LogicalOperator> children) const;
```
arianeziehn, line 100: what is unsafe?

nes-systests/operator/join/IntervalJoinBounds.test
```diff
@@ -0,0 +1,216 @@
...
+1004,10,4
+1005,10,5
+1006,14,6
+1007,2,7
+1008,12,8
+1009,12,9
+
+CREATE SINK p2Sink(start UINT64 NOT NULL, end UINT64 NOT NULL, f1_left UINT64 NOT NULL, f2_left UINT64 NOT NULL, timestampLeft UINT64 NOT NULL, f1_right UINT64 NOT NULL, f2_right UINT64 NOT NULL, timestampRight UINT64 NOT NULL) TYPE File;
+
+# PrivateTest2: future bounds [0 ms, 3 ms] with right tuples interleaved before/after left; join key f2.
```
arianeziehn, line 142: how is it possible with this bound? right tuples can only occur after their left, would itn't be interesting with a bound in both directions?

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinOperatorHandler.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+namespace NES
+{
+
+namespace
+{
+
+/// floor(x / divisor) rounding toward -inf (C's `/` rounds toward zero -> wrong slice index for negative bounds).
+/// Replace with std::div_floor once the toolchain ships C++26 <intdiv> (P3724).
+constexpr std::int64_t floorDiv(std::int64_t x, std::int64_t divisor)
```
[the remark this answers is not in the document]
arianeziehn: I mean here that comment might be irrelevant; however, if you use such a common function, can we not have a library of those we can use, and then x and div make sense if you use them in a specific class? I vote for terms common to this class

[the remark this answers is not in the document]
arianeziehn: I also saw h for handler where I would also suggest to call it ijhandler just for readability

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinProbePhysicalOperator.cpp
```diff
@@ -0,0 +1,332 @@
...
+    nautilus::val<bool> upperOk{false};
+    if (upperBoundMs >= 0)
+    {
+        const nautilus::val<std::uint64_t> offset{static_cast<std::uint64_t>(upperBoundMs)};
+        upperOk = (partnerTs <= (anchorTs + offset));
+    }
+    else
+    {
+        const nautilus::val<std::uint64_t> offset{static_cast<std::uint64_t>(-upperBoundMs)};
+        upperOk = ((partnerTs + offset) <= anchorTs);
```
arianeziehn: can that not be compressed ? (dup code)

nes-sql-parser/AntlrSQL.g4
```diff
@@ -291,6 +291,14 @@ windowClause
     : WINDOW windowSpec
     ;
 
+intervalClause
+    : INTERVAL '(' timestampParameter ',' LOWER lowerBound=signedTimeValue ',' UPPER upperBound=signedTimeValue ')'
+    ;
```
Artraxon, line 312: I really don't like this syntax, its something completely new that has no connection to SQL or our streaming dialect. In my opinion there are two options: Either we treat the window clause as "arguments" to the window, in which case we either pass positional or named arguments, or we infer the bounds from a predicate one-sql-to-rule-them-all style.

So for the first option we can  either just leave out lower and upper, and just infer from the position what they are.  Or, use our syntax for passing named parameters `0 ms AS upper`. I think the first option is much easier for now, the second one we will get for free over time anyway and currently is difficult to do ergonomically.


The alternative would be just writing out the join condition with a predicate `ts <= upper AND ts >= lower` (or using `BETWEEN`) and then extracting that. 
This is probably the cleanest interface, but its much harder to implement.

I would favor passing them as positional arguments without the upper and lower keywords.


Long-term, with more people adding more exotic, non-standard joins we will have to come to a common syntax, but I don't think you need to worry about it in this PR.

nes-logical-operators/include/Operators/Windows/IntervalJoinLogicalOperator.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+    explicit IntervalJoinLogicalOperator(
+        WeakLogicalOperator self,
+        std::array<LogicalOperator, 2> children,
+        LogicalFunction joinFunction,
+        JoinTimeCharacteristic timeCharacteristics,
+        IntervalBound lowerBound,
+        IntervalBound upperBound,
+        JoinLogicalOperator::JoinType joinType);
+
```
Artraxon, line 70: also add the create factory methods

Artraxon (changes requested):
Great that you are adding this, this should increase our expressivity greatly.

I am wondering, wouldn't we be able to in theory just remove the other windowed joins, imo this is a strict superset where you'd just need to filter out tuple afterwards.

Also, it would be nice if we could talk about this a bit more in person since it is already very close to my correlate operator.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
arianeziehn: Good point to add mixed window type tests if not yet existing

arianeziehn (commented):
LGTM from my side, for the pure IVJ tests I think you covered most, I just have 1 understanding question, given Lukas observation, it probably makes sense to add mixed window type tests

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -781,6 +784,29 @@ void AntlrSQLQueryPlanCreator::exitWindowClause(AntlrSQLParser::WindowClauseCont
...
+    /// windowTimestamp has already been populated while handling the timestamp parameter inside the interval clause.
+    if (!helpers.top().windowTimestamp.has_value())
+    {
+        throw InvalidQuerySyntax(
+            "Interval-join requires a timestamp parameter (e.g. INTERVAL (L.ts, lower 0 ms, upper 3 ms)) at {}", context->getText());
+    }
+
+    const auto parseSigned = [](AntlrSQLParser::SignedTimeValueContext* signedCtx) -> int64_t
+    {
+        const int magnitude = std::stoi(signedCtx->INTEGER_VALUE()->getText());
```
adrianmichalke: What about unsigned integers?

nes-logical-operators/src/Operators/Windows/IntervalJoinLogicalOperator.cpp
```diff
@@ -0,0 +1,321 @@
...
+    for (const auto& field : (*children)[0].getOutputSchema())
+    {
+        outputFields.emplace_back(field.unbound());
+    }
+    for (const auto& field : (*children)[1].getOutputSchema())
+    {
+        outputFields.emplace_back(field.unbound());
+    }
+    outputFields.emplace_back(startEndFields[0].getFullyQualifiedName(), startEndFields[0].getDataType());
+    outputFields.emplace_back(startEndFields[1].getFullyQualifiedName(), startEndFields[1].getDataType());
```
adrianmichalke, line 195: Currently NULL timestamps become zero. is this expected? What do we do in the other joins?

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinOperatorHandler.cpp
adrianmichalke, line 1: In the other joins we throw WindowingError 3014 for out-of-order tuples. Please apply it here as well.

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinProbePhysicalOperator.cpp
adrianmichalke: We should check here for overflows.

adrianmichalke (changes requested):
Some comments

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinProbePhysicalOperator.cpp
Artraxon: IMO there is no duplication here, all branches are different in a subtle way. I also like with the current state that it is very easy to read and reason about.

nes-logical-operators/include/Operators/Windows/IntervalJoinLogicalOperator.hpp
Artraxon: it's unsafe because it doesn't run local schema inference and so any LogicalFunction that references fields will point to the old, wrong children. 
Its only used by the query binder, and hopefully we can get rid of them soon.

nes-systests/operator/join/IntervalJoinBounds.test
keyseven123: MixedBounds (in line 54) already has a bound in both directions.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: Added mixed window type test

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinOperatorHandler.cpp
keyseven123: Done

nes-sql-parser/AntlrSQL.g4
keyseven123: Is it fine to create an issue for this then?

nes-logical-operators/include/Operators/Windows/IntervalJoinLogicalOperator.hpp
keyseven123: Done

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinProbePhysicalOperator.cpp
keyseven123: I will leave them as-is then

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: I have added an exception if the passed value does not fit into a signed i64

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinOperatorHandler.cpp
keyseven123: Done

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinProbePhysicalOperator.cpp
keyseven123: Done

nes-logical-operators/src/Operators/Windows/IntervalJoinLogicalOperator.cpp
keyseven123: We check if the timestamp is not nullable, as we expect timestamp fields to be not null. Added this check

nes-logical-operators/src/WindowTypes/Types/IntervalWindow.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include <WindowTypes/Measures/TimeMeasure.hpp>
+#include <fmt/format.h>
+#include <folly/hash/Hash.h>
+
+namespace NES::Windowing
+{
+
+/// No bound validation here. lowerBound < upperBound is enforced with a graceful error in
+/// IntervalJoinLogicalOperator.
+IntervalWindow::IntervalWindow(const IntervalBound lowerBound, const IntervalBound upperBound)
```
ls-1801, line 32: This constructor is the point where an `IntervalWindow` first exists, so it should enforce `lowerBound < upperBound` here. Deferring validation to `IntervalJoinLogicalOperator` allows invalid `IntervalWindow` instances through reflection or any other caller and makes `getSize()` depend on an invariant the class does not uphold.

nes-logical-operators/include/Operators/Windows/IntervalJoinLogicalOperator.hpp
```diff
@@ -0,0 +1,167 @@
...
+/// Bounds are signed ms offsets, so intervals may be past/future-anchored (e.g. `[-4, -1]`). Slice
+/// width `W = upperBound - lowerBound`. The constructor rejects `lowerBound >= upperBound` (W = 0 is degenerate).
+class IntervalJoinLogicalOperator final : public OriginIdAssigner, public ManagedByOperator
+{
+public:
+    explicit IntervalJoinLogicalOperator(
+        WeakLogicalOperator self,
+        LogicalFunction joinFunction,
+        JoinTimeCharacteristic timeCharacteristics,
+        IntervalBound lowerBound,
```
ls-1801, line 58: Could this operator store and accept an `IntervalWindow` instead of separate `lowerBound` and `upperBound` values? `JoinLogicalOperator` already keeps its `TimeBasedWindowType` separate from `JoinTimeCharacteristic`; doing the same here would preserve the bound-pair invariant established by `IntervalWindow` rather than repeatedly splitting and reassembling it.

nes-logical-operators/src/Plans/LogicalPlanBuilder.cpp
```diff
@@ -160,6 +165,73 @@ LogicalPlan LogicalPlanBuilder::addJoin(
...
+        if (itr.getChildren().size() == 2)
+        {
+            auto leftVisitingOp = itr.getChildren()[0];
+            if (leftVisitingOp.getChildren().size() == 1)
+            {
+                if (visitedFunctions.find(leftVisitingOp) == visitedFunctions.end())
+                {
+                    visitedFunctions.insert(leftVisitingOp);
+                    auto leftChild = leftVisitingOp.getChildren().at(0);
+                    auto rightChild = leftVisitingOp.getChildren().at(1);
```
ls-1801, line 206: The regular `addJoin` no longer performs this predicate traversal, and this copy is unsafe: after `leftVisitingOp.getChildren().size() == 1`, `getChildren().at(1)` necessarily throws before the intended `InvalidQuerySyntax`. Please remove this block. If interval joins require a distinct predicate restriction, enforce it during `IntervalJoinLogicalOperator::inferLocalSchema()` with a documented reason and a targeted error; otherwise constant-side predicates should not be rejected here.

nes-sql-parser/AntlrSQL.g4
```diff
@@ -307,6 +307,14 @@ windowClause
     : WINDOW windowSpec
     ;
 
+intervalClause
+    : INTERVAL '(' timestampParameter ',' lowerBound=signedTimeValue ',' upperBound=signedTimeValue ')'
```
ls-1801, line 311: Nit: I don't want to open a can of worms here, but this syntax differs from the existing window forms in two ways: it uses `INTERVAL (...)` instead of `WINDOW ...`, and its bounds are bare positional values while tumbling and sliding windows use named phrases such as `SIZE` and `ADVANCE BY`. Since this is represented as an `IntervalWindow`, could we align both the clause and its parameters with the existing natural-language window syntax?

Artraxon (approved):

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinProbePhysicalOperator.cpp
```diff
@@ -0,0 +1,321 @@
...
+            for (auto partnerIt = partnerPagedVector.begin(); partnerIt != partnerPagedVector.end(); ++partnerIt)
+            {
+                auto anchorRecord = *anchorIt;
+                auto partnerRecord = *partnerIt;
+                /// Check the join expression first (cheap key compare) and skip the interval check on failure.
+                const auto candidateRecord
+                    = createJoinedRecord(anchorRecord, partnerRecord, windowStart, windowEnd, anchorKeyFieldNames, partnerKeyFieldNames);
+                if (joinFunction.execute(candidateRecord, executionCtx.pipelineMemoryProvider.arena))
+                {
+                    if (intervalPredicateHolds(executionCtx, anchorRecord, partnerRecord))
```
ls-1801, line 309: Why does the probe apply the interval predicate to every same-key pair after partner slices have already been selected by interval bucket offsets? For bucket pairs that are fully inside the interval, this check appears redundant; could the trigger carry enough boundary information to skip it, leaving per-record timestamp filtering only for partially eligible bucket pairs?

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinOperatorHandler.cpp
```diff
@@ -0,0 +1,369 @@
...
+IntervalJoinOperatorHandler::IntervalJoinOperatorHandler(
+    std::vector<OriginId> anchorInputOriginsParam,
+    std::vector<OriginId> partnerInputOriginsParam,
+    const OriginId outputOriginIdParam,
+    const IntervalBound lowerBoundParam,
+    const IntervalBound upperBoundParam,
+    SliceCacheConfiguration sliceCacheConfigurationParam)
+    : lowerBound(lowerBoundParam)
+    , upperBound(upperBoundParam)
+    , sliceWidth(static_cast<std::uint64_t>(upperBoundParam.getRawValue() - lowerBoundParam.getRawValue()))
```
ls-1801, line 105: Consider an interval join with bounds `[-10 h, 0 s]`. This makes each anchor slice 10 hours wide, so an anchor tuple at `00:01` and an already-present matching partner tuple cannot produce output until the `10:00` slice boundary. A user of an interval join would not expect a matching result for this query to wait almost 10 hours. The anchor side needs eager emission rather than a slice-based trigger.

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinOperatorHandler.cpp
```diff
@@ -0,0 +1,369 @@
...
+    const std::scoped_lock guard{triggerMutex};
+    std::ignore = getBuildWatermark(side).updateWatermark(bufferMetaData.watermarkTs, bufferMetaData.seqNumber, bufferMetaData.originId);
+    tryTriggerAnchors(pipelineCtx);
+}
+
+void IntervalJoinOperatorHandler::notifyBufferDoneProbe(const BufferMetaData& bufferMetaData)
+{
+    std::ignore = watermarkProbe->updateWatermark(bufferMetaData.watermarkTs, bufferMetaData.seqNumber, bufferMetaData.originId);
+
+    const auto probeWatermark = watermarkProbe->getCurrentWatermark();
```
ls-1801, line 155: `MultiOriginWatermarkProcessor::updateWatermark()` already returns the current watermark. Could we assign that result directly to `probeWatermark` instead of discarding it and immediately calling `getCurrentWatermark()` again?

nes-logical-operators/src/WindowTypes/Measures/TimeCharacteristic.cpp
```diff
@@ -120,6 +122,25 @@ TimeCharacteristicWrapper::Type TimeCharacteristicWrapper::getType() const
...
+{
+/// A NULL timestamp would silently read as 0 and place the tuple at the start of time, so reject it up front.
+BoundEventTimeCharacteristic bindEventTime(const LogicalFunction& erasedFieldAccess, const TimeUnit& unit)
+{
+    const auto fieldAccess = erasedFieldAccess.tryGetAs<FieldAccessLogicalFunction>();
+    if (not fieldAccess.has_value())
+    {
+        throw CannotInferSchema("Event-time characteristic must resolve to a field access");
+    }
+    const auto field = (*fieldAccess)->getField();
```
ls-1801, line 135: This validation still accepts a `VARSIZED` event-time field, but `EventTimeFunction::getTs` later calls `getRawValueAs<val<uint64_t>>()`, which throws `UnknownOperation` for `VariableSizedData`; `BOOLEAN` is also silently treated as 0/1. Please reject unsupported timestamp data types here during schema inference and include the field name, actual type, and accepted types in the error.

nes-logical-operators/src/Operators/Windows/IntervalJoinLogicalOperator.cpp
```diff
@@ -0,0 +1,344 @@
...
+    const auto& inputSchema = inputSchemaOrCollisions.value();
+
+    this->joinFunction = this->joinFunction.withInferredDataType(inputSchema);
+
+    /// Bind left/right timestamp characteristics against the merged input schema by their field names.
+    this->timestampFields = std::visit(
+        [&](const auto& tsFields)
+        {
+            return std::array{
+                Windowing::TimeCharacteristicWrapper{tsFields[0]}.withInferredSchema(inputSchema),
```
ls-1801, line 179: Both timestamp characteristics are bound against the merged schema, so the first characteristic can resolve to a field produced only by the right child (and vice versa). Lowering treats array element 0 as the anchor timestamp and element 1 as the partner timestamp; a swapped characteristic therefore survives inference and later tries to read the other side’s field from each record. Please bind `tsFields[0]` against `children[0].getOutputSchema()` and `tsFields[1]` against `children[1].getOutputSchema()`, while retaining the merged schema for the join predicate.

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalIntervalJoin.cpp
```diff
@@ -0,0 +1,233 @@
...
+{
+
+/// Join-key field names that resolve against the given (physical) input schema.
+std::vector<Record::RecordFieldIdentifier>
+getJoinFieldNames(const Schema<QualifiedUnboundField, Ordered>& inputSchema, const LogicalFunction& joinFunction)
+{
+    return BFSRange(joinFunction)
+        | std::views::filter([](const auto& child) { return child.template tryGetAs<FieldAccessLogicalFunction>().has_value(); })
+        | std::views::transform([](const auto& child) { return child.template tryGetAs<FieldAccessLogicalFunction>().value()->getField(); })
+        | std::views::filter([&](const auto& field) { return inputSchema.contains(field.getLastName()); })
```
ls-1801, line 75: The predicate is lowered through `combinedFieldMapping`, but this helper filters and returns the logical last names. Here `id + 1 AS id` makes the left join key `7`, which should match the right key `7`, but the probe fails with `Field ID.NEW not found in record R_KEY, END, START`:

```sql
CREATE LOGICAL SOURCE mappedLeft(id UINT64 NOT NULL, l_ts UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR mappedLeft TYPE File;
ATTACH INLINE
6,100

CREATE LOGICAL SOURCE mappedRight(r_key UINT64 NOT NULL, r_ts UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR mappedRight TYPE File;
ATTACH INLINE
7,100

CREATE SINK mappedSink(start UINT64 NOT NULL, end UINT64 NOT NULL, id UINT64 NOT NULL, original_id UINT64 NOT NULL, l_ts UINT64 NOT NULL, r_key UINT64 NOT NULL, r_ts UINT64 NOT NULL) TYPE File;

SELECT * FROM (SELECT id + 1 AS id, id AS original_id, l_ts FROM mappedLeft) INNER JOIN (SELECT * FROM mappedRight) ON id = r_key INTERVAL (l_ts, r_ts, 0 ms, 3 ms) INTO mappedSink;
----
99 102 7 6 100 7 100
```

Please derive the key projections from each child's `FieldMappingTrait` and use `Field::getProducedBy()` to assign them to the correct side.

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalIntervalJoin.cpp
```diff
@@ -0,0 +1,233 @@
...
+        [tupleSizePartner](IntervalJoinOperatorHandler& operatorHandler, AbstractBufferProvider& bufferProvider)
+        {
+            const CreateNewIntervalJoinSliceArgs intervalSliceArgs{bufferProvider, tupleSizePartner};
+            return operatorHandler.getCreateNewSlicesFunction(intervalSliceArgs);
+        });
+
+    const IntervalJoinBuildPhysicalOperator anchorBuildOperator{
+        handlerId,
+        IntervalJoinBuildSide::Anchor,
+        TimeFunction::create(anchorTimeCharacteristic),
```
ls-1801, line 173: These time functions are created from logical bound characteristics without applying each child's physical field mapping. Here `ts + 1 AS ts` makes the event timestamp `100`, but the build fails with `Field TS not found in record T_KEY, TS.NEW, ORIGINAL_TS`:

```sql
CREATE LOGICAL SOURCE mappedTimeLeft(t_key UINT64 NOT NULL, ts UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR mappedTimeLeft TYPE File;
ATTACH INLINE
7,99

CREATE LOGICAL SOURCE mappedTimeRight(t_key_right UINT64 NOT NULL, t_right UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR mappedTimeRight TYPE File;
ATTACH INLINE
7,100

CREATE SINK mappedTimeSink(start UINT64 NOT NULL, end UINT64 NOT NULL, t_key UINT64 NOT NULL, ts UINT64 NOT NULL, original_ts UINT64 NOT NULL, t_key_right UINT64 NOT NULL, t_right UINT64 NOT NULL) TYPE File;

SELECT * FROM (SELECT t_key, ts + 1 AS ts, ts AS original_ts FROM mappedTimeLeft) INNER JOIN (SELECT * FROM mappedTimeRight) ON t_key = t_key_right INTERVAL (ts, t_right, 0 ms, 3 ms) INTO mappedTimeSink;
----
99 102 7 100 99 7 100
```

Please construct each side's build time function using its mapped physical timestamp identifier; the probe time functions need the same treatment.

nes-physical-operators/include/SliceStore/CachingSliceStoreRef.hpp
```diff
@@ -0,0 +1,163 @@
...
+    ~CachingSliceStoreRef() override = default;
+
+    NautilusBuffer getDataStructureRef(
+        const nautilus::val<Timestamp>& timestamp,
+        const nautilus::val<WorkerThreadId>& workerThreadId,
+        const nautilus::val<OperatorHandler*>& operatorHandler,
+        nautilus::val<AbstractBufferProvider*> bufferProvider) override
+    {
+        nautilus::val<Store*> sliceStoreVal{sliceStore};
+        return sliceCache->getDataStructureRef(
```
ls-1801, line 89: Are cached slices invalidated when they are triggered or garbage-collected? `getSlicesOrCreate()` checks `maxRetiredSliceEnd`, but it is only called on a cache miss; a cache hit returns the stored `TupleBuffer*` directly. Where is the equivalent retirement check enforced on the hit path, and what prevents a late tuple from writing through a cached entry after the slice’s join result was emitted?

nes-physical-operators/src/Join/IntervalJoin/IntervalJoinOperatorHandler.cpp
```diff
@@ -0,0 +1,369 @@
...
+{
+    /// Highest bucket whose end still fits into the unsigned timestamp range.
+    const std::uint64_t maxBucketIndex = (std::numeric_limits<std::uint64_t>::max() / sliceWidth) - 1;
+    for (auto& anchor : drivingSlices)
+    {
+        if (anchor->getNumberOfTuples() == 0)
+        {
+            continue;
+        }
+        anchor->combinePagedVectors();
```
ls-1801, line 292: With `[+10 h, +10.1 h]`, a partner-watermark advance can release a large prefix of 6-minute anchor slices at once. `assembleAndEmitTriggers()` merges every populated anchor before checking whether either candidate partner bucket exists, so an empty or sparse partner store turns that whole prefix into merging and lookup work while producing little or no output. Could partner existence be established before merging the anchor, or could trigger selection intersect the anchor and partner bucket ranges first?

ls-1801 (changes requested):
I think my biggest issue is waiting for window boundaries, which would limit real-world usage of the interval join if we always have to wait for the anchor window to complete. For very large intervals, this could mean days.

adrianmichalke (approved):

PR #1765, opened 2026-07-15
feat(optimizer) Extended and streamlined rule concept

### LINKED ISSUE #1740: [TASK] feat(Optimizer) Allow weak dependencies between rules
### Description

The current optimizer rules only allow strong dependencies. This means if A depends on B, A cannot run if B does not run. To enable the manual disabling of individual rules by the user, we need in addition, weak dependencies: If A weakly depends on B, A must run after B if B is enabled, but if B is disabled, A can run anyway. 

### Acceptance Criteria

* Update Rule concept to allow weak dependencies
* Change current dependencies between rules to weak if applicable
* Update RuleManager to support weak dependencies
* Update RuleManagerTests to verify behaviour

greenBene:

Streamlines the optimizer Rule/RuleManager concept and adds soft dependencies.
- Renamed dependsOn()/requiredBy() to needs()/neededBy().
- Added soft dependencies wants()/wantedBy(): ordering constraints that only apply if the referenced rule is registered.
- Removed getType(), getName(), operator==() as required boilerplate; now derived automatically or defaulted.
- needs()/neededBy()/wants()/wantedBy() are now optional, defaulting to empty sets.
- RuleManager builds its DAG from both hard and soft edges; explain() now also prints WANTS/WANTED_BY.
- Migrated all existing rules to the new interface

## Verifying this change

- Extended RuleManagerTest with SoftDependencies and HardAndSoftDependenciesMixed cases.
- Existing sequencing/cycle/missing-rule tests updated and passing.

- QueryOptimizer (nes-query-optimizer)

## Documentation

This PR closes #1740


Artraxon (approved):
We already talked about it, so it looks good to me

nes-query-optimizer/include/Rules/RuleManager.hpp
```diff
@@ -146,22 +119,97 @@ class RuleManager
...
 private:
-    /// Directed edges between rules. key points to rules that depend on it.
-    std::unordered_map<std::type_index, std::set<std::type_index>> directedEdges;
     std::unordered_map<std::type_index, Rule<U>> rules;
+
+    [[nodiscard]] std::unordered_map<std::type_index, std::set<std::type_index>> generateDownstreamEdges() const
+    {
+        std::unordered_map<std::type_index, std::set<std::type_index>> downstreamEdges;
+
+        for (const auto& [_, rule] : rules)
```
alepping, line 140: You could use modern c++ views/ranges to get rid of the loops and implement the code more modern-idiomat. Its generally r commended for safety. If you prefer loops, then go with it though.

nes-query-optimizer/include/Rules/RuleManager.hpp
```diff
@@ -146,22 +119,97 @@ class RuleManager
...
+        for (const auto& [type, _] : rules)
+        {
+            indegree[type] = 0;
+        }
+
+        for (const auto& [dependency, neededBy] : downstreamEdges)
+        {
+            if (!rules.contains(dependency))
+            {
+                throw InvalidOptimizerRuleset("unregistered rule in dependency graph: {}", dependency.name());
```
alepping: You might also want to log which rule had the dependency. Generally, I like if error msgs give as much potentially relevant info as possible

alepping (commented):
Really nice to see the clean up.

Two minor comments. Feel free to disregard them.

nes-query-optimizer/include/Rules/RuleManager.hpp
greenBene: Good point! I added the available context.

nes-query-optimizer/include/Rules/RuleManager.hpp
greenBene: Thanks for the suggestion! I tried it but find it personally way harder to understand what is going on (see below). Thus, I'd stick to the for loop for now. :)  

```cpp
            rule.needs()
                | std::ranges::for_each([&downstreamEdges, &rule](const auto& needed) { downstreamEdges[needed].insert(rule.getType()); });

            rule.neededBy()
                | std::ranges::for_each([&downstreamEdges, &rule](const auto& neededBy)
                                        { downstreamEdges[rule.getType()].insert(neededBy); });

            rule.wants() | std::views::filter([this](auto wanted) { return rules.contains(wanted); })
                | std::ranges::for_each([&downstreamEdges, &rule](const auto& wanted) { downstreamEdges[wanted].insert(rule.getType()); });

            rule.wantedBy() | std::views::filter([this](auto wantedBy) { return rules.contains(wantedBy); })
                | std::ranges::for_each([&downstreamEdges, &rule](const auto& wantedBy)
                                        { downstreamEdges[rule.getType()].insert(wantedBy); });
```

alepping (approved):

PR #1766, opened 2026-07-15
Systest .test file names do not need to be unique

adrianmichalke:
Fixes a limitation that discovered systest `.test` file names have to be unique.

I.e. the following crashes:
```
parser/regressionTests.test
operator/regressionTests.test
```

Stack PR to enable https://github.com/nebulastream/nebulastream/pull/1652.

### DIFF
--- a/nes-systests/systest/src/SystestState.cpp
+++ b/nes-systests/systest/src/SystestState.cpp
@@ -173,8 +173,7 @@ namespace NES::Systest
 std::filesystem::path
 SystestQuery::resultFile(const std::filesystem::path& workingDir, std::string_view testName, const SystestQueryId queryIdInTestFile)
 {
-    auto resultPath = workingDir / "results"
-        / std::filesystem::path(fmt::format("{}_{}.csv", testName, queryIdInTestFile));
+    auto resultPath = workingDir / "results" / std::filesystem::path(fmt::format("{}_{}.csv", testName, queryIdInTestFile));
     const auto resultDir = resultPath.parent_path();
     if (not is_directory(resultDir))
     {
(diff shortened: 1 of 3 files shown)


greenBene (approved):
The fix makes sense to me :)

I am slightly concerned that the nes-systests is becoming a bit unwieldy in its logic. Thus, I would suggest a spring-cleaning PR for the component in the near future. But this is out of scope for this PR.

adrianmichalke:
@greenBene agree, its already on the agenda :)

nes-systests/systest/src/SystestState.cpp
```diff
@@ -159,6 +159,35 @@ std::optional<std::string> getSkipReason(const NES::Systest::TestFile& testFile,
...
+    relativePath.replace_extension();
+    return relativePath.generic_string();
+}
+
+NES::Systest::TestName testNameStem(const NES::Systest::TestFile& testFile)
+{
+    return std::filesystem::path(testFile.name()).filename().string();
+}
+
+void shortenUniqueTestNames(NES::Systest::TestFileMap& testFiles)
```
keyseven123, line 175: I am missing here at least a minor comment what the intended goal/idea is on how to shorten test files to unique test names

keyseven123 (changes requested):
One minor stuff.

nes-systests/systest/src/SystestState.cpp
adrianmichalke: addressed

keyseven123 (approved):
Not blocking but you could add a `*.test` file for testing this exact error/bug.

adrianmichalke:
@keyseven123 The slt corpus has a lot of occurences of this type so we will see any errors of this type in the nightly after https://github.com/nebulastream/nebulastream/pull/1652 got merged. Plus this PR adds a unit test for it

PR #1767, opened 2026-07-15
ci: harden Docker image preparation

ls-1801:
## Summary
- preload all Docker images needed by CI jobs
- retry image pulls and manifest merges with bounded exponential backoff
- include and publish the runtime-base image with dependency images

## Test plan
- CI workflow validation
- Local YAML, shell, and retry checks

### DIFF
--- a/.github/workflows/get_dev_images.yml
+++ b/.github/workflows/get_dev_images.yml
@@ -122,6 +122,35 @@ jobs:
     steps:
       - run: echo "Acquired lock for ${{ needs.detect-dependency-changes.outputs.tag }}"
 
+  build-runtime-image:
+    name: "Build Runtime Base Image"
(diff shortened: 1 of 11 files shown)


.github/workflows/get_dev_images.yml
```diff
@@ -122,6 +122,35 @@ jobs:
     steps:
       - run: echo "Acquired lock for ${{ needs.detect-dependency-changes.outputs.tag }}"
 
+  build-runtime-image:
```
keyseven123, line 125: But without adding this to needs, we will already start the `systest-remnote-test` that requires the runtime docker image and might use an old one

keyseven123 (changes requested):
One small thing

keyseven123 (approved):

Artraxon (approved):

PR #1769, opened 2026-07-16
Fix(#1553) sql-parser: reject multi-argument type constructors

### LINKED ISSUE #1553: Type constructors silently accept multiple arguments
## Summary

Type constructors (`VARSIZED`, `UINT32`, `UINT64`, `INT32`, `FLOAT64`, etc.) silently accept multiple comma-separated arguments. The extra arguments are not rejected — instead the parser silently uses only the last argument and discards the rest.

## Reproduction

The following queries should all be rejected but currently succeed:

```
# name: datatype/TypeConstructorMultipleArguments.test
# description: Type constructors with multiple arguments should be rejected
# groups: [VariableSizedData]

CREATE LOGICAL SOURCE single(id UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR single TYPE File;
ATTACH INLINE
42

# Multiple string args to VARSIZED — silently outputs "world"
SELECT VARSIZED("hello", "world") as v FROM single INTO File();
----
ERROR 2000

# Multiple numeric args to UINT32 — silently outputs 2
SELECT UINT32(1, 2) as v FROM single INTO File();
----
ERROR 2000

# Multiple numeric args to UINT64 — silently outputs 3
SELECT UINT64(1, 2, 3) as v FROM single INTO File();
----
ERROR 2000

# Multiple numeric args to INT32 — silently outputs 2
SELECT INT32(1, 2) as v FROM single INTO File();
----
ERROR 2000

# Multiple numeric args to FLOAT64 — silently outputs 2.0
SELECT FLOAT64(1.0, 2.0) as v FROM single INTO File();
----
ERROR 2000
```

All 5 tests fail with `expected error 2000 but query succeeded`.

## Expected behavior

The parser should reject type constructors with more than one argument.

zeuchste:
## Root cause

In `AntlrSQLQueryPlanCreator::exitFunctionCall`, the datatype-constructor branch (`UINT32(...)`, `VARSIZED(...)`, ...) popped a single value from `constantBuilder.back()` without ever checking `context->argument.size()`. With `UINT32(1, 2)` the parser therefore built a `ConstantValueLogicalFunction` from the last argument only and silently discarded (and leaked into `constantBuilder`) the rest.

## Fix

Validate in the parser that a type constructor is called with exactly one argument and throw `InvalidQuerySyntax` (error 2000) otherwise, e.g. `Type constructor UINT32 expects exactly 1 argument, got 3 at UINT32(1,2,3)`. This also turns the zero-argument case `UINT32()` into a clear error instead of depending on whatever happened to be in the constant builder.

The deeper registry level is already safe: `RegisterCastToTypeLogicalFunction` rejects child counts != 1 with `CannotDeserialize`.

## Tests

New systest `nes-systests/datatype/TypeConstructorMultipleArguments.test` with the five negative cases from the issue (`VARSIZED('hello','world')`, `UINT32(1,2)`, `UINT64(1,2,3)`, `INT32(1,2)`, `FLOAT64(1.0,2.0)`), a zero-argument negative case, and a single-argument positive control.

Closes #1553

### DIFF
--- a/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
+++ b/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
@@ -1208,6 +1208,11 @@ void AntlrSQLQueryPlanCreator::exitFunctionCall(AntlrSQLParser::FunctionCallCont
             /// Check if the function is a constructor for a datatype
             if (const auto dataType = DataTypeProvider::tryProvideDataType(funcName); dataType.has_value())
             {
+                if (const auto numArgs = context->argument.size(); numArgs != 1)
+                {
+                    throw InvalidQuerySyntax(
+                        "Type constructor {} expects exactly 1 argument, got {} at {}", funcName, numArgs, context->getText());
+                }
                 if (helpers.top().constantBuilder.empty())
                 {
                     throw InvalidQuerySyntax("Expected constant, got nothing at {}", context->getText());
--- a/nes-systests/datatype/TypeConstructorMultipleArguments.test
+++ b/nes-systests/datatype/TypeConstructorMultipleArguments.test
@@ -1,0 +1,43 @@
+# name: datatype/TypeConstructorMultipleArguments.test
+# description: Type constructors with multiple arguments should be rejected
+# groups: [DataTypes, Parser]
+
+CREATE LOGICAL SOURCE single(id UINT64 NOT NULL);
+CREATE PHYSICAL SOURCE FOR single TYPE File;
+ATTACH INLINE
+42
+
+# Positive control: a type constructor with exactly one argument works
+SELECT UINT32(1) AS v FROM single INTO File();
+----
+1
+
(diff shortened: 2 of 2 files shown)


adrianmichalke (approved):

zeuchste:
CI failure analysis: the three aborting systest suites died with `Precondition violated: (!std::ranges::empty(schema)): We can not lower an empty schema!` in `LowerSchemaProvider::lowerSchema` (via `LowerToPhysicalProjection`). The culprit was the positive-control query `SELECT UINT32(1) AS v FROM single` — a constant-only projection accesses no source field, so the pruned scan schema is empty and the PRECONDITION aborts the binary. This is a pre-existing engine limitation unrelated to the arity fix; existing tests like `RawNumericLiteralTypeInference.test` avoid it by adding a `WHERE` on a source field. Adopted the same pattern (`WHERE id == 42`). The negative cases throw at parse time and are unaffected.

Artraxon (approved):

PR #1770, opened 2026-07-16
Task(#1692): drive CI clang-tidy-diff via the tidy-diff CMake target

### LINKED ISSUE #1692: [TASK] Migrate CI clang-tidy-diff bash to the tidy-diff CMake target (follow-up to #1609)
Follow-up to #1609 / PR #1685, which added the `tidy-diff`/`tidy-diff-fix` CMake targets but deferred migrating the CI stages to invoke them, to keep that PR focused and avoid risking CI.

## Goal
Replace the handwritten bash in the CI clang-tidy-diff stage with an invocation of the new `tidy-diff` CMake target, so the local developer command and the CI check exercise the same code path.

## Relevant
- New target/module: `cmake/ClangTidyDiff.cmake` + `scripts/tidy-diff.sh` (from PR #1685).
- CI workflow performing the current diff check: `.github/workflows/clang_tidy_diff.yml`.
- The target writes a report to `<build-dir>/clang-tidy-diff-report.txt` — wire that up as a CI artifact.

## Acceptance
- [ ] CI clang-tidy-diff stage calls `cmake --build --target tidy-diff` (with the appropriate `NES_TIDY_DIFF_BASE`)
- [ ] The report file is uploaded as a CI artifact
- [ ] The handwritten bash is removed

zeuchste:
Follow-up to #1609 / #1685: the CI clang-tidy-diff stage now invokes the `tidy-diff` CMake target instead of duplicating the `clang-tidy-diff` invocation in handwritten bash, so the local developer command and the CI check exercise the same code path (`scripts/tidy-diff.sh`).

## What changed

**`.github/workflows/clang_tidy_diff.yml`**
- The two handwritten `clang-tidy-diff-19.py` steps are replaced by a single step running `cmake --build build --target tidy-diff`, with `NES_TIDY_DIFF_BASE=${{ inputs.base_sha }}` selecting the PR base — the same `git diff -U0 <base>` selector (incl. the `.inc`/rust-bindings/vcpkg exclusions) the bash used, now maintained in one place.
- The pre-check step (`-checks='-*,readability-duplicate-include'`) is removed: the full run already enforces `readability-duplicate-include` as an error (it is enabled by `Checks: ['*', ...]` and not excluded from `WarningsAsErrors` in `.clang-tidy`), so no coverage is lost — only the fast-fail shortcut.
- Job and check names are untouched (`clang-tidy-diff / Clang-Tidy Diff (x64, libcxx, none, Debug)`), so branch protection keeps matching.

## Artifact wiring
The target writes `build/clang-tidy-diff-report.txt` (ANSI-free copy of the diagnostics). The existing `clang-tidy-result` artifact upload (`actions/upload-artifact@v5`, `!cancelled()`) now uploads both that report and `clang-tidy-result/fixes.yml`.

## Export-fixes design
The `platisd/clang-tidy-pr-comments` bot needs the `-export-fixes` YAML. `scripts/tidy-diff.sh` gains an optional `NES_TIDY_DIFF_EXPORT_FIXES=<path>` environment variable that is passed through to `clang-tidy-diff`'s `-export-fixes` (the directory is created if needed). CI sets it to `$(pwd)/clang-tidy-result/fixes.yml`; when unset — the local developer flow — no fixes file is written and behavior is unchanged. An env var (mirroring `NES_TIDY_DIFF_BASE`) was chosen over a positional argument because it flows through `cmake --build` to the custom target without touching the target definition. The resolved TODO in `cmake/ClangTidyDiff.cmake` is removed.

Exit-code semantics are preserved: the driver propagates the `clang-tidy-diff` status, so findings still fail the check while the artifact upload and the PR-comment bot run via `!cancelled()`.

Closes #1692

### DIFF
--- a/.github/workflows/clang_tidy_diff.yml
+++ b/.github/workflows/clang_tidy_diff.yml
@@ -73,27 +73,37 @@ jobs:
             cmake -GNinja -B build -DUSE_SANITIZER=${{ matrix.sanitizer }} -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DNES_LOG_LEVEL=DEBUG -DNES_SPLIT_TEST_BINARIES=OFF
             # We need to build the project, as some headers are only created during the build.
             cmake --build build -j  -- -k 0
-      - name: Create results directory
-        run: mkdir clang-tidy-result
       - uses: ./.github/steps/run-in-container
         name: Run Clang Tidy Pre-check
         with:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |
-            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml  -checks='-*,readability-duplicate-include' -j $(nproc)
+            # Fast-fail pass: only duplicate includes, so trivial mistakes surface
+            # before the full all-checks pass below.
+            printf "Checks: \"-*,readability-duplicate-include\"\nWarningsAsErrors: \"*\"\n" > build/clang-tidy-precheck.yaml
+            export NES_TIDY_DIFF_BASE=${{ inputs.base_sha }}
+            export NES_TIDY_DIFF_CONFIG_FILE=$(pwd)/build/clang-tidy-precheck.yaml
+            # Export fixes here too: if this pass fails the job stops, and the
+            # PR-comment bot still needs a fixes.yml to annotate from.
+            export NES_TIDY_DIFF_EXPORT_FIXES=$(pwd)/clang-tidy-result/fixes.yml
+            cmake --build build --target tidy-diff
       - uses: ./.github/steps/run-in-container
         name: Run Clang Tidy with all Checks
         with:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |
-            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml -j $(nproc)
+            export NES_TIDY_DIFF_BASE=${{ inputs.base_sha }}
+            export NES_TIDY_DIFF_EXPORT_FIXES=$(pwd)/clang-tidy-result/fixes.yml
+            cmake --build build --target tidy-diff
       - name: Upload Clang-Tidy Results
         if: ${{ !cancelled() && !github.event.act }}
         uses: actions/upload-artifact@v5
         with:
           if-no-files-found: warn
           name: clang-tidy-result
-          path: clang-tidy-result
+          path: |
+            clang-tidy-result
+            build/clang-tidy-diff-report.txt
       - name: Run clang-tidy-pr-comments action
         if: ${{ !cancelled() && github.event_name == 'pull_request' }}
         # requires python3-venv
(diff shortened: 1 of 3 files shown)


.github/workflows/clang_tidy_diff.yml
```diff
@@ -71,27 +71,26 @@ jobs:
...
-        with:
-          image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
-          run: |
-            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml -j $(nproc)
+            # Same code path as the local developer command (scripts/tidy-diff.sh via
+            # cmake/ClangTidyDiff.cmake). The diff base is the PR base commit; the
+            # exported fixes YAML feeds the PR-comment bot below.
+            export NES_TIDY_DIFF_BASE=${{ inputs.base_sha }}
+            export NES_TIDY_DIFF_EXPORT_FIXES=$(pwd)/clang-tidy-result/fixes.yml
+            cmake --build build --target tidy-diff
```
ls-1801, line 97: removing this step loses the fast-fail pre-check: obvious duplicate includes now wait for the entire all-checks pass. Please preserve both passes through the shared target by adding an optional environment variable such as `NES_TIDY_DIFF_CONFIG_FILE`, forwarding it as `-config-file` in `scripts/tidy-diff.sh`, and running `tidy-diff` first with the focused configuration and then with the default `.clang-tidy`.

.github/workflows/clang_tidy_diff.yml
```diff
@@ -71,27 +71,26 @@ jobs:
...
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |
-            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml  -checks='-*,readability-duplicate-include' -j $(nproc)
-      - uses: ./.github/steps/run-in-container
-        name: Run Clang Tidy with all Checks
-        with:
-          image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
-          run: |
-            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml -j $(nproc)
+            # Same code path as the local developer command (scripts/tidy-diff.sh via
```
ls-1801: avoid comments like these, its no really relevant here

cmake/ClangTidyDiff.cmake
```diff
@@ -17,8 +17,10 @@
 #   tidy-diff-fix -- check mode + `-fix` (applies suggested edits)
 #
 # The actual logic lives in scripts/tidy-diff.sh so the local developer command
-# and CI exercise the same code path. This module only discovers the tools and
-# wires up the targets.
+# and CI (.github/workflows/clang_tidy_diff.yml) exercise the same code path.
```
ls-1801: The comment is also not relevant

ls-1801 (changes requested):

.github/workflows/clang_tidy_diff.yml
zeuchste: Done as suggested: `scripts/tidy-diff.sh` now forwards an optional `NES_TIDY_DIFF_CONFIG_FILE` as `-config-file`, and the workflow runs the shared `tidy-diff` target twice — first the fast-fail pre-check with a focused config (`-*,readability-duplicate-include`, WarningsAsErrors: *), then the full pass with the default `.clang-tidy` (+ export-fixes for the PR-comment bot). Both passes go through the same code path as the local command.

.github/workflows/clang_tidy_diff.yml
zeuchste: Removed.

cmake/ClangTidyDiff.cmake
zeuchste: Trimmed to the pre-existing wording.

ls-1801 (approved):

keyseven123 (approved):

PR #1774, opened 2026-07-16
Fix(#1493) nes-sinks: distinguish SinkCatalog registration failures

### LINKED ISSUE #1493: SinkStatementHandler reports "sink already exists" for config validation failures

### LINKED ISSUE #1504: SinkCatalog::addSinkDescriptor silently ignores duplicate sink names

zeuchste:
## Root cause

`SinkCatalog::addSinkDescriptor` could not tell its callers *why* a registration failed, and silently tolerated duplicates:

- It ignored the `bool` returned by `lockedSinks->emplace(...)`: registering a duplicate name was a silent no-op, yet the method returned the freshly built descriptor as if it had been registered (#1504). Callers ended up holding a descriptor the catalog never stored.
- Although the signature already promises `std::expected<SinkDescriptor, Exception>`, a validation failure of a **known** sink type threw straight out of the method (`DescriptorConfig::validateAndFormat` throws, and the validation registry does not catch), while an **unknown** sink type and an invalid config were both collapsed into one generic `InvalidConfigParameter` (#1493).

## Error-type design

`addSinkDescriptor` keeps the `std::expected<SinkDescriptor, Exception>` shape (existing factories from `ExceptionDefinitions.inc`, no new codes) and now returns one distinct error per failure:

| failure | error |
| --- | --- |
| reserved all-digit name | `InvalidConfigParameter` (1000) |
| unknown sink type | `UnknownSinkType` (2016) |
| invalid config of a known type | the **underlying** validation error, caught and channeled into the `expected` (e.g. `Non-default parameter file_path not specified in config of File`) |
| duplicate name | `SinkAlreadyExists` (2034), `emplace`'s bool is checked; the existing registration stays untouched |

`StatementHandler` already forwards `created.error()` verbatim, so `CREATE SINK` now reports the real failure instead of "sink already exists".

## Downstream systest fixes (#1504)

- `SLTSinkFactory::registerSink` and its per-query sink provider no longer fabricate an unconditional `SinkAlreadyExists`; they propagate the catalog's real error.
- `createSink` no longer discards `registerSink`'s `bool`.
- **UB fix:** `setNamedSink` called `currentBuilder.setException(...)` on failure and then fell through to `sinkExpected.value()` — dereferencing an `expected` without a value. It now throws; the callers' existing `catch (Exception&)` blocks record the exception on the query builder, matching the surrounding idiom.
- Differential blocks (`====`) run `setSinks` twice with the same query number, so left and right plans previously registered the *same* assigned sink name and relied on the silent duplicate no-op (the right plan received a descriptor the catalog never stored). The right-hand plan now registers its named sinks under a distinct `...DIFFERENTIAL` assigned name.
- `Union.test` declared `sink5` twice with different schemas — a latent bug the no-op masked. The second declaration is dead (its query writes `INTO sink4`) and is removed.

## Tests

- New `nes-sinks/tests/SinkCatalogTest.cpp` (new test target for nes-sinks) covering successful registration, duplicate name (`SinkAlreadyExists`, original entry preserved), invalid config (`InvalidConfigParameter`, nothing registered), unknown sink type (`UnknownSinkType`), and the reserved all-digit name.
- No systest repro for #1493 was added: `ERROR` expectations in `.test` files only attach to query blocks, not to `CREATE SINK` statements (a failing `CREATE` aborts loading the file), so the invalid-config-vs-already-exists distinction is covered by the unit tests instead.

Closes #1493
Closes #1504

### DIFF
--- a/nes-frontend/apps/cli/tests/offline.bats
+++ b/nes-frontend/apps/cli/tests/offline.bats
@@ -365,7 +365,7 @@ bad_topology() {
   local f=$(bad_topology 's/type: Void/type: NonExistentSink/')
   run $NES_CLI -d -t "$f" dump
   [ "$status" -eq 1 ]
-  [[ "$output" == *"Invalid configuration for sink"* ]]
+  [[ "$output" == *"unknown sink type"* ]]
   [[ "$output" == *"NONEXISTENTSINK"* ]]
 }
 
--- a/nes-sinks/CMakeLists.txt
+++ b/nes-sinks/CMakeLists.txt
@@ -30,3 +30,5 @@ if (NES_ENABLE_PRECOMPILED_HEADERS)
 endif ()
 
 create_registries_for_component(Sink SinkValidation)
+
+add_tests_if_enabled(tests)
--- a/nes-sinks/src/SinkCatalog.cpp
+++ b/nes-sinks/src/SinkCatalog.cpp
@@ -25,6 +25,7 @@
 #include <utility>
 #include <variant>
 #include <vector>
+#include <Configurations/Descriptor.hpp>
 #include <DataTypes/Schema.hpp>
 #include <DataTypes/SchemaFwd.hpp>
 #include <DataTypes/UnboundField.hpp>
@@ -51,18 +52,35 @@ std::expected<SinkDescriptor, Exception> SinkCatalog::addSinkDescriptor(
         return std::unexpected{InvalidConfigParameter("Sink name '{}' is invalid: only-digit names are reserved", sinkName)};
     }
 
-    auto descriptorConfigOpt = SinkDescriptor::validateAndFormatConfig(sinkType.asCanonicalString(), std::move(config));
-    if (not descriptorConfigOpt.has_value())
+    auto descriptorConfig = [&]() -> std::expected<DescriptorConfig::Config, Exception>
     {
-        return std::unexpected{InvalidConfigParameter("Invalid configuration for sink '{}' of type '{}'", sinkName, sinkType)};
+        try
+        {
+            if (auto validated = SinkDescriptor::validateAndFormatConfig(sinkType.asCanonicalString(), std::move(config)))
+            {
+                return std::move(validated).value();
+            }
+            return std::unexpected{UnknownSinkType("{}", sinkType)};
+        }
+        catch (const Exception& configValidationError)
+        {
+            /// Validation of a known sink type throws on invalid configs; report what actually failed.
+            return std::unexpected{configValidationError};
+        }
+    }();
+    if (not descriptorConfig.has_value())
+    {
+        return std::unexpected{descriptorConfig.error()};
     }
 
     const auto lockedSinks = sinks.wlock();
     auto sinkDescriptor = SinkDescriptor{NamedSinkDescriptor{
-        sinkName, schema, sinkType.asCanonicalString(), std::move(host), formatConfig, std::move(descriptorConfigOpt.value())}};
+        sinkName, schema, sinkType.asCanonicalString(), std::move(host), formatConfig, std::move(descriptorConfig.value())}};
 
-    /// TODO #1504: duplicate sinks are not registered
-    lockedSinks->emplace(std::move(sinkName), sinkDescriptor);
+    if (const auto [existing, inserted] = lockedSinks->emplace(std::move(sinkName), sinkDescriptor); not inserted)
+    {
+        return std::unexpected{SinkAlreadyExists("{}", existing->first)};
+    }
     return sinkDescriptor;
 }
 
--- a/nes-sinks/tests/CMakeLists.txt
+++ b/nes-sinks/tests/CMakeLists.txt
@@ -1,0 +1,14 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#    https://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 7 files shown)


zeuchste:
CI follow-up (88a5e7caf3): the differential systest failures were not caused by the DIFFERENTIAL sink-name suffix itself but by a second pre-existing reliance on the silent duplicate no-op: a `====` block binds its **left** query twice (once via the parser's query callback, once inside the differential block callback, same query id), re-registering the same assigned sink name. Since an assigned name deterministically identifies its descriptor, `SLTSinkFactory`'s per-query provider now reuses the existing registration on `SinkAlreadyExists` instead of failing. The suffix stays so the right-hand plan registers its own descriptor pointing at the differential result file (reusing the left descriptor would make the comparison write to and read from the same file).

Also updated `offline.bats`: an unregistered sink type is now reported as `unknown sink type; <TYPE>` (2016) instead of the generic invalid-config message the test previously asserted.

greenBene (changes requested):
Is this PR coordinated with @yschroeder97 ? I'm not sure if this is the right time to change sth here when he is about to merge major logic changes to the catalogs :) 

If it is coordinated, please re-request once the tests pass.

zeuchste:
Second CI follow-up (48b916179a): the remaining three suite failures were another latent collision the silent duplicate no-op had been masking — the harness's assigned per-query sink names, `UPPER(name + queryNumber)`, can equal a *user-declared* sink name. Concretely, in `regression/ChainedFunctionsInWhere.test` query 2 writing `INTO sinkStream` registered `SINKSTREAM2`, and the later `CREATE SINK sinkStream2` then failed with `SinkAlreadyExists`; same for `INTO sink` (query 4) vs `CREATE SINK sink4` in `Union.test`. Assigned names now use a `#` separator (`UPPER(name)#queryNumber[suffix]`), which cannot occur in an unquoted SQL identifier, so they are disjoint from user-declared names by construction. A scan simulating the scheme across all `.test` files finds no remaining collisions; nothing consumes the name format (result files, EXPLAIN output, and differential pairing are all independent of it).

keyseven123 (changes requested):
Please clean up the commits and then re-request.

PR #1775, opened 2026-07-16
Feature(#1731): human-readable byte-size config options

### LINKED ISSUE #1731: Add ByteAmount TypeValidator / ScalarOption<Byte> for size configs (support 4KiB, 4gb, etc.)
## Problem

Configuration options that represent a number of bytes currently require the user to specify the raw byte count. Examples:

- `nes-query-compiler/interface/QueryExecutionConfiguration.hpp:70` — `operator_buffer_size` (default `4096`)
- `nes-query-compiler/interface/QueryExecutionConfiguration.hpp:55` — `page_size`
- `nes-runtime/interface/Configuration/WorkerConfiguration.hpp:46` — `number_of_buffers_in_global_buffer_manager` (a count of buffers, not bytes — but the *size* per buffer is a candidate too)

Writing `4096` instead of `4KiB`, or `1073741824` instead of `1GiB`, is error-prone and hard to read in YAML configs / CLI overrides.

## Proposal

Introduce a byte-quantity option so users can write human-readable size strings and the option resolves to `uint64_t` bytes.

Two implementation shapes to pick from:

1. **`ByteAmountValidation : ConfigurationValidation`** used with `UIntOption`, plus a shared parser used both in the validator and at parse time (post-processed string → bytes before assignment).
2. **`ScalarOption<Byte>`** where `Byte` is a strong-typedef wrapper around `uint64_t`, with `convertFromString<Byte>` in `ScalarOption.hpp` doing the parsing. Getter returns raw bytes.

Recommendation: (2), since existing validators only accept/reject strings and don't transform them; parsing `4KiB` → `4096` needs to happen during value assignment, which is what `convertFromString` already does per-type in `ScalarOption.hpp:63-95`.

## Unit convention

Follow Kubernetes' `resource.Quantity` grammar (also used by Prometheus, containerd):

- **Binary (power-of-2)** suffixes: `Ki`, `Mi`, `Gi`, `Ti`, `Pi`, `Ei` (1 KiB = 1024 B)
- **Decimal (power-of-10)** suffixes: `k`/`K`, `M`, `G`, `T`, `P`, `E` (1 KB = 1000 B)
- No suffix → bytes
- Case-insensitive `B` suffix optional (`4KiB` == `4Ki`)
- Integer or fractional value allowed (`1.5Gi` = `1610612736`)

Reference: https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/

Reject negative values and overflow.

## Scope

- Add `nes-configurations/include/Configurations/Validation/ByteAmountValidation.hpp` (or `Byte` type + specialization).
- Migrate the byte-typed options listed above.
- Unit tests in `nes-configurations/tests/UnitTests/Configurations/` covering: bare integers, all binary/decimal suffixes, fractional values, invalid strings, overflow, case sensitivity.
- Update `docs/` config examples once ported.

## Out of scope

- Retro-actively changing the default *values* — defaults stay numerically identical, only literal representation changes (e.g. `"4096"` → `"4KiB"`).

zeuchste:
## Summary

Byte-sized configuration options now accept human-readable size strings such as `4KiB`, `1.5Gi`, or `4gb` instead of only raw byte counts, via a new `Byte` strong typedef and `ScalarOption<Byte>` (`ByteOption`), as recommended in #1731 (shape 2).

## Grammar

`parseByteAmount(std::string_view) -> std::expected<uint64_t, Exception>` (shared by `convertFromString<Byte>`, `YAML::convert<Byte>`, and the new `ByteAmountValidation`) follows the Kubernetes [`resource.Quantity`](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/) suffix grammar:

- **Binary** suffixes `Ki`, `Mi`, `Gi`, `Ti`, `Pi`, `Ei` scale by powers of 1024
- **Decimal** suffixes `k`, `M`, `G`, `T`, `P`, `E` scale by powers of 1000
- No suffix means bytes; an optional trailing `B` is ignored (`4KiB` == `4Ki`, `4kb` == `4k`); suffix matching is case-insensitive overall (`4kib`, `4GB`)
- Fractional values are allowed (`1.5Gi` == 1610612736) but are **rejected if they do not scale to a whole number of bytes** (e.g. `1.0000000001Gi`)
- Rejected: signed values, overflow beyond `UINT64_MAX` (checked with 128-bit intermediates), bare suffixes, empty input, garbage

## Migrated options

- `worker.default_query_execution.page_size` (default unchanged: `1024`)
- `worker.default_query_execution.operator_buffer_size` (default unchanged: `4096`)

These are the only worker options that are true byte quantities today; queue sizes and `number_of_buffers_in_global_buffer_manager` are element counts and stay `UIntOption`.

## Backward compatibility

Plain numbers keep parsing as bytes, so all existing YAML configs, CLI overrides (`--worker.default_query_execution.page_size=4096`), and systest `GlobalConfiguration` lines behave exactly as before. `Byte` converts implicitly to `uint64_t`, so `getValue()` call sites are unchanged.

## Tests

- `byte-amount-tests` (nes-configurations): all binary/decimal suffixes, case variants, fractional values, plain byte counts, and rejection of negatives, overflow, non-integral fractions, bare suffixes, and garbage; plus `ByteOption` round-trips through defaults, command-line input, and YAML.
- `QueryExecutionConfigurationTest` (nes-query-compiler): migrated options parse `4KiB`/`1Mi` and keep their numeric defaults.

Closes #1731

### DIFF
--- a/nes-configurations/include/Configurations/ByteAmount.hpp
+++ b/nes-configurations/include/Configurations/ByteAmount.hpp
@@ -1,0 +1,88 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+#pragma once
+
+#include <cstdint>
+#include <expected>
+#include <ostream>
+#include <string_view>
+#include <fmt/ostream.h>
+#include <yaml-cpp/yaml.h>
+#include <ErrorHandling.hpp>
+
+namespace NES
+{
+
+/// Parses a human-readable byte amount following the Kubernetes resource.Quantity suffix grammar
+/// (https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/):
+/// - binary suffixes Ki, Mi, Gi, Ti, Pi, Ei scale by powers of 1024
+/// - decimal suffixes k, M, G, T, P, E scale by powers of 1000
+/// - no suffix means bytes, so plain numbers ("4096") keep working
+/// - suffix matching is case-insensitive and an optional trailing 'B' is ignored: "4KiB" == "4Ki" == "4kib", "4kb" == "4k"
+/// - fractional values are allowed ("1.5Gi" == 1610612736 bytes) but must scale to a whole number of bytes ("1.5" alone is rejected)
+/// Rejects signed values, results that are not a whole number of bytes, results exceeding UINT64_MAX, and anything else
+/// not matching the grammar (empty input, bare suffixes, unknown suffixes, more than 20 significant fractional digits).
+std::expected<uint64_t, Exception> parseByteAmount(std::string_view amount);
+
+/// Strong typedef around uint64_t that marks a configuration value as a byte quantity, so that ScalarOption<Byte>
+/// accepts human-readable size strings ("4KiB", "1.5Gi") in addition to plain byte counts (see parseByteAmount).
+struct Byte
+{
+    constexpr Byte() = default;
+
+    constexpr explicit Byte(const uint64_t bytes) : bytes(bytes) { }
+
+    /// Implicit on purpose: options migrated from UIntOption to ByteOption keep treating the value as a plain number of bytes,
+    /// so existing call sites (`uint64_t size = option.getValue();`, arithmetic, function arguments) stay unchanged.
+    constexpr operator uint64_t() const { return bytes; } /// NOLINT(google-explicit-constructor,hicpp-explicit-conversions)
+
+    friend std::ostream& operator<<(std::ostream& os, const Byte& byte) { return os << byte.bytes; }
+
+    uint64_t bytes{0};
+};
+
+}
+
+namespace fmt
+{
+template <>
+struct formatter<NES::Byte> : ostream_formatter
+{
+};
+}
+
+namespace YAML
+{
+/// Allows YAML-based option parsing (`node.as<NES::Byte>()`) to accept human-readable byte amounts.
+template <>
+struct convert<NES::Byte>
+{
+    static Node encode(const NES::Byte& byte) { return Node(byte.bytes); }
+
+    static bool decode(const Node& node, NES::Byte& result)
+    {
+        if (!node.IsScalar())
+        {
+            return false;
+        }
+        const auto parsed = NES::parseByteAmount(node.Scalar());
+        if (!parsed.has_value())
+        {
+            return false;
+        }
+        result = NES::Byte(parsed.value());
+        return true;
+    }
+};
+}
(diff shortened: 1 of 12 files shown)


nes-configurations/include/Configurations/ByteAmount.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+/// Parses a human-readable byte amount following the Kubernetes resource.Quantity suffix grammar
+/// (https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/):
+/// - binary suffixes Ki, Mi, Gi, Ti, Pi, Ei scale by powers of 1024
+/// - decimal suffixes k, M, G, T, P, E scale by powers of 1000
+/// - no suffix means bytes, so plain numbers ("4096") keep working
+/// - suffix matching is case-insensitive and an optional trailing 'B' is ignored: "4KiB" == "4Ki" == "4kib", "4kb" == "4k"
+/// - fractional values are allowed ("1.5Gi" == 1610612736 bytes) but must scale to a whole number of bytes ("1.5" alone is rejected)
+/// Rejects signed values, results that are not a whole number of bytes, results exceeding UINT64_MAX, and anything else
+/// not matching the grammar (empty input, bare suffixes, unknown suffixes, more than 20 significant fractional digits).
```
adrianmichalke: not a fan of these ai comments. they are formulated like a response to your prompt ('plain numbers keeps working', 'existing call sites stay unchanged').

adrianmichalke: Do we need the link? Isnt it general knowledge?

adrianmichalke (approved):
LGTM, not a fan of the ai comments though

nes-configurations/include/Configurations/ByteAmount.hpp
zeuchste: Fair point. Reworded them in a94c64b76c to state what the code does.

nes-configurations/include/Configurations/ByteAmount.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+/// Parses a byte amount following the Kubernetes resource.Quantity suffix grammar
+/// (https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/quantity/):
+/// - binary suffixes Ki, Mi, Gi, Ti, Pi, Ei scale by powers of 1024
+/// - decimal suffixes k, M, G, T, P, E scale by powers of 1000
+/// - no suffix means bytes
+/// - suffix matching is case-insensitive, an optional trailing 'B' is ignored: "4KiB" == "4Ki" == "4kib", "4kb" == "4k"
+/// - fractional values must scale to a whole number of bytes: "1.5Gi" == 1610612736, "1.5" is rejected
+/// Rejects signed values, non-integral results, results exceeding UINT64_MAX, empty input, bare or unknown suffixes,
+/// and more than 20 significant fractional digits.
```
ls-1801: It's not quite true; we should adapt the comment to say that it's inspired. It does not strictly follow the Kubernetes grammar. Kubernetes supports exponents, for example, which we don't need

ls-1801 (approved):
Just a minor change

Artraxon (changes requested):
I would prefer not to make the 'b' suffix optional. it the unit, i.e. the type, and we are requesting special parsing logic based on the type, so, while the k, Gi, etc. are just a parameter for that type, so it should be explicit to prevent accidental conversions.

nes-configurations/include/Configurations/ByteAmount.hpp
zeuchste: Reworded: the comment now says the grammar is *inspired by* Kubernetes' resource.Quantity and names the deviations (no exponent notation, case-insensitive suffixes); dropped the link.

PR #1776, opened 2026-07-16
Task(#1730): non-throwing GeneratorSource validation

### LINKED ISSUE #1730: [TASK] Make Generator Source Validation non-throwing
### Description

While reworking the config interface that is accessible from the frontend, I have noticed that the Generator Source Validation is a bit of a mess, and the fact that some function objects somewhere are throwing without that being visible is even worse. 
There is no good reason to throw there, they should have explicit error types.

### Acceptance Criteria

While it would in general be good to create more typed objects directly (which will be possible in the future), a first step would be to not throw exceptions in the validation of for example the generator fields, but rather return some error type (i.e. Exceptions or ideally something even more informative) via an optional, std::expected or variant.

zeuchste:
## Summary

GeneratorSource config validation no longer throws on malformed user config. All validation functions return `std::expected<T, Exception>` (using the existing exception factories from `ExceptionDefinitions.inc` as error payload, so errors keep their code + message); the single remaining throw sits visibly at the registry adapter, whose interface reports errors via exceptions.

## Throw-site inventory (before → after)

Validation path (all converted to `std::expected` error returns):

| Location | Throws before | After |
|---|---|---|
| `ConfigParametersGenerator::SEQUENCE_STOPS_GENERATOR` lambda | 2 | plain `tryGet`, informative error in `validateGeneratorSpecificConfig` |
| `ConfigParametersGenerator::GENERATOR_SCHEMA` lambda | 2 (empty schema, unknown field type) | delegates to `GeneratorFields::validateSchema` |
| `GeneratorFields::SequenceField::validate` | 4 (param count, bad type, VARSIZED/UNDEFINED, unparsable start/end/step) | `std::expected<void, Exception>`, monadic `and_then` chain |
| `GeneratorFields::NormalDistributionField::validate` | 4 | `std::expected<void, Exception>` |
| `GeneratorFields::WordListField::validate` | 4 | `std::expected<void, Exception>` |
| `GeneratorFields::RandomStrField::validate` | 6 | `std::expected<void, Exception>` |
| `DescriptorConfig::validateAndFormat` (framework) | 3 (unknown key, missing mandatory, failed validator) | new non-throwing `tryValidateAndFormat`; the throwing overload delegates to it (other sources unchanged) |

Total: **25 throw statements** removed from the validation path.

Kept as-is (post-validation construction path, i.e. internal invariants per the issue): `Generator` ctor, `Generator::parseRawSchemaLine`, `SequenceField`/`RandomStrField` ctors, and the existing `INVARIANT`s in `NormalDistributionField`/`WordListField` ctors.

## Error shape

- `std::expected<void, Exception>` for per-field validators and `std::expected<DescriptorConfig::Config, Exception>` for the plugin entry point. `Exception` already carries an error code and a formatted message, and `std::expected<SourceDescriptor, Exception>` is established precedent (`SourceCatalog`), so no new error type is introduced and no exception type is defined outside `ExceptionDefinitions.inc`.
- `GeneratorSource::validateAndFormat` first runs `validateGeneratorSpecificConfig` (informative errors: parameter name + offending value + expected values) and then composes with `DescriptorConfig::tryValidateAndFormat` via `and_then`. Returns the first error.
- The `ConfigParameter` validation lambdas keep the framework-mandated `std::optional<T>` signature but are now thin, non-throwing `tryGet` wrappers; the deep checks stay reachable through them as defense in depth.
- `RegisterGeneratorSourceValidation` unwraps the expected and throws the contained error at this single, documented boundary, since `SourceValidationRegistryReturnType` is `DescriptorConfig::Config` for all sources.

## Behavioral notes

- Valid configs validate exactly as before.
- Error messages are at least as informative as the old throw messages; enum parameters now additionally list the accepted values, and a missing `GENERATOR_SCHEMA` now reports "Generator schema cannot be empty!" instead of the generic missing-parameter message.

## Tests

New `generator_source_validation_test` (first unit test for this plugin, wired via `add_tests_if_enabled`) with 24 cases: 3 accepting valid configs (defaults filled in, multi-field schema, sinus rate) and one per distinct validation failure, asserting `ErrorCode::InvalidConfigParameter` plus a message substring including the offending value.

Closes #1730

### DIFF
--- a/nes-configurations/include/Configurations/Descriptor.hpp
+++ b/nes-configurations/include/Configurations/Descriptor.hpp
@@ -16,6 +16,7 @@
 
 #include <concepts>
 #include <cstdint>
+#include <expected>
 #include <memory>
 #include <optional>
 #include <ostream>
@@ -149,10 +150,12 @@ public:
     /// Iterates over all parameters in a user provided config and checks if they are supported by a specific config.
     /// Then iterates over all supported config parameters, validates and formats the strings provided by the user.
     /// Uses default parameters if the user did not specify a parameter.
-    /// @throws If a mandatory parameter was not provided, an optional parameter was invalid, or a not-supported parameter was encountered.
+    /// Returns an error if a mandatory parameter was not provided, an optional parameter was invalid, or a not-supported parameter
+    /// was encountered.
     template <typename SpecificConfiguration>
     requires DescriptorConfigurationConstraints::HasParameterMap<SpecificConfiguration>
-    static Config validateAndFormat(std::unordered_map<std::string, std::string> config, const std::string_view implementationName)
+    static std::expected<Config, Exception>
+    tryValidateAndFormat(std::unordered_map<std::string, std::string> config, const std::string_view implementationName)
     {
         auto validatedConfig = Config{};
 
@@ -161,7 +164,7 @@ public:
         {
             if (not SpecificConfiguration::parameterMap.contains(key))
             {
-                throw InvalidConfigParameter(fmt::format("Unknown configuration parameter: {}.", key));
+                return std::unexpected(InvalidConfigParameter("Unknown configuration parameter: {}.", key));
             }
         }
         /// Next, try to validate all config parameters.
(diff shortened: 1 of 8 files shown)


Artraxon:
I don't mind it, but just FIY I am redoing the config system and stop using the descriptor config altogether, and I am also adding correct passing of errors without throwing exceptions

nes-configurations/include/Configurations/Descriptor.hpp
```diff
@@ -175,19 +178,33 @@ class DescriptorConfig
...
             if (const auto validatedParameter = configParameter.validate(config); validatedParameter.has_value())
             {
                 validatedConfig.emplace(key, validatedParameter.value());
                 continue;
             }
-            throw InvalidConfigParameter(fmt::format("Failed validation of config parameter: {}, in: {}", key, implementationName));
+            return std::unexpected(InvalidConfigParameter("Failed validation of config parameter: {}, in: {}", key, implementationName));
         }
         return validatedConfig;
     }
```
Artraxon, line 203: So what if the generator source validation throws? 
One called the non-throwing version on purpose but got an exception.
This method warp the validation call in a try-catch block.

Artraxon (changes requested):

adrianmichalke (approved):

nes-configurations/include/Configurations/Descriptor.hpp
zeuchste: You're right — a plugin validator that still throws (legacy style) leaked the exception out of the non-throwing entry point. `tryValidateAndFormat` now wraps the per-parameter `validate()` call in try/catch and reports the thrown `Exception` as the `expected`'s error, so callers of the non-throwing API never see a throw regardless of plugin style. Added a regression test with a deliberately throwing validator (`ThrowingValidatorIsReportedAsError`). Doc comment states the guarantee.

PR #1781, opened 2026-07-16
Adds MQTT Source and Sink with preliminary bats testing

ls-1801:
## Summary

Adds MQTT source and sink plugins, the required Paho MQTT dependencies, and distributed Bats coverage.

## Commits

- `233854c019` **Fix source cleanup after a second exception.** Ensures a failure in `fillTupleBuffer` still closes the source, while catching and logging a second exception from `close()` so it does not terminate the process or hide the original failure.
- `59a561fc22` **Add `wait_until_status --require-healthy` for distributed Bats tests.** Lets tests fail early when services matching a regex are unhealthy, which makes worker crashes visible instead of waiting for a query-status timeout.
- `af0e17d68` **Add `assert_file_line_count` for Bats tests.** Provides a reusable assertion for output line counts, including an option to ignore empty lines.
- `fc5e6c68c9` **Add Paho MQTT dependencies.** Adds custom vcpkg ports for the Paho C and C++ clients and enables the MQTT feature by default. The ports include platform/linking fixes and a libc++ compatibility patch: Paho used `std::basic_string<unsigned char>`, which relies on the non-standard `std::char_traits<unsigned char>` specialization; the patch uses `char` instead. The dependency documentation records this reason.
- `863a80f453` **Add MQTT source and sink plugins.** Registers and implements the plugins, supports broker URI/topic/client ID/QoS configuration, handles sink backpressure and source payload buffering, and adds distributed Bats tests with a Mosquitto broker for normal and child-buffer payloads.


keyseven123 (approved):
I was requested as part of being on the DevOps team. I have not looked/reviewed the actual implementation. I will leave this to @alepping and @ThanGeo

alepping (changes requested):
Re-request when you added Nix.

Otherwise lgtm.

nes-plugins/Sinks/MQTTSink/tests/good/example.yaml
```diff
@@ -0,0 +1,67 @@
...
+#   │  worker-1  (cap: 10000) │  ──▶    │ mqtt-broker  │
+#   │  S: GENERATOR_SOURCE    │  publish│ topic:       │
+#   │  K: MQTT_SINK           │  qos=2  │ mqtt-sink-   │
+#   └─────────────────────────┘         │ test         │
+#                                       └──────────────┘
+#
+#   Single worker generates a UINT64 sequence and publishes it as CSV
+#   to the external MQTT broker. No downstream NES connections.
+
+query: |
```
ThanGeo, line 25: Why is the directory named "test/good"?

nes-plugins/Sinks/MQTTSink/tests/good/single-worker-with-4k-buffers.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+workers:
+  - host: worker-1:8080
+    data_address: worker-1:9090
```
ThanGeo, line 15: i'm missing a bit about how distributed configuration works. But what is ```data_address```?

nes-plugins/Sinks/MQTTSink/MQTTSink.hpp
```diff
@@ -0,0 +1,145 @@
...
+        Ok,
+        Full,
+        Closed,
+    };
+
+    PublishResult tryPublish(const TupleBuffer& buffer);
+
+    std::string serverURI;
+    std::string clientId;
+    std::string topic;
```
ThanGeo, line 77: So an MQTT sink can only be correlated with a single topic?

ThanGeo (changes requested):
Looks good overall, I just added some questions in the review.

When responded, squash the commits and re-request for approval

nes-plugins/Sinks/MQTTSink/tests/good/example.yaml
ls-1801: Good examples (i.e, those that are supposed to be working), whereas bad would contain examples that should not work.

nes-plugins/Sinks/MQTTSink/tests/good/single-worker-with-4k-buffers.yaml
ls-1801: This is the port used for worker-to-worker communication. The frontend connects to the host address (gRPC port) and sends queries via the data_address.

nes-plugins/Sinks/MQTTSink/MQTTSink.hpp
ls-1801: correct

nes-plugins/Sources/MQTTSource/MQTTSource.cpp
```diff
@@ -0,0 +1,225 @@
...
+            message = client->consume_message();
+        }
+
+        if (!message)
+        {
+            NES_DEBUG("Wakeup Signal received");
+            continue;
+        }
+
+        payload = message->get_payload();
```
ThanGeo: Have you thought of a way to parameterize the delimiter (if any) in the incoming messages? 
I propose that by default we expect NO DELIMITER, meaning that each mqtt message is expected to carry 1 tuple in its respective format (csv, json) and alternatively, in the case where an mqtt message is sending more than 1 tuples batched together and delimited, then we should have an easily parameterized interface that just sets the delimiter character. 

After all we only have this implementation with the '\n' because we were testing on test datasets and not real incoming mqtt messages, AFAIK sensors only emit one tuple at a time.

nes-plugins/Sources/MQTTSource/MQTTSource.hpp
```diff
@@ -0,0 +1,124 @@
...
+#include <Sources/SourceDescriptor.hpp>
+#include <Util/Logger/Logger.hpp>
+#include <Util/Strings.hpp>
+#include <mqtt/client.h>
+#include <PayloadStash.hpp>
+
+namespace NES
+{
+
+class MQTTSource : public Source
```
ThanGeo, line 40: It would be really useful to add some interface for printing/logging incoming mqtt messages in a readable format.

ThanGeo (changes requested):

nes-plugins/Sources/MQTTSource/MQTTSource.cpp
ls-1801: added a config option: `IMPLICIT_MESSAGE_DELIMITER` which defaults to '\n'. The tests now only send the data without newline at the end. I also added a test that opts out by setting the `IMPLICIT_MESSAGE_DELIMITER` to `''`, that way tuple can span multiple messages

ThanGeo: But how is it handled internally if a tuple spans multiple messages? Is it being reconstructed after all messages have been received? Does this mean that it gets buffered until completed?

ls-1801: Yes, that's what the `SequenceShredder` does. It extracts all tuples that are ready to be processed from a buffer and fixes tuples that span multiple buffers. This is not done by the source. Our sources only ingest bytes into the engine.

nes-plugins/Sources/MQTTSource/MQTTSource.hpp
ls-1801: I added a `LOG_MESSAGES` config that defaults to false, but you can enable to see the messages received on the mqtt source.

alepping (approved):

ThanGeo (approved):

PR #1783, opened 2026-07-16
Fix sql parser issues to support slt corpus

adrianmichalke:
Fixes a set of issues in SQL parsing blocking #1652.

Now validated via `nes-systests/parser/SQLParserRegressions.test`:
```
CREATE LOGICAL SOURCE predicate_input(
    pk UINT64 NOT NULL,
    col0 INT64 NOT NULL,
    col1 FLOAT64,
    col2 INT64,
    col3 INT64 NOT NULL,
    col4 FLOAT64 NOT NULL,
    flag BOOLEAN NOT NULL
);

CREATE PHYSICAL SOURCE FOR predicate_input TYPE File;
ATTACH INLINE
1,1,10.5,,761,500.0,true
2,2,20.0,7,100,563.88,true
3,3,,8,922,600.0,false
4,1,40.0,,600,100.0,false
5,5,27.89,9,34,700.0,true

CREATE LOGICAL SOURCE alias_input(pk UINT64 NOT NULL, value INT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR alias_input TYPE File;
ATTACH INLINE
1,10
2,20

# Parses a parenthesized BETWEEN predicate.
SELECT pk FROM predicate_input WHERE (col1 BETWEEN FLOAT64(10.0) AND FLOAT64(30.0)) INTO File();
----
1
2
5

# Parses a parenthesized IN predicate.
SELECT pk FROM predicate_input WHERE (col0 IN (INT64(1), INT64(2))) INTO File();
----
1
2
4

# Parses a parenthesized IS NULL predicate.
SELECT pk FROM predicate_input WHERE (col1 IS NULL) INTO File();
----
3

# Preserves precedence in nested AND and OR predicates.
SELECT pk
FROM predicate_input
WHERE col4 <= FLOAT64(563.88) AND (col3 >= INT64(922) OR (col0 BETWEEN INT64(1) AND INT64(2)))
INTO File();
----
1
2
4

# Parses deeply nested predicate parentheses.
SELECT pk
FROM predicate_input
WHERE ((col3 > INT64(700))) AND (((col3 BETWEEN INT64(700) AND INT64(950))))
INTO File();
----
1
3

# Combines IN and null predicates.
SELECT pk
FROM predicate_input
WHERE col0 IN (INT64(1), INT64(2)) AND (col1 IS NULL OR col2 IS NOT NULL)
INTO File();
----
2

# Uses a parenthesized predicate as a comparison operand.
SELECT pk
FROM predicate_input
WHERE flag = (col0 = INT64(1) OR col1 = FLOAT64(20.0))
INTO File();
----
1
2

# Accepts a named source alias with unqualified fields.
SELECT pk FROM alias_input AS source_alias WHERE value = INT64(10) INTO File();
----
1

# Resolves alias-qualified fields.
SELECT source_alias.pk FROM alias_input AS source_alias WHERE source_alias.value = INT64(10) INTO File();
----
1

# Resolves an alias-qualified asterisk.
SELECT source_alias.* FROM alias_input AS source_alias INTO File();
----
1,10
2,20

# Resolves fields qualified by the source name.
SELECT alias_input.pk FROM alias_input INTO File();
----
1
2

# Rejects an unknown source qualifier.
SELECT wrong_alias.pk FROM alias_input AS source_alias INTO File();
----
ERROR 2000

# Rejects trailing tokens after a complete query.
SELECT pk FROM alias_input INTO File(); trailing_tokens;
----
ERROR 2000
```

### DIFF
--- a/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
+++ b/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
@@ -228,23 +228,11 @@ LogicalFunction createInFunction(const LogicalFunction& valueFunction, std::vect
     return function;
 }
 
-Identifier collapseQualifiedField(std::vector<LogicalFunction>& functions, const std::string& expressionText)
+bool isSourceQualifier(AntlrSQLParser::IdentifierContext* context)
 {
-    if (functions.size() < 2)
-    {
-        throw InvalidQuerySyntax("Qualified field reference is missing an identifier at {}", expressionText);
-    }
-
-    const auto qualifierFunction = functions.at(functions.size() - 2).tryGetAs<UnboundFieldAccessLogicalFunction>();
-    const auto fieldFunction = functions.back().tryGetAs<UnboundFieldAccessLogicalFunction>();
-    if (!qualifierFunction.has_value() || !fieldFunction.has_value())
-    {
-        throw InvalidQuerySyntax("Only source-qualified field references are supported at {}", expressionText);
-    }
-
-    auto qualifier = qualifierFunction->get().getFieldName();
-    functions.erase(functions.end() - 2);
-    return qualifier;
+    const auto* column = dynamic_cast<AntlrSQLParser::ColumnReferenceContext*>(context->parent);
+    const auto* dereference = column == nullptr ? nullptr : dynamic_cast<AntlrSQLParser::DereferenceContext*>(column->parent);
+    return dereference != nullptr && dereference->base == column;
 }
 
 void validateSourceQualifiers(const AntlrSQLHelper& helper)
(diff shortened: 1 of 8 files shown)


nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -228,6 +228,74 @@ LogicalFunction createInFunction(const LogicalFunction& valueFunction, std::vect
...
+    const auto fieldFunction = functions.back().tryGetAs<UnboundFieldAccessLogicalFunction>();
+    if (!qualifierFunction.has_value() || !fieldFunction.has_value())
+    {
+        throw InvalidQuerySyntax("Only source-qualified field references are supported at {}", expressionText);
+    }
+
+    auto qualifier = qualifierFunction->get().getFieldName();
+    functions.erase(functions.end() - 2);
+    return qualifier;
+}
```
greenBene, line 236: I am not too happy with this solution since it is basically (if I understand correctly) a workaround for ignoring qualifiers after they are parsed. Could we fix it closer to the problem's source? E.g. by skipping them during the AST traversal?

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -228,6 +228,74 @@ LogicalFunction createInFunction(const LogicalFunction& valueFunction, std::vect
...
+
+    const auto firstOperandIndex = functions.size() - operandCount;
+    auto function = std::move(functions[firstOperandIndex]);
+    for (auto operandIndex = firstOperandIndex + 1; operandIndex < functions.size(); ++operandIndex)
+    {
+        function = createLogicalBinaryFunction(std::move(function), std::move(functions[operandIndex]), tokenType);
+    }
+    functions.resize(firstOperandIndex);
+    functions.emplace_back(std::move(function));
+}
```
greenBene, line 291: I struggled a bit to understand what this function and its arguments are doing. While I know that the other functions in this file are not documented, I'd still ask for a short inline documentation explaining what is happening here and why it is needed. 

Also:  I am unsure about the name "operandCount". Could we rename it to be more self-explanatory?

greenBene (changes requested):
Thank you for improving our SQL interoperability! :)

Artraxon (approved):
I admire your persistence

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
adrianmichalke: addressed. The source qualifiers are now detected in enterIdentifier and skipped when building field functions. I also removed collapseQualifiedField and the exitDereference override.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
adrianmichalke: addressed. I renamed the arguments and indices and added a short explanation that child predicates append functions first

greenBene (approved):

yschroeder97 (approved):

PR #1787, opened 2026-07-17
Add configurable combined unit-test binary

### LINKED ISSUE #1785: [TASK] Add configurable unit-test binary grouping
## Description

Unit tests are currently built largely as individual executables. Because NebulaStream links static libraries, this produces many large binaries (approximately 200 MB each in affected configurations), repeated link work, and substantial disk consumption.

Add a CMake-controlled unit-test layout that can link multiple test suites into one binary. The primary target is **one binary for all unit tests**. Investigate and benchmark the following layouts:

- One binary for all unit tests (first implementation target)
- One binary per module (fallback)
- The existing one-binary-per-test-suite layout

A single all-unit-tests binary appears technically feasible: the common `add_nes_test()` helper already supports multiple source files and uses `gtest_discover_tests`, so individual GoogleTest cases can still be exposed to and filtered by CTest. However, it will require consolidating target-specific link dependencies, include paths, compile definitions, and test-data setup.

Per-module binaries should only be selected if a single binary has a concrete technical blocker or unacceptable performance characteristics. The existing per-suite layout should remain available for focused local development.

## Acceptance Criteria

- [ ] Implement and validate one aggregate binary containing all unit tests first.
- [ ] Expose the layout with a documented CMake cache option, including values equivalent to:
  - `single-binary`
  - `per-module`
  - `per-suite` (current behavior)
- [ ] Set CI to `single-binary` once validated.
- [ ] Use `per-module` only if the aggregate binary cannot cleanly preserve target-specific configuration or has unacceptable build/link performance.
- [ ] Keep `per-suite` available on request for tightly focused local development.
- [ ] Preserve CTest test names, labels, filtering, JUnit output, sanitizer behavior, and required test data.
- [ ] Benchmark clean and incremental builds, disk footprint, and execution time for all layouts; document any reason to choose `per-module` instead.

## Implementation Notes

- The existing `add_nes_test()` helper in `cmake/NebulaStreamTest.cmake` accepts multiple source files and discovers GoogleTest cases individually.
- Most project libraries are static by default, which explains the repeated size and link cost.
- CI currently builds the full build graph before invoking CTest, so the option must be passed during CMake configuration in CI workflows.

ls-1801:
## Summary
- add `NES_SPLIT_TEST_BINARIES` to select per-suite or combined unit-test binaries
- use the combined `nes-unit-tests` binary in PR and nightly CI
- fix test-only ODR and duplicate-symbol conflicts exposed by combined linking

Closes #1785

## Test plan
- [x] Debug build with `NES_SPLIT_TEST_BINARIES=ON`
- [x] Aggregate unit tests with `NES_SPLIT_TEST_BINARIES=OFF` (855 passed, 1 disabled)
- [x] Formatting check

### DIFF
--- a/cmake/NebulaStreamTest.cmake
+++ b/cmake/NebulaStreamTest.cmake
@@ -22,15 +22,14 @@ else ()
     set(RAPIDCHECK_MAX_SUCCESS 20)
 endif ()
 
+option(NES_SPLIT_TEST_BINARIES "Build each unit-test suite as a separate executable" ON)
+
 # Target to build all integration tests
 add_custom_target(integration_tests)
 # Target to build all e2e tests
 add_custom_target(e2e_tests)
 
-# This function registers a test with gtest_discover_tests
-function(add_nes_test)
-    add_executable(${ARGN})
-    set(TARGET_NAME ${ARGV0})
+function(configure_nes_test_target TARGET_NAME)
     target_link_libraries(${TARGET_NAME} nes-test-util rapidcheck rapidcheck_gtest)
     if (NES_ENABLE_PRECOMPILED_HEADERS)
         target_precompile_headers(${TARGET_NAME} REUSE_FROM nes-common)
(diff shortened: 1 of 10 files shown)


cmake/NebulaStreamTest.cmake
```diff
@@ -22,15 +22,14 @@ else ()
     set(RAPIDCHECK_MAX_SUCCESS 20)
 endif ()
 
+option(NES_SPLIT_TEST_BINARIES "Build each unit-test suite as a separate executable" ON)
```
keyseven123, line 25: When should we ever want this? What advantage does it have?

keyseven123 (approved):

alepping (approved):

PR #1789, opened 2026-07-17
Skip ci on text file changes

keyseven123:
This PR adds a `detect-changes` step that skips the CI run for tests if there are no code changes. Thus, we reduce the time to merge and the CI usage.

## Verifying this change
This change is tested by running our CI on this pr. Also, the PRs #1790 and #1795 show if these changes work for (non-)code changes. 

- CI Workflow


.github/ci/full-ci-skip-paths.txt
```diff
@@ -0,0 +1,5 @@
+*.md
+**/*.md
+*.txt
+**/*.txt
```
oczerwia-2010, line 4: Could this prevent commits that only chage CMakeLists.txt from beeing tested?

oczerwia-2010 (approved):
LGTM, comment discussed already

.github/ci/full-ci-skip-paths.txt
keyseven123: Done

ls-1801 (approved):
We can do this in a follow-up. 

I would pull apart the docs-change and the code-change PR workflows. And while we are at we can also get rid of the pr-fork workflow, it only causes headaches...

keyseven123:
> We can do this in a follow-up.
> 
> I would pull apart the docs-change and the code-change PR workflows. And while we are at we can also get rid of the pr-fork workflow, it only causes headaches...

I have created issue #1800 for this.

PR #1793, opened 2026-07-17
Use raw literals in systests and documentation

adrianmichalke:


docs/guide/query_api.md
```diff
@@ -38,10 +38,10 @@ Queries can be submitted as YAML specifications or SQL statements. Below is a co
 ```yaml
 query: |
   SELECT start, end, highway, direction, positionDiv5280, AVG(speed) AS avgSpeed
-  FROM (SELECT creationTS, highway, direction, position / INT32(5280) AS positionDiv5280, speed FROM lrb)
+  FROM (SELECT creationTS, highway, direction, position / 5280 AS positionDiv5280, speed FROM lrb)
   GROUP BY (highway, direction, positionDiv5280)
   WINDOW SLIDING(creationTS, SIZE 5 MINUTES, ADVANCE BY 1 SEC)
-  HAVING avgSpeed < FLOAT32(40)
+  HAVING avgSpeed < 40
```
greenBene, line 44: Since it is a float, should it be 40.0?

greenBene: Similar cases for other FLOAT32(*)

greenBene (approved):
I am so incredibly happy that we can now use constants with automatically inferred types! Thank you! I am slightly unsure about some of the changes where we replaced floats with integers, but I´d assume that the casting is done automatically and correctly? 

FYI, the tests were apparently not triggered.

Artraxon (approved):

PR #1794, opened 2026-07-17
Systest: Regex support for explain result lines 

adrianmichalke:
To simplify e2e testing of the optimizer, this PR adds support for verifying the results of Systest `Explain` statements using regular expressions.

Supported are:
 ```text
----
 <REGEX>pattern should exist</REGEX>
 <!REGEX>pattern should not exist</!REGEX>
 ```

 and multiline:

 ```text
----
<REGEX>
SINK\(SINK[0-9]+\)
SOURCE\(stream_[0-9]+\) 
</REGEX>
```

Multiple patterns per result are allowed and each pattern is applied on the whole result text.

### CODE REVIEW

File: docs/development/systests.md [hunk 1]
```diff
@@ -171,6 +171,45 @@ FROM input
...
+An `EXPLAIN` test can compare the complete normalized plan verbatim:
+
+```sql
+EXPLAIN (LOGICAL) FORMAT TEXT
+SELECT id FROM input INTO File();
+----
+== Initial Logical Plan ==
+SINK(FILE)
+  PROJECTION(fields: [ID])
+    SOURCE(INPUT)
```
greenBene, line 185:
```suggestion
EXPLAIN (LOGICAL) FORMAT TEXT
SELECT id FROM input INTO File();
----
== Initial Logical Plan ==
SINK(FILE)
  PROJECTION(fields: [ID])
    SOURCE(INPUT)
==END==
```

Thanks for adding the documentation that I forgot to add 👀 
In the example is the leading `==END==` missing :)

nes-systests/systest/src/SystestResultCheck.cpp
```diff
@@ -767,6 +769,153 @@ QueryCheckResult checkQuery(const NES::Systest::RunningQuery& runningQuery)
...
+    return std::nullopt;
+}
+
+std::string explainRegexSyntaxError(const size_t line, const std::string_view message)
+{
+    return fmt::format("\n\nInvalid Explain Regex Assertion (line {}): {}", line + 1, message);
+}
+
+std::expected<ExplainRegexAssertion, std::string>
+parseInlineExplainRegex(const std::string_view line, const ExplainRegexTags tags, const size_t assertionLine)
```
greenBene: I feel the term 'inline' might be a bit misleading. What do you think about `parseSingleLineExplainRegex`?

greenBene (changes requested):
Thanks for extending the functionality of the Explain statement validation in the system-level tests! :) I just have two minor suggested improvement points.

docs/development/systests.md
adrianmichalke: addressed

nes-systests/systest/src/SystestResultCheck.cpp
adrianmichalke: addressed

greenBene (approved):

nes-systests/systest/src/SystestResultCheck.cpp
yschroeder97, line 1: I must say I really don't like how this file is now over 1K lines long and keeps growing.

nes-systests/systest/tests/SystestResultCheckTests.cpp
```diff
@@ -0,0 +1,164 @@
...
+#include <ErrorHandling.hpp>
+#include <SystestParser.hpp>
+#include <SystestResultCheck.hpp>
+#include <SystestState.hpp>
+
+namespace
+{
+NES::Systest::RunningQuery makeExplainRunningQuery(std::vector<std::string> expectedResultLines, std::string actualExplainOutput)
+{
+    NES::Systest::SystestQuery query{
```
yschroeder97, line 39: This is one example of a historically grown struct that has way too much data IMO.
 
A lot of the fields are an `std::optional`, and these are set at non-obvious scattered points during the lifetime of such a query. 
Therefore, parts of the systest code are littered with checks or invariants to ensure that some of these are not nullopt.

This should, in my opinion, be refactored, and we should introduce several types that represent a systest query at different points in time, ensuring that, for each type, the data it stores actually exists.

yschroeder97 (approved):
Thank you for the feature addition :) 
I do approve the changes themselves, but I don't like the direction the whole Systest framework is going from a code quality perspective. 

New features are introduced on a bi-weekly basis, but the code has historically grown and has deep issues; for example: 
- The types, like `SystestQuery` and `RunningQuery` are used everywhere, for everything, and have way too many members and optionals. At one point, a `RunningQuery` is returned after a failure; it still has type `RunningQuery`, which screams that something is fishy. 
- Some of the cpp files have more than 1K lines, and the code should be decomposed into smaller, more digestible chunks. 

..and some other issues that I would be happy to discuss and propose alternatives. 
In my opinion, it would be time to refactor the library before adding new features.

adrianmichalke:
@yschroeder97 aggree, lets plan this refactor together in a meeting

Issue #1756, opened 2026-07-13
[TASK] Allow instantiating executable query plans with multiple sinks and adjust backpressure handling


fkuzun:
### Description

Instantiate multi-sink executable query plans, `ExecutableQueryPlan::instantiate` must handle all entries in `compiledQueryPlan.sinks`.
In addition the `BackpressureController` must be adjusted to handle backpressure based on the performance of all sinks in the plan

### Acceptance Criteria

- Handle all sinks in  `ExecutableQueryPlan::instantiate`
- BackpressureController supports multiple sinks

PR #1797, opened 2026-07-18
feat(QueryCompiler): lower multi-sink logical plans into a shared physical plan

### LINKED ISSUE #1754: [TASK] Support lowering of multi-sink logical plans
### Description

Multi-sink plans share operator instances between sink roots (the same operator is reachable through more than one parent). To ensure that each operator is lowered only once, the lowering process must keep track of visited nodes.
- Track set of visited  operators during the lowering process in`LowerToPhysicalOperators.cpp`

### Acceptance Criteria

Lowering multi-sink plans produces correct output

fkuzun:
Support lowering of multi-sink plans.
A multi-sink logical plan visits subplans multiple times when iterating from different sink roots; lowering must translate the shared logical operators into ONE shared physical subgraph, or the shared part (including the source, i.e. reading the input) is duplicated per sink. Two fixes:

- LowerToPhysicalOperators: memoize lowerOperatorRecursively with an OperatorId -> lowered-wrapper map, consulted before lowering and filled on every return path (including the operator-bypass path), so a shared operator lowers exactly once and all lowered parents link to the same wrapper instance. apply() now adds ALL lowered sink roots to the PhysicalPlanBuilder instead of silently dropping everything but newRootOperators[0].
- PhysicalPlanBuilder::flip: run the node collection over all sink roots. 

## Verifying this change
This change is tested by
- Unit tests for lowering multi-root plans
- Unit tests for flipping multi root plans

- QueryCompiler

## Documentation

This PR closes #1754 #1804 #1756


nes-query-compiler/src/Phases/LowerToPhysicalOperators.cpp
```diff
@@ -97,10 +107,13 @@ lowerOperatorRecursively(const LogicalOperator& logicalOperator, const LoweringR
                 logicalOperator.getChildren().size() == 1,
                 "Empty lowering results of operators with multiple keys are not supported for {}",
```
greenBene: Not your change, but this should say "Empty lowering result of operators with multiple children are not supported", right?

nes-query-compiler/src/PhysicalPlanBuilder.cpp
```diff
@@ -151,23 +151,28 @@ PhysicalPlanBuilder::Roots PhysicalPlanBuilder::flip(const Roots& rootOperators)
 
     auto collectNodes = [&visitStateMap, &allNodes](const PhysicalOpPtr& node, auto&& self) -> void
     {
-        auto [it, inserted] = visitStateMap.try_emplace(node, VisitState::Unvisited);
+        /// Bind a reference, not an iterator: the recursive calls below insert into visitStateMap, and an
+        /// insertion that rehashes invalidates every iterator into it. References to elements stay valid.
+        VisitState& visitState = visitStateMap.try_emplace(node, VisitState::Unvisited).first->second;
         PRECONDITION(
-            it->second != VisitState::InProgress,
+            visitState != VisitState::InProgress,
```
greenBene, line 160: We are ignoring the result of `inserted` now completely. Your change maintains the same check as before, but shouldn't we check whether we successfully inserted the node? Or is this a conscious decision to ignore `inserted` ?

greenBene (changes requested):
This is really cool! Thank you for implementing it :) 

I am missing a test to validate that the multi-sink plan can actually be executed. Please add such a test.

nes-query-compiler/src/Phases/LowerToPhysicalOperators.cpp
keyseven123: Done

nes-query-compiler/src/PhysicalPlanBuilder.cpp
keyseven123: I think that the success of insertion does not matter. Either way, we have now a reference to the node. I added a comment though

adrianmichalke (approved):

nes-query-optimizer/include/PlanRewriteUtils.hpp
```diff
@@ -25,4 +31,17 @@ namespace NES
 /// recursively replaces field accesses of given LogicalFunction if they are a key in given field map with the field the key points to
 LogicalFunction replaceFieldAccesses(const LogicalFunction& function, const std::unordered_map<Field, Field>& fields);
 
+/// Returns the ids of all operators that are reachable through more than one parent edge in the (possibly DAG-shaped) plan.
+/// Duplicate edges from the same parent (e.g. a union reading the same input twice) count as sharing as well.
+[[nodiscard]] std::unordered_set<OperatorId> getSharedOperatorIds(const LogicalPlan& plan);
```
Artraxon: Don't use OperatorIds, just return a set of LogicalOperator

nes-query-optimizer/src/Placement/BottomUpPlacement.cpp
```diff
@@ -63,21 +63,33 @@ size_t operatorCapacityDemand(const LogicalOperator& op)
     return 1;
 }
 
-LogicalOperator addPlacementTrait(const LogicalOperator& op, const std::unordered_map<OperatorId, NetworkTopology::NodeId>& placement)
+/// @param placed operators already placed, keyed by their id before the assignment. An operator shared between sinks is placed
+/// once and every parent must end up with that same instance.
+LogicalOperator addPlacementTrait(
+    const LogicalOperator& op,
+    const std::unordered_map<OperatorId, NetworkTopology::NodeId>& placement,
+    std::unordered_map<OperatorId, LogicalOperator>& placed)
```
Artraxon: I know it was before like this as well, but just change it to LogicalOperators here as well

nes-query-optimizer/include/PlanRewriteUtils.hpp
```diff
@@ -25,4 +31,17 @@ namespace NES
...
+
+/// Rewrites all operators of a (possibly DAG-shaped) plan bottom-up.
+/// The rewrite function is invoked exactly once per unique operator (identified by the input plan's OperatorId) with the original
+/// operator and its already-rewritten children. The result is memoized and reused for every additional parent, so operators shared
+/// between multiple parents stay shared in the rewritten plan. Children are rewritten left-to-right and roots in order, making the
+/// invocation order deterministic for stateful rewrite functions.
+[[nodiscard]] LogicalPlan rewritePlanBottomUp(
+    const LogicalPlan& plan,
+    const std::function<LogicalOperator(const LogicalOperator& original, std::vector<LogicalOperator> rewrittenChildren)>& rewrite);
+
```
Artraxon: Why is this function needed instead of using the new Planvisitor from Bene?

nes-logical-operators/src/Plans/LogicalPlan.cpp
```diff
@@ -137,24 +137,57 @@ std::vector<LogicalOperator> getParents(const LogicalPlan& plan, const LogicalOp
     return parents;
 }
 
-std::string explain(const LogicalPlan& plan, ExplainVerbosity verbosity)
+std::vector<LogicalOperator> planOperators(const LogicalPlan& plan)
 {
-    std::stringstream stringstream;
-    if (verbosity == ExplainVerbosity::Short)
+    std::vector<LogicalOperator> operators;
+    std::unordered_set<OperatorId> visited;
```
Artraxon: just store LogicalOperators here as well

nes-logical-operators/src/Plans/LogicalPlan.cpp
```diff
@@ -137,24 +137,57 @@ std::vector<LogicalOperator> getParents(const LogicalPlan& plan, const LogicalOp
...
+{
+/// Marks an operator an earlier sink already rendered; repeating its sub-plan would read as if each sink had a copy.
+constexpr std::string_view SharedOperatorMarker = " [shared]";
+
+void dumpOperator( /// NOLINT(misc-no-recursion)
+    const LogicalOperator& op,
+    const std::size_t level,
+    std::ostream& out,
+    const ExplainVerbosity verbosity,
+    std::unordered_set<OperatorId>& alreadyDumped)
```
Artraxon: and use LogicalOperators here as well

nes-query-optimizer/src/Placement/QueryDecomposition.cpp
```diff
@@ -57,15 +64,25 @@ namespace
 struct DecompositionContext
 {
     std::unordered_map<NetworkTopology::NodeId, std::vector<LogicalPlan>> plansByNode;
+    /// Roots collected per node while decomposing, merged into plans once decomposition is done.
+    std::unordered_map<NetworkTopology::NodeId, std::vector<LogicalOperator>> rootsByNode;
+    /// Network source per (operator, consuming node): an operator shared by sinks is sent to a node once, not once per consumer. It
+    /// may still feed several nodes, one channel each — the subtree below it is decomposed once and deployed once all the same.
+    std::map<std::pair<OperatorId, NetworkTopology::NodeId>, LogicalOperator> networkChannels;
```
Artraxon: aand, in this file as well

nes-query-optimizer/src/Placement/QueryDecomposition.cpp
```diff
@@ -138,19 +171,24 @@ Bridge connect(const DecompositionContext& context, const NetworkChannel& channe
...
     upstreamTs.insert(PlacementTrait{channel.upstreamNode});
     downstreamTs.insert(PlacementTrait{channel.downstreamNode});
+    downstreamTs.insert(OutputOriginIdsTrait{channelOriginIds});
 
     return Bridge{
         SourceDescriptorLogicalOperator::create(networkSourceDescriptor)->withTraitSet(downstreamTs),
-        SinkLogicalOperator::create(channel.upstreamOp, networkSinkDescriptor.value())->withTraitSet(upstreamTs).withInferredSchema()};
+        /// withChildren re-infers only this sink's own schema; the recursive withInferredSchema() would deep-copy the subtree below
+        /// it, so two network sinks fed by one shared subplan would each end up with a private copy of it, sources included.
+        SinkLogicalOperator::create(channel.upstreamOp, networkSinkDescriptor.value())
```
Artraxon: IMO redundant comment, withInferredSchema should never be used inside a traversal and we actively try to get rid of these calls (and eventually the whole function)

nes-runtime/src/ExecutableQueryPlan.cpp
```diff
@@ -69,25 +70,29 @@ ExecutableQueryPlan::instantiate(CompiledQueryPlan& compiledQueryPlan, const Sou
...
     }
 
-    auto& [pipelineId, descriptor, predecessors] = compiledQueryPlan.sinks.front();
+    /// Every sink throttles the query's sources on its own; the channel is only destroyed with the last of them.
+    auto [backpressureControllers, backpressureListener] = createBackpressureChannel(compiledQueryPlan.sinks.size());
 
-    auto sink = ExecutablePipeline::create(pipelineId, lower(std::move(backpressureController), descriptor), {});
-    compiledQueryPlan.pipelines.push_back(sink);
-    for (const auto& predecessor : predecessors)
+    for (auto&& [sinkToInstantiate, backpressureController] : std::views::zip(compiledQueryPlan.sinks, backpressureControllers))
```
Artraxon, line 81: @ls-1801 is there any reason why the CompiledQueryPlan would need to store operator IDs instead of just storing OperatorIds?

Artraxon (changes requested):
I cannot comment too much on the placement, decomposition and backpressure since this is @ls-1801 domain, but I would ideally like to see no usage of OperatorIds at all in any optimizer component. Its usage should be strictly limited to serialization/deserialization. LogicalOperators should now just do pointer eq/hash and therefor be as fast. If you don't want to take ownership, use WeakLogicalOperator.

And I think some remnants of the previous multi-sink-optimizer-traversal slipped through?

Artraxon:
@ls-1801 do you see any issues in the composition?

nes-query-compiler/src/Phases/LowerToPhysicalOperators.cpp
```diff
@@ -95,12 +104,15 @@ lowerOperatorRecursively(const LogicalOperator& logicalOperator, const LoweringR
...
+                "Empty lowering result of operators with multiple children are not supported {}",
                 logicalOperator);
-            return lowerOperatorRecursively(logicalOperator.getChildren()[0], registryArgument);
+            auto loweredChild = lowerOperatorRecursively(logicalOperator.getChildren()[0], registryArgument, memo);
+            memo.emplace(logicalOperator, loweredChild);
+            return loweredChild;
         }
         return {};
     }
+    memo.emplace(logicalOperator, root);
```
ls-1801, line 115: Nit: Could `lowerOperatorRecursively` add its result to `memo` in one place before returning? Maintaining `memo.emplace` on each return path makes it easy for a path to omit the update; for example, the empty leaf path currently returns `{}` without memoizing.

nes-query-optimizer/src/Placement/QueryDecomposition.cpp
```diff
@@ -99,8 +129,31 @@ Bridge connect(const DecompositionContext& context, const NetworkChannel& channe
         sourceConfig.emplace(Identifier::parse("receiver_queue_size"), std::to_string(context.config.receiverQueueSize.getValue()));
     }
 
+    /// The channel carries origins of its own on the receiving side: the sink stamps its buffers with them and the source below
```
ls-1801, line 132: Could we move responsibility for reassigning origin IDs out of decomposition? Decomposition seems like the wrong abstraction level to decide that a stream has acquired a new identity. When the optimizer splits the execution graph, it could assign distinct origin IDs to the resulting branches and represent that decision in the graph, for example with a trait. During physical lowering, a scan/emit operator could then stamp the new origin when it observes that trait. This would let us remove the `origin_id_map` protocol between the network sink and source and keep the network operators focused on transport.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -341,11 +350,17 @@ void AntlrSQLQueryPlanCreator::enterSinkClause(AntlrSQLParser::SinkClauseContext
         throw InvalidQuerySyntax("INTO must be followed by at least one sink-identifier.");
     }
     /// Store all specified sinks.
+    std::unordered_set<Identifier> sinkNames;
```
ls-1801, line 353: Should sink clauses be permitted inside individual UNION operands? The parser collects those sinks globally, and `getQueryPlan()` attaches every collected sink to the completed union rather than to the operand where it was written. Thus `SELECT a FROM inputStreamA INTO sinkA UNION SELECT a FROM inputStreamB INTO sinkB` means that both sinks consume the full union, which is surprising; using the same sink name in both operands also bypasses this clause-local duplicate check and produces two writes to that sink. Could we reject operand-local sink clauses and require `INTO` on the outer query instead? A regression test for the duplicate case is:

```cpp
TEST_F(StatementBinderTest, RepeatedSinkAcrossUnionTerms)
{
    const std::string query
        = "SELECT a FROM inputStreamA INTO outputStream UNION SELECT a FROM inputStreamB INTO outputStream";
    const auto statement = binder->parseAndBindSingle(query);
    ASSERT_FALSE(statement.has_value());
    ASSERT_EQ(statement.error().code(), ErrorCode::InvalidQuerySyntax);
}
```

nes-query-compiler/tests/UnitTests/MultiRootLoweringTest.cpp
```diff
@@ -0,0 +1,189 @@
...
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+/// Multi-sink lowering into a shared physical DAG. Optimize a single-sink tree, graft a second sink onto
+/// the optimized plan (the deployment model: extra sinks attached after optimization), compile, and assert
+/// the sink/source counts. Without the fix apply() dropped all but the first sink, and the un-memoized
```
ls-1801, line 17: This file-level comment narrates the implementation that this change replaces (`"Without the fix"` and `"the un-memoized lowerOperatorRecursively"`). Could we keep it focused on the current contract—both sink roots must survive lowering and shared logical operators must map to one physical subplan—so the documentation does not depend on removed code?

nes-executable/src/BackpressureChannel.cpp
```diff
@@ -15,90 +15,130 @@
...
+    return opened;
 }
 
 void BackpressureListener::wait(const std::stop_token& stopToken) const
 {
     auto state = channel->stateMtx.lock();
-    /// If the channel is open, backpressureListener can proceed
-    if (*state == Channel::State::OPEN)
+    /// If no controller applies backpressure, the backpressureListener can proceed
+    if (state->isOpen())
```
ls-1801, line 116: Please check `isDestroyed()` before this open fast path. When the last controller is destroyed, its destructor leaves both counters at zero, so `isOpen()` is also true and a listener first used afterward returns successfully instead of enforcing the documented controller-outlives-listener contract. A regression test is:

```cpp
TEST_F(BackpressureChannelTest, ListenerDetectsControllerDestroyedBeforeWait)
{
    SKIP_IF_TSAN();
    GTEST_FLAG_SET(death_test_style, "threadsafe");

    auto listener = []
    {
        auto [controller, listener] = createBackpressureChannel();
        return listener;
    }();

    EXPECT_DEATH_DEBUG(listener.wait({}), "");
}
```

nes-executable/include/BackpressureChannel.hpp
```diff
@@ -17,31 +17,36 @@
...
+std::pair<std::vector<BackpressureController>, BackpressureListener> createBackpressureChannel(size_t numberOfControllers);
+
+/// A Backpressure Controller controls a backpressure channel. It allows the user to apply and release backpressure, which blocks
+/// or unblocks all connected Ingestions. With more than one controller, Ingestions are blocked while any one of them applies it.
 class BackpressureController
 {
     explicit BackpressureController(std::shared_ptr<Channel> channel);
 
     std::shared_ptr<Channel> channel;
+    /// Applying twice from the same controller is a no-op
```
ls-1801, line 42: This move assignment overwrites an existing controller without decrementing its `controllers` count or releasing its `controllersApplyingPressure` contribution. With two controllers, moving the unpressured controller over a pressured one leaves an unowned pressure count, so the listener remains blocked permanently. Please implement move assignment with the same cleanup as destruction (or delete it). A regression test is:

```cpp
TEST_F(BackpressureChannelTest, MoveAssignmentReleasesPreviousControllerState)
{
    auto [controllers, listener] = createBackpressureChannel(2);
    ASSERT_TRUE(controllers[0].applyPressure());
    controllers[0] = std::move(controllers[1]);

    std::atomic waitReturned{false};
    std::jthread waiter([&](const std::stop_token& stopToken) {
        listener.wait(stopToken);
        waitReturned = true;
    });
    std::this_thread::sleep_for(std::chrono::milliseconds(100));
    EXPECT_TRUE(waitReturned);
}
```

ls-1801 (changes requested):
I think my biggest issue personally is moving the origin ID reassignment into the network source/sink.
I would much rather prefer if we handle the assignment of new origin IDs in the optimizer. I.e., we probably want a dedicated split operator that assigns new origin IDs to its successors. During lowering, this operator needs to become a scan/emit operator that can update the buffer's origin ID.

Also, I am very sorry for the long wait time.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
fkuzun: I drafted how "require INTO on the outer query instead" could look, on the sink-clause-at-statement-level branch: it moves sinkClause off querySpecification and onto a new statement-level rule, so the bad positions become unparseable. But that looks like it might need more discussion and maybe a separate PR.

For now, on this branch I added a check that a sink clause can only appear on the query the statement itself runs (see https://github.com/nebulastream/nebulastream/pull/1797/changes#diff-c988fd8a65592ff4c43aaa43a921e125e1a3372e5d72b0d966b10a136ace8768R446). It's a whitelist over the route from the clause up to the statement, so it rejects an INTO on any earlier UNION operand, in a parenthesised or FROM subquery, and in EXISTS/IN subqueries.

What it still has to accept is an INTO on the last operand of a UNION, because that is where the grammar currently puts a trailing sink clause, e.g. from 
`nes-systests/guide/guide5.test`:

```
SELECT a, b FROM s                                                         
UNION
SELECT a, b FROM t                                                         
INTO sink;
```

The INTO here parses as part of the right operand.

PR #1802, opened 2026-07-20
fix(CI): Restore the latest-tag retag of dev images on merge

keyseven123:
The job compared the string output `build-dependency` against the boolean `true`, which GitHub never matches, so it has not run since 40f6522941.

Fixing that exposed the retag loop: the stdlib-less shortcut pulled from nes-<image>:<hash>-<sanitizer>, a tag get_dev_images.yml never creates. It now resolves to libcxx and runs once per sanitizer. Dropped a duplicated retag.

## Verifying this change
This change is tested by our CI and the pr-merge workflow.

### DIFF
--- a/.github/workflows/pr-merge.yml
+++ b/.github/workflows/pr-merge.yml
@@ -31,7 +31,8 @@ jobs:
     # updated dependencies.
     runs-on: [ self-hosted, linux, Build ]
     needs: [ get-dev-images ]
-    if: ${{ github.event.pull_request.merged == true && needs.get-dev-images.outputs.build-dependency == true }}
+    # `build-dependency` is a job output, i.e. a string.
+    if: ${{ github.event.pull_request.merged == true && needs.get-dev-images.outputs.build-dependency == 'true' }}
     steps:
       - uses: actions/checkout@v5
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-merge.yml
```diff
@@ -31,7 +31,9 @@ jobs:
     # updated dependencies.
     runs-on: [ self-hosted, linux, Build ]
     needs: [ get-dev-images ]
-    if: ${{ github.event.pull_request.merged == true && needs.get-dev-images.outputs.build-dependency == true }}
+    # `build-dependency` is a job output, i.e. a string. Comparing it against the boolean `true`
+    # casts the string to a number (NaN) and never matches, which silently disabled this job.
```
ls-1801: `, which silently disabled this job.` this part of the comment should be removed. The rest can stay if you want to

ls-1801 (approved):

Artraxon (approved):

.github/workflows/pr-merge.yml
keyseven123: Changed.

PR #1803, opened 2026-07-22
feat(optimizer) created plugin support for optimization rules

greenBene:
This PR introduces a plugin registry for the rule-based optimizer.

* Created `PlanRuleRegistry` <- Plan rule, because in the future, we also want to introduce OperatorRules. 
* All existing rules are updated to register themself as a plugin
* The `RuleBasedOptimizer` uses the newly introduced registry to instantiate all registered rules.
* The `SemanticAnalysier` is folded into the `RuleBasedOptimizer`. 
* The semantic rules are ordered into the rule dependency graph using a new `SemanticBarrierRule`. 
* Move `RedundantUnionRemovalRule` to `nes-plugins` to have there a simple example that is actually used. 

## Verifying this change
* All existing tests continue to run

- Optimizer 

## Documentation
* A walkthrough, similar to the ones for source and sink plugins, was created

This PR closes #1522


docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
+# How to add an Optimizer `Rule`
+
+The query optimization in NebulaStream is rule-based.
```
yschroeder97:
```suggestion
Query optimization in NebulaStream is rule-based.   
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+├── Sinks/
+├── Sources/
+└── ...
+```
+
+## 2. Plugin Registration 
+
+Registering a rule plugin consists of:
+1) The registration of the rule plugin
+2) The registration of the defined unit tests (optional, but encouraged)
```
yschroeder97: You could add that it is mandatory if desired that the rule becomes part of the main system.

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+3) The activation of the rule plugin
+
+Following is the commented `CMakeLists.txt` file for the RedundantUnionRemovalRule plugin.
```
yschroeder97:
```suggestion
The following is the commented `CMakeLists.txt` file for the RedundantUnionRemovalRule plugin. 
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+
+# 2) Register Tests
+if (NES_ENABLES_TESTS)
+    add_nes_unit_test(RedundantUnionRemovalRuleTest RedundantUnionRemovalRuleTest.cpp)
+    target_link_libraries(RedundantUnionRemovalRuleTest nes-query-optimizer nes-logical-operators nes-test-util nes-query-optimizer-test-utils-lib)
+    target_include_directories(RedundantUnionRemovalRuleTest PRIVATE .)
+endif ()
+```
+
+To activate the plugin, you must at the line `activate_optional_plugin("Rules/RedundantUnionRemovalRule" ON)` to `nes-plugins/CMakeLists.txt`.
```
yschroeder97:
```suggestion
To activate the plugin, you must add the line `activate_optional_plugin("Rules/RedundantUnionRemovalRule" ON)` to `nes-plugins/CMakeLists.txt`.  
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+Additionally and optionally, the following methods can be defined to overwrite the default behavior. 
+
+```cpp
+// returns a set of rules that must be applied before the given rule
+std::set<std::type_index> needs() const;
+
+// returns a set of rules that must be applied after the given rule
+std::set<std::type_index> neededBy() const;
+
+// soft version of needs(): returned rules must be applied before the given rule if the rules are registered
```
yschroeder97: You could also say "relaxed".

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+
+> Note: While the `Rule.hpp` defines rules using templates (`Rule<U>`) for future extensibility, 
+> in our example we assume that rules always works on `LogicalPlan` (Rule<LogicalPlan>).
+
+
+## 4. Rule Dependencies
+
+To ensure the rules are applied in a valid order, rules can define hard and soft directed dependencies between each other.
+Hard dependencies (`needs/neededBy`) define that a rule MUST run before another rule and will fail if that other rule is not available. 
+E.g. if rule A needs rule B, rule A can only run after rule B ran.
```
yschroeder97:
```suggestion
E.g., if rule A needs rule B, rule A can only run after rule B has run. 
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+> Note: While the `Rule.hpp` defines rules using templates (`Rule<U>`) for future extensibility, 
+> in our example we assume that rules always works on `LogicalPlan` (Rule<LogicalPlan>).
+
+
+## 4. Rule Dependencies
+
+To ensure the rules are applied in a valid order, rules can define hard and soft directed dependencies between each other.
+Hard dependencies (`needs/neededBy`) define that a rule MUST run before another rule and will fail if that other rule is not available. 
+E.g. if rule A needs rule B, rule A can only run after rule B ran. 
+If rule B is not run (e.g. because it is not registered), an error is thrown when the rule sequence is generated.
```
yschroeder97:
```suggestion
If rule B is not run (e.g., because it is not registered), an error is thrown when the rule sequence is generated.
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+Soft dependencies (`wants/wantedBy`) define that a rule must run before another rule but the dependency is ignored if that other rule is not available.
```
yschroeder97:
```suggestion
Soft dependencies (`wants/wantedBy`) define that a rule must run before another rule, but the dependency is ignored if that other rule is not available.
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+E.g. if rule A wants rule B, rule A must run after rule B if rule B is also registered,
```
yschroeder97:
```suggestion
E.g., if rule A wants rule B, rule A must run after rule B if rule B is also registered, 
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+
+## 4. Rule Dependencies
+
+To ensure the rules are applied in a valid order, rules can define hard and soft directed dependencies between each other.
+Hard dependencies (`needs/neededBy`) define that a rule MUST run before another rule and will fail if that other rule is not available. 
+E.g. if rule A needs rule B, rule A can only run after rule B ran. 
+If rule B is not run (e.g. because it is not registered), an error is thrown when the rule sequence is generated.
+Soft dependencies (`wants/wantedBy`) define that a rule must run before another rule but the dependency is ignored if that other rule is not available.
+E.g. if rule A wants rule B, rule A must run after rule B if rule B is also registered, 
+but if rule B is not registered, A can still run and the `wants` dependency to B is ignored.
```
yschroeder97:
```suggestion
but if rule B is not registered, A can still run, and the `wants` dependency on B is ignored.    
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+
+To ensure the rules are applied in a valid order, rules can define hard and soft directed dependencies between each other.
+Hard dependencies (`needs/neededBy`) define that a rule MUST run before another rule and will fail if that other rule is not available. 
+E.g. if rule A needs rule B, rule A can only run after rule B ran. 
+If rule B is not run (e.g. because it is not registered), an error is thrown when the rule sequence is generated.
+Soft dependencies (`wants/wantedBy`) define that a rule must run before another rule but the dependency is ignored if that other rule is not available.
+E.g. if rule A wants rule B, rule A must run after rule B if rule B is also registered, 
+but if rule B is not registered, A can still run and the `wants` dependency to B is ignored.    
+These dependencies are then used to create a directed acyclic graph.
+If the graph is cycle free, a topological sort is used to generate a valid and ordered rule sequence.
```
yschroeder97:
```suggestion
If the graph is cycle-free, a topological sort is used to generate a valid and ordered rule sequence.
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+#include <typeinfo>
+#include <Plans/LogicalPlan.hpp>
+#include <Rules/Rule.hpp>
+
+namespace NES
+{
+
+/**
+ * @brief This pass removes redundant unions with only a single child.
+ */
```
yschroeder97: I think we agreed using only `///` for comments.

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+class RedundantUnionRemovalRule
+{
+public:
+    static constexpr std::string_view NAME = "RedundantUnionRemovalRule";
+
+    [[nodiscard]] LogicalPlan apply(LogicalPlan queryPlan) const;
+    [[nodiscard]] std::set<std::type_index> needs() const;
```
yschroeder97, line 160: A general note: how do I know which other rules I potentially might have dependencies on without knowing them all by heart?

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+static_assert(RuleConcept<RedundantUnionRemovalRule, LogicalPlan>);
+}
+```
+
+We first declare the required constant `NAME` and the required method `LogicalPlan apply(LogicalPlan)`. 
+Then, because we want to control where the rule is placed within the dependency graph, we declare the methods
+`std::set<std::type_index> needs() const` and 
+`std::set<std::type_index> neededBy() const`. 
+To assert we correctly declared the rule, we assert that it follows the RuleConcept using the call
+`static_assert(RuleConcept<RedundantUnionRemovalRule, LogicalPlan>);`.
```
yschroeder97:
```suggestion
To assert that we correctly declared the rule, we assert that it follows the RuleConcept using the call
`static_assert(RuleConcept<RedundantUnionRemovalRule, LogicalPlan>);`.
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+
+Next, we define the actual behavior of the rule in `RedundantUnionRemovalRule.cpp`. 
+To keep the example readable, we'll only focus on individual snippets. 
+
+First, we define the `apply` method. 
+We first ensure that the given plan only has one root operator.
+Then we start a (yet to define) recursive function call at that root operator,
```
yschroeder97:
```suggestion
Then we start a (yet-to-be-defined) recursive function call at that root operator, 
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+    queryPlan = queryPlan.withRootOperators({recur(queryPlan.getRootOperators().front().withInferredSchema())});
+    return queryPlan;
+}
+```
+
+Next, we define what is happening in the recursion.
+The algorithm works bottom up. Thus, we first continue the recursion for all child operators of the current operator. 
+It will stop at Source operators because source operators are the only operators that do not have children. 
+Then, if the current operator is a UNION operator and has only one child, we return that child operator and thus remove the unnecessary union operator. 
+If the current operator is not a UNION operator or if it is but has more than one child, we leave it as is,
```
yschroeder97:
```suggestion
Then, if the current operator is a UNION operator with only one child, we return that child operator, thereby removing the unnecessary UNION operator. 
If the current operator is not a UNION operator or if it is but has more than one child, we leave it as is, 
```

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,249 @@
...
+}
+
+std::set<std::type_index> RedundantUnionRemovalRule::neededBy() const
+{
+    return {typeid(FixedPlanStructureBarrier)};
+}
+```
+
+To ensure that the NebulaStream optimizer is able to instantiate the newly created rule, we define the registration function.
+While the `PlanRuleRegistryArguments` gives us access to multiple catalogs and also the optimizer configuration,
```
yschroeder97, line 234: Not really that important, would remove.

docs/guide/how_to_add_an_optimizer_rule.md
yschroeder97, line 1: Cool guide!

nes-query-optimizer/include/Rules/Barriers/SemanticAnalysisBarrier.hpp
yschroeder97, line 1: This seems unrelated to the PR.

nes-query-optimizer/include/Rules/Semantic/CalcTargetOrderRule.hpp
yschroeder97, line 1: Also unrelated, or why do you need this change to enable plugin support?

yschroeder97 (changes requested):
Hey, very cool to have this part of the plugin system as well!

The changes are mostly things in the docs; the actual registry infra looks like in the other registries and runs.

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: added

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: I removed the 'optional, but encouraged' to imply that it is necessary :)

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: applied

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: good catch!

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: applied

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: applied

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: applied

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: added

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: added

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: fixed

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: fixed

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: changed

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: The general idea is that, by default, you just depend on the barrier rules defining the phase. If necessary, we can define a shorthand for this (e.g., you add a phase "tag" and the barrier rule dependencies are automatically added). For now, we wanted to keep it simple and avoid unnecessary complexity. 

But, in some cases, such as the ProjectionPushdown and the RedundantProjectionRemoval, it makes sense to order them specifically. To know such dependencies, you have to have an understanding of optimization rules. In the future, once the system is more stable, it might make sense to list and explain the default optimization rules in the documentation, but I don't think we are there yet.

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: added

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: added

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: fixed

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: I would leave it in case someone wants to write a semantic analysis rule for some reason. Otherwise, we don't document in this guide why the arguments are there at all.

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: Thank you! :) And thank you for fixing my grammar :D

nes-query-optimizer/include/Rules/Barriers/SemanticAnalysisBarrier.hpp
greenBene: The first commit of this PR moves all rules into one rule-based optimizer. Before, it was split into "Semantic Analysis" and "Rule-based". This merging was done to ensure all rules can be initialized as part of a single optimizer and that they don't have to specify which optimizer they should be initialized in. Since all semantic rules still need to be run before the "logic" optimizations, this barrier was introduced. 

We could move this merging of optimizers into a separate PR, but I prefer to keep it in here.

nes-query-optimizer/include/Rules/Semantic/CalcTargetOrderRule.hpp
greenBene: Just a small code cleanup to ensure `apply` is always the first declared function and `needs` is before `neededBy`. IMO, it gives better readability. The change changes no functionality.

nes-plugins/Rules/RedundantUnionRemovalRule/RedundantUnionRemovalRule.cpp
adrianmichalke, line 1: I guess we plan to move it at some point in time back out of the plugins once we have a better example?

nes-query-optimizer/src/Phases/RuleBasedOptimizer.cpp
```diff
@@ -15,43 +15,52 @@
...
-#include <Rules/Static/RedundantUnionRemovalRule.hpp>
-#include <Rules/Static/WatermarkAssignerPushdownRule.hpp>
+#include <Sinks/SinkCatalog.hpp>
+#include <Sources/SourceCatalog.hpp>
 #include <Util/Logger/Logger.hpp>
 #include <Util/PlanRenderer.hpp>
+#include <ModelCatalog.hpp>
+#include <PlanRuleRegistry.hpp>
 #include <QueryOptimizerConfiguration.hpp>
+#include "ErrorHandling.hpp"
```
adrianmichalke: use <>

nes-plugins/Rules/RedundantUnionRemovalRule/RedundantUnionRemovalRuleTest.cpp
```diff
@@ -0,0 +1,135 @@
...
+    auto op1 = op0.getChildren().at(0);
+    ASSERT_TRUE(op1.tryGetAs<SelectionLogicalOperator>());
+    auto op2 = op1.getChildren().at(0);
+    ASSERT_TRUE(op2.tryGetAs<SourceDescriptorLogicalOperator>());
+}
+
+TEST_F(RedundantUnionRemovalRuleTest, TwoUnionsToRemove)
+{
+    /// BEFORE: Sink < Selection < Union < Source
+    /// AFTER: Sink < Selection < Source
```
adrianmichalke, line 76: I wonder if such a plan can actually get created via our current SQL syntax? Do you have a SQL example for redundant unions?

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,248 @@
+# How to add an Optimizer `Rule`
+
+Query optimization in NebulaStream is rule-based.   
+This includes the semantic analysis, the logical optimization, and the trait selection.
```
adrianmichalke:
```suggestion
This includes the semantic analysis, the logical optimization, and the physical optimization (trait selection). 
```
people might be more familiar with that term

adrianmichalke: I guess we also allow to add barriers now as they are now modelled as rules

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,248 @@
...
+For our example, we use the following structure:
+
+```
+nes-plugins/
+├── Rules/
+│   ├── RedundantUnionRemovalRule/
+│   │   ├── CMakeLists.txt
+│   │   ├── RedundantUnionRemovalRule.cpp
+│   │   ├── RedundantUnionRemovalRule.hpp
+│   │   └── RedundantUnionRemovalRuleTest.cpp
```
adrianmichalke, line 22: comment: I wonder if we should also add support for .test files in the future here, that are run via systest once the plugin is activated

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,248 @@
...
+    add_nes_unit_test(RedundantUnionRemovalRuleTest RedundantUnionRemovalRuleTest.cpp)
+    target_link_libraries(RedundantUnionRemovalRuleTest nes-query-optimizer nes-logical-operators nes-test-util nes-query-optimizer-test-utils-lib)
+    target_include_directories(RedundantUnionRemovalRuleTest PRIVATE .)
+endif ()
+```
+
+To activate the plugin, you must add the line `activate_optional_plugin("Rules/RedundantUnionRemovalRule" ON)` to `nes-plugins/CMakeLists.txt`.  
+
+
+For a detailed explanation of the plugin system, CMake macros, and how registries work, see `guide/extensibility.md`.
```
adrianmichalke: nit: this could be a markdown link

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -0,0 +1,248 @@
...
+
+## 3. Interface
+
+Generally speaking, a rule can be any object that defines the following constant and method:
+
+```cpp
+// Human-readable name of rule
```
adrianmichalke:
```suggestion
/// Human-readable name of rule 
```

adrianmichalke: other occurences as well

docs/guide/how_to_add_an_optimizer_rule.md
adrianmichalke: @greenBene cool, i like the idea of having the means via sql or command line to print the rules/rule dependency tree

adrianmichalke (approved):
approved once comments are addressed

nes-plugins/Rules/RedundantUnionRemovalRule/RedundantUnionRemovalRule.cpp
greenBene: Ideally, yes :) I tried to find a rule that does sth. useful, is easy to understand, and kinda optional. If we have a better fit in the future that is actually even more optional, switching them would make sense!

nes-query-optimizer/src/Phases/RuleBasedOptimizer.cpp
greenBene: Thanks!

nes-plugins/Rules/RedundantUnionRemovalRule/RedundantUnionRemovalRuleTest.cpp
greenBene: We don't create such plans via SQL, but they can happen as an artefact of the `LogicalSourceExpansionRule.cpp:72ff`. There, a logical source is replaced by a union operator with all physical sources as children. If it has only one physical source, we have the above-mentioned case.

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: I applied the first suggestion. I wouldn't mention barrier rules here already, since they are a very specific concept used to define the phases. In the introduction of this guide, I want to keep it lightweight. The Barrier rules are mentioned further down in section 4.

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: Fully agree! :) Isn´t this something @keyseven123 or @alepping started to investigate a few months back? :thinking:

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: Good idea! Changed it :)

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: Changed it

yschroeder97 (approved):

PR #1805, opened 2026-07-23
Implement Median as Nautilus Function

keyseven123:

The compilation of physical aggregation functions is expensive because each aggregation's `lift`/`combine`/`lower` body is traced and **inlined** into the build/probe pipeline functions. For median this is the worst case: its `lower` is a doubly-nested loop, and a query with many `MEDIAN` aggregations inlines that loop once per aggregation, blowing up the traced pipeline and slowing compilation.

This PR extracts the median bodies into dedicated **nautilus functions** (`NautilusFunction`, i.e. cross-calls compiled once and called from the pipeline) so they are no longer inlined. Since nautilus deduplicates functions by name, naming them purely by data type means all medians of the same type in a pipeline share a single compiled instantiation.

Change log:
- Added a global, statically-allocated templated table `MedianFunctionTable<InputT, ResultT, Nullable>` holding the four extracted nautilus functions (`lift`, `combine`, `lower`, `reset`) with clear, type-derived names (e.g. `Median_lower_INT64_FLOAT64_X`). `combine`/`reset` depend only on nullability and are named/shared accordingly.
- The bodies are free, type-parametric function templates (no per-instance capture) so they can be global statics shared across all medians of the same type.
- The virtual `lift`/`combine`/`lower`/`reset` are now thin wrappers that keep the column-specific work in the main pipeline trace (compute the input value via `inputFunction` into a scratch slot; read the lowered result from a scratch slot before writing the result field) and forward to the selected functions via a `MedianFunctions` pointer struct.
- The paged vector now stores **only the single aggregated value** per row (a one-column buffer derived from the input type) instead of the whole input record, so `lower` reads that value directly rather than re-applying `inputFunction`. `getSizeOfStateInBytes()` is unchanged, so the hash-map layout and the lowering rule are untouched.
- Scope is **median only**; the same pattern can be applied to Sum/Min/Max/Count/Avg later. The change is self-contained in `MedianAggregationPhysicalFunction.hpp/.cpp`.

## Verifying this change

This change should be covered by the existing aggregation systests (no behavior change intended):
- `nes-systests/operator/aggregation/WindowAggregationNull.test`, `WindowAggregationDifferentDataTypes*.test`, `WindowAggregationStackedAggregations.test`, `ExpressionInsideAggregation.test`, `ColumnNames.test`, `AggregationOfUnion.test`
- `nes-systests/differential/AggregationEquiValence.test`, `nes-systests/guide/guide8.test`

The intended win can be confirmed by compiling a query with many `MEDIAN`s and checking that the probe module emits a single `Median_lower_*` per distinct type instead of one inlined copy per aggregation.

- ExecutionEngine (physical aggregation functions / nautilus code generation)
- QueryCompiler (only indirectly; the windowed-aggregation lowering rule is unchanged)

## Documentation

- The change is reflected in in-code documentation (class and helper comments in `MedianAggregationPhysicalFunction`).

### DIFF
--- a/nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
+++ b/nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
@@ -1,0 +1,184 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <Aggregation/Function/MedianNautilusFunctions.hpp>
+
+#include <cstdint>
+#include <memory>
+#include <Aggregation/Function/AggregationPhysicalFunction.hpp>
+#include <DataTypes/DataType.hpp>
+#include <DataTypes/VarVal.hpp>
+#include <Interface/NautilusBuffer.hpp>
+#include <Interface/PagedVector/PagedVector.hpp>
+#include <Interface/PagedVector/PagedVectorRef.hpp>
+#include <Interface/Record.hpp>
+#include <Runtime/AbstractBufferProvider.hpp>
+#include <Runtime/TupleBuffer.hpp>
+#include <nautilus/function.hpp>
+#include <ErrorHandling.hpp>
+#include <val_arith.hpp>
+#include <val_bool.hpp>
+#include <val_ptr.hpp>
+
+namespace NES
+{
+
+namespace
+{
+uint64_t medianNumberOfRecords(const TupleBuffer* buffer)
+{
+    const auto pagedVector = PagedVector::load(*buffer);
+    const auto numberOfEntries = pagedVector.getTotalNumberOfRecords();
+    INVARIANT(numberOfEntries > 0, "The number of entries in the paged vector must be greater than 0");
+    return numberOfEntries;
+}
+
+void medianCombinePagedVectors(AbstractBufferProvider* provider, TupleBuffer* vector1Buffer, const TupleBuffer* vector2Buffer)
+{
+    auto vector1 = PagedVector::load(*vector1Buffer);
+    const auto vector2 = PagedVector::load(*vector2Buffer);
+    vector1.copyPagesFrom(*provider, vector2);
+}
+
+void medianInitPagedVectorState(AggregationState* pagedVectorMemArea, AbstractBufferProvider* provider, uint64_t sizeOfTuple)
+{
+    /// NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast): aggregation state stores a TupleBuffer at this slot.
+    auto* pagedVectorBufferMemArea = reinterpret_cast<TupleBuffer*>(pagedVectorMemArea);
+    if (auto pagedVectorBuffer = provider->getUnpooledBuffer(PagedVector::getMainBufferSize()))
+    {
+        /// initialize paged vector buffer
+        PagedVector::init(pagedVectorBuffer.value(), provider->getBufferSize(), sizeOfTuple);
+        /// @warning: this will be refactored again during the ChainedHashMap refactor
+        new (pagedVectorBufferMemArea) TupleBuffer(pagedVectorBuffer.value());
+        return;
+    }
+    throw BufferAllocationFailure("No unpooled TupleBuffer available for median aggregation paged vector!");
+}
+
+}
+
+nautilus::val<double> medianOf(
+    const nautilus::val<TupleBuffer*>& pagedVectorBuffer,
+    const std::shared_ptr<PagedVectorTupleLayout>& tupleLayout,
+    const Record::RecordFieldIdentifier& valueFieldName)
+{
+    const auto numberOfEntries = nautilus::invoke(medianNumberOfRecords, pagedVectorBuffer);
+
+    /// Iterating in two nested loops over all the records in the paged vector to get the median.
+    /// We pick a candidate and then count for each item, if the candidate is smaller and also if the candidate is less than the item.
+    const nautilus::val<int64_t> medianPos1 = (numberOfEntries - 1) / 2;
+    const nautilus::val<int64_t> medianPos2 = numberOfEntries / 2;
+    nautilus::val<uint64_t> medianItemPos1 = 0;
+    nautilus::val<uint64_t> medianItemPos2 = 0;
+    nautilus::val<bool> medianFound1(false);
+    nautilus::val<bool> medianFound2(false);
+
+    /// Picking a candidate and counting how many items are smaller or equal to the candidate.
+    /// Iterator-based scan: the page lookup is amortized per page-crossing instead of paid per access,
+    /// which matters here because the outer/inner pair is O(N^2) over the paged vector.
+    const PagedVectorRef pagedVectorRef(BorrowedNautilusBuffer::from(pagedVectorBuffer), tupleLayout);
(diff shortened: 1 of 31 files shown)


nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+namespace NES
+{
+
+nautilus::val<double> medianOf(
+    const nautilus::val<TupleBuffer*>& pagedVectorBuffer,
+    const std::shared_ptr<PagedVectorTupleLayout>& tupleLayout,
+    const Record::RecordFieldIdentifier& valueFieldName)
+{
+    const auto numberOfEntries = nautilus::invoke(
```
PhilippGrulich: I think you could define this as a free function and don't use the lambda here. I guess that you would also get a better name in the IR and the code would maybe be a bit simpler

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
```diff
@@ -0,0 +1,185 @@
...
+    nautilus::val<uint64_t> medianItemPos1 = 0;
+    nautilus::val<uint64_t> medianItemPos2 = 0;
+    nautilus::val<bool> medianFound1(false);
+    nautilus::val<bool> medianFound2(false);
+
+    /// Picking a candidate and counting how many items are smaller or equal to the candidate.
+    /// Iterator-based scan: the page lookup is amortized per page-crossing instead of paid per access,
+    /// which matters here because the outer/inner pair is O(N^2) over the paged vector.
+    const PagedVectorRef pagedVectorRef(BorrowedNautilusBuffer::from(pagedVectorBuffer), tupleLayout);
+    const auto pagedVectorEnd = pagedVectorRef.end();
```
PhilippGrulich, line 91: In general, I'm wondering if it not should be possible to implement the whole median computation in a runtime function. There should be some better data structures to efficiently calculate the median and the main question is how to integrate them in nautilus. Do we only consider numeric sequences?

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
```diff
@@ -0,0 +1,185 @@
...
+        if (not medianFound2 && countLessThan <= medianPos2 && medianPos2 < countLessThan + countEqual)
+        {
+            medianItemPos2 = candidatePos;
+            medianFound2 = true;
+        }
+        candidatePos = candidatePos + 1;
+    }
+
+    VarVal medianValue = VarVal(nautilus::val<double>(0));
+    if (medianFound1 and medianFound2)
```
PhilippGrulich, line 129: sometimes you use  && and sometime and. I prefer && but this is a bit mixed :)

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
```diff
@@ -0,0 +1,185 @@
...
+        const auto containsNull2 = AggregationPhysicalFunction::readNull(aggregationState2);
+        AggregationPhysicalFunction::storeNull(aggregationState1, containsNull1 and containsNull2);
+
+        /// Skipping the first byte (null)
+        memArea1 += nautilus::val<uint64_t>{1};
+        memArea2 += nautilus::val<uint64_t>{1};
+    }
+
+    /// Calling the copyFrom function of the paged vector to combine the two paged vectors by copying the content of the second paged vector to the first paged vector
+    nautilus::invoke(
```
PhilippGrulich: same, I would define the inner lambda as a free function and just call it via invoke

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
```diff
@@ -0,0 +1,185 @@
...
+        memArea2);
+}
+
+void medianResetState(
+    const nautilus::val<AggregationState*>& aggregationState,
+    const nautilus::val<AbstractBufferProvider*>& bufferProvider,
+    const nautilus::val<uint64_t>& tupleSize,
+    const bool nullable)
+{
+    nautilus::invoke(
```
PhilippGrulich, line 174: same

nes-runtime/include/CompilationContext.hpp
```diff
@@ -74,6 +77,11 @@ class CompilationContext
     /// We assume that a compilation context never outlives the module; both live in CompiledExecutablePipelineStage::start()
     nautilus::engine::NautilusModule& module; /// NOLINT(cppcoreguidelines-avoid-const-or-ref-data-members)
     std::vector<std::function<void(nautilus::engine::CompiledModule&)>> pendingResolvers;
+    /// Type-erased owners of the functions handed out by getOrRegisterNautilusFunction(), keyed by their name.
+    /// In interpreted mode this map is looked up at runtime by every worker thread executing the pipeline, so it
+    /// is guarded. The map is node-based, hence the references handed out stay valid past the lock.
+    std::unordered_map<std::string, std::shared_ptr<void>> nautilusFunctions;
```
PhilippGrulich: it need to be ensured that the names are unique within one pipeline

nes-systests/regression/2026-07-23_BenchmarkModeIgnoresGlobalConfiguration.test
```diff
@@ -0,0 +1,37 @@
...
+GlobalConfiguration worker.default_query_execution.page_size: [16384]
+
+CREATE LOGICAL SOURCE stream(value INT64 NOT NULL, ts UINT64 NOT NULL);
+CREATE PHYSICAL SOURCE FOR stream TYPE File;
+ATTACH INLINE
+1,100
+3,150
+5,200
+
+SELECT start, end, MEDIAN(value) as out0, MEDIAN(value) as out1, MEDIAN(value) as out2, MEDIAN(value) as out3,
```
PhilippGrulich, line 19: do you have tested that median functions this generates in the IR? there should only be one I guess?

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
keyseven123: I prefer the other :) 
It is now consistent

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
keyseven123: Yes, we should think about if there is a better datastructure. For now, the performance of the median is not the problem but rather the compilation time. Thus, I would like to do the data structure optimization at a later stage as the current pr really helps in the compilation time.

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
keyseven123: Good idea, done.

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
keyseven123: Done

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
keyseven123: Done

nes-runtime/include/CompilationContext.hpp
keyseven123: Adapted the comment to highlight this

nes-physical-operators/src/Aggregation/Function/MedianNautilusFunctions.cpp
PhilippGrulich: In general if you move more code like this to the runtime instead of using nautilus then your compile time would also improve.

keyseven123: Yes, we are currently reworking the paged vector + buffer + memory anyway and I would like to do it as part of this rework and not here. To not introduce a lot of conflicts.

keyseven123:
@ls-1801 will do this via a sort in the PagedVector

PR #1806, opened 2026-07-23
Fixes SLT Benchmarking Mode

keyseven123:
This PR includes two fixes for the benchmarking mode `-b` of our system level test. 
1. Passing exceptions to cout
2. Passing GlobalConfiguration in `.test`files to the worker

## Verifying this change
This change is tested by existing and newly added tests.

- System Level Tests

### DIFF
--- a/nes-systests/systest/src/SystestRunner.cpp
+++ b/nes-systests/systest/src/SystestRunner.cpp
@@ -26,6 +26,7 @@
 #include <filesystem>
 #include <fstream>
 #include <iostream>
+#include <iterator>
 #include <memory>
 #include <optional>
 #include <ostream>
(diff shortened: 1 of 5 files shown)


nes-systests/systest/src/SystestRunner.cpp
```diff
@@ -604,6 +496,60 @@ std::vector<RunningQuery> runQueriesAtLocalWorker(
...
+             .time = executionTimeInSeconds,
+             .bytesPerSecond = static_cast<double>(runningQuery.bytesProcessed.value_or(NAN)) / executionTimeInSeconds,
+             .tuplesPerSecond = static_cast<double>(runningQuery.tuplesProcessed.value_or(NAN)) / executionTimeInSeconds});
+        return fmt::format(" in {} ({})", runningQuery.getElapsedTime(), runningQuery.getThroughput());
+    };
+
+    /// Benchmarking runs one query at a time so that the timings are not skewed by concurrently running queries.
+    return runQueriesAtLocalWorker(queries, 1, clusterConfig, configuration, progressTracker, benchmarkQuery);
+}
+
```
greenBene, line 552: Just to double-check: The previous implementation started the query using the call

```cpp
auto catalog = std::make_shared<WorkerCatalog>(clusterConfig.workers);
auto worker = std::make_unique<QueryManager>(std::move(catalog), createEmbeddedBackend(configuration));
QuerySubmitter submitter(std::move(worker));
/// [...]
const auto startResult = submitter.startQuery(queryToRun.planInfoOrException.value().queryPlan);
```
The new implementation start them it via 

```cpp
return runQueriesAtLocalWorker(queries, 1, clusterConfig, configuration, progressTracker, benchmarkQuery);
```

The above-mentioned function is defined as 

```cpp
std::vector<RunningQuery> runQueriesAtLocalWorker(
    const std::vector<SystestQuery>& queries,
    const uint64_t numConcurrentQueries,
    const SystestClusterConfiguration& clusterConfig,
    const SingleNodeWorkerConfiguration& configuration,
    SystestProgressTracker& progressTracker,
    const QueryPerformanceMessageBuilder& queryPerformanceMessage)
{
    auto catalog = std::make_shared<WorkerCatalog>(clusterConfig.workers);

    QuerySubmitter submitter(std::make_unique<QueryManager>(std::move(catalog), createEmbeddedBackend(configuration)));
    return runQueries(queries, numConcurrentQueries, submitter, progressTracker, queryPerformanceMessage);
}
```

Given all that, are we sure we "just" forward any exception to `std::cout` (as stated in the commit message), or do we introduce a generally quite different behaviour and change quite fundamentally the setup of the started query?

greenBene (changes requested):
See the comment. If you say it's fine, I am happy to have the PR merged as is :)

nes-systests/systest/src/SystestRunner.cpp
keyseven123: You are right. I have changed the commit message. The actual code run did not change. It is still the same embedded worker backend and runs one query-at-a-time.

greenBene (approved):

adrianmichalke (approved):

PR #1807, opened 2026-07-23
fix: preserve quoted identifiers end to end

ls-1801:
## Summary

- Preserve quotes when qualified identifiers are serialized and deserialized.
- Accept quoted identifiers as window timestamp parameters and retain explicit quoted aggregation aliases.
- Make systest result-schema checks respect the expected identifier's case sensitivity even though CSV headers do not retain quote metadata.
- Add mixed-case quoted-identifier coverage for schemas, projections, filters, windows, aggregations, unions, joins, anonymous sources/sinks, JSON input, and the distributed CLI and REPL.
- Verify that an uppercase quoted identifier such as `"A"` is equivalent to unquoted `A` in systest and distributed CLI/REPL execution.
- Document identifier case sensitivity, CLI YAML quoting, rendered output names, and the current qualified-field limitation in the frontend reference.

## Motivation

Unquoted SQL identifiers are canonicalized to uppercase, while double-quoted identifiers must preserve their exact spelling. Several paths were losing that distinction:

- qualified-identifier reflection formatted the canonical value rather than the original quoted spelling;
- window timestamp grammar bypassed the shared identifier rule;
- aggregation alias projection reconstruction discarded the explicit alias metadata;
- systest reparsed CSV header names as unquoted identifiers, even though CSV headers contain no quote metadata.

Together, these made mixed-case quoted identifiers fail or made systest incapable of checking them correctly.

## Impact

Quoted mixed-case identifiers now survive parsing, logical-plan serialization, distributed deployment, execution, and result validation. Existing unquoted identifiers remain case-insensitive. An uppercase quoted identifier such as `"A"` has the same canonical name as unquoted `A` and therefore resolves to the same object.

Relation-qualified field expressions such as `"stream"."field"` remain outside this change because the current logical field-access representation is unqualified.

## Validation

- Built `nes-cli`, `nes-repl`, `nes-single-node-worker`, and `systest`.
- New quoted-identifier systest: 7/7 queries pass in interpreter and compiler modes.
- Distributed CLI quoted-identifier and `"A"`/`A` compatibility BATS test passes.
- Distributed REPL quoted-identifier and `"A"`/`A` compatibility BATS test passes, including exact schema response assertions.
- Parser, binder, and systest unit selection: 83/83 pass.
- Registered non-large interpreter systest configuration: 621/621 queries pass.
- `git diff --check` passes.


docs/guide/nebulastream-frontend.md
```diff
@@ -29,6 +30,48 @@ All interfaces support JSON output for programmatic access.
...
+CREATE LOGICAL SOURCE "quotedSource" ("mixedValue" UINT64);
+CREATE SINK "quotedSink" ("projectedValue" UINT64) TYPE Void;
+
+SELECT "mixedValue" AS "projectedValue"
+FROM "quotedSource"
+INTO "quotedSink";
+```
+
+Double quotes delimit identifiers; single quotes delimit string values. For example, `"status"` refers to a field,
+whereas `'status'` is a string literal.
```
ThanGeo, line 59: string literals in what context?

ThanGeo (approved):

docs/guide/nebulastream-frontend.md
ls-1801: What we currently call varsized. `SELECT name FROM names WHERE name = 'ThanGeo' ` 'ThanGeo' is a string literal

Artraxon (approved):
LGTM, but we maybe should also serialize the `Identifier` in the same way

keyseven123 (approved):

ls-1801:
@Artraxon you are right, I changed the Identifier serialization as well

PR #1810, opened 2026-07-24
Handle multiple children in pipelining phase

### LINKED ISSUE #1755: [TASK] Break pipelines when operator has multiple children
### Description

The current implementation of `PipeliningPhase.cpp` does not produce correct results for operators with multiple children. Instead of recursing into all children, the pipeline should be broken up when multiple children are encountered. This is needed to support multi-sink plans.

### Acceptance Criteria

The pipelining phase splits pipelines that fan out to multiple children

fkuzun:

`buildPipelineRecursively` recursed into all children within the *same* pipeline, which is only correct for a single consumer: with multiple consumers, they were appended sequentially into one linear pipeline — at runtime the second consumer would read the first consumer's output instead of the shared operator's output — with one emit stacked per consumer, and the shared operator's successor list pointing at the same downstream pipeline twice. 

- New `recurseIntoChildren` helper around all child-recursion sites: a single child continues in the current pipeline as before; multiple children form a fan-out point — every consumer must read the operator's full output, so the pipeline is closed with a single emit (unless the fan-out operator is itself a custom emit) and each consumer starts its own successor pipeline.
- New `PipelinePolicy::ForceNewClosed`, carried into the consumers of a fan-out point: like `ForceNew`, but records that the predecessor pipeline is already closed, so none of the emit-adding sites may add a second emit (which would duplicate the shared output). All decision sites that previously keyed only on `prevOpWrapper`'s pipeline location now also respect this policy; the DAG merge-point (fan-in) upgrade applies only to `Continue`, so it cannot downgrade a `ForceNewClosed` edge.
- Sink consumers of a fan-out point: a native-format sink consumes the shared emit's buffers directly; any other format gets its own formatting pipeline (scan + formatting emit), since the shared emit must stay native for the sibling consumers. Both paths register the sink pipeline in the pipeline map, so further parents converge on the same sink pipeline.
- Defensive invariants that cross-check the carried intent against the actually built state: `endsInEmit` verifies that `ForceNewClosed` edges really start from a closed pipeline, and `addDefaultEmit`/`addOutputFormattingEmit` reject pipelines that already end in an emit — a divergence now fails loudly at the guilty site instead of producing a silently wrong plan.

The test plans are single-root (the fan-out re-merges or converges on one sink), because lowering multi-sink logical plans and the multi-root flip are handled separately (#1754, PR #1797); once that lands, an end-to-end multi-sink variant through the QueryCompiler becomes possible.

## Verifying this change

This change is tested by

- A new `PipeliningMultiConsumerTest`: a hand-built diamond (`src -> F -> {A, B} -> merge -> sink` in pipelining orientation) verifying that the shared operator's pipeline is closed with exactly one emit and fans out into one *distinct* successor pipeline per consumer — on main, `A` and `B` are appended sequentially into `F`'s pipeline (`[scan|F|A|emit|B|emit]`) and `F`'s successor entries are the same merge pipeline twice.
- A new `PipeliningPhaseFanOutTest` covering the pipeline shapes: source fan-out to two consumers that re-merge (distinct consumer pipelines converging on the *same* merge pipeline), fan-out into a non-native sink plus an operator branch reaching the same sink (own formatting pipeline for the sink, shared emit stays native, both paths converge on the same sink pipeline), and a fan-in regression test (merge points still produce a single merged pipeline that both input pipelines have as successor).
- All existing query-compiler tests pass unchanged, covering the single-consumer behavior of every touched code path.

- QueryCompiler

## Documentation

- The fan-out semantics (`recurseIntoChildren`, `ForceNewClosed`) and the intent-vs-state cross-checking (`endsInEmit`) are documented in-code at their definitions.
- Both test suites document the expected pipeline shapes and the pre-fix failure mode they guard against.

This PR closes #1755


nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -190,10 +205,38 @@ void addOutputFormattingEmit(
...
+        return currentPipeline;
+    }
+    const auto formattingPipeline = std::make_shared<Pipeline>(
+        createScanOperator(*currentPipeline, sinkWrapper.getInputSchema(), sinkWrapper.getInputMemoryLayoutType(), configuredBufferSize));
+    currentPipeline->addSuccessor(formattingPipeline, currentPipeline);
+    /// The output format is treated in a case sensitive manner, since it serves as a key to the output formatter registry.
+    /// That's why we cannot pass the upper case sinkFormat.
+    addOutputFormattingEmit(
+        formattingPipeline, sinkWrapper, configuredBufferSize, std::string(sinkFormat), sink.getDescriptor().getOutputFormatterConfig());
+    INVARIANT(formattingPipeline->getRootOperator().getChild().has_value(), "Scan operator requires at least an emit as child.");
```
greenBene: The text of the invariants states that "Scan operator requires at least an emit as child.", but the test just verifies that the root operator has any child. 
Should we ensure that the child is of type `EmitPhysicalOperator`?

nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -203,12 +246,72 @@ void buildPipelineRecursively(
...
+    const std::shared_ptr<Pipeline>& pipeline,
+    OperatorPipelineMap& pipelineMap,
+    const PipelinePolicy childPolicy,
+    const uint64_t configuredBufferSize,
     const MergePointSet& mergePoints)
 {
+    const auto children = opWrapper->getChildren();
+    if (children.size() <= 1)
+    {
+        for (const auto& child : children)
```
greenBene: I feel like this for loop is a bit misleading if you ensure before that `children` contains at most one element. What do you think about using an if clause instead?

nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -203,12 +246,72 @@ void buildPipelineRecursively(
     OperatorPipelineMap& pipelineMap,
     PipelinePolicy policy,
     uint64_t configuredBufferSize,
+    const MergePointSet& mergePoints);
+
+/// Recurses into the children of an operator that was just placed into pipeline. A single child continues with
+/// childPolicy as before. Multiple children form a fan-out point: every consumer must read the operator's full
+/// output, so the pipeline is closed with a single emit and each consumer starts its own successor pipeline.
+void recurseIntoChildren(
```
greenBene, line 260: Not sure if this is possible, but can the children of `opWrapper` form a diamond shape? 
Sth. like 
```
        B > C
      /       \
A >              > F 
      \       /
        D > E
```

If so, wouldn't this algorithm visit the operator `F` twice? And if it does, is it okay?

nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -203,12 +246,72 @@ void buildPipelineRecursively(
...
+        addDefaultEmit(pipeline, *opWrapper, configuredBufferSize);
+    }
+    for (const auto& child : children)
+    {
+        buildPipelineRecursively(
+            child, opWrapper, pipeline, pipelineMap, PipelinePolicy::ForceNewClosed, configuredBufferSize, mergePoints);
+    }
+}
+
+void buildPipelineRecursively(
```
greenBene, line 296: Probably out of scope for this PR, but this function is really long and complex. Maybe we could break each of the 5 cases into separate functions to make it more readable?

greenBene (changes requested):
Looks good :) My only bigger concern is the comment about `recurseIntoChildren`.

nes-query-compiler/src/Phases/PipeliningPhase.cpp
fkuzun: Created an issue https://github.com/nebulastream/nebulastream/issues/1945

nes-query-compiler/src/Phases/PipeliningPhase.cpp
fkuzun: done

nes-query-compiler/src/Phases/PipeliningPhase.cpp
fkuzun: done

nes-query-compiler/src/Phases/PipeliningPhase.cpp
fkuzun: I am not sure if such a plan shape will become used in the future. Currently it is not possible to construct this kind of query via SQL. But the current solution does support it: 
  1. findMergePoints walks the DAG from before pipelining and records every operator reached a second time. F lands in that set.
  2. In buildPipelineRecursively, a merge point downgrades Continue to ForceNew, so F gets a pipeline of its own instead of being fused into whichever branch
  arrived first. 
  3. On the second arrival (via the other branch) the pipelineMap.find(opId) guard hits, wires that branch's pipeline as a predecessor of F's
  existing pipeline, and returns without recursing. So F's subtree is traversed exactly once, and whatever sits below F isn't duplicated.

I added a test showing that it works to create such a pipeline: https://github.com/nebulastream/nebulastream/pull/1810/changes#diff-74654d83cbfa786edf0d3e61b981bab766f41b71a9cd1801a0da8c0e6d5b9e75R291
The execution of such a pipeline leads to a failure though because F sees duplicate sequence numbers with the same origin id. So getting something like this to actually run would require making the origin ids for each path distinguishable

adrianmichalke: hm, if such cases cannot be constructed we shouldnt add a test for it

nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -150,6 +150,19 @@ std::shared_ptr<Pipeline> createNewPipelineWithScan(
     return newPipeline;
 }
 
+/// True iff the pipeline's operator chain already ends in an EmitPhysicalOperator. Used only by invariants
+/// that cross-check the carried intent (pipeline closed / not) against reality. A pipeline closed by a
+/// CUSTOM emit does not end in one — that case is covered by the prevOpWrapper location checks.
+[[maybe_unused]] bool endsInEmit(const Pipeline& pipeline)
```
adrianmichalke: Hm, as we have custom emits like a join build I would prefer to call it for example `endsInEmitPhysicalOperator` to make clear that we refer to the explicit operator.

adrianmichalke: nit: add [[nodiscard]]

nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -346,9 +457,11 @@ void buildPipelineRecursively(
                         sink->getDescriptor().getOutputFormatterConfig());
                 }
 
-                INVARIANT(sourcePipeline->getRootOperator().getChild().has_value(), "Scan operator requires at least an emit as child.");
-                const auto emitOperatorId = sourcePipeline->getRootOperator().getChild().value().getId();
-                pipelineMap[emitOperatorId] = sourcePipeline;
+                const auto emitOperator = sourcePipeline->getRootOperator().getChild();
+                INVARIANT(
+                    emitOperator.has_value() and emitOperator->tryGet<EmitPhysicalOperator>().has_value(),
+                    "Scan operator requires an emit as child");
```
adrianmichalke: nit:
```suggestion
                    "Formatting pipeline are expected to have shape ScanPhysicalOperator -> EmitPhysicalOperator"");
```

nes-query-compiler/src/Phases/PipeliningPhase.cpp
```diff
@@ -412,20 +535,19 @@ void buildPipelineRecursively(
...
+    auto targetPipeline = currentPipeline;
     if (prevOpWrapper->getPipelineLocation() == PhysicalOperatorWrapper::PipelineLocation::EMIT)
     {
         /// If the current operator is a fusible operator and the prev operator was an emit operator, we need to add a scan before the
-        /// current operator to create a new pipeline.
-        createNewPipelineWithScan(currentPipeline, pipelineMap, *opWrapper, configuredBufferSize);
+        /// current operator to create a new pipeline. Everything below must then target that new pipeline, not the closed one.
+        /// @note No dispatch path currently gets here: Case 2 returns for every EMIT-located operator, so the Continue policy is
+        /// only ever passed together with a SCAN- or INTERMEDIATE-located prevOpWrapper. Kept correct in case that changes.
+        targetPipeline = createNewPipelineWithScan(currentPipeline, pipelineMap, *opWrapper, configuredBufferSize);
```
adrianmichalke: I guess its a leftover code path that can be removed

adrianmichalke (changes requested):

nes-query-compiler/src/Phases/PipeliningPhase.cpp
fkuzun: removed

nes-query-compiler/src/Phases/PipeliningPhase.cpp
fkuzun: done

nes-query-compiler/src/Phases/PipeliningPhase.cpp
fkuzun: applied

nes-query-compiler/src/Phases/PipeliningPhase.cpp
fkuzun: done

greenBene (approved):
Thank you for addressing my comments :) The changes look good!

adrianmichalke (approved):

Issue #1804, opened 2026-07-22
[TASK]  Handle multi-sink plans in query plan serialization


fkuzun:
### Description

  `QueryPlanSerializationUtil` currently cannot handle plans with more than one sink root.
  `serializeQueryPlan` asserts a single root and then serializes only
  `rootOperators.front()`: the `BFSRange` walk starts at that one root, so operators
  reachable only from a second sink are never written, and only one entry is added to
  `rootOperatorIds`. `deserializeQueryPlan` builds every id in `rootOperatorIds`
  correctly, but then rejects the result with "Plan contains multiple root operators!",
  validates only `rootOperators.at(0)`, and collapses the plan to that single sink.

  Related: #1753, #1755, #1756

### Acceptance Criteria

  - `serializeQueryPlan` walks every root operator and emits every root id
  - `deserializeQueryPlan` accepts and rebuilds plans with multiple root operators
  - Sink validation applies to every root
  - A multi-sink plan survives a serialize/deserialize round trip with all sinks and shared operators intact

PR #1813, opened 2026-07-24
feat(QueryOptimizer): handle multi-sink plans in optimization rules 

### LINKED ISSUE #1753: [TASK] Handle multi-sink plans in optimization rules
### Description

Optimization rules must be adjusted to allow handling multi-sink plans.
Multi-sink plans share operator instances between sink roots (the same operator is reachable through more than one parent). To create correct output and avoid creating copies of already visited operators, optimization rules must be adjusted to keep track of the operators they have already visited. Because operator ids are not stable during the application of the rule (changed operators get a new id), this might require constructing an external mapping to track visited nodes.

### Acceptance Criteria

All optimization rules return correct results for multi-sink plans

fkuzun:

Multi-sink plans share operator instances between sink roots (the same operator is reachable through more than one parent), so the logical plan is a DAG. Previously, every rule in nes-query-optimizer either asserted a single root or recursed from each root without visited-tracking, cloning the shared subtree once per parent — e.g. a source shared by two sinks came out as two unrelated copies with two different origin ids.

Visited-tracking is only needed by rules whose rewrite traverses the whole plan; rules that only touch the sink roots or only read the plan get by without it. Where tracking is needed, it uses an external mapping keyed on the *input* plan's `OperatorId`s — those are stable during a rule application, whereas every rewrite creates a new operator with a new id (and operator hashing/equality are structural, so operators themselves cannot serve as visited-set keys). The change list is as follows:

- `PlanRewriteUtils` gains the shared machinery (plain free functions; each rule opts in inside its own `apply()`, the `Rule` interface is unchanged):
  - `rewritePlanBottomUp(plan, rewrite)`: memoized bottom-up rewrite that invokes `rewrite` exactly once per unique operator with the original operator and its already-rewritten children; the result is reused by every additional parent, so shared operators stay shared in the output.
  - `getSharedOperatorIds(plan)`: ids of operators reachable through more than one parent edge (duplicate edges from the same parent count).
  - `PushdownBarrier`: visited-tracking restricted to shared operators, for rules that keep their own traversal; rewrites a shared operator at most once and re-introduces each parent's pending state above the cached rewrite.
- Eleven rules rewrite the whole plan bottom-up and now do so through `rewritePlanBottomUp`, with their existing per-operator logic as the callback: `DecideJoinTypes`, `DecideMemoryLayout`, `DecideFieldMappings`, `DecideFieldOrder`, `OriginIdInference`, `RedundantUnionRemoval`, `RedundantProjectionRemoval`, `TypeInference`, `InferModelResolution`, `AnonymousSourceBinding`, and `LogicalSourceExpansion`.
  - `DecideFieldOrder` additionally collects each sink's target order up front, keyed by the sink child's id: sinks sharing a child must agree on the order (a shared operator can only carry one `FieldOrderingTrait`), an agreed order is stamped once, and conflicting orders throw `UnsupportedQuery`.
- The three pushdown rules (`PredicatePushdown`, `WatermarkAssignerPushdown`, `ProjectionPushdown`) keep their own top-down recursion, because the state they thread downward (predicates, watermark assigners, required fields) doesn't fit a bottom-up callback. They use `PushdownBarrier` to track only the shared operators, which act as pushdown barriers: downward-flowing state must not enter a subtree another parent also reads, so pending state is materialized above the barrier per parent edge and the shared subtree is rewritten exactly once with empty pending state. `ProjectionPushdown` rewrites shared operators requiring their full output schema, so one sink's narrowing cannot affect another sink.
  - These rules also no longer call the recursive `withInferredSchema()` after `withChildren(...)`: `withChildren` already re-infers the local schema (children are guaranteed inferred at that point), while the recursive variant deep-copies the just-wired — potentially shared — subtrees and would break sharing again.
- The remaining rules need no visited-tracking at all: `CalcTargetOrder` now infers the target schema order per root instead of asserting a single root (its traversal is read-only, so sharing is not at risk), and the two sink binding rules already iterated all roots without touching the subtrees — they are unchanged, but now covered by tests.

Out of scope, tracked by the sibling issues of this series: lowering (#1754, PR #1797), pipelining (#1755), executable plans/backpressure (#1756), and plan serialization (#1804).

## Verifying this change

This change is tested by

- A new `MultiSinkPlanTest` suite (19 unit tests): one sharing-preservation test per rule, the DAG-specific edge cases (branch predicate/assigner kept above the shared boundary, no narrowing of a shared subtree, one origin id for a shared assigner, agreed vs. conflicting sink field orders, target order feeding another sink's branch), and the full `RuleBasedOptimizer` sequence over a shared-chain plan. Sharing is asserted through operator ids, since structural operator equality cannot detect duplication.
- New `PlanRewriteUtilsTest` cases for `getSharedOperatorIds` (tree, multi-sink, duplicate-edge) and `rewritePlanBottomUp` (once-per-unique-operator invocation, sharing preservation).
- All existing optimizer rule tests pass unchanged, covering the single-sink behavior of every touched rule.

- Query Optimizer

## Documentation

- The new `PlanRewriteUtils` helpers and `PushdownBarrier` are documented in-code, including the invariants they rely on (id-based visited-tracking, bottom-up invocation order).
- The non-obvious decisions carry comments at the point of use: why the pushdown rules must not call the recursive `withInferredSchema()` on rewritten children, and how `DecideFieldOrder` resolves target orders for shared sink children.

This PR closes #1753

### DIFF
--- a/nes-logical-operators/include/Plans/LogicalPlan.hpp
+++ b/nes-logical-operators/include/Plans/LogicalPlan.hpp
@@ -66,22 +66,18 @@ private:
 /// Get all parent operators of the target operator
 [[nodiscard]] std::vector<LogicalOperator> getParents(const LogicalPlan& plan, const LogicalOperator& target);
 
+/// All operators reachable from any root, each one exactly once. BFSRange alone yields a shared operator once per path.
+[[nodiscard]] std::vector<LogicalOperator> planOperators(const LogicalPlan& plan);
+
 [[nodiscard]] LogicalPlan addRootOperators(const LogicalPlan& plan, const std::vector<LogicalOperator>& rootsToAdd);
 
+/// Every operator of the given type, each one exactly once even if it is shared between sink roots.
 template <LogicalOperatorConcept T>
 [[nodiscard]] std::vector<TypedLogicalOperator<T>> getOperatorByType(const LogicalPlan& plan)
 {
-    std::vector<TypedLogicalOperator<T>> operators;
-    std::ranges::for_each(
-        plan.getRootOperators(),
-        [&operators](const auto& rootOperator)
-        {
-            auto typedOps = BFSRange(rootOperator)
-                | std::views::filter([&](const LogicalOperator& op) { return op.tryGetAs<T>().has_value(); })
-                | std::views::transform([](const LogicalOperator& op) { return op.getAs<T>(); });
-            std::ranges::copy(typedOps, std::back_inserter(operators));
-        });
-    return operators;
+    return planOperators(plan) | std::views::filter([](const LogicalOperator& op) { return op.tryGetAs<T>().has_value(); })
+        | std::views::transform([](const LogicalOperator& op) { return op.getAs<T>(); })
+        | std::ranges::to<std::vector<TypedLogicalOperator<T>>>();
 }
 
 [[nodiscard]] std::optional<LogicalOperator> getOperatorById(const LogicalPlan& plan, OperatorId operatorId);
--- a/nes-logical-operators/src/Plans/LogicalPlan.cpp
+++ b/nes-logical-operators/src/Plans/LogicalPlan.cpp
@@ -137,24 +137,57 @@ std::vector<LogicalOperator> getParents(const LogicalPlan& plan, const LogicalOp
     return parents;
 }
 
+std::vector<LogicalOperator> planOperators(const LogicalPlan& plan)
+{
+    std::vector<LogicalOperator> operators;
+    std::unordered_set<OperatorId> visited;
+    for (const auto& root : plan.getRootOperators())
+    {
+        for (const auto& op : BFSRange(root))
+        {
+            if (visited.insert(op.getId()).second)
+            {
+                operators.push_back(op);
+            }
+        }
+    }
+    return operators;
+}
+
+namespace
+{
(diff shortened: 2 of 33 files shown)


greenBene (changes requested):
Hi Felix :) Thank you for implementing this! While I agree with your general approach, I am not really happy with some of the decisions here, such as the reliance on OperatorId or the assumption that you only need to visit each operator once in rules such as PredicatePushdown.

I would like to take over this issue because the optimizer is my active research area, and I want to design it in a principled way. I am sorry for the miscommunication regarding this. I should have prioritized enabling multi-sink plan optimization earlier, but the priority of this in the context of the statistics wasn't quite on my radar. 

I will prioritize this issue and try to have a generalizable solution ASAP, roughly the middle of next week.

Artraxon (changes requested):
Independently of who continues with this PR in what way, I would like to not use OperatorIDs as the memoization key. 
The intention was always that you can just use the LogicalOperators directly, and I was a bit surprised that their operator== also checks structural equality.

I created a PR so that LogicalOperators only use ptr eq https://github.com/nebulastream/nebulastream/pull/1828 , when rebasing it on top of that this PR can just use LogicalOperators directly.

greenBene:
Update: I started to implement #1835 today and will continue on Monday :)

keyseven123:
Obsolete with #1835 .

PR #1814, opened 2026-07-26
fix: Paged Vector fallback to Unpooled buffers for large VarSized

ThanGeo:
This PR implements a fix for the Paged Vector for when a single varsized value is too big and cannot fit on a single pooled buffer. The Paged Vector then reverts to an unpooled buffer with the varsized's size. The data structure's architecture is not affected by this change.
- Added a clause in paged vector that checks if the varsized data fits in a single buffer
- Uses pooled buffer if it fits, unpooled buffer otherwise 

## Verifying this change
This change is tested by property testing unit tests
- Adjusted the way random varsized data is generated against the default buffer size
- added dedicated test for oversized varsized data

### DIFF
--- a/nes-nautilus/src/Interface/PagedVector/PagedVectorRef.cpp
+++ b/nes-nautilus/src/Interface/PagedVector/PagedVectorRef.cpp
@@ -143,7 +143,21 @@ auto makeVarSizedAllocFunction(const NautilusBuffer& lastPageBuffer, const nauti
                                                              .data());
                     }
                 }
-                TupleBuffer newVarSizedBuffer = bufferProvider->getBufferBlocking();
+                TupleBuffer newVarSizedBuffer;
+                if (allocationSize <= bufferProvider->getBufferSize())
+                {
+                    newVarSizedBuffer = bufferProvider->getBufferBlocking();
+                }
+                else
+                {
+                    /// The pooled buffer size can't hold this value; fall back to an unpooled buffer sized to fit it exactly.
+                    auto unpooledBuffer = bufferProvider->getUnpooledBuffer(allocationSize);
+                    if (not unpooledBuffer.has_value())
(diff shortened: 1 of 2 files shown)


ls-1801 (approved):
Were you able to reproduce this in the unit test?

ThanGeo:
> Were you able to reproduce this in the unit test?

Yeah, if you undo the patch then the unit tests fail with very high probability.

keyseven123 (changes requested):
I think we should also test the exception deterministically. Something like this here:

```c++
TEST(PagedVectorTest, oversizedVarSizedThrowsWhenUnpooledUnavailable)
{
    Logger::setupLogging("PagedVectorPropertyTest.log", LogLevel::LOG_DEBUG);
    constexpr uint64_t bufferSize = 4096;
    constexpr size_t numberOfBuffers = 16;
    NoOversizedUnpooledBufferProvider bufferManager{BufferManager::create(
        10 * numberOfBuffers * bufferSize,
        UNPOOLED_MEMORY_FRACTION,
        BUFFER_ALIGNMENT,
        static_cast<uint32_t>(bufferSize),
        std::make_shared<NesDefaultMemoryAllocator>())};

    const std::vector<DataType> fieldTypes{DataType{DataType::Type::VARSIZED, DataType::NULLABLE::NOT_NULLABLE}};
    TestablePagedVector pagedVector(fieldTypes, bufferManager, EngineMode::Interpreter);

    AnyVec record;
    record.emplace_back(std::string(bufferSize * 2, 'x')); /// oversized -> unpooled fallback -> nullopt -> throw
    ASSERT_EXCEPTION_ERRORCODE(pagedVector.pushBack(record), ErrorCode::BufferAllocationFailure);
}
```

keyseven123 (approved):

PR #1815, opened 2026-07-26
Runtime Plugin Resolution

Artraxon:
Our current plugin/registry system tries to resolve all registry entries at configure time.
There are some major downsides and some smaller annoyances:

### Major Downsides
1. Plugins get added to the target of the registry, not the target of the plugin. If one were to add a small function, it's just unexpected behavior, but it leads to multiple more serious issues:
2. If one wants to use dependencies, they have to be reachable in two ways: All .cpps that include them need to have them in their include path; all targets that contain such files need to be linked to the binary of the library containing the implementation. 
But, if you create a plugin, say the simdjson plugin, in nes-plugins, but the source file gets added to nes-input-formatters, how is the dependency resolved?
First, the header is resolved because every .cpp file's compilation includes all headers from all dependencies from vcpkg. It's not a big issue per se, but it's unclean, and (one of) the reasons why CLion always likes to include completely random headers.
But the linking part is much worse: In this example, nes-input-formatters would fail to link because we didn't provide the simdjson dependency. The way it currently gets fixed is to create a new target for the simdjson input formatter, add the source files to it, compile it a _second time_, and link simdjson to that target. The commands currently written out suggest that the dependency gets resolved because we linked the library to the target containing the source files. However, no, it only resolves because linking with `PRIVATE` only hides the header, but still pulls in all the transitive links into all targets that are linked against it.
In other words, _the only reason this works is that the registry is linked against its plugins, and the plugins leak their dependencies, instead of the plugins linking against the registry, and protecting their dependencies._
This is very confusing and unintuitive, and only works because the CMake commands you run in one target mess with completely unrelated targets.
3. This also makes this a walking ODR violation. ODR is literally the only reason this works, because we compiled the same TU twice and linked it in twice, and let the linker throw out one of them. Yes, the source files are identical, but that doesn't mean that the compiler output is the same. Both targets have different include path orderings, because the plugin target includes its own include dirs first. Also, a plugin author might think it's safe to add some `target_compile_definitions`, because after all, it's their own target, which also can lead to the compiler emitting different binaries.
4. Loading plugins via dlopen is impossible. It's not a big issue on its own, because for a workable, community-based plugin system you need lots more infrastructure, but no one could even start building this with the current build system. 
Also, we cannot use GPL-licensed libraries in plugins, because we would have to include them in-tree and link our core components against them (which makes all of our code GPL), instead of just having a bit of glue code that links both the library and us.
5. Adding new registries in plugins is not possible, making it impossible to be both extensible with regards to operators and phases (also in other places, but here it bothers me the most). You cannot implement a rule in terms of: If an operator implements this type trait, call this function on them. You either have to modify the operator by making it inherit from some super class, which is not extensible with regards to rules, or hard-code the operator-specific logic directly in your rule (what @greenBene did in his phases), which is not extensible with regards to operators. 
I would also like to have this for the uniform config system, so that for example the input-formatter component can say: If the systest-default-options registry is present, register 'CSV' AS "INPUT_FORMATTER.TYPE" as a default value.

### Minor Annoyances:
1. I mentioned this already, but it's very hard to understand because CMake functions you call in one place apply non-trivial logic in another place.
2. Adding new registries requires a header + a registrar.inc template, where the latter is done in a bespoke templating language, but always contains the same logic across, just with some types and imports swapped out. The amount of bespoke string templating should be kept to the absolute minimum.
3. All registries are forced to implement the generated key-suffixed functions, even if the entries are already implemented by some other method (for example, template specialization like std::hash). In these cases, the registry should not require additional boilerplate rerouting to a mandatory function; instead, you should be able to point the registry to that, as is already the case with the unreflection registry.

## Requirements
I will use the term `plugin` both for statically linked components and dynamically loaded libraries containing registries and entries.

1. Entries depend on registries, not the other way around.
2. Any plugin containing entries gets compiled and linked only once.
3. Any external dependency of a plugin is hidden from upstream dependent components.
4. Any plugin can not only introduce entries, but also registries.
5. Entries can be marked as optional, so they are only loaded when the registry is also present.
7. The required code-gen in plugins is kept to a minimum.
8. The registration logic is contained in C++, not CMake, which is only used to generate boilerplate code.
9. The unreflection registry becomes a wrapper around the new system.

### Loading Plugins
Anything that wants to introduce a new registry or wants to add registry entries needs to produce a `PluginDescriptor` (which just contains a vector of registries and entries with a name for debugging).
~For dynamically loaded plugins, the code gen generates one for the whole plugin;~ for the statically linked components, it generates one per entry, and one for each registry, because any grouping would be arbitrary and a bit difficult because of link cycles.

Additionally, static "plugins" register themselves during static initialization in a `BuiltinPlugins` catalog.
During startup, all PluginDescriptors in that catalog get loaded at once.

~Dynamic Plugins instead implement an `extern C` interface that is called during dynamic loading and allows the .so to return a PluginDescriptor.~

When any number of PluginDescriptors get loaded (independently of their origin), first all registries are registered in one meta RegistryCatalog.
Then, all entries are put into the appropriate registries. 
Loading will fail with an exception if a non-optional entry registry was not found.

### CMake
We replace the `add_plugin` function with a `add_registry_entry` function for less ambiguity.
~We also introduce `add_dynamic_registry_entry` for use in dynamically loaded libraries.~

Both take an `OPTIONAL` flag and a `KEY` and `HEADER` argument if an entry wants to overwrite those (LogicalFunctions overwrite KEY to register the same function multiple times under the same name).

For static plugins, nothing changes for adding entries.

~`add_dynamic_registry_entry` takes as a first argument the CMake target it should be added to; the other arguments are the same.
 Dynamic plugins need to ensure that their `add_dynamic_registry_entry` call are between the `add_library` call and `finalize_dynamic_plugin`.~

To create a minimal new registry, use `create_runtime_registry(<registry name> <target> ENTRY_TEMPLATE <entry template>)`, whereas <entry template> will be what's included in the auto-generated registration glue-TU. There are several optional parameters, which are documented in the `RuntimeRegistrationUtil.cmake`, but to give an example, for input formatters we write:

```cmake
create_runtime_registry(InputFormatIndexer nes-input-formatters
        ENTRY_TEMPLATE "makeInputFormatterFactory<\${PLUGIN_NAME}InputFormatIndexer>()"
        HEADER_TEMPLATE "\${PLUGIN_NAME}InputFormatIndexer.hpp"
        INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}/private)
```
Whereas `makeInputFormatterFactory<T>` is a helper function defined in pure C++ that turns a `create` method defined in the specific InputFormatterIndexer into an entry. `HEADER_TEMPLATE` are additional `#include` statements that the glue-TUs need, and the `INCLUDE_DIRS` argument is the directory where to find them.

```c++
template <typename IndexerImpl>
InputFormatIndexerFactoryFn makeInputFormatterFactory()
{
    return [](InputFormatIndexerRegistryArguments arguments) -> InputFormatIndexerRegistryReturnType
    {
        return arguments.createInputFormatterWithIndexer(
            IndexerImpl::create(arguments.getInputFormatterConfig(), arguments.getInputMemoryProvider()));
    };
}

``` 
In CSVInputFormatterIndexer.hpp
```c++
    static std::unique_ptr<CSVInputFormatIndexer> create(const InputFormatterDescriptor& config, const TupleBufferRef& tupleBufferRef)
    {
....
    }
```

## Verifying this change

Of course, all tests run and so on, but I also created a visualization of the build graph:

The old build graph:
nes-depgraph-old

The new build graph:
nes-depgraph-new

What you can see is that there are no longer any public dependencies of the registries on the (static) plugins.
There still is an INTERFACE dependency, which doesn't expose headers, gives include dirs, or similar, and is mostly there to simplify the linking with the tests.
If we had some more fine-grained components and test targets, we could get rid of these as well, but that is for future PRs.

CMake and every component that has registries.

## Documentation
- I updated the document for extending NES

## Open Questions:
I am not happy with the LogicalFunctions yet; the functions that I am registering are very ugly because of the inclusion of the key.

I am open to suggestions here; maybe have a factory struct like this?
```c++
template <>
struct LogicalFunctionFactory<"Month_Of">
{
    LogicalFunctionRegistryReturnType operator()(LogicalFunctionRegistryArguments arguments)
    {
        return ExtractFromTimestampLogicalFunction(TimestampUnit::Month, arguments.children[0]);
    }
};
```

Edit: After talking to @ls-1801 I removed the dynamic loading and shared library aspects of this PR. We can revisit this topic once the coordinator PRs are through and the frontends are much thinner.


cmake/RuntimeRegistrationUtil.cmake
```diff
@@ -0,0 +1,450 @@
...
+
+# Generic CMake helpers for runtime registries (see nes-common/include/Util/RuntimeRegistry.hpp).
+#
+# Plugins only DESCRIBE what they provide — registries (RegistryProvision) and entries
+# (EntryProvision), see nes-common/include/Plugins/PluginDescriptor.hpp. The host's
+# PluginLoader performs the actual registration and owns all policy (duplicate keys,
+# missing registries, optional entries, type checks). Built-in and runtime-loaded plugins share
+# this protocol; only the transport differs:
+#
+#  - Built-in: each registry/entry gets a generated glue TU whose static initializer merely
```
greenBene: What is a TU? Please add the name once (maybe including a short description)

nes-logical-operators/src/Functions/LogicalFunctionProvider.cpp
```diff
@@ -35,6 +35,10 @@ LogicalFunction provide(const std::string& functionName, std::vector<LogicalFunc
 
 std::optional<LogicalFunction> tryProvide(const std::string& functionName, std::vector<LogicalFunction> arguments)
 {
-    return LogicalFunctionRegistry::instance().create(functionName, LogicalFunctionRegistryArguments{.children = std::move(arguments)});
+    if (const auto factory = LogicalFunctionRegistry::instance().find(functionName))
+    {
+        return (*factory)(LogicalFunctionRegistryArguments{.children = std::move(arguments)});
```
greenBene, line 40: **nit** I am not sure if "factory" is the right term here. But I don't have a better suggestion, so I'm not being helpful here. 🤷

greenBene (approved):
I quite like this change and think it makes our registry system better! 

One point I want to raise for consideration, but not a blocking one: 

I am not too happy with the need to append the create functions with the function names. E.g. `FromBase64LogicalFunction::createFROM_BASE64()`. While we have a few cases where we create multiple functions from a single class, I think it is an exception. Would it be possible without too much overhead to only having to append the function name in these cases, and by default allow sth like `FromBase64LogicalFunction::create()`?

CMakeLists.txt
```diff
@@ -27,6 +27,7 @@ option(ENABLE_LARGE_TESTS "Runs testcases with larger input data" OFF)
 set(LARGE_TEST_DOWNLOAD_TIMEOUT 1200 CACHE STRING "Absolute timeout in seconds for downloading large test data files")
 option(ENABLE_DOCKER_TESTS "Runs testcases that require docker" ON)
 
+option(NES_BUILD_STATIC_WORKER "Additionally build nes-single-node-worker-static, a self-contained worker without runtime plugin support" OFF)
```
adrianmichalke: Interesting, but what would be the use case for it?

adrianmichalke (commented):
The dependency graph looks much better this way. 

I looked over the other changes but I don’t have a strong opinion on them. I’d be interested to hear @alepping’s thoughts on this.

CMakeLists.txt
Artraxon: With this flag we just get one big statically linked binary as before. Otherwise, you always get the executable + the shared library that has to be reachable by the OS. So in that case, you'd have the nes-single-node-worker and libnes, and you then have to also deal with all of the dynamic linking stuff at the OS level.

Both @ls-1801 @alepping voiced to me that being able to ship a statically linked binary that just contains everything simplifies their lives a lot, and I figured that it doesn't hurt to have the option.

ls-1801 (changes requested):
I would like to suggest a simpler alternative

The main idea is that a statically enabled plugin is compiled exactly once, directly as part of its owning component. Registration is expressed in C++, while CMake only decides whether the plugin's sources and dependencies are present.

For example, enabling the MQTT source can be as simple as:

```cmake
# nes-plugins/CMakeLists.txt
add_subdirectory(Sources/MQTTSource)
```

```cmake
# nes-plugins/Sources/MQTTSource/CMakeLists.txt
target_sources(nes-sources PRIVATE
        MQTTSource.cpp)

target_link_libraries(nes-sources PRIVATE
        PahoMqttCpp::paho-mqttpp3)
```

To disable the plugin, do not add its subdirectory or sources. There is no generated registration translation unit, additional plugin archive, or indirect linking through the registry target.

The registration itself stays next to the implementation:

```cpp
SourceValidationRegistryReturnType RegisterMQTTSourceValidation(SourceValidationRegistryArguments arguments)
{
    return MQTTSource::validateAndFormat(std::move(arguments.config));
}

SourceRegistryReturnType RegisterMQTTSource(SourceRegistryArguments arguments)
{
    return std::make_unique<MQTTSource>(arguments.sourceDescriptor);
}

ADD_PLUGIN("MQTT")
{
    SourceValidationRegistry::registerPlugin("MQTT", RegisterMQTTSourceValidation);
    SourceRegistry::registerPlugin("MQTT", RegisterMQTTSource);
}
```

`ADD_PLUGIN` places one initialization hook in a linker-populated array. At startup, NES iterates over that array and calls every hook. The linker retention is handled generically for the final executables.

This has a few properties I find valuable:

- Plugin implementations are compiled once.
- There is no registry-to-plugin dependency or duplicate translation unit.
- CMake only models ordinary source and library dependencies.
- Registration logic is normal C++ where it can be inspected, stepped through, and debugged with breakpoints.
- A plugin can register with several registries through one initialization hook.
- A future dynamic plugin can export the same initialization hook from an `.so`.

I would avoid making plugin-defined registries, type-erased registry catalogs, optional cross-plugin entries, and a generated descriptor protocol foundational requirements without a concrete use case.

A plugin can already introduce a registry using ordinary C++ dependencies: it exposes the registry header, and interested plugins include it and register from their initialization hook. If initialization ordering eventually matters, the hook descriptor can carry a small dependency list and startup can topologically sort the hooks. That does not require generated C++ or a registry-of-registries.

This deliberately chooses explicit dependencies over runtime discovery. It may not cover every hypothetical plugin composition, but it keeps the common case straightforward and leaves room to add a specific capability once NES actually needs it.

The linker-based registration prototype is implemented in [commit 452aee9d1f](https://github.com/nebulastream/nebulastream/commit/452aee9d1f9acb5d25df89f4d1cfa1de1a0c7cb6). An example of loading a dynamic plugin with the same approach is available in [commit 7afa6bb2b0](https://github.com/nebulastream/nebulastream/commit/7afa6bb2b06126c2418e492e0e22f73a983f164b).
The size of the prototype change mostly comes from two mechanical migrations. First, it deletes the configure-time registry-registration machinery: the CMake functions that collected registry entries, the generated registrar templates and includes, and the additional plugin-library setup that could compile an implementation translation unit more than once. Second, every previously generated registration is made explicit in the plugin's C++ implementation. Most added lines are therefore ordinary calls to `Registry::registerPlugin`, rather than new plugin-framework infrastructure.
The prototype also needs the `WHOLE_ARCHIVE` workaround, so a small amount of CMake remains for each component that may contain plugins.

nes-common/tests/CMakeLists.txt
```diff
@@ -16,9 +16,13 @@ include(GoogleTest)
 #-------------------------------##############----------------------------------
 #------------------------------- # Nes Utils for Test Suite --------------------------------
 
+# TestMain.cpp provides the shared main for all test binaries: it loads the built-in plugins
+# before running the tests. It shadows gtest_main/gmock_main (nes-test-util precedes them on
+# every test's link line, so main resolves from here first).
```
ls-1801: That seems sketchy. Isn't the proper way to not link against gtest_main?

ls-1801 (approved):

cmake/RuntimeRegistrationUtil.cmake
Artraxon: done

nes-common/tests/CMakeLists.txt
Artraxon: thanks, removed the linking

PR #1816, opened 2026-07-26
remove unnecessary usage of static_val to improve tracing performance

PhilippGrulich:
This change remove some unnecessary usages of static val in the operator implementations. 
The usage of static_val is explained here: https://github.com/nebulastream/nautilus/blob/main/docs/static-val.md

### DIFF
--- a/nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
+++ b/nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
@@ -70,7 +70,7 @@ Record FieldOffsetRawBufferIndex::readSpanningRecord(
 {
     Record record;
     const auto indexBufferPtr = nautilus::invoke(getIndexValuesProxy, rawBufferIndex);
-    const auto numberOfFields = nautilus::static_val{bufferRef.getAllDataTypes().size()};
+    const auto numberOfFields = bufferRef.getAllDataTypes().size();
     for (nautilus::static_val<uint64_t> i = 0; i < numberOfFields; ++i)
     {
         const auto fieldName = bufferRef.getAllFieldNames().at(i);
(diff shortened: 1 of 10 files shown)


nes-input-formatters/src/FieldOffsetRawBufferIndex.cpp
```diff
@@ -70,7 +70,7 @@ Record FieldOffsetRawBufferIndex::readSpanningRecord(
 {
     Record record;
     const auto indexBufferPtr = nautilus::invoke(getIndexValuesProxy, rawBufferIndex);
-    const auto numberOfFields = nautilus::static_val{bufferRef.getAllDataTypes().size()};
```
PhilippGrulich, line 73: we only should use static_val as loop iteration variable. Every unnecessary is just unnecessary overhead during tracing

nes-nautilus/src/Interface/HashMap/ChainedHashMap/ChainedHashMapRef.cpp
```diff
@@ -88,7 +88,7 @@ nautilus::val<int8_t*> ChainedHashMapRef::ChainedEntryRef::getValueMemArea() con
     /// We call this method solely, if we actually need the value memory area and not a VarVal.
     /// Therefore, we do not store the valueOffset in the ChainedEntryRef or the ChainedEntryMemoryProvider
     /// During tracing the offset is calculated and should be stored as a constant in the compiled code
-    nautilus::static_val<uint64_t> valueMemAreaOffset(0);
```
PhilippGrulich, line 91: in this function body, we don't trace any code, as a result all the static val loops are not necessary.

ls-1801 (approved):

keyseven123 (approved):

PR #1824, opened 2026-07-28
feat(Version): expose the worker build info over gRPC and via SHOW VERSION

keyseven123:

`nes-single-node-worker --version` prints a build report (version, git commit, build timestamp, build
type, sanitizer, compiler + flags, stdlib, log level, vcpkg baseline), but that string was only
reachable by running the binary locally with a flag. There was no way to ask a *running* worker what
build it is — which matters as soon as workers run in containers and you want to confirm the worker
you talk to actually runs the code you think it does.

This PR exposes that exact string over gRPC and makes it readable from `nes-repl`. The change list is
as follows

Example (`nes-repl -f TEXT`):

```
version
=====================================================================
worker localhost:8080
nes-single-node-worker TBA
  commit:         c9be00290422095d2e7234902a179c8e52b96cf4+dirty
  built:          2026-07-28 15:03:50 +0000
  build:          Debug
  ...
  vcpkg baseline: 4334d8b4c8916018600212ab4dd4bbdc343065d1
```

## Verifying this change

This change is tested by
- *Added `repl/tests/distributed.bats` case that byte-compares the string returned over gRPC against
  `docker compose exec worker-node nes-single-node-worker --version`, so the RPC and the flag cannot
  diverge unnoticed.*
- *Added `repl/tests/distributed.bats` case that stops the worker and asserts the statement still
  succeeds while reporting `error:` for that worker.*
- *Added `repl/tests/embedded.bats` case covering the embedded backend path.*
- *Added `StatementBinderTest.ShowVersionStatementTest` for binding, `FORMAT JSON`, and rejection of
  `SHOW VERSION WHERE ...`.*
- *Added a `--version` smoke test to `nes-single-node-worker/tests/offline.bats`

- gRPC interface (adds one RPC and one message; wire- and source-compatible in both directions)
- SingleNodeWorker
- Frontend (`QueryManager`, both query submission backends, statement layer)
- SQL parser — **note:** `VERSION` becomes a reserved keyword, so an unquoted identifier named
  `version` no longer parses (`"version"` still does). No fixture or systest uses that name.

### DIFF
--- a/nes-frontend/include/Statements/StatementOutputAssembler.hpp
+++ b/nes-frontend/include/Statements/StatementOutputAssembler.hpp
@@ -40,6 +40,7 @@
 #include <DistributedQuery.hpp>
 #include <InputFormatterDescriptor.hpp>
 #include <QueryStatus.hpp>
+#include <Version.hpp>
 
 namespace NES
 {
@@ -439,8 +440,8 @@ struct StatementOutputAssembler<WorkerStatusStatementResult>
     }
 };
 
-using WorkerVersionOutputRowType = std::tuple<Host, std::string>;
-constexpr std::array<std::string_view, 2> workerVersionOutputColumns{"worker", "version"};
+using WorkerVersionOutputRowType = std::tuple<Host, std::optional<VersionInfo>, std::optional<std::string>>;
+constexpr std::array<std::string_view, 3> workerVersionOutputColumns{"worker", "version", "error"};
 
 template <>
 struct StatementOutputAssembler<ShowVersionStatementResult>
@@ -453,7 +454,14 @@ struct StatementOutputAssembler<ShowVersionStatementResult>
         output.reserve(result.versions.size());
         for (const auto& [host, version] : result.versions)
         {
-            output.emplace_back(host, version.has_value() ? *version : fmt::format("error: {}", version.error().what()));
+            if (version.has_value())
+            {
+                output.emplace_back(host, *version, std::nullopt);
+            }
+            else
+            {
+                output.emplace_back(host, std::nullopt, version.error().what());
+            }
         }
         return std::make_pair(workerVersionOutputColumns, output);
     }
--- a/nes-frontend/include/Statements/StatementOutputAssembler.hpp
+++ b/nes-frontend/include/Statements/StatementOutputAssembler.hpp
@@ -439,6 +439,26 @@ struct StatementOutputAssembler<WorkerStatusStatementResult>
     }
 };
 
+using WorkerVersionOutputRowType = std::tuple<Host, std::string>;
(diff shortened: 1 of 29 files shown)


nes-frontend/include/Statements/StatementOutputAssembler.hpp
```diff
@@ -439,6 +439,25 @@ struct StatementOutputAssembler<WorkerStatusStatementResult>
...
+        for (const auto& [host, version] : result.versions)
+        {
+            output.emplace_back(
+                version.has_value() ? fmt::format("worker {}\n{}", host, *version)
+                                    : fmt::format("worker {}\n  error: {}\n", host, version.error().what()));
+        }
+        return std::make_pair(std::to_array<std::string_view>({"version"}), output);
+    }
+};
+
```
Artraxon, line 469: I already don't like the fact that we just pass the version stuff as a string instead of a typed objects, but at least make the row type here a tuple of host and string and let the output formatter decide on how to format them into strings.

Artraxon (changes requested):

nes-frontend/include/Statements/StatementOutputAssembler.hpp
keyseven123: I have used a strong type and returned host and typed object instead of the simple string

Artraxon (approved):

nes-frontend/src/QueryManager/GRPCQuerySubmissionBackend.cpp
```diff
@@ -151,6 +152,24 @@ std::expected<WorkerStatus, Exception> GRPCQuerySubmissionBackend::workerStatus(
...
+std::expected<VersionInfo, Exception> GRPCQuerySubmissionBackend::version() const
+{
+    grpc::ClientContext context;
+    const google::protobuf::Empty request;
+    VersionResponse response;
+    const auto responseCode = stub->RequestVersion(&context, request, &response);
+    if (responseCode.error_code() == grpc::StatusCode::UNIMPLEMENTED)
+    {
+        return std::unexpected(
+            NotImplemented("worker {} does not support the RequestVersion RPC; it was built before this client", workerConfig.host));
```
oczerwia-2010, line 164: Is this tested somehow? how is this kind of worker version mismatch handled in other cases?

oczerwia-2010 (changes requested):

nes-frontend/src/QueryManager/GRPCQuerySubmissionBackend.cpp
keyseven123: Good catch. Added a test case for this

oczerwia-2010 (approved):

PR #1826, opened 2026-07-30
fix(Deps): Pinning vcpkg scripts now as well

keyseven123:
**(for example:)** 

## Verifying this change
This change is tested by
*(for example:)*

*(for example:)*
- ExecutionEngine
- QueryCompiler
- QueryEngine
- RestAPI

## Documentation

### DIFF
--- a/docker/dependency/Dependency.dockerfile
+++ b/docker/dependency/Dependency.dockerfile
@@ -76,6 +76,7 @@ RUN --mount=type=secret,id=VCPKG_CACHE_ACCESS_KEY \
     \
     cd /vcpkg_input; \
     git clone https://github.com/microsoft/vcpkg.git vcpkg_repository; \
+    git -C vcpkg_repository checkout "$(grep -o "\"builtin-baseline\": \"[0-9a-f]*\"" vcpkg.json | cut -d\" -f4)"; \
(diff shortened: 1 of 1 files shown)


ls-1801 (changes requested):
Why? This shouldn't be required vcpkg will manage the repository based on the baseline

keyseven123:
> Why? This shouldn't be required vcpkg will manage the repository based on the baseline

```
Building a dependency image from scratch failed on the first port with:

  CMake Error at scripts/cmake/z_vcpkg_spdx.cmake:15 (string):
    string sub-command JSON got an invalid mode 'STRING_ENCODE'
  error: building libuuid:arm64-linux-none-libcxx failed with: BUILD_FAILED
```

  Dependency.dockerfile cloned microsoft/vcpkg without a revision, so only half of vcpkg was
  pinned: port versions came from builtin-baseline in vcpkg/vcpkg.json (4334d8b, Sep 2025),
  while the tool and scripts/ came from upstream HEAD. Upstream's SPDX helper now uses
  string(JSON ... STRING_ENCODE), which the CMake 3.31.11 pinned in Base.dockerfile does not
  support — every port build calls into that helper, so nothing could be built from source.

oczerwia-2010 (approved):

ls-1801 (approved):

PR #1827, opened 2026-07-30
fix(ci): notify on failure of any nightly job

oczerwia-2010:
## Summary

Only the failures of specific jobs created a message for the ci channel.
E.g. one recent [nighlty run](https://github.com/nebulastream/nebulastream/actions/runs/30420091276) failed silently

## Fix

Make the ci message trigger on any nighly job failure.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -152,7 +152,18 @@ jobs:
       dockerhub-tag: "latest"
 
   send-zulip-msg-on-failure:
-    needs: [ compile-time-regression, create-nightly-image, large-tests, full-clang-tidy ]
+    needs:
+      - nix-build
+      - full-clang-tidy
+      - compile-time-regression
+      - check-format
+      - stress-test
+      - build-and-test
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -152,7 +152,20 @@ jobs:
...
+      - full-clang-tidy
+      - compile-time-regression
+      - check-format
+      - stress-test
+      - build-and-test
+      - run_large_scale_tests_random_config
+      - large-tests
+      - build-and-benchmark
+      - benchmark-compile-time
+      - create-nightly-image
```
keyseven123: As some of them depend on each other, we can get rid of them.

keyseven123 (changes requested):

.github/workflows/nightly.yml
oczerwia-2010: I removed the two intermediate nodes so we now only check leaves

ls-1801 (approved):

keyseven123 (approved):

PR #1828, opened 2026-07-30
fix(LogicalOperators): Ptr equality in LogicalOperators

Artraxon:
Currently, Logical Operators only use ptr equality to short-circuit the comparison, but if it is false, still a structural comparison with the == of the concrete operator is done.
That also means, that if you want to have ptr equality, you'd have to use the operator ID, which is only supposed to be used for serialization.

This was never the intended behaviour, its part of the semantics of a `LogicalOperator` that it is heap-allocated, and PRs like #1813 should be able to just use LogicalOperators as keys.
We already do this in some rules, which also means that the current behaviour is a latent bug, because it would swallow structurally equivalent operators when building up a set or map, even if the intent was, to have them twice.

If you want to have structural comparison, I added `StructuralOperatorEqual` and `StructuralOperatorHash` which you can pass to std::unordered_set or std::unordered_map. Currently, the `flatten` method in `LogicalPlan.hpp` uses it.

## Verifying this change
All tests run.

- LogicalOperators
- Optimizer

### DIFF
--- a/nes-logical-operators/include/Operators/LogicalOperator.hpp
+++ b/nes-logical-operators/include/Operators/LogicalOperator.hpp
@@ -278,24 +278,15 @@ struct TypedLogicalOperator
 
     [[nodiscard]] OperatorId getId() const { return self->getOperatorId(); }
 
+    /// Identity comparison: two wrappers are equal if they refer to the same underlying operator instance,
+    /// analogous to comparing pointers. Compare dereferenced operators (`*lhs == *rhs`) for structural equality.
     [[nodiscard]] bool operator==(const TypedLogicalOperator& other) const
     requires(!std::is_same_v<Checked, NES::detail::ErasedLogicalOperator>)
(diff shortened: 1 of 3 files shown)


adrianmichalke:
@Artraxon comparing ops strucutrally can lead to exceptions if we havent run schema inference before (e.g. getOutputSchema() throws). This seems a bit unsave. Can we assume that we don't compare before it?

Artraxon:
@adrianmichalke yes, we do not do that. In general, we currently do not have any code outside of testing, that actually requires structural eq, there where just some places using it by accident.

At the same time, comparing two logical operators now becomes saves even before running schema inference, so it should reduce unexpected crashes and would also allow us to use more normal optimizer infrastructure in binding.

adrianmichalke (approved):
lgtm

greenBene (approved):

PR #1829, opened 2026-07-30
Use a B-tree with indexed sorted access for median aggregation

ls-1801:
## Summary

Replace the attempted in-place PagedVector sorting design with a B-tree and use it to keep median aggregation values sorted.

- Add a schema-independent C++ `BTree` over fixed-size byte entries. It supports append, size, and direct access by sorted position, and owns storage used by variable-sized values.
- Add `BTreeRef`, which converts between NES `Record` values and the byte entries stored by the tree. Only its record comparator is compiled through Nautilus.
- Add dynamic Nautilus stack allocation so a record can be serialized into temporary entry storage without a persistent insertion buffer.
- Migrate median aggregation from sorting a PagedVector to inserting values into the B-tree and directly reading the middle element or elements.
- Add property tests against a sorted `std::vector<AnyVec>` in compiler and interpreter modes, including nullable values, duplicates, variable-sized data, and forced page splits.
- Share variable-sized test-memory budgeting with the PagedVector properties and remove sorting concerns from the PagedVector tests.

## Why

Sorting a PagedVector in place is the wrong abstraction for variable-sized records. Moving tuple bytes also requires managing references into child buffers and compaction. Median only needs insertion and access to an element by its position in sorted order, which the B-tree provides directly.

The complete record is stored in the tree. Its comparator may use only the fields that determine sorting, while all other fields remain attached to the stored value. The API therefore does not need separate key and value objects.

## Design

`BTree` remains ordinary C++ with a small byte-oriented interface for initialization, append, size, indexed access, and variable-sized allocation. Tree traversal, splitting, storage, and entry counting stay outside Nautilus.

`BTreeRef` provides the record-based Nautilus interface. Comparator functions are registered once per compilation context under stable keys and reused during tracing.

Supporting changes include dynamic `NautilusStackAllocation`, preserving `char` signedness in Nautilus, compiled-function caching by key, and one-time median comparator initialization.

## Verification

The B-tree properties use a sorted `std::vector<AnyVec>` as their expected result and check every indexed entry. They cover compiler and interpreter execution, fixed and variable-sized schemas, nullable values, duplicates, empty trees, generated variable-sized memory budgets, and forced page splits.


nes-nautilus/include/Interface/PagedVector/PagedVector.hpp
```diff
@@ -27,6 +27,10 @@
 namespace NES
 {
 
+struct Arena;
```
keyseven123: I think we prefer includes. If this is not possible, please add a comment why we need to forward declare here.

nes-nautilus/include/Interface/PagedVector/PagedVectorComparator.hpp
```diff
@@ -0,0 +1,53 @@
...
+class PagedVectorComparator
+{
+private:
+    using Function = std::function<bool(TupleBuffer*, int8_t*, TupleBuffer*, int8_t*)>;
+
+    explicit PagedVectorComparator(Function function) : function(std::move(function)) { }
+
+    static std::shared_ptr<PagedVectorComparator> create(Function function)
+    {
+        return std::shared_ptr<PagedVectorComparator>{new PagedVectorComparator(std::move(function))};
```
keyseven123: I think this new leaks memory

nes-nautilus/include/Interface/PagedVector/PagedVectorRef.hpp
```diff
@@ -36,6 +41,8 @@
 
 namespace NES
 {
+struct Arena;
```
keyseven123: Please add why we need this forward declaration here

nes-nautilus/src/Interface/PagedVector/PagedVector.cpp
```diff
@@ -147,6 +200,122 @@ void PagedVector::movePagesFrom(PagedVector& other)
...
+    {
+        pages.emplace_back(buffer.loadChildBuffer(VariableSizedAccess::Index{pageIndex}));
+    }
+
+    const auto recordAddress = [tupleSize](TupleBuffer& page, const uint64_t tupleIndex) -> int8_t*
+    {
+        auto* tuple = page.getAvailableMemoryArea<>().data() + Page::getHeaderSize() + (tupleIndex * tupleSize);
+        return reinterpret_cast<int8_t*>(tuple); /// NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)
+    };
+    const auto recordLess
```
keyseven123: I think this is in general a comparator

nes-nautilus/src/Interface/PagedVector/PagedVector.cpp
```diff
@@ -147,6 +200,122 @@ void PagedVector::movePagesFrom(PagedVector& other)
...
+
+    /// Keep the page handles alive while sorting and merging.
+    std::vector<TupleBuffer, Arena::Allocator<TupleBuffer>> pages{Arena::Allocator<TupleBuffer>{arena}};
+    pages.reserve(numberOfPages);
+    for (size_t pageIndex = 0; pageIndex < numberOfPages; ++pageIndex)
+    {
+        pages.emplace_back(buffer.loadChildBuffer(VariableSizedAccess::Index{pageIndex}));
+    }
+
+    const auto recordAddress = [tupleSize](TupleBuffer& page, const uint64_t tupleIndex) -> int8_t*
```
keyseven123: I do not like this here. Why are we not using the `PagedVectorRef::readRecord`?

keyseven123 (changes requested):
One high-level comment. We always do record/tuples in Nautilus, and now we are mixing both together a lot. Why are we not implementing the sort in nautilus directly?

nes-physical-operators/src/Aggregation/Function/MedianAggregationPhysicalFunction.cpp
```diff
@@ -50,9 +69,21 @@ MedianAggregationPhysicalFunction::MedianAggregationPhysicalFunction(
...
 }
 
+void MedianAggregationPhysicalFunction::setup(CompilationContext& compilationContext)
+{
+    comparatorOwners.emplace_back(PagedVectorRef::registerComparator(
+        compilationContext,
+        comparatorIdentifier,
+        tupleLayout,
+        [inputFieldIdentifier = inputFieldIdentifier](const Record& lhs, const Record& rhs) -> nautilus::val<bool>
+        { return (lhs.read(inputFieldIdentifier) < rhs.read(inputFieldIdentifier)).getRawValueAs<nautilus::val<bool>>(); }));
```
keyseven123: This comparator is a bare IEEE `<`, which is not a strict weak ordering once the input column contains a NaN, and it is fed straight into `std::sort` in `PagedVector::stableSortRun`.

For a NaN `n` and values 3 and 5, `<` is false in both directions, so the predicate at `PagedVector.cpp:54-66` falls through to the index tie-break and reports `n ~ 3` and `n ~ 5` as equivalent while `3 < 5` — equivalence is not transitive. `std::sort` relies on the ordering to bound its insertion-sort/partition loops, so it can run past the end of `indexes` and past the page's tuple area. `MEDIAN` over a FLOAT32/FLOAT64 column with a single NaN (e.g. `0.0/0.0`) then corrupts memory instead of returning a value.

Worth noting the unit tests don't catch this: `compareNonNullRecordValues` in `PagedVectorTest.cpp` adds explicit `std::isnan` handling, so the property test never exercises the production comparator's NaN behaviour.

Could we make the ordering NaN-total here (e.g. order NaNs last), and drop the `isnan` special case from the test comparator so the property test uses the same ordering the engine does?

nes-nautilus/src/Interface/PagedVector/PagedVector.cpp
```diff
@@ -147,6 +200,122 @@ void PagedVector::movePagesFrom(PagedVector& other)
...
+        return reinterpret_cast<int8_t*>(tuple); /// NOLINT(cppcoreguidelines-pro-type-reinterpret-cast)
+    };
+    const auto recordLess
+        = [&comparator, &recordAddress](TupleBuffer& lhsPage, const uint64_t lhsIndex, TupleBuffer& rhsPage, const uint64_t rhsIndex)
+    { return comparator.compare(&lhsPage, recordAddress(lhsPage, lhsIndex), &rhsPage, recordAddress(rhsPage, rhsIndex)); };
+
+    /// Stable-sort every page independently. Original indexes break ties so std::sort produces stable-equivalent ordering.
+    const auto pageCapacity = getPageCapacity();
+    PRECONDITION(pageCapacity <= std::numeric_limits<size_t>::max() / sizeof(uint64_t), "PagedVector page sort allocation size overflow");
+    std::vector<uint64_t, Arena::Allocator<uint64_t>> indexes(pageCapacity, Arena::Allocator<uint64_t>{arena});
```
keyseven123: `indexes` is sized by `getPageCapacity()` — the maximum tuples a page *can* hold — not by the actual record count, and `Arena::Allocator::deallocate` is a no-op, so nothing is reclaimed until the `Arena` from `CompiledExecutablePipelineStage::execute` dies at the end of the whole `execute()` call.

`AggregationProbePhysicalOperator::open` calls `lower()` for every entry of `finalHashMap` (`AggregationProbePhysicalOperator.cpp:150`), and the new `lower()` sorts per group. With a 4 KiB page and an INT32 input, `pageCapacity` is ~1020, so `indexes` alone is ~8 KiB per group. For `SELECT MEDIAN(v) FROM s GROUP BY k` with a few thousand distinct keys per window, one `execute()` accumulates thousands of buffers and can exhaust the pool ("Global buffer pool could not allocate buffer before timeout") even though each group holds only a handful of records.

Sizing `indexes` by the actual per-page record count would shrink it a lot; releasing the scratch per group (or a scoped/reset arena around the sort) would fix the accumulation.

nes-nautilus/src/Interface/PagedVector/PagedVector.cpp
```diff
@@ -147,6 +200,122 @@ void PagedVector::movePagesFrom(PagedVector& other)
...
+    std::priority_queue<RunCursor, HeapStorage, decltype(belongsAfter)> heap{belongsAfter, std::move(heapStorage)};
+    for (uint64_t pageIndex = 0; pageIndex < numberOfPages; ++pageIndex)
+    {
+        if (pages[pageIndex].getNumberOfTuples() > 0)
+        {
+            heap.push(RunCursor{.pageIndex = pageIndex, .tupleIndex = 0});
+        }
+    }
+
+    auto sortedBuffer = bufferProvider.getUnpooledBuffer(getMainBufferSize());
```
keyseven123: At this point the whole input is still resident — the source pages are pinned for the entire call by the `pages` vector (line 212-217, needed for the k-way merge), and the source main buffer is only released once `sortProxy` does `*pagedVectorBuffer = <new buffer>` after the call returns (`PagedVectorRef.cpp:405`). So peak memory is ~2x the PagedVector for every median group, where the old O(N^2) implementation used O(1) extra.

For a long window over a high-rate source this can turn into `BufferAllocationFailure("No unpooled TupleBuffer available!")` from `addNewPage` at emit time, for window sizes that passed before this PR. Is releasing each source page as soon as its records have been merged out an option?

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
```diff
@@ -82,91 +87,27 @@ nautilus::engine::NautilusEngine makeEngine(EngineMode mode)
...
 }
 
-/// Marker pattern written into freshly handed-out buffers so tests notice if the PagedVector
-/// reads uninitialised memory: a dirty page surfaces as 0xDEADBEEF instead of zero.
-constexpr uint32_t DIRTY_FILL_PATTERN = 0xDEADBEEF;
 constexpr NES::BufferAlignment BUFFER_ALIGNMENT{64};
 constexpr double UNPOOLED_MEMORY_FRACTION = 0.9;
 
-struct DirtyBufferProvider : AbstractBufferProvider
+std::shared_ptr<BufferManager>
```
keyseven123: Dropping `DirtyBufferProvider` also dropped the 0xDEADBEEF fill of every handed-out buffer, which was the only thing making reads of uninitialised page memory observable in these property tests.

If the new merge path ever leaves a tuple slot unwritten — the last partially filled output page, or padding not covered by the `memcpy(..., tupleSize)` in `stableSortRecords` — it now reads back as plausible zeros from a fresh `BufferManager` buffer, so `anyVecsEqual` can pass while the engine emits records whose fields silently read 0/false/'\0'. Keeping the dirty-fill (a thin wrapper over the real manager) would preserve that oracle for exactly the code this PR adds.

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalWindowedAggregation.cpp
```diff
@@ -100,6 +94,9 @@ getAggregationPhysicalFunctions(const WindowedAggregationLogicalOperator& logica
         auto physicalFinalType = descriptor.function->getAggregateType();
         auto aggregationInputFunction = QueryCompilation::FunctionProvider::lowerFunction(
             fieldAccessFunction, *logicalOperator.getChild().getTraitSet().get<FieldMappingTrait>());
+        const auto inputFieldIdentifier = aggregationInputFunction.getAs<FieldAccessPhysicalFunction>()->getFieldIdentifier();
+        auto tupleLayout = std::make_shared<DefaultPagedVectorTupleLayout>(
```
keyseven123: `tupleLayout` is consumed only by `MedianAggregationPhysicalFunction` (grep over `nes-physical-operators/src/Aggregation` shows no other reader), but it's built and `make_shared`'d for every aggregation descriptor — ten SUMs build ten single-field layouts that are immediately dropped — and every aggregation function type has to carry a parameter it ignores.

Building it lazily inside the median registrar would keep the cost and the parameter where they belong.

nes-nautilus/include/Arena.hpp
```diff
@@ -42,6 +81,9 @@ struct Arena
     /// 3. The required size is smaller than the last buffer size. In this case, we return the pointer to the address in the last buffer.
     std::span<std::byte> allocateMemory(size_t sizeInBytes);
 
+    /// Allocates memory with the requested alignment.
+    std::span<std::byte> allocateMemory(size_t sizeInBytes, size_t alignment);
+
     std::shared_ptr<AbstractBufferProvider> bufferProvider;
     std::vector<TupleBuffer> fixedSizeBuffers;
     std::vector<TupleBuffer> unpooledBuffers;
```
keyseven123, line 47: (about `lastAllocationSize` just below — commenting here since the member line is outside the diff hunk)

`lastAllocationSize` is write-only after the `allocateMemory` rewrite: `Arena.cpp:58` and `Arena.cpp:89` still assign it, but the Case-2 comparison that read it was deleted and there's no remaining reader in the repo. Worth deleting the member and both assignments so nobody has to work out whether it constrains the allocation logic.

keyseven123 (changes requested):

nes-nautilus/include/Interface/PagedVector/PagedVectorComparator.hpp
ls-1801: How? The shared ptr takes ownership and will call delete.

ls-1801:
Because the sorting algorithm is not schema dependent only the comparison is. There is no benefit of jitting it, and we can rely on proven implementations (like std sort).

Once we have the inlining for nautilus there wouldn't even be an additional function call.

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
ls-1801: The BufferManager was changed in a previous commit (2 months ago) to fill unused bytes with a pattern. So the DirtyBufferProvider is no longer required.

nes-nautilus/src/Interface/BTree/BTreeRef.cpp
```diff
@@ -0,0 +1,285 @@
...
+{
+    PRECONDITION(this->tupleLayout != nullptr, "BTree TupleLayout must not be null");
+}
+
+Record BTreeRef::read(const nautilus::val<int8_t*>& address) const
+{
+    return tupleLayout->readRecord(address, makeBTreeVarSizedLoadFunction(btreeBuffer));
+}
+
+void BTreeRef::appendRecord(
```
keyseven123, line 230: I think this can be an anonymous function. Also, we usually add `proxy` at the end to highlight that this is a proxy functionc all but this is not blocking.

nes-nautilus/tests/TestUtils/include/DataStructureTestUtils.hpp
```diff
@@ -74,6 +74,14 @@ constexpr size_t pooledBufferCountFor(uint64_t bufferSize)
     return std::clamp<size_t>(fitsInBudget, MIN_POOLED_BUFFER_COUNT, POOLED_BUFFER_COUNT);
 }
 
+/// Sizes the test buffer manager so its unpooled-memory allowance can absorb the generated variable-sized payloads
+/// and the allocator's rolling-average chunk preallocation.
+constexpr size_t pooledBufferCountFor(uint64_t bufferSize, uint64_t varSizedMemoryBudget)
+{
+    const auto requiredForVarSized = ((8 * varSizedMemoryBudget) + (9 * bufferSize) - 1) / (9 * bufferSize);
+    return std::max<size_t>(pooledBufferCountFor(bufferSize), requiredForVarSized);
```
keyseven123, line 82: **medium** — this `std::max` discards the `MAX_POOL_BYTES` cap whose stated purpose (line 57) is "so a large buffer size doesn't allocate gigabytes". The cap only applies inside the 1-arg overload; taking the max of it and `requiredForVarSized` bypasses it entirely.

With `bufferSize == 64` (PagedVector's smallest pool entry) and a drawn budget near `MAX_VARSIZED_MEMORY_BUDGET` (64 MiB), `requiredForVarSized ≈ 932,000`, so `createBufferManager` asks for a `10 * 932000 * 64 ≈ 596 MB` total budget and pre-allocates ~932k individual 64-byte pooled buffers plus their control blocks.

That happens per rapidcheck case (~100 per property), across three test binaries, run concurrently under `ctest -jN`. Large runtime/RSS regression and a plausible CI OOM.

nes-nautilus/tests/TestUtils/src/DataStructureTestUtils.cpp
```diff
@@ -103,11 +103,19 @@ rc::Gen<std::vector<DataType>> genDataTypeSchema(std::span<const DataType::Type>
...
     return rc::gen::exec(
-        [types = std::move(types)]()
+        [types = std::move(types), varSizedMemoryBudget = std::move(varSizedMemoryBudget)]()
         {
+            const auto generateString = [&varSizedMemoryBudget]
+            {
+                const auto characters = rc::gen::inRange<char>(PRINTABLE_ASCII_MIN, PRINTABLE_ASCII_MAX);
+                const auto length = *rc::gen::inRange<uint64_t>(0, *varSizedMemoryBudget + 1);
+                *varSizedMemoryBudget -= length;
+                return std::string(length, *characters);
```
keyseven123, line 116: **medium** — the character generator is sampled once, so every generated VARSIZED payload is a run of one repeated character. The replaced code used `rc::gen::container<std::string>(charGen)`, which randomised each byte.

Content-sensitive paths lose their coverage: `BTree`'s comparator only ever differs at byte 0 or by length, and `ChainedHashMap` varsized key hashing/equality now sees a degenerate value space. The change was meant to be about *sizing* the payloads, not flattening their content.

```cpp
return *rc::gen::container<std::string>(length, characters);
```

nes-physical-operators/src/Aggregation/Function/MedianAggregationPhysicalFunction.cpp
```diff
@@ -41,16 +48,45 @@
...
+    return Identifier;
+}
+
+std::shared_ptr<BTreeTupleLayout> createMedianTupleLayout(const DataType& inputType)
+{
+    return std::make_shared<DefaultBTreeTupleLayout>(
+        Schema<QualifiedUnboundField, Ordered>{QualifiedUnboundField{getMedianValueFieldIdentifier(), inputType}});
+}
+
+BTreeComparator::RecordComparator getMedianComparator()
```
keyseven123, line 65: **low** — this comparator is a bare `lhs < rhs` with no NaN handling, but `BTreeRef`'s contract (BTreeRef.hpp:346) is "The comparator must define a strict weak ordering."

For FLOAT32/FLOAT64, a NaN compares false in both directions, so it is "equivalent" to every other value and `findInsertPosition` always places it rightmost — `MEDIAN(f64)` over a window containing NaN silently returns the wrong element.

The property test cannot catch this: `TestableBTree::compareNonNullValues` (TestableBTree.cpp:1473-1484) deliberately *adds* NaN handling to make its comparator a strict weak ordering, so the tested comparator and the shipped one differ exactly here.

nes-physical-operators/src/Aggregation/Function/MedianAggregationPhysicalFunction.cpp
```diff
@@ -41,16 +48,45 @@
...
-    , tupleLayout(std::move(tupleLayout))
+    DataType inputType, DataType resultType, PhysicalFunction inputFunction, Record::RecordFieldIdentifier resultFieldIdentifier)
+    : AggregationPhysicalFunction(inputType, std::move(resultType), std::move(inputFunction), std::move(resultFieldIdentifier))
+    , tupleLayout(createMedianTupleLayout(inputType))
+{
+}
+
+void MedianAggregationPhysicalFunction::setup(CompilationContext& compilationContext)
 {
+    std::call_once(
```
keyseven123, line 83: **low** — `std::call_once` registers the comparator into whichever `CompilationContext` calls `setup()` first. `LowerToPhysicalWindowedAggregation.cpp:210-212` hands the *same* `aggregationPhysicalFunctions` vector to both `AggregationBuildPhysicalOperator` and `AggregationProbePhysicalOperator`, which live in different pipelines and therefore different nautilus modules.

So one of the two pipelines always calls the comparator cross-module, through `std::function` → `ModuleFunction::operator()` (atomic version check + `InvocationScope`) on *every* B-tree key comparison, instead of as a direct in-module call.

It also couples handle readiness to the other pipeline finishing `compile()`: if that pipeline's `compile()` throws, `comparator` is left holding a permanently unresolved handle and the surviving pipeline fails later at `PipelineFunction`'s "invoked before its pipeline module was compiled" invariant rather than at query setup.

`registerFunction` is already keyed and idempotent per context, so registering per context (rather than once per function object) would avoid both.

nes-physical-operators/src/Aggregation/Function/MedianAggregationPhysicalFunction.cpp
```diff
@@ -121,31 +161,39 @@ void MedianAggregationPhysicalFunction::combine(
...
+        treeBuffer1.asArg(),
         static_cast<nautilus::val<uint32_t*>>(memArea1),
         parentBuffer2,
+        treeBuffer2.asArg(),
         static_cast<nautilus::val<uint32_t*>>(memArea2));
+
+    const BTreeRef destination{BorrowedNautilusBuffer::from(treeBuffer1.asArg()), tupleLayout};
+    const BTreeRef source{BorrowedNautilusBuffer::from(treeBuffer2.asArg()), tupleLayout};
+    const auto sourceSize = source.size();
+    for (nautilus::val<uint64_t> index = 0; index < sourceSize; index = index + 1)
```
keyseven123, line 187: **low** (perf) — `combine` replaced `PagedVector::copyPagesFrom` (an O(pages) buffer-level copy) with a per-entry loop that reads each source record, re-serialises it through `writeRecord` (re-allocating var-sized storage per value), and B-tree-inserts it.

Slice merging goes from O(n) memcpy to O(n log n) comparator invocations, each of which is a compiled-function call through the cross-module indirection noted above. For large windows this is the hot path of the probe phase — worth confirming against a median benchmark before merge.

nes-physical-operators/src/Aggregation/Function/MedianAggregationPhysicalFunction.cpp
```diff
@@ -159,93 +207,32 @@ Record MedianAggregationPhysicalFunction::lower(
...
-            const auto medianValue1 = inputFunction.execute(medianRecord1, pipelineMemoryProvider.arena);
-            const auto medianValue2 = inputFunction.execute(medianRecord2, pipelineMemoryProvider.arena);
-            const VarVal two = nautilus::val<uint64_t>(2);
-            medianValue
-                = (medianValue1.castToType(resultType.type) + medianValue2.castToType(resultType.type)) / two.castToType(resultType.type);
-        }
+        const BTreeRef tree{BorrowedNautilusBuffer::from(treeBuffer.asArg()), tupleLayout};
+        const auto numberOfEntries = tree.size();
+
+        const auto medianPos1 = (numberOfEntries - 1) / 2;
```
keyseven123, line 224: **low** — the old code guarded this with an explicit `INVARIANT(numberOfEntriesVal > 0, ...)`, which is gone.

`numberOfEntries` is `val<uint64_t>`, so on an empty tree `(numberOfEntries - 1) / 2` evaluates to `0x7FFF'FFFF'FFFF'FFFF` and both `tree.at()` calls run unconditionally. In Debug/Release that hits `BTree::at`'s `PRECONDITION(index < size())` with a misleading message; in Benchmark builds (asserts off) `at()` proceeds to `node(header().rootIndex)` with `rootIndex == INVALID_CHILD_INDEX` and `loadChildBuffer` runs off the end of the children vector.

I couldn't construct a reachable path today (`reset` is always immediately followed by `lift`, and the nullable path is gated by the "no value seen yet" flag), but the guard that documented that assumption was removed rather than re-expressed.

nes-nautilus/src/Interface/BTree/BTreeRef.cpp
```diff
@@ -0,0 +1,285 @@
...
+                reinterpret_cast<int8_t*>(const_cast<std::byte*>(lhs.data())),
+                reinterpret_cast<int8_t*>(const_cast<std::byte*>(rhs.data()))); /// NOLINT(cppcoreguidelines-pro-type-const-cast)
+        });
+}
+
+void BTreeRef::append(
+    const Record& record, const nautilus::val<AbstractBufferProvider*>& bufferProvider, const BTreeComparator& comparator) const
+{
+    const auto entrySize = tupleLayout->getSizeInBytes();
+    const NautilusStackAllocation entry{entrySize};
```
keyseven123, line 257: Mention here why we need a stack allocation here, if we are adding it to the BTree anyway

nes-physical-operators/include/Aggregation/Function/AggregationPhysicalFunction.hpp
```diff
@@ -46,6 +47,8 @@ class AggregationPhysicalFunction
     AggregationPhysicalFunction(
         DataType inputType, DataType resultType, PhysicalFunction inputFunction, Record::RecordFieldIdentifier resultFieldIdentifier);
 
+    virtual void setup(CompilationContext&) { }
```
keyseven123, line 50: Comment when this gets called and what should be done in its method

nes-nautilus/src/Interface/BTree/BTreeRef.cpp
```diff
@@ -0,0 +1,285 @@
...
+    comparator = [compiledComparator = std::move(compiledComparator)](TupleBuffer* buffer, int8_t* lhs, int8_t* rhs)
+    { return compiledComparator(buffer, lhs, rhs); };
+}
+
+bool BTreeComparator::compare(TupleBuffer* treeBuffer, int8_t* lhs, int8_t* rhs) const
+{
+    return comparator(treeBuffer, lhs, rhs);
+}
+
+Record DefaultBTreeTupleLayout::readRecord(const nautilus::val<int8_t*> recordMemAddress, LoadVarSized loadVarSized) const
```
keyseven123, line 127: We should extract this method or part of it into a Util as free function(s) and then use this method in the RowTupleBufferRef and BTreeRef. In the near future, I assume that we will have for each data structure also a tuple layout with a lot of same common functionality.

nes-nautilus/include/Interface/BTree/BTree.hpp
```diff
@@ -0,0 +1,107 @@
...
+#include <Runtime/AbstractBufferProvider.hpp>
+#include <Runtime/TupleBuffer.hpp>
+
+namespace NES
+{
+/// A schema-agnostic, append-only order-statistic B-tree.
+///
+/// Every key occupies a fixed-size slot. Variable-sized payloads can be allocated through
+/// allocateSpaceForVarSized() and referenced from such a slot (for example with VariableSizedAccess).
+/// BTreeRef provides the record-layout-facing interface.
```
keyseven123, line 32: Maybe also add here how it looks with the tuple buffers. So one root buffer and then one buffer per node etc

nes-nautilus/include/Interface/BTree/BTree.hpp
```diff
@@ -0,0 +1,107 @@
...
+    [[nodiscard]] uint32_t allocateNode(AbstractBufferProvider* bufferProvider, bool leaf);
+    [[nodiscard]] uint32_t ensureRoot(AbstractBufferProvider* bufferProvider);
+    [[nodiscard]] bool isLeaf(uint32_t nodeIndex) const;
+    [[nodiscard]] bool isFull(uint32_t nodeIndex) const;
+    [[nodiscard]] uint32_t childIndex(uint32_t nodeIndex, uint64_t childPosition) const;
+    void incrementSubtreeSize(uint32_t nodeIndex);
+    void insertKey(uint32_t nodeIndex, uint64_t keyPosition, std::span<const std::byte> entry);
+    void splitRoot(AbstractBufferProvider* bufferProvider);
+    void splitChild(uint32_t parentIndex, uint64_t childPosition, AbstractBufferProvider* bufferProvider);
+
```
keyseven123, line 104: Highly appreciate some comments here. Not for the simple methods, such as getter, setter, header() etc

nes-nautilus/include/Interface/BTree/BTree.hpp
```diff
@@ -0,0 +1,107 @@
...
+    explicit BTree(TupleBuffer buffer) : buffer(std::move(buffer)) { }
+
+    [[nodiscard]] uint64_t getEntrySize() const;
+    [[nodiscard]] uint64_t getPageSize() const;
+    [[nodiscard]] uint64_t getMaxKeysPerNode() const;
+    [[nodiscard]] Header& header();
+    [[nodiscard]] const Header& header() const;
+    [[nodiscard]] static uint64_t calculateMaxKeys(uint64_t pageSize, uint64_t entrySize);
+    [[nodiscard]] TupleBuffer node(uint32_t nodeIndex) const;
```
keyseven123, line 85: I think this should be renamed to `getNode()`. `node()` especially without a comment, is difficult to grasp what it does

nes-nautilus/src/Interface/BTree/BTree.cpp
```diff
@@ -0,0 +1,406 @@
...
+    PRECONDITION(static_cast<bool>(comparator), "BTree comparator must not be empty");
+
+    auto current = ensureRoot(bufferProvider);
+    if (isFull(current))
+    {
+        splitRoot(bufferProvider);
+        current = header().rootIndex;
+    }
+
+    while (not isLeaf(current))
```
keyseven123, line 254: Comment here what happens and maybe also why the while loop does not turn into a while(true)

nes-nautilus/src/Interface/BTree/BTree.cpp
```diff
@@ -0,0 +1,406 @@
...
+    PRECONDITION(isFull(oldRoot), "BTree root must be full before splitting");
+    const auto newRoot = allocateNode(bufferProvider, false);
+    auto newRootBuffer = node(newRoot);
+    setChild(newRootBuffer, 0, oldRoot);
+    nodeHeader(newRootBuffer).subtreeSize = nodeHeader(node(oldRoot)).subtreeSize;
+    header().rootIndex = newRoot;
+    splitChild(newRoot, 0, bufferProvider);
+}
+
+void BTree::splitChild(const uint32_t parentIndex, const uint64_t childPosition, AbstractBufferProvider* bufferProvider)
```
keyseven123, line 288: Please add comments in this method here

nes-nautilus/src/Interface/BTree/BTree.cpp
```diff
@@ -0,0 +1,406 @@
...
+    leftHeader.subtreeSize = calculateSubtreeSize(left);
+    rightHeader.subtreeSize = calculateSubtreeSize(right);
+    parentHeader.subtreeSize = calculateSubtreeSize(parent);
+}
+
+std::span<std::byte> BTree::at(uint64_t index) const
+{
+    PRECONDITION(index < size(), "BTree index {} out of bounds for size {}", index, size());
+    auto current = node(header().rootIndex);
+    while (true)
```
keyseven123, line 338: Not a fan of while(true). Can we not get rid of it? I also do not like this goto at all as an exit point of the for loop.

nes-nautilus/src/Interface/BTree/BTree.cpp
```diff
@@ -0,0 +1,406 @@
...
+        for (uint64_t keyIndex = 0; keyIndex < currentHeader.numberOfKeys; ++keyIndex)
+        {
+            if (not currentHeader.leaf)
+            {
+                auto leftChild = node(child(current, keyIndex));
+                const auto leftSize = nodeHeader(leftChild).subtreeSize;
+                if (index < leftSize)
+                {
+                    current = std::move(leftChild);
+                    goto next_node; /// NOLINT(cppcoreguidelines-avoid-goto): exits the key scan into the next tree level.
```
keyseven123, line 350: Can we not get rid of this goto and replace it by extracting the for loop in its own function?

nes-nautilus/src/Interface/BTree/BTree.cpp
```diff
@@ -0,0 +1,406 @@
...
+            page.setNumberOfTuples(offset + storageSize);
+            return {
+                .childIndex = ChildBufferIndex{header().varSizedPageIndex},
+                .offset = offset,
+                .memory = page.getAvailableMemoryArea<>().subspan(offset, storageSize)};
+        }
+    }
+
+    TupleBuffer page;
+    if (storageSize <= bufferProvider->getBufferSize())
```
keyseven123, line 387: This is now the same as in PagedVectorRef and ChainedHashmapRef and already lead to one error/bug. I think a free function for this is better.

nes-nautilus/tests/TestUtils/src/TestableBTree.cpp
```diff
@@ -0,0 +1,223 @@
...
+    const auto projections
+        = schema | std::views::transform([](const auto& field) { return field.getFullyQualifiedName(); }) | std::ranges::to<std::vector>();
+    auto layout = std::make_shared<DefaultBTreeTupleLayout>(schema);
+    const auto recordComparator = makeComparator(projections, dataTypes);
+
+    engine = std::make_unique<nautilus::engine::NautilusEngine>(makeEngine(mode));
+    CompilationContext compilationContext{engine->createModule()};
+    comparator = std::make_shared<BTreeComparator>(compilationContext, layout, "testableBTreeComparator", recordComparator);
+    tree = bufferManager.getUnpooledBuffer(BTree::getMainBufferSize()).value();
+    BTree::init(tree, bufferManager.getBufferSize(), schema.getSizeInBytes());
```
keyseven123, line 146: But we are now testing our own implementation with itself. Why are we not using a sorted container like `std::map`?

keyseven123 (changes requested):
I have implemented a draft of the function extration on the branch `nls/paged-vector-sort`. We can also have a quick chat about it, if you like.

nes-plugins/InputFormatters/JSONInputFormatter/SIMDJSONInputFormatIndexer.hpp
yschroeder97: Why are these changes included?

README.md
yschroeder97: This is also unrelated.

nes-nautilus/include/Interface/BTree/BTree.hpp
ThanGeo: agreed

nes-nautilus/include/Interface/BTree/BTreeRef.hpp
```diff
@@ -0,0 +1,111 @@
...
+    using AllocateVarSized
+        = std::function<nautilus::val<int8_t*>(nautilus::val<int8_t*> fieldSlot, nautilus::val<uint64_t> allocationSize)>;
+
+    virtual ~BTreeTupleLayout() = default;
+    [[nodiscard]] virtual uint64_t getSizeInBytes() const = 0;
+    [[nodiscard]] virtual Record readRecord(nautilus::val<int8_t*> recordMemAddress, LoadVarSized loadVarSized) const = 0;
+    virtual void writeRecord(const Record& record, nautilus::val<int8_t*> memoryForRecord, AllocateVarSized allocateVarSized) = 0;
+};
+
+struct DefaultBTreeTupleLayout final : BTreeTupleLayout
```
ThanGeo, line 51: Maybe add a comment for this layout describing it for reference

nes-nautilus/src/Interface/BTree/BTree.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+TupleBuffer BTree::node(const uint32_t nodeIndex) const
+{
+    PRECONDITION(nodeIndex != INVALID_CHILD_INDEX, "Invalid BTree node index");
+    return buffer.loadChildBuffer(ChildBufferIndex{nodeIndex});
+}
+
+BTree::NodeHeader& BTree::nodeHeader(TupleBuffer& nodeBuffer)
+{
+    return *nodeBuffer.getAvailableMemoryArea<NodeHeader>().data();
```
ThanGeo, line 108: dont you need to check if node is valid before getting the header? or do you check that via the header?

nes-nautilus/src/Interface/BTree/BTree.cpp
```diff
@@ -0,0 +1,406 @@
...
+        }
+        ++position;
+    }
+    return position;
+}
+
+uint32_t BTree::allocateNode(AbstractBufferProvider* bufferProvider, const bool leaf)
+{
+    PRECONDITION(bufferProvider != nullptr, "BTree buffer provider must not be null");
+    auto nodeBuffer = bufferProvider->getUnpooledBuffer(getPageSize());
```
ThanGeo, line 178: getUnpooledBuffer(getPageSize()) is this really what we want here? Won't pooled buffers work? How is page size set?

nes-nautilus/tests/UnitTests/BTreeTest.cpp
```diff
@@ -0,0 +1,357 @@
...
+    appendAndReadByIndexProperty(TestUtils::EngineMode::Compiler);
+}
+
+RC_GTEST_PROP(BTreePropertyTest, appendAndReadByIndexInterpreter, ())
+{
+    Logger::setupLogging("BTreePropertyTest.log", LogLevel::LOG_DEBUG);
+    appendAndReadByIndexProperty(TestUtils::EngineMode::Interpreter);
+}
+
+TEST(BTreeTest, DistinctComparatorsInOneCompilation)
```
ThanGeo, line 296: what is this test about?

ThanGeo (changes requested):

PR #1830, opened 2026-07-31
Fix(CI) clang-tidy-diff: create the results dir before exporting fixes

zeuchste:
The PR-CI split (6d666a0db7) dropped the "Create results directory" step from the clang-tidy-diff workflow. The full all-checks pass writes `-export-fixes clang-tidy-result/fixes.yml` inline, so clang-tidy-diff.py now dies with

```
FileNotFoundError: [Errno 2] No such file or directory: 'clang-tidy-result/fixes.yml'
```

after analysis, failing the check regardless of findings (and masking any real findings behind the crash). First observed on #1776, reproduced deterministically on rerun.

One-line fix: `mkdir -p clang-tidy-result` before the pass. (The pre-check step is unaffected — it goes through `scripts/tidy-diff.sh`, which already creates the directory for its export path.)

### DIFF
--- a/.github/workflows/clang_tidy_diff.yml
+++ b/.github/workflows/clang_tidy_diff.yml
@@ -92,6 +92,9 @@ jobs:
         with:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |
+            # clang-tidy-diff crashes with FileNotFoundError when the -export-fixes directory is missing,
+            # masking the actual findings behind the crash.
+            mkdir -p clang-tidy-result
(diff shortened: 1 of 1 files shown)


ls-1801 (approved):

.github/workflows/clang_tidy_diff.yml
```diff
@@ -92,6 +92,9 @@ jobs:
         with:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |
+            # clang-tidy-diff crashes with FileNotFoundError when the -export-fixes directory is missing,
+            # masking the actual findings behind the crash.
```
ls-1801, line 96: Remove this comment. Or at least modify it, the commit message can say why this was added and what it fixes.

ls-1801 (approved):

PR #1832, opened 2026-07-31
feat(Formatter): Using nautilus function in formatter/parsing to reduce compilation/tracing time

keyseven123:

During tracing, the formatter and parser inline the per-field parsing and formatting logic at every column of the schema. As a result, query compilation time grows superlinearly with schema width. This pull request traces each per-type body once into a shared nautilus function and calls it per column. In particular, the compilation time of `MaxNumberOfSupportedFields` (196 fields) drops from 51.3s to 9.8s, and the two longest taking systests `systest_compiler_sliceCache_true`and `systest_benchmark_small`are 60s and 70s faster, respectively.

## Verifying this change

This change is tested by existing tests.

- ExecutionEngine (nes-nautilus, nes-physical-operators)
- InputFormatters and OutputFormatters

## Documentation

- All new methods carry in-code documentation.


Artraxon (approved):
Non-blocking: I see the benefit of this, but I don't like the fact that you essentially have to write out the signature of the function 5 times. Once for the nautilus::invoke, once for the function itself, once for `using Signature`, once for the lambda, and once the invocation of the registered function. Ideally, we'd have some wrapper registerInvoke that does all of what has to be declared manually here automatically. But it is also possible that this would only be possible with c++26 reflection, but maybe with some smart templating its already possible now.

nes-input-formatters/src/RawValueParser.cpp
```diff
@@ -24,114 +25,206 @@
...
+    /// Capture nullValues by pointer, not by copy: the proxy bakes &nullValues into the compiled code as a constant, so it
+    /// must outlive the pipeline. A captured copy lives only in this lambda and would dangle once tracing is done.
+    /// The type alone identifies the shared function because the other captures are constant within a pipeline: a pipeline
+    /// has a single input formatter, so nullValues is one object and trimQuotes is set only for CHAR. Should a pipeline ever
+    /// parse with two formatters, both would have to enter the name -- see registerTracedFunction().
+    if (dataType.nullable)
+    {
+        using ParseSignature = nautilus::val<ParseResult<T>*>(
+            nautilus::val<int8_t*>, nautilus::val<const FieldIndex*>, nautilus::val<uint64_t>, nautilus::val<uint64_t>);
+        auto& parseFunction = compilationContext.registerTracedFunction<ParseSignature>(
```
PhilippGrulich: why is this a nautilus function if it anyway just calls a runtime function with invoke? 
Could this not be inside that function: if (trimQuotes)
                {
                    address = address + nautilus::val<uint32_t>(1);
                    size = size - nautilus::val<uint32_t>(2);
                 }
                 
                 
                Also you can write address + 1;

nes-input-formatters/src/RawValueParser.cpp
```diff
@@ -24,114 +25,206 @@
...
-                    record.write(fieldName, varVal);
-                    return;
-                }
-            }
-            std::unreachable();
-        }
         case DataType::Type::VARSIZED: {
+            /// Varsized fields stay inline: the record references the raw buffer directly, so there is no compact
+            /// runtime value that could travel back across a shared function boundary.
+            const auto [fieldAddress, fieldSize]
```
PhilippGrulich: Maybe also use a nautilus function here. There is no performance difference as the backend can still inline the function of beneficial. Then Öl the code stays the same

nes-nautilus/include/CompilationContext.hpp
```diff
@@ -117,6 +125,35 @@ class CompilationContext
...
+    /// The name is the identity: nautilus itself interns traced functions by name while tracing, so two bodies
+    /// registered under one name collapse into the first one, silently. The name must therefore encode everything
+    /// the body captures -- not only its signature, but every configuration value baked into it.
+    ///
+    /// The functions live as long as this context, i.e. as long as the pipeline stage: in interpreted mode the
+    /// bodies are not compiled but invoked directly, every time the pipeline runs. That is also why, unlike
+    /// registerFunction(), this may be called after the module has been compiled -- it hands out nothing that
+    /// would still need resolving, and why it must be safe to call from several worker threads at once.
+    template <typename Signature>
+    nautilus::NautilusFunction<std::function<Signature>>& registerTracedFunction(const std::string& name, std::function<Signature> body)
```
PhilippGrulich: I have the feeling that this should be in nautilus, but I'm not completely sure about the requirements.

keyseven123: Do you mean in the nautilus repo? I could create a patch for this, if You like

nes-input-formatters/src/RawValueParser.cpp
keyseven123: Done

nes-input-formatters/src/RawValueParser.cpp
keyseven123: Done

keyseven123:
> Non-blocking: I see the benefit of this, but I don't like the fact that you essentially have to write out the signature of the function 5 times. Once for the nautilus::invoke, once for the function itself, once for `using Signature`, once for the lambda, and once the invocation of the registered function. Ideally, we'd have some wrapper registerInvoke that does all of what has to be declared manually here automatically. But it is also possible that this would only be possible with c++26 reflection, but maybe with some smart templating its already possible now.

I have reduced the amount of times you have write the signature.

ls-1801 (changes requested):
Is this now ready for review?

keyseven123:
> Is this now ready for review?

Now, it is ready for review.

nes-physical-operators/include/ScanPhysicalOperator.hpp
```diff
@@ -33,6 +33,9 @@ class ScanPhysicalOperator final : public PhysicalOperatorConcept
     explicit ScanPhysicalOperator(std::shared_ptr<TupleBufferRef> bufferRef, std::vector<Record::RecordFieldIdentifier> projections);
 
     void open(ExecutionContext& executionCtx, RecordBuffer& recordBuffer) const override;
+    /// Resolves the buffer ref's shared nautilus functions, then continues down the pipeline -- scan is not a leaf,
+    /// so dropping the setupChild() the default does would silently skip every operator below it.
```
ls-1801: We could eliminate this footgun by just having the wrapper iterate over all children and call setup. There is currently nothing to pass along, and I don't see a reason why an operator would ever not want to call the setup of its child operator?

nes-runtime/include/Pipelines/CompiledExecutablePipelineStage.hpp
```diff
@@ -53,10 +54,14 @@ class CompiledExecutablePipelineStage final : public ExecutablePipelineStage
...
-    void registerPipelineFunction(nautilus::engine::NautilusModule& module) const;
+    void registerPipelineFunction(nautilus::engine::NautilusModule& module, CompilationContext& compilationContext) const;
 
     nautilus::engine::NautilusEngine engine;
-    /// Both are created lazily in start(); neither type is default-constructible.
+    /// All created lazily in start(); none of the types is default-constructible. The module and its compilation
+    /// context outlive start() because in interpreted mode the pipeline function is not compiled but invoked
+    /// through the traced lambdas themselves, which still resolve shared nautilus functions out of the context.
+    std::optional<nautilus::engine::NautilusModule> module;
+    std::optional<CompilationContext> compilationContext;
```
ls-1801: nit: I’m not a big fan of keeping the module and compilation context as separate `optional`s. They conceptually belong together, and this representation permits partially initialized states even though their lifetimes depend on each other.

I would prefer an `OperatorFunction` that owns both the module and its `CompilationContext`, ensuring they are always constructed together. `CompiledExecutablePipelineStage` could then own a single `optional<OperatorFunction>`: absent before pipeline setup and created as part of setup.

nes-nautilus/include/CompilationContext.hpp
```diff
@@ -0,0 +1,249 @@
...
+        }
+        return *std::static_pointer_cast<Function>(stored.function);
+    }
+
+    /// registerTracedFunction() for a body that is nothing but a single nautilus::invoke: signature and body are
+    /// both derived from the proxy, so the signature is spelled out once. Anything the shared function should do
+    /// besides the call belongs in the proxy, where it is plain C++ and costs no trace operations.
+    template <typename R, typename... Args>
+    /// NOLINTNEXTLINE(fuchsia-default-arguments-declarations) almost every proxy wants the default attributes
+    auto& registerTracedInvoke(const std::string& name, R (*proxy)(Args...), const nautilus::FunctionAttributes attributes = {})
```
ls-1801: `nautilus::invoke` already emits a single external `CALL` while tracing and does not trace the proxy body. `registerTracedInvoke` wraps that call in a separately traced named function, so once the formatter/parser logic has moved into the plain proxy there is no substantial traced body left to share and the wrapper adds another function and call boundary. Since this adds a function and call boundary without reducing the traced proxy body, I do not expect it to improve total compilation time. Could we drop `registerTracedInvoke` and call the proxies directly with `nautilus::invoke`?

nes-nautilus/include/Interface/BufferRef/TupleBufferRef.hpp
```diff
@@ -88,6 +89,21 @@ class TupleBufferRef
         const nautilus::val<AbstractBufferProvider*>& bufferProvider) const
         = 0;
 
+    /// As above, for callers that can offer a CompilationContext. Only OutputFormatterBufferRef needs it, to let
+    /// its formatter register shared nautilus functions. With TODO #1831, we will get rid of this overload.
+    virtual WriteRecordResult writeRecord(
```
ls-1801: Could we propagate `CompilationContext&` through all remaining `writeRecord` call paths, including the currently context-free utility callers, and then keep only this overload? `OutputFormatterBufferRef` uses the context, while the row and column layouts can ignore it; once every caller supplies it, the context-free virtual overload and the delegating implementations are unnecessary.

nes-physical-operators/src/ScanPhysicalOperator.cpp
```diff
@@ -85,6 +87,12 @@ void ScanPhysicalOperator::open(ExecutionContext& executionCtx, RecordBuffer& re
     }
 }
 
+void ScanPhysicalOperator::setup(ExecutionContext& executionCtx, CompilationContext& compilationContext) const
+{
+    bufferRef->setup(compilationContext);
+    setupChild(executionCtx, compilationContext);
```
ls-1801: Could we move setup propagation out of the individual operator overrides and into `PhysicalOperator::setup`? Every operator in the pipeline needs to be set up, so the wrapper could walk the child chain and call a non-recursive local setup hook, removing the requirement for every override to remember `setupChild`. We would need to define the traversal order explicitly because `ScanPhysicalOperator` currently performs local setup before its child, whereas `WindowProbePhysicalOperator` sets up its child first.

alepping (approved):

PR #1835, opened 2026-07-31
feat(optimizer) enable multi sink optimization

greenBene:
Enabled optimization of multi-sink plans.

* Adds `PlanVisitor<>`, a generic top-down/bottom-up traversal helper for `LogicalPlan` that treats the plan as a DAG: a node reachable through multiple parents is visited exactly once, with contexts from all parents aggregated before descending further, and its rebuilt result shared rather than duplicated. 
* Existing rules are updated to use PlanVisitor

## Verifying this change
* All existing tests continue to pass
* Created new unit tests where necessary for the existing rules. 

- nes-query-optimizer

## Documentation
* PlanVisitor has in-code documentation explaining how it works and how to use it
* Updated `how_to_add_an_optimizer_rule.md` to reflect changes.

### DIFF
--- a/nes-query-optimizer/include/Rules/PlanVisitor.hpp
+++ b/nes-query-optimizer/include/Rules/PlanVisitor.hpp
@@ -1,0 +1,341 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#pragma once
+
+#include <cstdint>
+#include <functional>
+#include <ranges>
+#include <stack>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include <Operators/LogicalOperatorFwd.hpp>
+#include <Plans/LogicalPlan.hpp>
+#include <Util/Overloaded.hpp>
+#include <ErrorHandling.hpp>
+
+/**
+ * @file PlanVisitor.hpp
+ * @brief Generic two-pass traversal/rewrite utility for LogicalPlan DAGs.
+ *
+ * PlanVisitor<OperatorContext, DownContext, UpContext> walks a LogicalPlan first top-down
+ * (sinks to sources) and then bottom-up (sources to sinks), threading user-defined context
+ * through the traversal. It is used to implement rules that need to accumulate information
+ * while descending the plan and then rebuild operators using that information while ascending.
+ *
+ * A plan may have multiple root operators (e.g. several sinks sharing part of a subplan),
+ * so the traversal handles a DAG, not just a tree: each operator is visited exactly once
+ * per pass, even if it is shared by multiple parents. The top-down pass only visits a node
+ * once all its parents have propagated their DownContext to it, and the bottom-up pass only
+ * rebuilds a node once all its children have been rebuilt; the resulting operator is then
+ * reused, not duplicated, by every parent that shares it.
+ *
+ * ## Callbacks
+ * - `FunctionDown`: `(LogicalOperator op, std::vector<DownContext> downContexts) -> DownResult`
+ *   Computes op's own OperatorContext (later handed to FunctionUp) from the DownContexts
+ *   propagated by its parents (one entry per parent edge, empty for a root), and a per-child
+ *   DownContext map to propagate further down — letting different children of the same
+ *   operator receive different contexts.
+ * - `FunctionUp`: `(LogicalOperator op, std::vector<LogicalOperator> newChildren, OperatorContext context,
+ *   std::unordered_map<LogicalOperator, UpContext> upContexts) -> UpResult`
+ *   Rebuilds/rewrites op using its already-rebuilt children, the OperatorContext computed for
+ *   it during the top-down pass, and the UpContext each child produced (keyed by the rebuilt
+ *   child), then returns the new operator together with the UpContext to forward to its own
+ *   parent(s).
+ *
+ * ## Alternative callback signatures
+ * The constructors also accept reduced-arity variants for rules that do not need the full
+ * signature, so callers do not have to accept and ignore parameters they never use:
+ * - `FunctionDownAlt`: `(LogicalOperator op) -> DownResult`
+ *   Drops the DownContexts parameter, for rules that only depend on the operator itself.
+ * - `FunctionUpAlt1`: `(LogicalOperator op, std::vector<LogicalOperator> newChildren, OperatorContext context) -> UpResult`
+ *   Drops the UpContexts map, for rules that do not need per-child UpContext.
+ * - `FunctionUpAlt2`: `(LogicalOperator op, std::vector<LogicalOperator> newChildren,
+ *   std::unordered_map<LogicalOperator, UpContext> upContexts) -> UpResult`
+ *   Drops the OperatorContext, for rules that do not need information from the top-down pass.
+ * - `FunctionUpAlt3`: `(LogicalOperator op, std::vector<LogicalOperator> newChildren) -> UpResult`
+ *   Drops both OperatorContext and UpContexts, for rules that only rebuild op from its
+ *   already-rebuilt children.
+ *
+ * Each constructor takes a `std::variant` of the full callback and its alternatives, so any
+ * of these forms can be passed directly; PlanVisitor normalizes them internally to the full
+ * `FunctionDown`/`FunctionUp` signature.
+ */
+
+namespace NES
+{
+template <typename OperatorContext = std::monostate, typename DownContext = std::monostate, typename UpContext = std::monostate>
+class PlanVisitor
+{
+public:
+    struct DownResult
+    {
+        OperatorContext operatorContext;
+        std::unordered_map<LogicalOperator, DownContext> downContexts;
+    };
+
+    struct UpResult
+    {
+        /// NOLINTNEXTLINE(google-explicit-constructor)
+        UpResult(LogicalOperator op) : op(std::move(op)), upContext({}) { };
+        UpResult(LogicalOperator op, UpContext upContext) : op(std::move(op)), upContext(std::move(upContext)) { };
+        LogicalOperator op;
+        UpContext upContext;
+    };
+
+    /// Main callbacks
+    using FunctionDown = std::function<DownResult(LogicalOperator, std::vector<DownContext>)>;
+    using FunctionUp = std::function<UpResult(
+        LogicalOperator, std::vector<LogicalOperator>, OperatorContext, std::unordered_map<LogicalOperator, UpContext>)>;
+
+    /// Alternative callbacks
+    using FunctionDownAlt = std::function<DownResult(LogicalOperator)>;
+    using FunctionUpAlt1 = std::function<UpResult(LogicalOperator, std::vector<LogicalOperator>, OperatorContext)>;
+    using FunctionUpAlt2
+        = std::function<UpResult(LogicalOperator, std::vector<LogicalOperator>, std::unordered_map<LogicalOperator, UpContext>)>;
+    using FunctionUpAlt3 = std::function<UpResult(LogicalOperator, std::vector<LogicalOperator>)>;
+
+    explicit PlanVisitor(
+        std::variant<FunctionDown, FunctionDownAlt> fnDown, std::variant<FunctionUp, FunctionUpAlt1, FunctionUpAlt2, FunctionUpAlt3> fnUp)
+        : fnDown(normalizeFnDown(fnDown)), fnUp(normalizeFnUp(fnUp)) { };
+
+    explicit PlanVisitor(std::variant<FunctionDown, FunctionDownAlt> fnDown) : PlanVisitor(fnDown, noOpUp) { };
+    explicit PlanVisitor(std::variant<FunctionUp, FunctionUpAlt1, FunctionUpAlt2, FunctionUpAlt3> fnUp) : PlanVisitor(noOpDown, fnUp) { };
+
+    [[nodiscard]] LogicalPlan apply(LogicalPlan plan)
+    {
+        Instance instance{fnDown, fnUp};
+
+        instance.init(plan);
+        instance.traverseDown(plan);
+        instance.traverseUp();
+        const auto newRoots = instance.getNewRoots(plan);
+
+        return plan.withRootOperators(newRoots);
+    }
+
+private:
+    struct Instance
+    {
+        Instance(FunctionDown fnDown, FunctionUp fnUp) : fnDown(std::move(fnDown)), fnUp(std::move(fnUp)) { };
+
+        void init(const LogicalPlan& plan)
+        {
+            std::stack<LogicalOperator> toVisit;
+            std::unordered_set<LogicalOperator> visited;
+
+            for (const auto& op : plan.getRootOperators())
+            {
+                toVisit.push(op);
+            }
+
+            while (!toVisit.empty())
+            {
+                auto op = toVisit.top();
+                toVisit.pop();
+                if (visited.contains(op))
+                {
+                    continue;
+                }
+                visited.insert(op);
+
+                childCounter.emplace(op, op.getChildren().size());
+
+                for (const auto& child : op.getChildren())
+                {
+                    parentCounter[child]++;
+                    parentMap[child].emplace_back(op);
+                    toVisit.push(child);
+                }
+            }
+        }
+
+        void traverseDown(const LogicalPlan& plan)
+        {
+            std::stack<LogicalOperator> toVisit;
+
+            for (const auto& root : plan.getRootOperators() | std::views::reverse)
+            {
+                toVisit.push(root);
+            }
+
+            while (!toVisit.empty())
+            {
+                auto op = toVisit.top();
+                toVisit.pop();
+
+                auto [opCon, childCon] = fnDown(op, std::move(downContextMap[op]));
+
+                operatorContextMap.insert_or_assign(op, opCon);
+
+                for (auto child : op.getChildren())
+                {
+                    auto itr = childCon.find(child);
+                    downContextMap[child].emplace_back(itr == childCon.end() ? DownContext{} : itr->second);
+                }
+
+                if (op.getChildren().empty())
+                {
+                    leaves.emplace_back(op);
+                }
+
+                for (const auto& child : op.getChildren() | std::views::reverse)
+                {
+                    parentCounter[child]--;
+                    if (parentCounter[child] == 0)
+                    {
+                        toVisit.push(child);
+                    }
+                }
+            }
+        }
+
+        void traverseUp()
+        {
+            std::stack<LogicalOperator> toVisit;
+
+            for (const auto& leaf : leaves | std::views::reverse)
+            {
+                toVisit.push(leaf);
+            }
+
+            while (!toVisit.empty())
+            {
+                auto op = toVisit.top();
+                toVisit.pop();
+
+                auto context = operatorContextMap.at(op);
+                std::vector<LogicalOperator> newChildren;
+
+                std::unordered_map<LogicalOperator, UpContext> upContext;
+
+                for (const auto& child : op.getChildren())
+                {
+                    auto newChild = newOperatorMap.at(child);
+                    newChildren.emplace_back(newChild);
+                    upContext.emplace(newChild, upContextMap.at(child));
+                }
+
+                auto [newOp, newUpContext] = fnUp(op, newChildren, context, upContext);
+
+                INVARIANT(
+                    !newOperatorMap.contains(op),
+                    "each operator can only be inserted once into newOperatorMap since it is only visited once");
+                newOperatorMap.insert_or_assign(op, std::move(newOp));
+                INVARIANT(
+                    !upContextMap.contains(op), "each operator can only be inserted once into upContextMap since it is only visited once");
+                upContextMap.insert_or_assign(op, std::move(newUpContext));
+
+                if (auto it = parentMap.find(op); it != parentMap.end())
+                {
+                    for (const auto& parent : it->second)
+                    {
+                        childCounter[parent]--;
+                        if (childCounter[parent] == 0)
+                        {
+                            toVisit.push(parent);
+                        }
+                    }
+                }
+            }
+        }
+
+        std::vector<LogicalOperator> getNewRoots(const LogicalPlan& plan)
+        {
+            std::vector<LogicalOperator> newRoots;
+            for (const auto& root : plan.getRootOperators())
+            {
+                newRoots.emplace_back(newOperatorMap.at(root));
+            }
+
+            return newRoots;
+        }
+
+        FunctionDown fnDown;
+        FunctionUp fnUp;
+        std::unordered_map<LogicalOperator, uint64_t> parentCounter;
+        std::unordered_map<LogicalOperator, std::vector<LogicalOperator>> parentMap;
+        std::unordered_map<LogicalOperator, uint64_t> childCounter;
+
+        std::vector<LogicalOperator> leaves;
+        std::unordered_map<LogicalOperator, std::vector<DownContext>> downContextMap{};
+        std::unordered_map<LogicalOperator, UpContext> upContextMap{};
+        std::unordered_map<LogicalOperator, OperatorContext> operatorContextMap{};
+
+        std::unordered_map<LogicalOperator, LogicalOperator> newOperatorMap;
+    };
+
+    FunctionDown fnDown;
+    FunctionUp fnUp;
+
+    FunctionDown normalizeFnDown(std::variant<FunctionDown, FunctionDownAlt> function)
+    {
+        return std::visit(
+            Overloaded{
+                [](FunctionDown fnDown) -> FunctionDown { return fnDown; },
+                [](FunctionDownAlt fnDown) -> FunctionDown
+                { return [fnDown](LogicalOperator op, std::vector<DownContext>) -> DownResult { return fnDown(op); }; }},
+            function);
+    }
+
+    FunctionUp normalizeFnUp(std::variant<FunctionUp, FunctionUpAlt1, FunctionUpAlt2, FunctionUpAlt3> function)
+    {
+        return std::visit(
+            Overloaded{
+                [](FunctionUp fnUp) -> FunctionUp { return fnUp; },
+                [](FunctionUpAlt1 fnUp) -> FunctionUp
+                {
+                    return [fnUp](
+                               LogicalOperator op,
+                               std::vector<LogicalOperator> children,
+                               OperatorContext operatorContext,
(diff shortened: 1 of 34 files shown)


nes-query-optimizer/include/Rules/Semantic/CalcTargetOrderRule.hpp
```diff
@@ -37,6 +46,12 @@ class CalcTargetOrderRule
     [[nodiscard]] LogicalPlan apply(LogicalPlan plan) const;
     [[nodiscard]] std::set<std::type_index> needs() const;
     [[nodiscard]] std::set<std::type_index> neededBy() const;
+
+private:
+    using Visitor = PlanVisitor<std::monostate, std::monostate, Schema<Field, Ordered>>;
+
+    [[nodiscard]] static Visitor::UpResult calcTargetOrder(
+        LogicalOperator op, std::vector<LogicalOperator> children, std::unordered_map<LogicalOperator, Schema<Field, Ordered>> upContexts);
 };
```
Artraxon, line 55: Can't this all be moved to the .cpp?

nes-query-optimizer/src/Rules/Semantic/CalcTargetOrderRule.cpp
```diff
@@ -50,26 +51,67 @@ namespace NES
...
+        /// avoid overwrite sink orders if they are predefined
+        if (hasOrder(op))
+        {
+            return {op, {}};
+        }
 
+        PRECONDITION(children.size() == 1, "Sinks can only have one child");
 
-    if (const auto reorderer = visiting.tryGetAs<Reorderer>())
+        const auto newTargetSchema = upContexts.at(children.at(0)) | std::views::transform(Unbinder<Field>{})
```
Artraxon: btw you can just do | RangeUnbinder{} | ...

nes-query-optimizer/src/Rules/Static/DecideFieldOrder.cpp
```diff
@@ -57,73 +64,133 @@ namespace NES
...
+struct OperatorContext
 {
-    /// For now we reuse the semantic field order of the operators as a heuristic for deciding the FieldOrdering trait
-    /// Other strategies may be explored for better physical optimization
-    const auto oldWithNewChildren = visiting.getChildren()
-        | std::views::transform([&](const auto& child) { return std::pair{child, applyRecur(child)}; }) | std::ranges::to<std::vector>();
-    const auto newChildren
-        = oldWithNewChildren | std::views::transform(&std::pair<LogicalOperator, LogicalOperator>::second) | std::ranges::to<std::vector>();
+    Schema<UnqualifiedUnboundField, Ordered> requiredFieldOrdering;
+    uint64_t parentCount = 0;
```
Artraxon: It's effectively only used as a bool and thus should be one, and one comment on why its needed.

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
```diff
@@ -156,27 +169,35 @@ LogicalOperator pushBeyondSelection(
         predicateSet.emplace_back(newPredicate);
     }
 
-    return predicatePushdown(op->getChild(), std::move(predicateSet), std::move(fields));
+    const std::unordered_map<LogicalOperator, Projections> toPush{{op->getChild(), {.predicateSet = predicateSet, .fields = fields}}};
```
Artraxon: make it non const and move it into the return struct

nes-query-optimizer/src/Rules/Semantic/AnonymousSourceBindingRule.cpp
```diff
@@ -67,20 +67,23 @@ LogicalOperator AnonymousSourceBindingRule::bindAnonymousSourceLogicalOperators(
             throw InvalidConfigParameter("Could not create an anonymous source descriptor because of invalid config parameters");
         }
         const auto& descriptor = descriptorOpt.value();
-        return SourceDescriptorLogicalOperator::create(descriptor);
+        return LogicalOperator{SourceDescriptorLogicalOperator::create(descriptor)};
     }
```
Artraxon: make the descriptorOpt non const and 
```c++
return LogicalOperator{SourceDescriptorLogicalOperator::create(std::move(descriptorOpt).value())};
```

nes-query-optimizer/src/Rules/Semantic/CalcTargetOrderRule.cpp
```diff
@@ -95,47 +137,26 @@ Schema<Field, Ordered> applyRecursive(const LogicalOperator& visiting)
     {
         outputOrder.push_back(field);
     }
-    return outputOrder | std::ranges::to<Schema<Field, Ordered>>();
-}
 
+    auto schema = outputOrder | std::ranges::to<Schema<Field, Ordered>>();
```
Artraxon: `auto schema = std:move(outputOrder) | std::ranges::to<Schema<Field, Ordered>>();`

nes-query-optimizer/src/Rules/Static/DecideMemoryLayoutRule.cpp
```diff
@@ -31,27 +34,27 @@
 namespace NES
 {
 
+namespace
+{
+PlanVisitor<>::UpResult decideMemoryLayout(const LogicalOperator& op, const std::vector<LogicalOperator>& children)
+{
+    auto traitSet = op.getTraitSet();
+    tryInsert(traitSet, MemoryLayoutTypeTrait{MemoryLayoutType::ROW_LAYOUT});
+    return op.withChildren(children).withTraitSet(traitSet);
```
Artraxon: move traitset

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
```diff
@@ -156,27 +169,35 @@ LogicalOperator pushBeyondSelection(
...
     {
         const auto childFields = fieldsWithNewOperator(child, fields);
-        newChildren.emplace_back(predicatePushdown(child, predicateSet, childFields));
+        downContext[child] = {.predicateSet = predicateSet, .fields = childFields};
     }
 
-    return op.withChildren(newChildren).withInferredSchema();
+    const OperatorContext operatorContext{.toApply = {}, .pushed = downContext};
+
+    return {.operatorContext = operatorContext, .downContexts = downContext};
```
Artraxon: make non const and move

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
```diff
@@ -234,12 +255,21 @@ LogicalOperator pushBeyondProjection(
...
+    {
+        downContext[child] = {.predicateSet = pushable, .fields = fieldsWithNewOperator(child, fields)};
+    }
+
+    const Projections toApply{.predicateSet = nonPushable, .fields = fields};
+
+    const OperatorContext operatorContext{.toApply = toApply, .pushed = downContext};
+
+    return {.operatorContext = operatorContext, .downContexts = downContext};
 }
```
Artraxon: non const and move

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
```diff
@@ -281,25 +311,34 @@ LogicalOperator pushBeyondJoin(
...
-    return addSelectionIfRequired(newOp, std::move(nonPushable), fields);
+    downContexts[left] = {.predicateSet = leftPushable, .fields = leftFields};
+    downContexts[right] = {.predicateSet = rightPushable, .fields = rightFields};
+
+    const Projections toApply = {.predicateSet = nonPushable, .fields = fields};
+
+    const OperatorContext operatorContext{.toApply = toApply, .pushed = downContexts};
+
+    return {.operatorContext = operatorContext, .downContexts = downContexts};
 }
```
Artraxon: non const and move

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
```diff
@@ -281,25 +311,34 @@ LogicalOperator pushBeyondJoin(
...
 
-    const auto newFields = fieldsWithNewOperator(op->getChildren().at(0), fields);
-    return applyToAllChildren(op, predicateSet, newFields);
+    PRECONDITION(op.getChildren().size() == 1, "WatermarkAssigners must have exactly one child");
+
+    const auto child = op.getChildren().at(0);
+    const auto childFields = fieldsWithNewOperator(op->getChildren().at(0), fields);
+
+    return {.operatorContext = {}, .downContexts = {{child, {.predicateSet = predicateSet, .fields = fields}}}};
 }
```
Artraxon: here as well

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
```diff
@@ -344,61 +383,106 @@ LogicalOperator pushBeyondWindowedAggregation(
...
+    std::unordered_map<LogicalOperator, Projections> downContexts;
+
+    downContexts[op->getChild()] = {.predicateSet = pushable, .fields = fieldsWithNewOperator(op->getChild(), fields)};
+
+    const Projections toApply = {.predicateSet = nonPushable, .fields = fields};
+    const OperatorContext operatorContext{.toApply = toApply, .pushed = downContexts};
 
-    return addSelectionIfRequired(newOp, std::move(nonPushable), fields);
+    return {.operatorContext = operatorContext, .downContexts = downContexts};
 }
```
Artraxon: here as well

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
```diff
@@ -344,61 +383,106 @@ LogicalOperator pushBeyondWindowedAggregation(
...
+    else if (auto windowedAggOp = op.tryGetAs<WindowedAggregationLogicalOperator>())
     {
-        return pushBeyondWindowedAggregation(windowedAggOp.value(), predicateSet, std::move(fields));
+        downResult = pushBeyondWindowedAggregation(windowedAggOp.value(), predicateSet, fields);
     }
+    else
+    {
+        downResult
+            = {.operatorContext
+               = {.toApply = {.predicateSet = predicateSet, .fields = fields}, .pushed = {}, .hasMultipleParents = hasMultipleParents},
```
Artraxon: make predicateSet non const and move, but don't move the final result because of name return value optimization

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -122,13 +145,13 @@ LogicalOperator pushBeyondSource(TypedLogicalOperator<SourceDescriptorLogicalOpe
         {
             projections.emplace_back(field.getLastName(), UnboundFieldAccessLogicalFunction{field.getLastName()});
         }
-
-        return TypedLogicalOperator<ProjectionLogicalOperator>{op, projections, ProjectionLogicalOperator::Asterisk{false}};
+        return {.operatorContext{projections}, .downContexts = {}};
```
Artraxon: move projections

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -181,11 +204,10 @@ LogicalOperator pushBeyondProjection(const TypedLogicalOperator<ProjectionLogica
         }
     }
 
-    auto newChild = projectionPushdown(op->getChild(), newRequired);
-    return TypedLogicalOperator<ProjectionLogicalOperator>{newChild, newProjections, ProjectionLogicalOperator::Asterisk{false}};
+    return {.operatorContext = {newProjections}, .downContexts = {{op->getChild(), newRequired}}};
```
Artraxon: move new projections

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -259,19 +281,19 @@ LogicalOperator pushBeyondJoin(const TypedLogicalOperator<JoinLogicalOperator>&
...
+    std::unordered_map<LogicalOperator, std::unordered_set<Field>> downContexts;
 
-    auto newLeft = projectionPushdown(left, requiredLeft);
-    auto newRight = projectionPushdown(right, requiredRight);
+    downContexts[left] = requiredLeft;
+    downContexts[right] = requiredRight;
 
-    return TypedLogicalOperator<JoinLogicalOperator>{
-        std::array{newLeft, newRight}, predicate, op->getWindowType(), op->getJoinType(), op->getJoinTimeCharacteristics()};
+    return {.operatorContext = {}, .downContexts = downContexts};
```
Artraxon: move downContexts

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -259,19 +281,19 @@ LogicalOperator pushBeyondJoin(const TypedLogicalOperator<JoinLogicalOperator>&
         }
     }
 
+    std::unordered_map<LogicalOperator, std::unordered_set<Field>> downContexts;
 
-    auto newLeft = projectionPushdown(left, requiredLeft);
-    auto newRight = projectionPushdown(right, requiredRight);
+    downContexts[left] = requiredLeft;
+    downContexts[right] = requiredRight;
```
Artraxon: also move assign here

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -281,13 +303,13 @@ LogicalOperator pushBeyondUnion(const TypedLogicalOperator<UnionLogicalOperator>
             INVARIANT(newField.has_value(), "the given field must be available in the plan");
             newRequired.insert(newField.value());
         }
-        newChildren.emplace_back(projectionPushdown(child, newRequired));
+        downContexts[child] = newRequired;
     }
 
-    return op.withChildren(newChildren).withInferredSchema();
+    return {.operatorContext = {}, .downContexts = downContexts};
```
Artraxon, line 302: move downContexts

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -415,23 +433,24 @@ LogicalOperator pushBeyondDefault(const LogicalOperator& op, const std::unordere
...
     std::vector<ProjectionLogicalOperator::UnboundProjection> newProjections;
     for (const auto& field : required)
     {
-        auto function = FieldAccessLogicalFunction{Field(newOp, field.getLastName(), field.getDataType())};
+        auto function = UnboundFieldAccessLogicalFunction{field.getLastName()};
         newProjections.emplace_back(field.getLastName(), function);
     }
 
-    return TypedLogicalOperator<ProjectionLogicalOperator>{newOp, newProjections, ProjectionLogicalOperator::Asterisk{false}};
+    return {.operatorContext = {newProjections}, .downContexts = downContext};
```
Artraxon: move newProjections and downContext

nes-query-optimizer/src/Rules/Static/WatermarkAssignerPushdownRule.cpp
```diff
@@ -170,21 +185,13 @@ LogicalOperator pushBeyondProjection(
...
+    auto childContexts = getChildContexts(op, {.ingestionTime = context.ingestionTime, .eventTime = pushFurther});
 
-    for (const auto& [onField, unit] : std::ranges::reverse_view(applyNow))
-    {
-        auto field = newOp.getOutputSchema()[onField.getLastName()];
-        PRECONDITION(field.has_value(), "the operator must have the event time field");
-        newOp = TypedLogicalOperator<EventTimeWatermarkAssignerLogicalOperator>(newOp, FieldAccessLogicalFunction{field.value()}, unit);
-    }
-    return newOp;
+    return {.operatorContext = {.ingestionTime = false, .eventTime = applyNow}, .downContexts = childContexts};
```
Artraxon: move childContexts

Artraxon (changes requested):
lgtm since we already talked about it a lot, mostly just move a bit more

nes-query-optimizer/include/Rules/Semantic/CalcTargetOrderRule.hpp
greenBene: Sadly, no, because we create a new `SinkDescriptor` within the function, which is only possible because the class `SinkDescriptor` is befriended with this rule.

nes-query-optimizer/src/Rules/Semantic/CalcTargetOrderRule.cpp
greenBene: Applied

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Semantic/AnonymousSourceBindingRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/DecideFieldOrder.cpp
greenBene: done

nes-query-optimizer/src/Rules/Semantic/CalcTargetOrderRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/DecideMemoryLayoutRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/PredicatePushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: done

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: done

Artraxon (approved):

nes-query-optimizer/include/Rules/PlanVisitor.hpp
```diff
@@ -0,0 +1,334 @@
...
+    using FunctionUpAlt1 = std::function<UpResult(LogicalOperator, std::vector<LogicalOperator>, OperatorContext)>;
+    using FunctionUpAlt2
+        = std::function<UpResult(LogicalOperator, std::vector<LogicalOperator>, std::unordered_map<LogicalOperator, UpContext>)>;
+    using FunctionUpAlt3 = std::function<UpResult(LogicalOperator, std::vector<LogicalOperator>)>;
+
+    explicit PlanVisitor(
+        std::variant<FunctionDown, FunctionDownAlt> fnDown, std::variant<FunctionUp, FunctionUpAlt1, FunctionUpAlt2, FunctionUpAlt3> fnUp)
+        : fnDown(normalizeFnDown(fnDown)), fnUp(normalizeFnUp(fnUp)) { };
+
+    explicit PlanVisitor(std::variant<FunctionDown, FunctionDownAlt> fnDown) : PlanVisitor(fnDown, noopUp) { };
```
keyseven123:
```suggestion
    explicit PlanVisitor(std::variant<FunctionDown, FunctionDownAlt> fnDown) : PlanVisitor(fnDown, noOpUp) { };

```

nes-query-optimizer/include/Rules/PlanVisitor.hpp
```diff
@@ -0,0 +1,334 @@
...
+ *   already-rebuilt children.
+ *
+ * Each constructor takes a `std::variant` of the full callback and its alternatives, so any
+ * of these forms can be passed directly; PlanVisitor normalizes them internally to the full
+ * `FunctionDown`/`FunctionUp` signature.
+ */
+
+namespace NES
+{
+template <typename OperatorContext = std::monostate, typename DownContext = std::monostate, typename UpContext = std::monostate>
```
keyseven123, line 82: Not blocking and as you are the codeowner your decision. 
The .hpp file is quite large, you could think about moving the implementation to a .cpp and then add into the .cpp the actual instantiation set.

nes-query-optimizer/include/Rules/PlanVisitor.hpp
```diff
@@ -0,0 +1,334 @@
...
+                {
+                    auto newChild = newOperatorMap.at(child);
+                    newChildren.emplace_back(newChild);
+                    upContext.emplace(newChild, upContextMap.at(newChild));
+                }
+
+                auto [newOp, newUpContext] = fnUp(op, newChildren, context, upContext);
+
+                newOperatorMap.insert_or_assign(op, newOp);
+                upContextMap.insert_or_assign(newOp, newUpContext);
```
keyseven123: What happens if the newOp is already in the map, if a rule returns op directly , then we would overwrite its context. I think this should be fine but not 100% sure.

nes-query-optimizer/include/Rules/PlanVisitor.hpp
```diff
@@ -0,0 +1,334 @@
...
+        instance.traverseUp();
+        const auto newRoots = instance.getNewRoots(plan);
+
+        return plan.withRootOperators(newRoots);
+    }
+
+private:
+    struct Instance
+    {
+        Instance(FunctionDown fnDown, FunctionUp fnUp) : fnDown(fnDown), fnUp(fnUp) { };
```
keyseven123: Std::move here? I think we are here copying any function twice. Once passing it to the constructor and then in the constructor again

nes-query-optimizer/include/Rules/PlanVisitor.hpp
```diff
@@ -0,0 +1,334 @@
...
+            {
+                toVisit.push(root);
+            }
+
+            while (!toVisit.empty())
+            {
+                auto op = toVisit.top();
+                toVisit.pop();
+
+                auto contexts = downContextMap[op];
```
keyseven123: Copying the full context always. Also we are here creating an empty vector, if thge op was not part of the downContextMap

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
```diff
@@ -45,21 +45,45 @@
...
+{
+    std::unordered_set<Field> result;
+
+    for (auto requiredFields : downContexts)
+    {
+        result.insert(requiredFields.begin(), requiredFields.end());
+    }
+
+    return result;
+}
```
keyseven123:
```suggestion
    return downContexts | std::views::join | std::ranges::to<std::unordered_set<Field>>();
```

keyseven123 (changes requested):
I have added some tests that have failed for me locally.

nes-query-optimizer/include/Rules/PlanVisitor.hpp
greenBene: applied for `noopUp` and `noopDown`

nes-query-optimizer/include/Rules/PlanVisitor.hpp
greenBene: Since most of it uses templating for the contexts, I don't think we can hide much in a .cpp file. Thus, I'd leave it as is for now. If we find a better way in the future, I am always happy to refactor :)

nes-query-optimizer/include/Rules/PlanVisitor.hpp
greenBene: While I think it should be okay, it could have caused some weird behaviour. 
I slightly changed the logic so that upContextMap uses the original operator (and not the new ones) and added invariants to ensure that newOperatorMap and newContextMap can only be set once.

nes-query-optimizer/include/Rules/PlanVisitor.hpp
greenBene: done

nes-query-optimizer/include/Rules/PlanVisitor.hpp
greenBene: 1) The creation of the empty vector is required. E.g., if you visit a sink, it has no parents. It still needs an empty vector as input. 
2) Agree about the unnecessary copying. downContextMap is only read once per operator and not written to afterward. It can be moved here. I changed it accordingly.

nes-query-optimizer/src/Rules/Static/ProjectionPushdownRule.cpp
greenBene: applied

greenBene:
@keyseven123 thanks for the thorough review! I applied most of your suggested changes. Your suggested test cases helped to make incorrect behaviour visible, and I fixed the underlying issues. The only exception was the Projection Pushdown test: I don´t agree that two projections were the expected behaviour (as indicated by the test). I changed the test to what I expect the behaviour to be and ensured that it works as intended. Do you agree with the changed behaviour, or did I misunderstand something?

keyseven123:
> @keyseven123 thanks for the thorough review! I applied most of your suggested changes. Your suggested test cases helped to make incorrect behaviour visible, and I fixed the underlying issues. The only exception was the Projection Pushdown test: I don´t agree that two projections were the expected behaviour (as indicated by the test). I changed the test to what I expect the behaviour to be and ensured that it works as intended. Do you agree with the changed behaviour, or did I misunderstand something?

Sounds good for now.

keyseven123 (approved):

PR #1837, opened 2026-08-01
refactor(test): bind compose workdirs directly

ls-1801:
## What changed

- replace Docker named-volume staging with direct bind mounts for CLI, REPL, and remote systest Compose suites
- use deterministic, inspectable directories under `build/test-tmp/<suite>/<test>`
- write generated worker configuration and container output directly into those directories
- remove Alpine helper containers, `docker cp`, copy-back synchronization, and named-volume cleanup
- bind remote systest data and repository paths directly instead of copying them through helper containers
- automatically disable Docker tests when CMake detects that the host Docker daemon cannot see the build directory at the same absolute path
- document the identical host/container workspace-path requirement

## Why

These tests use Docker-out-of-Docker: Docker commands run in the development container but Compose creates sibling containers through the host daemon. Named volumes previously required temporary Alpine containers and repeated `docker cp` operations. Besides adding an undeclared alpine image dependency, failures in that staging path surfaced as opaque Bats/CTest failures.

The CI and documented development-container invocation already mount the repository at the same absolute host and container path. Direct bind mounts make test files and logs immediately visible and remove the helper-container dependency entirely.

## Path eligibility

When configuration runs inside a container, CMake writes a random token below the build directory, discovers the enclosing container through the host daemon, and verifies the token through a fresh `--pull=never` bind-mounted container. Docker tests are disabled when that round trip fails.

## Validation

- Bash syntax checks for all changed Bats and Compose scripts
- `git diff --check`
- generated CLI, REPL, and remote-systest Compose configurations validated with `docker compose config --quiet`
- confirmed no `TEST_VOLUME`, `TESTDATA_VOLUME`, `TESTCONFIG_VOLUME`, `docker cp`, Alpine helper-container, or `sync_workdir` references remain in the affected suites
- exercised the path-discovery and bind-mount approach against the local host Docker daemon

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -92,7 +92,7 @@ jobs:
           docker_password: ${{ secrets.DOCKER_SECRET }}
           run: |
             source .github/.env/test-${{ matrix.sanitizer }}.env
-            ctest --test-dir build -j -LE "Systest|DockerCompose" --output-on-failure --output-junit build/junit-unit.xml --timeout 2400 ${ADDITIONAL_CTEST_ARGS}
+            ctest --test-dir build --no-tests=error -j -LE "Systest|DockerCompose" --output-on-failure --output-junit build/junit-unit.xml --timeout 2400 ${ADDITIONAL_CTEST_ARGS}
 
       - uses: ./.github/steps/run-in-container
         name: Run System Tests
@@ -101,7 +101,7 @@ jobs:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{ matrix.sanitizer }}
           run: |
             source .github/.env/test-${{ matrix.sanitizer }}.env
-            ctest --test-dir build -j -L Systest --output-on-failure --output-junit build/junit-systest.xml --timeout 4800 ${ADDITIONAL_CTEST_ARGS}
+            ctest --test-dir build --no-tests=error -j -L Systest --output-on-failure --output-junit build/junit-systest.xml --timeout 4800 ${ADDITIONAL_CTEST_ARGS}
 
       - uses: ./.github/steps/run-in-container
         name: Run Docker Tests
@@ -110,7 +110,7 @@ jobs:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{ matrix.sanitizer }}
           run: |
             source .github/.env/test-${{ matrix.sanitizer }}.env
-            ctest --test-dir build -j 1 -L DockerCompose --output-on-failure --output-junit build/junit-docker.xml --timeout 2400 ${ADDITIONAL_CTEST_ARGS}
+            ctest --test-dir build --no-tests=error -j 1 -L DockerCompose --output-on-failure --output-junit build/junit-docker.xml --timeout 2400 ${ADDITIONAL_CTEST_ARGS}
 
       - name: Upload Test Logs on Failure
         uses: actions/upload-artifact@v5
@@ -170,7 +170,7 @@ jobs:
               cmake -GNinja -B build -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DNES_LOG_LEVEL=$LOG_LEVEL
               sccache --show-stats
               cmake --build build --target nes-common-tests -j -- -k 0
-              ctest --test-dir build -j --output-on-failure --output-junit build/junit-log-${{ matrix.build_type }}-${LOG_LEVEL}.xml --timeout 600 -R "testLogLevel"
+              ctest --test-dir build --no-tests=error -j --output-on-failure --output-junit build/junit-log-${{ matrix.build_type }}-${LOG_LEVEL}.xml --timeout 600 -R "testLogLevel"
               cmake --build build --target clean
             done
             sccache --show-stats
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -42,6 +42,13 @@ jobs:
         uses: actions/checkout@v5
         with:
           ref: ${{ inputs.head_sha }}
+      - uses: ./.github/steps/prepare-github
+        name: Preparing Github Runners
+        with:
+          images: |
+            nebulastream/nes-ci:${{ inputs.dev_image_tag }}
+          docker_username: ${{ secrets.DOCKER_USER_NAME }}
+          docker_password: ${{ secrets.DOCKER_SECRET }}
       - name: Build coverage and export as lcov
         uses: ./.github/steps/run-in-container
         with:
--- a/.github/workflows/code_coverage.yml
+++ b/.github/workflows/code_coverage.yml
@@ -24,19 +24,13 @@ on:
 
 jobs:
   build-code-coverage-pr-branch:
-    container:
-      image: nebulastream/nes-ci:${{ inputs.dev_image_tag }}
-      volumes:
-        - /var/run/docker.sock:/var/run/docker.sock
-      env:
-        SCCACHE_BUCKET: ${{ secrets.SCCACHE_CACHE_BUCKET }}
-        SCCACHE_ENDPOINT: ${{ secrets.SCCACHE_CACHE_ENDPOINT }}
-        SCCACHE_REGION: auto
-        SCCACHE_S3_USE_SSL: "true"
-        AWS_ACCESS_KEY_ID: ${{ secrets.SCCACHE_CACHE_ACCESS_KEY }}
-        AWS_SECRET_ACCESS_KEY: ${{ secrets.SCCACHE_CACHE_SECRET_KEY }}
-      # TODO #401 Investigate rootless docker containers
-      options: --user root
+    env:
+      SCCACHE_BUCKET: ${{ secrets.SCCACHE_CACHE_BUCKET }}
+      SCCACHE_ENDPOINT: ${{ secrets.SCCACHE_CACHE_ENDPOINT }}
+      SCCACHE_REGION: auto
+      SCCACHE_S3_USE_SSL: "true"
+      AWS_ACCESS_KEY_ID: ${{ secrets.SCCACHE_CACHE_ACCESS_KEY }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.SCCACHE_CACHE_SECRET_KEY }}
     timeout-minutes: 90
     runs-on: [ self-hosted, linux, Build, "${{matrix.platform}}" ]
     strategy:
@@ -49,10 +43,13 @@ jobs:
         with:
           ref: ${{ inputs.head_sha }}
       - name: Build coverage and export as lcov
-        run: |
-          cmake -DCODE_COVERAGE=ON -B build
-          cmake --build build -j -- -k 0
-          cmake --build build --target ccov-all-export
+        uses: ./.github/steps/run-in-container
+        with:
+          image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}
+          run: |
+            cmake -DCODE_COVERAGE=ON -DENABLE_DOCKER_TESTS=ON -B build
+            cmake --build build -j -- -k 0
+            cmake --build build --target ccov-all-export
       - uses: codecov/codecov-action@v7
         with:
           slug: nebulastream/nebulastream
--- a/.github/workflows/large_test.yml
+++ b/.github/workflows/large_test.yml
@@ -140,7 +140,7 @@ jobs:
             cmake --build build --target systest nes-single-node-worker -j
             sccache --show-stats
             sccache --stop-server
-            ctest --test-dir build --output-on-failure --timeout 2400 -R systest-remote-test
+            ctest --test-dir build --no-tests=error --output-on-failure --timeout 2400 -R systest-remote-test
       - name: Upload Test Logs on Failure
         uses: actions/upload-artifact@v5
         if: failure() && !github.event.act
--- a/.github/workflows/large_test.yml
+++ b/.github/workflows/large_test.yml
@@ -39,15 +39,6 @@ on:
 jobs:
   build-and-test-linux:
     name: "Run large scale tests on ${{matrix.arch}} with ${{matrix.stdlib}} in ${{ matrix.build_type }}"
-    container:
-      image: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{ matrix.sanitizer }}
-      volumes:
-        - test-file-cache:/test-file-cache
-        - /var/run/docker.sock:/var/run/docker.sock
-      env:
-        MOLD_JOBS: 1
-      # TODO #401 Investigate rootless docker containers
-      options: --user root
     timeout-minutes: 40
     runs-on: [ self-hosted, linux, Build, "${{matrix.arch}}" ]
     strategy:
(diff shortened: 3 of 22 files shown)


greenBene (approved):
Thank you for implementing this! :) I think it makes the whole data handling easier to reason about. I also appreciate that you directly updated the documentation!  🙌

.github/workflows/build_and_test.yml
```diff
@@ -217,7 +217,7 @@ jobs:
           run: |
             source .github/.env/test-${{ matrix.sanitizer }}.env
             rm -rf build
-            cmake -GNinja -B build -DUSE_SANITIZER=${{ matrix.sanitizer }} -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DNES_LOG_LEVEL=DEBUG
+            cmake -GNinja -B build -DUSE_SANITIZER=${{ matrix.sanitizer }} -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DNES_LOG_LEVEL=DEBUG -DENABLE_DOCKER_TESTS=ON
```
keyseven123: I think we now need to add this flag to the `pr_matrix.yml`.

.github/workflows/large_test.yml
```diff
@@ -137,19 +127,20 @@ jobs:
...
-          cmake -GNinja -B build -DUSE_SANITIZER=${{ matrix.sanitizer }} -DCMAKE_BUILD_TYPE=RelWithDebInfo -DENABLE_LARGE_TESTS=ON -DExternalData_OBJECT_STORES=/test-file-cache -DLARGE_TEST_TOTAL_MEMORY_BYTES=76911476736
-      - name: Build NebulaStream
-        run: |
-          sccache --show-stats
-          cmake --build build --target systest nes-single-node-worker -j
-          sccache --show-stats
-          sccache --stop-server
-      - name: Run systest remote test
-        run: ctest --test-dir build --output-on-failure --timeout 2400 -R systest-remote-test
+      - name: Configure, build, and run systest remote test
```
keyseven123, line 130: Not sure if this should work. We run `build-and-test-linux`job with ´--user root`. So I assume tht this configure will just fail.

.github/workflows/code_coverage.yml
```diff
@@ -49,10 +43,13 @@ jobs:
         with:
           ref: ${{ inputs.head_sha }}
       - name: Build coverage and export as lcov
-        run: |
-          cmake -DCODE_COVERAGE=ON -B build
-          cmake --build build -j -- -k 0
-          cmake --build build --target ccov-all-export
+        uses: ./.github/steps/run-in-container
```
keyseven123, line 53: This does not pull a new image so one runners that have this imag enot cached, it iwll lead ot an error.

nes-systests/systest/CMakeLists.txt
```diff
@@ -37,6 +37,58 @@ ExternalData_Expand_Arguments(
...
+# the large case in the remote Compose suite.
+set(NES_DOCKER_EXTERNAL_DATA_VOLUME "" CACHE STRING
+        "Docker volume exposing ExternalData_OBJECT_STORES to Docker-based large systests")
+set(ENABLE_DOCKER_LARGE_TESTS OFF)
+set(NES_DOCKER_EXTERNAL_DATA_MOUNT_TYPE "")
+set(NES_DOCKER_EXTERNAL_DATA_SOURCE "")
+set(NES_DOCKER_EXTERNAL_DATA_TARGET "")
+if (ENABLE_LARGE_TESTS AND ENABLE_DOCKER_TESTS)
+    list(LENGTH ExternalData_OBJECT_STORES _external_data_store_count)
+    if (_external_data_store_count EQUAL 0)
```
keyseven123, line 52: The build-tree store is only reachable in-container because `TESTCONFIG_DIR` happens to cover it when `<build>` sits inside the source tree. With an out-of-source build dir the `/data` symlinks into `<build>/ExternalData/Objects` dangle — the old `tar -ch` dereference is gone. Worth mounting `<build>/ExternalData` explicitly instead of relying on the checkout bind.

scripts/testing/distributed_bats_lib.bash
```diff
@@ -301,27 +302,11 @@ nes_distributed_setup() {
...
-  local volume_host_container
-  volume_host_container=$(docker run -d --rm -v $TEST_VOLUME:/data alpine sleep infinite)
-  docker cp $volume_host_container:/data/. .
-  docker stop -t0 $volume_host_container
+  export TEST_DIR="$TMP_DIR"
 }
 
 nes_distributed_teardown() {
-  sync_workdir || true
   docker compose down -v || true
```
keyseven123, line 323: Workers now write into the bind-mounted `$TMP_DIR` as root, and `working_dir: /workdir/$HOST_NAME` makes Docker create root-owned dirs there. Nothing cleans `$TMP_DIR` anymore (the old path copied out via `docker cp` as the runner uid), so the next `rm -rf build` on a persistent runner hits EPERM. Add `rm -rf "$TMP_DIR"` to teardown.

nes-plugins/Sources/MQTTSource/tests/util/create_compose.sh
```diff
@@ -94,7 +94,9 @@ services:
     working_dir: /workdir
     command: ["sleep", "infinity"]
     volumes:
-      - $TEST_VOLUME:/workdir
+      - type: bind
+        source: "$TEST_DIR"
```
keyseven123, line 103: Same as MQTTSink: `$TEST_DIR` is used as a bind source without the `ERROR: TEST_DIR is not set` guard the other generators got.

cmake/CheckTestUtilities.cmake
```diff
@@ -10,6 +10,45 @@
...
 # limitations under the License.
 
+# Docker commands in the development container use the host daemon, which cannot
+# necessarily access paths from the container's filesystem. Write a unique token
+# into the path, identify the development image through the container hostname,
+# then start a sibling with the requested mount and verify that it reads the token.
+# A native host uses the local filesystem directly and needs no probe.
+function(verify_host_path_is_accessible_from_docker PATH MOUNT_SOURCE RESULT_VARIABLE)
+    set(${RESULT_VARIABLE} TRUE PARENT_SCOPE)
+    if (NOT EXISTS "/.dockerenv")
```
keyseven123, line 23: `/.dockerenv` misses Podman/rootless, which uses `/run/.containerenv`. There the probe is skipped and returns TRUE, so the new hard guard passes and the failure resurfaces later as an unexplained Compose bind-mount error at test time.

.github/workflows/large_test.yml
```diff
@@ -137,19 +127,20 @@ jobs:
...
+        uses: ./.github/steps/run-in-container
```
keyseven123, line 131: The removed `container:` block also carried `MOLD_JOBS: 1` (and `ccache:/ccache` + `CCACHE_DIR`). `run-in-container` forwards only `SCCACHE_*`/`AWS_*`, so mold link serialization is lost on this large-memory RelWithDebInfo build — `build-and-test-linux` still keeps it.

nes-plugins/Sinks/MQTTSink/tests/util/create_compose.sh
```diff
@@ -95,7 +95,9 @@ services:
     working_dir: /workdir
     command: ["sleep", "infinity"]
     volumes:
-      - $TEST_VOLUME:/workdir
+      - type: bind
+        source: "$TEST_DIR"
```
keyseven123, line 104: `$TEST_DIR` is now a bind source here, but this generator skipped the `ERROR: TEST_DIR is not set` guard added to the cli/repl/remote-test versions. Unset yields `source: ""` and an opaque Compose validation error.

keyseven123 (changes requested):

.github/workflows/large_test.yml
yschroeder97: Three separate things were lost when this job moved off `container:` onto `run-in-container`. The first two look blocking.

**1. The image is never pulled.** Same as in `code_coverage.yml`: `run-in-container` runs `docker run --pull=never` (`.github/steps/run-in-container/action.yml:51`) and this job has no `prepare-github` step. The suffixed tag is pulled by the sibling `build-and-test-linux` job, but that one runs on `[self-hosted, linux, Build, x64]` while this one requires the `large-memory` label, so a warm cache on the other pool does not help.

**2. The job lost `--user root` but kept the root-owned volume.** The old `container:` block ran with `options: --user root`. `run-in-container` hardcodes `-u "$(id -u):$(id -g)"` (`.github/steps/run-in-container/action.yml:57`), so this job now runs unprivileged against `test-file-cache`, which is a Docker named volume and therefore root-owned 0755. The sibling `build-and-test-linux` job at line 42 still runs as root with the same volume mounted, so it is what populates the store. Two failures follow, in order:

- Configure aborts. `nes-systests/systest/CMakeLists.txt:60` creates the store directory and then calls `verify_host_path_is_accessible_from_docker`, whose first act is `file(WRITE "${PATH}/docker-bind-probe-...")` at `cmake/CheckTestUtilities.cmake:29`. On the volume route only the mount source becomes the volume name; the probe path is still `/test-file-cache`. A non-root `file(WRITE)` there is a hard CMake error. This failure mode is new in this PR.
- Even without the probe, ExternalData downloads objects into the first entry of `ExternalData_OBJECT_STORES`, which is the unwritable `/test-file-cache` (line 139), so building `systest` cannot materialize the large test data.

Options: give the action a way to pass `--user root` for this job, pre-chown the volume, or point `ExternalData_OBJECT_STORES` at a writable path under `$(pwd)` and expose the shared cache as a read-only second store. The last one conflicts with the at-most-one-store rule added at `nes-systests/systest/CMakeLists.txt:84`, so it may need a rethink there.

**3. `MOLD_JOBS: 1` is gone (non-blocking).** The removed `container:` block also carried `env: MOLD_JOBS: 1` and `CCACHE_DIR: /ccache`. `run-in-container` forwards only `SCCACHE_*` and `AWS_*`, so `MOLD_JOBS` is now unset for a `RelWithDebInfo` link of `systest` and `nes-single-node-worker`. That cap is still deliberately set on the sibling job at line 48 and in `smoke-tests.yml:33` to bound link-time memory, which makes its disappearance here look accidental rather than intended. The 40-minute timeout leaves little room for an OOM-killed link and retry.

.github/workflows/code_coverage.yml
yschroeder97: Blocking: the image is never pulled.

This job dropped its `container:` block, but `run-in-container` runs `docker run --pull=never` (`.github/steps/run-in-container/action.yml:51`) and never pulls. For a `container:` job the runner pulled the image implicitly; now nothing does.

Every other consumer of this action pairs it with `prepare-github`, which pulls with retry before the first `run-in-container` step: `build_and_test.yml:59`, `clang_tidy_diff.yml:62`, `endless-mode.yml:47`, `nightly_slt_corpus.yml:48`, and an explicit `docker pull` in `create_executable_images.yml:59`.

This job is the more exposed of the two, because it asks for the unsuffixed tag `nebulastream/nes-ci:${{ inputs.dev_image_tag }}`, while every `prepare-github` call pulls the suffixed variant. `pr_code.yml:48` gives the job only `needs: [get-dev-images, smoke-tests]`, so nothing in its path guarantees the tag is present on the runner it lands on. It fails with "Unable to find image locally" on the first run after a new dev-image tag, on a fresh runner, or after an image prune.

Adding a `prepare-github` step before this one, with the same tag, would close it.

.github/workflows/large_test.yml
yschroeder97: Three separate things were lost when this job moved off `container:` onto `run-in-container`. The first two look blocking.

**1. The image is never pulled.** Same as in `code_coverage.yml`: `run-in-container` runs `docker run --pull=never` (`.github/steps/run-in-container/action.yml:51`) and this job has no `prepare-github` step. The suffixed tag is pulled by the sibling `build-and-test-linux` job, but that one runs on `[self-hosted, linux, Build, x64]` while this one requires the `large-memory` label, so a warm cache on the other pool does not help.

**2. The job lost `--user root` but kept the root-owned volume.** The old `container:` block ran with `options: --user root`. `run-in-container` hardcodes `-u "$(id -u):$(id -g)"` (`.github/steps/run-in-container/action.yml:57`), so this job now runs unprivileged against `test-file-cache`, which is a Docker named volume and therefore root-owned 0755. The sibling `build-and-test-linux` job at line 42 still runs as root with the same volume mounted, so it is what populates the store. Two failures follow, in order:

- Configure aborts. `nes-systests/systest/CMakeLists.txt:60` creates the store directory and then calls `verify_host_path_is_accessible_from_docker`, whose first act is `file(WRITE "${PATH}/docker-bind-probe-...")` at `cmake/CheckTestUtilities.cmake:29`. On the volume route only the mount source becomes the volume name; the probe path is still `/test-file-cache`. A non-root `file(WRITE)` there is a hard CMake error. This failure mode is new in this PR.
- Even without the probe, ExternalData downloads objects into the first entry of `ExternalData_OBJECT_STORES`, which is the unwritable `/test-file-cache` (line 139), so building `systest` cannot materialize the large test data.

Options: give the action a way to pass `--user root` for this job, pre-chown the volume, or point `ExternalData_OBJECT_STORES` at a writable path under `$(pwd)` and expose the shared cache as a read-only second store. The last one conflicts with the at-most-one-store rule added at `nes-systests/systest/CMakeLists.txt:84`, so it may need a rethink there.

**3. `MOLD_JOBS: 1` is gone (non-blocking).** The removed `container:` block also carried `env: MOLD_JOBS: 1` and `CCACHE_DIR: /ccache`. `run-in-container` forwards only `SCCACHE_*` and `AWS_*`, so `MOLD_JOBS` is now unset for a `RelWithDebInfo` link of `systest` and `nes-single-node-worker`. That cap is still deliberately set on the sibling job at line 48 and in `smoke-tests.yml:33` to bound link-time memory, which makes its disappearance here look accidental rather than intended. The 40-minute timeout leaves little room for an OOM-killed link and retry.

nes-systests/systest/CMakeLists.txt
yschroeder97: This branch enables `ENABLE_DOCKER_LARGE_TESTS` with no mount at all, on the grounds that the build-tree store was already verified by `CheckTestUtilities.cmake`. That check establishes host and container path identity, not reachability from inside the compose containers.

For a store under `CMAKE_BINARY_DIR`, CMake's `_ExternalData_link_or_copy` emits a relative symlink, because the path from `relative_top` does not start with `../`. The testdata directory is mounted at `/data` in the containers, not at its build path, so `/data/large/<file>` resolves through `../../../ExternalData/Objects/...` and dangles. Anyone configuring with `-DENABLE_DOCKER_TESTS=ON -DENABLE_LARGE_TESTS=ON` and no explicit `ExternalData_OBJECT_STORES` gets a large case that runs against broken input files rather than skipping.

The CI path is unaffected, since `-DExternalData_OBJECT_STORES=/test-file-cache` yields absolute symlinks.

cmake/CheckTestUtilities.cmake
```diff
@@ -44,14 +84,28 @@ endif ()
...
-            "  Set -DENABLE_DOCKER_TESTS=OFF to suppress this warning\n"
-            "Docker tests will be automatically disabled."
+            "  For host system: Ensure docker is installed and running"
         )
-        set(ENABLE_DOCKER_TESTS OFF CACHE BOOL "Runs testcases that require docker" FORCE)
+    elseif (EXISTS "/.dockerenv")
+        # Docker commands use the host daemon through its socket. Prove that
+        # the build directory is mounted at the same absolute path on the host
+        # before registering tests that use it as a Compose bind source.
+        verify_host_path_is_accessible_from_docker(
```
yschroeder97: Only `CMAKE_BINARY_DIR` is probed, but the remote suite also binds the source tree: `nes-systests/systest/remote-test/distributed.bats:36` exports `TESTCONFIG_DIR=$(realpath "$NES_DIR")` and `create_compose.sh:112` uses it as a bind source.

With an out-of-source build where only the build directory is mounted at a matching path (`cmake -S /src/repo -B /work/build`, `-v /work:/work`), the probe passes, then Docker silently creates an empty host directory for the unresolvable source bind and every container sees an empty `$NES_DIR`. That surfaces as an opaque systest failure, which is the class of failure this PR sets out to remove. Probing `CMAKE_SOURCE_DIR` as well would keep the guarantee complete.

nes-systests/systest/remote-test/create_compose.sh
```diff
@@ -88,9 +104,16 @@ services:
...
     working_dir: $CONTAINER_WORKDIR
     volumes:
-      - $TESTDATA_VOLUME:/data
-      - $TESTCONFIG_VOLUME:$NES_DIR
-      - $TEST_VOLUME:$CONTAINER_WORKDIR
+      - type: bind
+        source: "$TESTDATA_DIR"
+        target: /data
+$TESTDATA_CACHE_MOUNT
+      - type: bind
```
yschroeder97: `TESTCONFIG_DIR` is the real repository checkout, and it is bind-mounted read-write here and again at line 151 into every worker container, all of which run as root. Previously this was a throwaway copy in a named volume, so stray writes were discarded with the volume.

Anything a worker or the systest binary writes under `$NES_DIR` now lands root-owned in the checkout. On a self-hosted runner the next `actions/checkout` cleanup fails with permission denied, and locally it leaves root-owned files in the working tree. The mount is only read from (`--clusterConfig $NES_DIR/...` and test discovery), so `read_only: true` on both entries would match the intent.

nes-systests/systest/CMakeLists.txt
ls-1801: Out-of-tree builds are currently not supported for the Docker Compose-based tests, and our documentation does not claim otherwise. These tests require the build directory to be available to the host Docker daemon at the same path.

Rather than adding another mount for this unsupported configuration, we will make the limitation explicit. The CMake accessibility check will disable the Docker tests when the build directory is unavailable and report a clearer message explaining that Docker Compose cannot access an out-of-tree build directory.

.github/workflows/large_test.yml
ls-1801: `MOLD_JOBS=1` is already defined by the CI development image, so `docker run` inherits it without requiring explicit forwarding. The removed `ccache` volume and `CCACHE_DIR` are no longer needed because this workflow uses sccache. Therefore, no build-cache or mold serialization behavior is lost here.

cmake/CheckTestUtilities.cmake
ls-1801: Our documented and supported setup uses Docker. Podman and other container runtimes are currently outside the scope of this change, so detecting `/run/.containerenv` is not required here. If Podman support is requested, we can add and test it comprehensively in a follow-up.

scripts/testing/distributed_bats_lib.bash
ls-1801: Keeping `$TMP_DIR` is intentional so failed-test artifacts remain available for debugging. The documented setup runs containers either as the locally installed user or as container root under rootless Docker, where root is mapped to the invoking host user. In both supported configurations, the generated files remain removable by the user running the build, so a later `rm -rf build` should not fail with `EPERM`.

Adding `rm -rf "$TMP_DIR"` would remove the debugging artifacts we deliberately retain, so I do not think a teardown change is needed here.

nes-plugins/Sinks/MQTTSink/tests/util/create_compose.sh
ls-1801: added the check

nes-plugins/Sources/MQTTSource/tests/util/create_compose.sh
ls-1801: added the check

.github/workflows/large_test.yml
ls-1801: Good point. I changed `build-and-test-linux` to use the same `run-in-container` action as the remote-test job. It now configures and builds as the runner UID/GID instead of using a job-level container with `--user root`, avoiding a root-owned build directory on the persistent runner.

.github/workflows/build_and_test.yml
ls-1801: This is now covered by `.github/ci/matrix.yaml`: `ENABLE_DOCKER_TESTS=ON` is included in `default_cmake_flags`, so it is applied to the PR matrix as well as the nightly matrix.

.github/workflows/code_coverage.yml
ls-1801: Good catch. I added the `prepare-github` step before `run-in-container`, including the coverage build image and Docker Hub credentials. The image is therefore pulled before the action invokes it with `--pull=never`.

keyseven123 (approved):

PR #1663, opened 2026-06-08
added reflect to reflectionContext

gustafsa:
## Purpose of the Change and Brief Change Log

Following up on the recent `Unreflector` changes (#1412), this PR applies the same context-aware architecture to the serialization side (`Reflector`).

For a detailed explanation of why our IR requires a `ReflectionContext` and why we cannot rely on context-less `reflectcpp` calls, please refer to the previous `Unreflector` PR (#1412).

Changes in this PR:

* Replaced all global, context-less `reflect(...)` calls with `context.reflect(...)` across the entire codebase.
* Updated all `Reflector` specializations to mandate a `const ReflectionContext&` as their second argument.
* Moved template specializations for wrappers (like `TypedLogicalFunction`) strictly into headers to fix linker issues.
* Ensured that our custom context is correctly propagated through the entire serialization chain without falling back to default `reflectcpp` behavior.

## What components does this pull request potentially affect?

`nes-common`, `nes-logical-operators`, `nes-frontend`

PR #1847, opened 2026-08-06
ci: add weekly LLM-based docs drift check for merged PRs

### LINKED ISSUE #1838: [TASK] Create a documentation CI workflow that uses LLMs to analyze PR diffs
To keep the documentation up to date, we should add a CI workflow that on each PR feeds the diffs and current documentation to a locally hosted LLM. On detecting a mismatch between the implementation and the documentation, the CI should notify a maintainer that is responsible for the documentation so that the documentation gets manually updated similar to a hotfix

oczerwia-2010:
Adds a workflow that runs weekly and checks every PR merged into the default branch in the last 7 days. It only cares whether code changes have made docs/ stale: PRs that only touch docs/ are skipped outright.

For each remaining PR: a locally-hosted LLM (Qwen2.5-Coder, no external API key) reads the diff and produces a short semantic summary of what changed, or decides it's purely internal and stops there. That summary is embedded and used to query a RAG index built fresh every run from every docs/ section. The retrieved doc excerpts go back to the LLM along with the diff, which decides whether the diff makes any specific statement in those excerpts incorrect or outdated, and if so, explains exactly which statement and why. 

## Verifying this change
This change is tested by
- Performed 5 non trivial changes with varying impact on the documentation and optimized for Recall

This PR closes Issue #1838

### DIFF
--- a/.github/workflows/docs-diff-check.yml
+++ b/.github/workflows/docs-diff-check.yml
@@ -1,0 +1,75 @@
+name: Docs Drift Check
+description: |
+  Runs weekly (Sunday night, so results are ready Monday morning), checking
+  every PR merged in the last 7 days for whether it has made docs/ stale. A
+  PR touching docs/ itself is not skipped - that doesn't guarantee the docs
+  update was correct or complete for the change made.
+
+  For each PR, a locally-hosted LLM (Qwen2.5-Coder, served from the
+  cached image built by build-docs-drift-llm-image.yml) summarizes what the
+  diff changes; that summary is embedded (mxbai-embed-large, same image) and
+  matched against a RAG index built fresh this run from every docs/ section
+  (fresh because jobs land on a random self-hosted worker each time, so
+  there's nothing to cache between runs); the LLM then cross-references the
+  diff against the retrieved excerpts and flags it, with a detailed reason,
+  only if a specific documented statement has gone stale. Flags are posted to
+  the same Zulip "ci" stream/bot used by nightly.yml.
+
+on:
+  schedule:
+    - cron: '0 22 * * 0'
+  workflow_dispatch: {}
+
+concurrency:
+  group: docs-drift-check
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  pull-requests: read
+  packages: read
+
+jobs:
+  check-docs-drift:
+    runs-on: [ self-hosted, linux, Build, x64 ]
+    # Runtime is dominated by one generation call per PR (~230-450s on
+    # CPU-only inference in local testing; the model/embeddings are already
+    # cached, only per-PR generation calls scale with batch size), so this
+    # leaves headroom for a week with a larger-than-usual number of merged PRs.
+    timeout-minutes: 240
+    container:
+      image: ghcr.io/${{ github.repository_owner }}/docs-drift-llm:latest
+      credentials:
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+    env:
+      GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      DOCS_ROOT: ${{ github.workspace }}/docs
+      LLM_API_BASE: http://localhost:11434/v1
+      LLM_MODEL: qwen2.5-coder:7b
+      EMBED_MODEL: mxbai-embed-large
+      ZULIP_SITE: https://nebulastream.zulipchat.com
+      ZULIP_BOT_EMAIL: pr-reminder-bot@nebulastream.zulipchat.com
+      ZULIP_BOT_API_KEY: ${{ secrets.PRBOT_API_KEY }}
+      ZULIP_STREAM: ci
+      ZULIP_TOPIC: docs drift warnings
+      ZULIP_NOTIFY_USERS: oczerwia-2010,keyseven123
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - name: Start local LLM
+        run: |
+          nohup ollama serve > /tmp/ollama.log 2>&1 &
+          for i in $(seq 1 60); do
+            ollama list >/dev/null 2>&1 && exit 0
+            sleep 2
+          done
+          echo "::error::local LLM did not become ready in time"
+          echo "::group::ollama serve log"
+          cat /tmp/ollama.log || true
+          echo "::endgroup::"
+          exit 1
+
+      - name: Run docs drift check
+        run: python3 scripts/docs-diff-check/check_docs_impact.py
--- a/docker/docs-drift-llm/Dockerfile
+++ b/docker/docs-drift-llm/Dockerfile
@@ -1,0 +1,50 @@
+# Ollama image with both the docs-drift-check generation model and the
+# embedding model pre-pulled into the image layer, so the CI job never has to
+# download either at run time - only pull this image (cached by the container
+# registry / runner) and start it. The docs corpus itself is still embedded
+# fresh on every run, since jobs land on a random self-hosted worker each
+# time and there's no persistent cache to rely on.
+#
+# This image is also used as the docs-diff-check.yml job's own container
+# (not just a services: sidecar) - ollama serve is the ENTRYPOINT, which
+# doubles as the long-running process that keeps a GitHub Actions job
+# container alive while steps are exec'd into it. gh/git/python3 are here
+# so the job's steps (checkout, gh pr list/diff, the check script) can run
+# in the same container instead of depending on tooling the self-hosted
+# runner host happens to have.
+FROM ollama/ollama:latest
+
+ARG MODEL=qwen2.5-coder:7b
+ARG EMBED_MODEL=mxbai-embed-large
+ENV OLLAMA_MODEL=${MODEL}
+ENV OLLAMA_EMBED_MODEL=${EMBED_MODEL}
+# Ollama stores/looks up models under $HOME/.ollama/models by default. HOME
+# is /root at build time (where the models below get pulled to), but GitHub
+# Actions job containers force HOME=/github/home at runtime - without this,
+# ollama looks in the wrong, empty directory and reports the model missing
+# even though it's sitting right here in the image.
+ENV OLLAMA_MODELS=/root/.ollama/models
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends curl ca-certificates gnupg git python3 && \
+    curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg -o /usr/share/keyrings/githubcli-archive-keyring.gpg && \
+    chmod go+r /usr/share/keyrings/githubcli-archive-keyring.gpg && \
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" > /etc/apt/sources.list.d/github-cli.list && \
+    apt-get update && \
+    apt-get install -y --no-install-recommends gh && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN ollama serve & \
+    pid=$! && \
+    for i in $(seq 1 60); do \
+        ollama list >/dev/null 2>&1 && break; \
+        kill -0 $pid 2>/dev/null || { echo "ollama serve died on startup" >&2; exit 1; }; \
+        sleep 1; \
+    done && \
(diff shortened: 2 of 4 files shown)


ls-1801 (changes requested):
I am a bit lost, does that even work? can we test that?

scripts/docs-diff-check/check_docs_impact.py
```diff
@@ -0,0 +1,348 @@
...
+        "pr", "list",
+        "--search", f"is:pr is:merged sort:created-asc merged:>={since}",
+        "--json", "number,title,url,author,mergedAt,files",
+        "--limit", "200",
+    )
+    return json.loads(out)
+
+
+def touches_docs(pr):
+    return any(f["path"].startswith(DOCS_PREFIX) for f in pr.get("files", []))
```
keyseven123: `gh pr list --json files` returns only the first 100 files per PR (GraphQL `files(first: 100)`); `--limit 200` bounds the PR count, not the file count. A PR that did update `docs/` but changed more than 100 files falls through this gate and gets a Zulip message publicly stating it "was merged without touching `docs/`", with `@**author**` attached.

Verified against this repo: PR #1663 has 180 changed files, `--json files` returns 100.

Suggest fetching the file list per candidate PR (`gh pr view <n> --json files` paginates no better — `gh api --paginate repos/{owner}/{repo}/pulls/{n}/files` does), or deriving it from the diff you already fetch.

scripts/docs-diff-check/check_docs_impact.py
```diff
@@ -0,0 +1,348 @@
...
+    prs = list_prs_merged_last_week()
+    print(f"Found {len(prs)} PR(s) merged in the last 7 days")
+
+    docs_index = build_docs_header_index()
+    print(f"Indexed {len(docs_index)} doc section(s)")
+    chunk_vectors = embed_texts([get_excerpt_text(h) for h in docs_index]) if docs_index else []
+    print(f"Embedded {len(chunk_vectors)} doc section(s) with {EMBED_MODEL}")
+
+    flagged = 0
+    for pr in prs:
```
keyseven123, line 667: Nothing in this loop is guarded: `gh()` runs with `check=True`, and `call_llm` / `embed_texts` / `post_drift_warning` all let `HTTPError`/`URLError` escape. One PR whose diff GitHub refuses to render (406 on very large or binary-heavy diffs), or one transient Ollama/Zulip blip, raises straight out of `main` — every PR after it in the week is never examined, and the only signal is a red weekly cron.

A `try/except Exception` around the loop body with `::warning::` and `continue` keeps the rest of the week's PRs covered.

scripts/docs-diff-check/check_docs_impact.py
```diff
@@ -0,0 +1,348 @@
...
+            path = os.path.relpath(os.path.join(root, fname), DOCS_ROOT)
+            path = os.path.join(DOCS_PREFIX.rstrip("/"), path).replace(os.sep, "/")
+            with open(os.path.join(root, fname), encoding="utf-8", errors="replace") as f:
+                lines = f.readlines()
+
+            headers = []
+            for i, line in enumerate(lines, start=1):
+                stripped = line.lstrip("#")
+                level = len(line) - len(stripped)
+                if 1 <= level <= 6 and stripped.startswith(" "):
```
keyseven123, line 305: This treats every `#`-prefixed line as a header, including `#` comments inside ``` fences. Running this exact loop over the current `docs/` yields 551 sections, 76 of them inside code fences.

Concrete case — `docs/guide/nebulastream-frontend.md`: the bash block contains `# Embedded mode (single-node)`, which is indexed as an H1. The real `## NES-REPL` section is then cut off at line 85 so its body is never embedded, and a section titled `Distributed mode (multi-node)` (lines 89-297) that does not exist in the docs enters the index — and can end up named in a Zulip warning.

Tracking an `in_fence` flag toggled on lines starting with ``` or ~~~ fixes both halves.

.github/workflows/docs-diff-check.yml
```diff
@@ -0,0 +1,64 @@
...
+  contents: read
+  pull-requests: read
+
+jobs:
+  check-docs-drift:
+    runs-on: [ self-hosted ]
+    timeout-minutes: 120
+    services:
+      llm:
+        image: ghcr.io/${{ github.repository_owner }}/docs-drift-llm:latest
```
keyseven123: This service pulls from GHCR anonymously: no `credentials:` block here, and the job's `permissions:` grant only `contents: read` / `pull-requests: read`. Container packages published by `build-docs-drift-llm-image.yml` with `secrets.GITHUB_TOKEN` are **private** by default until someone flips visibility in the package settings, so the pull returns `denied` and the service never starts. The failure then surfaces only as "local LLM did not become ready in time" after the wait loop burns its 120 s — misleading, on a weekly cron nobody watches.

Adding `packages: read` to `permissions:` plus a `credentials:` block (`username: ${{ github.actor }}`, `password: ${{ secrets.GITHUB_TOKEN }}`) makes it work regardless of package visibility.

.github/workflows/docs-diff-check.yml
```diff
@@ -0,0 +1,64 @@
...
+  group: docs-drift-check
+  cancel-in-progress: false
+
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  check-docs-drift:
+    runs-on: [ self-hosted ]
```
keyseven123: The runner pool here is mixed — `.github/ci/pr_matrix.yaml` defines both `self-hosted-x64: [self-hosted, linux, Build, x64]` and `self-hosted-arm64: […, arm64]` — and every other self-hosted job in this repo pins the arch. This job and `build-docs-drift-llm-image.yml:26` both use the bare label, and the image build passes no `platforms:`, so the image is single-arch for whichever worker happened to build it.

When the two land on different arches the service container dies with `exec /bin/ollama: exec format error`, and the job reports it 120 s later as "local LLM did not become ready in time" — failing on a random subset of weeks.

Pin both jobs to the same arch, or build multi-arch.

scripts/docs-diff-check/check_docs_impact.py
```diff
@@ -0,0 +1,348 @@
...
+def gh(*args):
+    result = subprocess.run(
+        ["gh", *args, "--repo", REPO],
+        capture_output=True, text=True, check=True,
+    )
+    return result.stdout
+
+
+def list_prs_merged_last_week():
+    since = (datetime.now(timezone.utc) - timedelta(days=7)).strftime("%Y-%m-%d")
```
keyseven123: `merged:>=YYYY-MM-DD` resolves to 00:00 UTC of that date, but the cron fires at 22:00 UTC (`docs-diff-check.yml:20`). Every PR merged between 00:00 and 22:00 on a cron day therefore falls inside two consecutive windows, and since nothing dedupes against already-posted PRs, `post_drift_warning` sends a second identical message — the topic accumulates repeat warnings for the same PR.

`strftime("%Y-%m-%dT%H:%M:%SZ")` gives GitHub the full timestamp and closes the overlap.

docker/docs-drift-llm/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+FROM ollama/ollama:latest
+
+ARG MODEL=qwen2.5-coder:7b
+ARG EMBED_MODEL=mxbai-embed-large
+ENV OLLAMA_MODEL=${MODEL}
+ENV OLLAMA_EMBED_MODEL=${EMBED_MODEL}
+
+RUN ollama serve & \
+    pid=$! && \
+    until ollama list >/dev/null 2>&1; do sleep 1; done && \
```
keyseven123: `until ollama list >/dev/null 2>&1; do sleep 1; done` never terminates if `ollama serve` dies at startup (port already bound, non-writable `/root/.ollama`, base-image change) — `$pid` is captured but never checked. The image-build workflow sets no `timeout-minutes`, so a dead server pins a self-hosted runner until GitHub's 6-hour cap kills the job, and the only artifact is a cancelled run with no error text.

Either add a timeout here or to the CI job.

scripts/docs-diff-check/check_docs_impact.py
```diff
@@ -0,0 +1,348 @@
...
+    except json.JSONDecodeError:
+        print(f"::warning::could not parse LLM response for PR #{pr['number']}: {content}")
+        return False, "unparseable LLM response"
+    return bool(parsed.get("needs_docs_update")), parsed.get("reason", "")
+
+
+def post_drift_warning(pr, reason):
+    content = (
+        f"**Possible docs drift**: [#{pr['number']} {pr['title']}]({pr['url']}) "
+        f"by @**{pr['author']['login']}** was merged without touching `docs/`.\n"
```
keyseven123: For now, assign yourself and myself only.

scripts/docs-diff-check/check_docs_impact.py
```diff
@@ -0,0 +1,348 @@
...
+    )
+    return [h for h, _ in scored[:MAX_SELECTED_SECTIONS]]
+
+
+def cross_reference_with_docs(pr, diff, sections):
+    excerpt_parts = []
+    total = 0
+    for h in sections:
+        text = get_excerpt_text(h)
+        if total + len(text) > MAX_TOTAL_SECTION_CHARS:
```
keyseven123, line 451: Should this not be a continue? We are stopping here as soon as a section is larger than 1.5k chars

scripts/docs-diff-check/check_docs_impact.py
```diff
@@ -0,0 +1,348 @@
...
+    dot = sum(x * y for x, y in zip(a, b))
+    norm_a = math.sqrt(sum(x * x for x in a))
+    norm_b = math.sqrt(sum(y * y for y in b))
+    if norm_a == 0 or norm_b == 0:
+        return 0.0
+    return dot / (norm_a * norm_b)
+
+
+def select_relevant_sections_by_embedding(summary, index, chunk_vectors):
+    query_vector = embed_texts([summary])[0]
```
keyseven123, line 437: `summarize_diff_for_retrieval` returns `parsed.get("summary", "")`, so a reply like `{"user_facing": true}` with no summary reaches this line and `""` goes into `embed_texts`. Ollama's `/api/embed` rejects empty input, so `urlopen` raises inside the unguarded loop and the run dies at that PR; if a degenerate vector comes back instead, cosine similarity is 0.0 for all sections and the top-8 is arbitrary — the LLM then cross-references the diff against unrelated docs and can emit a confident but wrong warning.

Treating an empty summary as "not user-facing" (or falling back to the PR title) avoids both.

keyseven123 (changes requested):
On the branch `nls/testing/docs-drift-on-every-pr`, I have the commit ce84bac0b05451b7a7c055b7b206e73ba7c7e75d that enables this new workflow for every pr run for testing purposes.

scripts/docs-diff-check/check_docs_impact.py
oczerwia-2010: removed this as we look at the individual diffs later on in the code.

scripts/docs-diff-check/check_docs_impact.py
oczerwia-2010: Included guards with verbose error messages that pass the error message and stage of failure

scripts/docs-diff-check/check_docs_impact.py
oczerwia-2010: Included the proposed fence mechanism

.github/workflows/docs-diff-check.yml
oczerwia-2010: moved everything to a single container instead of having a serving container and a query container.

scripts/docs-diff-check/check_docs_impact.py
oczerwia-2010: included the proposed solution to prevent accidental overlap / blindspots

docker/docs-drift-llm/Dockerfile
oczerwia-2010: set different timeouts at different stages of the pipeline to catch errors in a more fine grained manner

scripts/docs-diff-check/check_docs_impact.py
oczerwia-2010: doubled MAX_TOTAL_SECTION_CHARS and removed the break to prevent early stopping

scripts/docs-diff-check/check_docs_impact.py
oczerwia-2010: have different error handling for malformed json output by the model and empty output. empty output is now skipped and treated as not user-facing.

PR #1848, opened 2026-08-06
Guard potentially throwing Nautilus invokes

ls-1801:
## Problem

On `main`, `nautilus::invoke` emits a direct call for every function pointer, regardless of its exception specification. If the invoked native function throws, the exception crosses the generated frame. The trace only contains the normal continuation, so cleanups after the invoke are not executed and traced destructors leak.

For example, consider a traced `val<StructResult>` whose writer throws:

```cpp
nautilus::val<StructResult> result;
nautilus::invoke(writeStructOrThrow, &result, nautilus::val<int32_t>{42});
const auto value = result.get(&StructResult::value);
*output = value;
```

The relevant trace on `main` has no failure edge:

```text
entry(output):
    struct = ALLOCA StructResult
    CALL construct(struct)
    CALL writeStructOrThrow(struct, 42)  // [M1] exception crosses the generated frame
    value = LOAD struct.value            // [M2] trace models only normal continuation
    STORE output, value
    CALL destruct(struct)                // [M3] skipped when the invoke throws
    RETURN
```

- **[M1]** The potentially throwing function is called directly.
- **[M2]** No generated control-flow edge represents failure.
- **[M3]** The cleanup is only reachable from the success path, so the destructor leaks when the call throws.

## Fix

Invokes are now guarded by default. A function opts out of guarding by declaring its function type `noexcept`:

```cpp
void guardedByDefault(StructResult*, int32_t);
void directCall(StructResult*, int32_t) noexcept;
```

For a function without `noexcept`, the generated code calls a `noexcept` proxy. The proxy catches the native exception, parks its `std::exception_ptr` in thread-local state, and reports failure through a trace-local stack holder. The generated failure branch exits the trace through normal C++ stack unwinding, which emits destructor calls before returning from the generated frame. The original exception is rethrown at the outermost native `ModuleFunction` boundary.

The throwing example becomes:

```text
entry(output):
    struct = ALLOCA StructResult
    holder = ALLOCA ExceptionHolder                         // [F1] one shared holder per traced function
    STORE holder.exception, nullptr
    CALL guardedRuntimeProxy(holder, writeStructOrThrow,    // [F2] proxy catches and parks exceptions
                             struct, 42)
    failed = LOAD holder.exception != nullptr
    BRANCH failed, invoke_failure, invoke_success           // [F3] failure is explicit generated control flow

invoke_failure:
    CALL destruct(struct)                                   // [F4] emitted while the trace stack unwinds
    RETURN                                                  // [F5] exception never crosses the generated frame

invoke_success:
    value = LOAD struct.value                               // [F6] sentinel is unreachable on failure
    STORE output, value
    CALL destruct(struct)
    RETURN
```

- **[F1]** All guarded invokes in one traced function reuse a single stack allocation. The generated trace does not capture the tracing thread's TLS address.
- **[F2]** The proxy is `noexcept`; it catches the original exception and parks it for later rethrow.
- **[F3]** The holder turns the parked exception into an explicit failure edge.
- **[F4]** Trace termination unwinds the trace-building C++ stack, causing traced destructors to be emitted on the failure edge.
- **[F5]** Generated callers observe the parked state and return through their own cleanup edges. The outermost native boundary rethrows the original exception.
- **[F6]** Operations that consume the invoke result exist only on the success edge.

Functions explicitly marked `noexcept` retain the direct-call path:

```text
body:
    CALL writeStructNoexcept(struct, 42)  // [N1] direct call: no proxy or post-call failure branch
    value = LOAD struct.value
    STORE output, value
    CALL destruct(struct)                 // [N2] destructor helper is also noexcept and direct
    RETURN
```

- **[N1]** The `noexcept` function-pointer overload preserves the low-overhead direct invoke.
- **[N2]** Destruction is explicitly `noexcept`, so cleanup cannot recursively enter guarded exception handling while the trace is already unwinding.

The parked state is propagated across generated `NautilusFunction` calls and genuinely nested compiled invocations. Only the outermost `ModuleFunction` rethrows, ensuring every generated frame first returns through its cleanup path.

## Tests

- All 14 `InvokeExceptionTest` cases pass with exception-based and lazy tracing.
- `systest_compiler_sliceCache_true` passes.

### DIFF
--- a/nes-nautilus/tests/UnitTests/ExceptionTest.cpp
+++ b/nes-nautilus/tests/UnitTests/ExceptionTest.cpp
@@ -1,0 +1,332 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <array>
+#include <cstdint>
+#include <filesystem>
+#include <fstream>
+#include <functional>
+#include <iterator>
+#include <memory>
+#include <stdexcept>
+#include <string>
+#include <utility>
+#include <variant>
+
+#include <gtest/gtest.h>
+#include <nautilus/nautilus_function.hpp>
+#include <DataStructureTestUtils.hpp>
+#include <function.hpp>
+#include <val_arith.hpp>
+#include <val_ptr.hpp>
+#include <val_std.hpp>
+
+namespace NES
+{
+namespace
+{
+thread_local uint32_t destructorCalls = 0;
+thread_local std::array<int32_t, 4> destructorValues{};
+
+struct ExceptionResult
+{
+    int32_t value = 0;
+
+    ~ExceptionResult() noexcept
+    {
+        if (destructorCalls < destructorValues.size())
+        {
+            destructorValues[destructorCalls] = value; ///NOLINT(cppcoreguidelines-pro-bounds-constant-array-index) bounds are checked
+        }
+        ++destructorCalls;
+    }
+};
+
+void writeResult(ExceptionResult* result, int32_t value) noexcept
+{
+    result->value = value;
+}
+
+void throwWhileWriting(ExceptionResult*, int32_t)
+{
+    throw std::runtime_error("invoke failed");
+}
+
+void throwWithoutCleanup()
+{
+    throw std::runtime_error("invoke failed without cleanup");
+}
+
+nautilus::NautilusFunction nestedThrowingFunction{
+    "nested_throwing_function",
+    []
+    {
+        nautilus::val<ExceptionResult> inner;
+        nautilus::invoke(writeResult, &inner, nautilus::val<int32_t>{2});
+        nautilus::invoke(throwWhileWriting, &inner, nautilus::val<int32_t>{42});
+    }};
+
+class InvokeExceptionTest : public testing::TestWithParam<std::string>
+{
+protected:
+    template <bool ThrowsAtInnerLevel, bool ThrowsAtMiddleLevel, bool ThrowsAtOuterLevel>
+    static void runThreeLevelPermutation(const std::string& traceMode)
+    {
+        constexpr uint8_t throwingLevels = static_cast<uint8_t>(ThrowsAtInnerLevel) | static_cast<uint8_t>(ThrowsAtMiddleLevel) << 1U
+            | static_cast<uint8_t>(ThrowsAtOuterLevel) << 2U;
+        SCOPED_TRACE(testing::Message() << "throwing levels mask: " << static_cast<uint32_t>(throwingLevels));
+
+        nautilus::NautilusFunction inner{
+            "three_level_inner_" + std::to_string(throwingLevels),
+            []() noexcept(!ThrowsAtInnerLevel)
+            {
+                nautilus::val<ExceptionResult> value;
+                nautilus::invoke(writeResult, &value, nautilus::val<int32_t>{3});
+                if constexpr (ThrowsAtInnerLevel)
+                {
+                    nautilus::invoke(throwWhileWriting, &value, nautilus::val<int32_t>{42});
+                }
+            }};
+        nautilus::NautilusFunction middle{
+            "three_level_middle_" + std::to_string(throwingLevels),
+            [&inner]() noexcept(!(ThrowsAtInnerLevel || ThrowsAtMiddleLevel))
+            {
+                nautilus::val<ExceptionResult> value;
+                nautilus::invoke(writeResult, &value, nautilus::val<int32_t>{2});
+                inner();
+                if constexpr (ThrowsAtMiddleLevel)
+                {
+                    nautilus::invoke(throwWhileWriting, &value, nautilus::val<int32_t>{42});
+                }
+            }};
+        nautilus::NautilusFunction outer{
+            "three_level_outer_" + std::to_string(throwingLevels),
+            [&middle]() noexcept(!(ThrowsAtInnerLevel || ThrowsAtMiddleLevel || ThrowsAtOuterLevel))
+            {
+                nautilus::val<ExceptionResult> value;
+                nautilus::invoke(writeResult, &value, nautilus::val<int32_t>{1});
+                middle();
+                if constexpr (ThrowsAtOuterLevel)
+                {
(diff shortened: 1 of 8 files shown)


keyseven123:
Re-request once CI is green

vcpkg/vcpkg-registry/ports/nautilus/0002-auto-guard-throwing-invokes.patch
```diff
@@ -0,0 +1,832 @@
...
++	static void construct_with(ValueType* ptr,
++	                           RawArgs... args) noexcept(std::is_nothrow_constructible_v<ValueType, RawArgs...>) {
+ 		new (ptr) ValueType(args...);
+ 	}
+ 
+@@ -214,6 +236,7 @@ private:
+ 			return;
+ 		}
+ 		if constexpr (!std::is_trivially_destructible_v<ValueType>) {
++			unregister_destructor();
```
keyseven123, line 237: **correctness** — `val<T>::operator=(val&&)` was not updated for the new cleanup registry, so a moved-from object's destructor entry is never unregistered and the moved-to object is never re-registered.

The patch adds `unregister_destructor()` to `release_storage()` and `register_destructor()` to every constructor, but leaves `operator=(val<ValueType>&&)` untouched. That operator does `release_storage(); value_ptr = other.value_ptr; moved_ = false; other.moved_ = true;` — `release_storage()` unregisters the destination's own entry, `val<T*>::operator=` (val_ptr.hpp:302) keeps `this->state` unchanged so `value_ptr.getState()` still returns the destination's original ref, and no `register_destructor()` follows; meanwhile the source keeps its entry because its own `release_storage()` returns early on `moved_` before reaching `unregister_destructor()`.

So in a traced function doing `nautilus::val<std::string> a; nautilus::val<std::string> b; a = std::move(b);`, once the scope ends `~a` destroys the shared storage while the source's stale entry stays in `activeDestructors` forever. Any subsequent potentially-throwing `invoke` in that same traced function is lowered with a landing pad that calls `~std::string` a second time on that already-destroyed storage, so the first exception that unwinds out of the compiled pipeline causes a double free / heap corruption crash instead of propagating cleanly.

nes-nautilus/tests/UnitTests/ExceptionTest.cpp
```diff
@@ -0,0 +1,314 @@
...
+namespace
+{
+thread_local uint32_t destructorCalls = 0;
+thread_local std::array<int32_t, 4> destructorValues{};
+
+struct ExceptionResult
+{
+    int32_t value = 0;
+
+    ~ExceptionResult() noexcept { destructorValues.at(destructorCalls++) = value; }
```
keyseven123: **correctness** — `~ExceptionResult` is `noexcept` but indexes a 4-element array with an unbounded counter via `.at()`, so a regression that runs a 5th destructor calls `std::terminate` instead of failing the test.

`destructorValues` is `std::array<int32_t, 4>` and the destructor does `destructorValues.at(destructorCalls++)`. `ThreeNestedFunctionsCoverEveryThrowingPermutation` already drives three destructor calls per permutation with only one reset per permutation, and the whole point of the suite is to catch the exact regression class where cleanups run too many times (landing pad plus normal path both destructing, or a stale `activeDestructors` entry firing twice).

When that regression occurs, `destructorCalls` reaches 4, `.at()` throws `std::out_of_range` from a `noexcept` destructor, and the process calls `std::terminate`. CI then reports the test binary as a crash/abort with no assertion message, and every test scheduled after it in the same process never runs — precisely the case where a clear `EXPECT_EQ(destructorCalls, 1)` failure was needed to diagnose the bug.

nes-nautilus/tests/UnitTests/ExceptionTest.cpp
```diff
@@ -0,0 +1,314 @@
...
+    {
+        nautilus::val<ExceptionResult> inner;
+        nautilus::invoke(writeResult, &inner, nautilus::val<int32_t>{2});
+        nautilus::invoke(throwWhileWriting, &inner, nautilus::val<int32_t>{42});
+    }};
+
+class InvokeExceptionTest : public testing::TestWithParam<std::string>
+{
+protected:
+    static nautilus::engine::NautilusEngine makeEngine(const std::string& traceMode, bool dumpAfterTracing = false)
```
keyseven123: **cleanup** — `makeEngine` re-implements the shared `NES::TestUtils::makeEngine` helper the test already links against.

`nes-nautilus/tests/TestUtils/include/DataStructureTestUtils.hpp` already declares `nautilus::engine::NautilusEngine makeEngine(EngineMode)` and `DataStructureTestUtils.cpp:65-72` sets the identical `engine.Compilation` / `engine.backend`=mlir / `engine.compilationStrategy`=legacy / `mlir.enableMultithreading`=false quartet; the new target links `nes-nautilus-test-util`, so it is already available.

This is a fifth copy of the engine-option boilerplate (`NautilusBufferTest.cpp:63`, `NautilusTestUtils.cpp:114` and `:256` are the others) — when a backend or compilation-strategy option name changes upstream, every copy has to be found and fixed, and the ones missed silently fall back to a different backend. Extend the shared helper with the `traceMode`/`dump` parameters instead.

nes-nautilus/tests/UnitTests/ExceptionTest.cpp
```diff
@@ -0,0 +1,314 @@
...
+    {
+        nautilus::engine::Options options;
+        options.setOption("engine.Compilation", true);
+        options.setOption("engine.backend", std::string("mlir"));
```
keyseven123: **cleanup** — constructor temporaries written with parentheses (`std::string("mlir")`) violate the repo's strict brace-initialization convention: always `Type{args}`, never `Type(args)`, including temporary construction.

Affects lines 81, 82 and 312: `std::string("mlir")`, `std::string("legacy")`, `std::string("exceptionBasedTracing")`, `std::string("lazyTracing")`. Should be `std::string{"mlir"}` etc.

keyseven123 (changes requested):
Small stuff

vcpkg/vcpkg-registry/ports/nautilus/0002-auto-guard-throwing-invokes.patch
ls-1801: addressed

nes-nautilus/tests/UnitTests/ExceptionTest.cpp
ls-1801: addressed

nes-nautilus/tests/UnitTests/ExceptionTest.cpp
ls-1801: addressed

nes-nautilus/tests/UnitTests/ExceptionTest.cpp
ls-1801: addresed

keyseven123 (approved):

alepping (approved):

PR #1856, opened 2026-08-10
feat(nautilus): Move static hash map config out of the ChainedHashMap header

keyseven123:
This PR extracts static hash map config out of the header
  
 ## Verifying this change
  This change is tested by running our existing tests.


ls-1801 (approved):

nes-physical-operators/src/Aggregation/AggregationOperatorHandler.cpp
```diff
@@ -57,8 +53,7 @@ AggregationOperatorHandler::getCreateNewSlicesFunction(const CreateNewSlicesArgu
 {
     PRECONDITION(
         numberOfWorkerThreads > 0, "Number of worker threads not set for window based operator. Was setWorkerThreads() being called?");
-    auto newHashMapArgs = dynamic_cast<const CreateNewHashMapSliceArgs&>(newSlicesArguments);
-    newHashMapArgs.numberOfBuckets = std::clamp(rollingAverageNumberOfKeys.rlock()->getAverage(), 1UL, maxNumberOfBuckets);
+    const auto& newHashMapArgs = dynamic_cast<const CreateNewHashMapSliceArgs&>(newSlicesArguments);
```
yschroeder97, line 57: The bucket count now comes only from `number_of_partitions`, which defaults to 100.
The handler previously sized each slice from the rolling average of observed key counts, capped at `max_number_of_buckets` of 10000.

For an aggregation over 1M distinct keys that is 128 chains instead of 16384, so a chain holds about 7800 entries instead of 61.
Every lookup walks that chain, and the probe merges every worker map into a final map with the same 128 chains, so the merge alone costs roughly K^2/128 key comparisons.
The hash join build takes the same change.

If dropping the adaptive sizing is intended, `number_of_partitions` needs a default near the old ceiling, and the PR description should say so.

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMapConfig.hpp
```diff
@@ -0,0 +1,62 @@
...
+    /// Empty when this map runs without an in-map BloomFilter.
+    std::optional<Nautilus::Interface::BloomFilterParams> bloomFilterParams;
+    std::vector<FieldOffsets> fieldKeys;
+    std::vector<FieldOffsets> fieldValues;
+    std::shared_ptr<const HashFunction> hashFunction;
+
+    /// Turning numberOfBuckets into a chain count, a mask or a buffer size needs the map's memory layout, so
+    /// those live on ChainedHashMap itself: calculateNumberOfChains() and calculateBufferSize(). Only the
+    /// derivations that need nothing but this struct stay here.
+    [[nodiscard]] uint64_t entriesPerPage() const { return pageSize / entrySize; }
```
yschroeder97, line 56: `entriesPerPage()` divides by `entrySize`, and the struct default-constructs that to 0.
The options type this replaces checked entrySize, pageSize, numberOfBuckets and entriesPerPage in its constructor, plus that entrySize exceeded keySize plus valueSize.
Nothing checks any of it now.

Two paths reach that.
`number_of_partitions: 0` passes NumberValidation, which matches `^\d+$`, and now reaches calculateNumberOfChains directly.
The clamp that floored the bucket count at 1 is gone, so a build without preconditions gets capacity 1 and mask 0, and every key goes into a single chain.
A default-constructed config handed to the map view divides by zero in insert().

The view constructor already checks the hash function, so the sizing checks belong next to it.

nes-physical-operators/include/HashMapSlice.hpp
```diff
@@ -33,34 +33,17 @@
...
+        : config(std::move(config)), bufferProvider(bufferProvider)
     {
     }
 
     ~CreateNewHashMapSliceArgs() override = default;
-    uint64_t keySize;
-    uint64_t valueSize;
-    uint64_t pageSize;
-    uint64_t numberOfBuckets;
+    ChainedHashMapConfig config;
```
yschroeder97, line 46: A slice reads only entrySize, pageSize, numberOfBuckets and the bloom byte count, but it now stores the field layout and the hash function as well.
Each slice construction copies two `std::vector<FieldOffsets>`, and every element holds a record field identifier string.

A join with a 1s slide over a 10min window keeps about 600 slices alive.
At four key fields that is 1200 vector allocations and 4800 strings, against four integers and an optional before.
The handler then copies the args struct a second time into the std::function it returns.

The deleted `static_assert(std::is_trivially_copyable_v<ChainedHashMapConfig>)` recorded that constraint and nothing replaces it.
Consider keeping a sizing-only struct for the slice path.

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
```diff
@@ -160,83 +154,47 @@ class ChainedHashMap final : public HashMap
...
-        /// NOLINTEND(cppcoreguidelines-pro-type-reinterpret-cast, cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        /// NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast, cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        return reinterpret_cast<ChainedHashMapEntry**>(data + sizeof(Header));
     }
 
     /// The in-map BloomFilter bit area, sitting inline right behind the chains array. Empty when the filter
-    /// is disabled. allocationByteCount() rounds to whole words, so the span is always word-aligned in size.
-    [[nodiscard]] std::span<uint64_t> bloomBits()
+    /// is disabled. allocationByteCount() rounds to whole words, so the area is always word-aligned in size.
+    [[nodiscard]] uint64_t* bloomBitsBegin(const uint64_t numberOfChains)
```
yschroeder97: The comment above says the area is empty when the filter is disabled, but the function returns a raw pointer and cannot express emptiness.
With zero bloom bytes the address is one past the end of the buffer calculateBufferSize allocated.

Both call sites check `bloomFilterParams` first, so nothing dereferences it today.
Reword the comment to say the caller has to check that the filter is enabled.

yschroeder97 (commented):
Read the full diff and the branch sources. I did not build or run tests.

The extraction reads well. The header drops to four fields, the key functions move to the build operators where they belong, and taking the mask out of the proxy turns it into an immediate instead of a per-probe header load. That also removes the case where the traced probe sized its filter from one bucket count while the map allocated its chains from another. The comments that state what the change gives up are good.

One change is larger than the title suggests. The rolling average bucket sizing is gone, so every hash map is fixed at `number_of_partitions`, which defaults to 100 against the removed ceiling of 10000. Details inline.

One more, outside the diff: `RollingAverage` in nes-common now has no production caller. `git grep RollingAverage` on the branch returns the header and its own unit test. Either delete both or say why it stays.

nes-physical-operators/include/HashMapSlice.hpp
keyseven123: The two `std::vector<FieldOffsets>` copies happen once per slice construction, never per tuple and never on the build or probe hot path, and the `shared_ptr<const HashFunction>` copy is a refcount bump. So the cost is real but it is allocation-time, not throughput.

The `static_assert(std::is_trivially_copyable_v<...>)` cannot come back in any form: `std::vector<FieldOffsets>` and `shared_ptr<const HashFunction>` are load-bearing members of the config, so the type is inherently non-trivial. There is nothing left for that assert to protect.

Keeping a second sizing-only struct for the slice path would re-introduce exactly the two-options-types split this PR set out to delete, so I would rather leave it. If the per-slice allocations show up in a profile for a high-slice-count join, narrowing `HashMapSlice` to the four sizing fields it actually reads is a contained follow-up.

keyseven123:
Thanks for the careful read. Summary of what changed and what did not.

**Bucket sizing (`AggregationOperatorHandler`).** You are right that the sizing is now fixed, and that is deliberate: the mask is a query-compile-time constant folded into the traced probe, so per-slice resizing is exactly the build/probe mismatch this PR removes. What was wrong was the default. `number_of_partitions` now defaults to **1024** (2048 chains, ~16 KB of chain pointers per hash map, per worker thread, per slice) instead of 100.

I did not go to the old ceiling of 10000: that is 16384 chains, i.e. 128 KB per map/worker/slice, and the old code never actually paid it — the chains came from the rolling average, only the bloom area was max-sized. 1024 is the compromise. The option help text now states that the count is fixed at compile time and what each direction costs.

**Missing validation (`ChainedHashMapConfig`).** `ChainedHashMap::init()` does still check `entrySize`, `pageSize / entrySize`, the chain count, power-of-two and the buffer size — but behind `IF_PRECONDITION`, so a Benchmark build strips them, which is your real point. Fixed at the layer where it belongs: `number_of_partitions` now carries `NonZeroValidation` in addition to `NumberValidation`, so `0` is rejected at config parse instead of reaching a precondition. No production path default-constructs a `ChainedHashMapConfig`, so the `entriesPerPage()` division is unreachable, but a bad config value no longer is.

**Bloom bit area.** Better than a comment fix: `bloomBitsBegin()` is now `bloomBits(numberOfChains, bloomBytes)` returning `std::span<uint64_t>`, which is genuinely empty when the filter is disabled, so `init()`s zeroing becomes a no-op instead of relying on a count. The traced path still needs a raw pointer, so `getBloomFilterMemArea()` keeps one but now takes `bloomBytes` and asserts it is non-zero — the one raw handout is guarded rather than documented.

**`RollingAverage`.** It still has a production caller:
- `nes-memory/include/Runtime/UnpooledChunksManager.hpp:31,86`
- `nes-memory/UnpooledChunksManager.cpp:45,93,118`

The member is named `rollingAverage` (lowercase), which is probably why the grep missed it. It stays.

Verified locally: build clean, 929/929 ctest pass, clang-format clean.

nes-physical-operators/src/Aggregation/AggregationOperatorHandler.cpp
keyseven123: I set it to 1024. This work goes in the direction of adaptive optimization via the optimizer of @greenBene  so I suspect that we will implement a "better" version later via adaptive optimization

nes-nautilus/include/Interface/HashMap/ChainedHashMap/ChainedHashMap.hpp
keyseven123: Changed to a span

yschroeder97 (approved):

PR #1859, opened 2026-08-10
refactor(#1831): rename buffer abstractions (MemoryLayout, TaskBufferRef, Buffer)

### LINKED ISSUE #1831: Renaming / Refactoring of Nautilus Wrappers around TupleBuffer
### Current state
  
 In our current code, there are  three overlapping buffer abstractions in `nes-nautilus` with similar/or even the same methods / functionality. This makes them unclear to use. 

  - `NautilusBuffer` (`Interface/NautilusBuffer.hpp`) — the general buffer abstraction, handles ownership (`Owned`/`Borrowed`). Use it whenever a
  buffer is involved.
  - `RecordBuffer` (`Interface/RecordBuffer.hpp`) — the `TupleBuffer` view (sequence number, watermark, tuples), but does not use `NautilusBuffer`:
  it keeps a raw `nautilus::val<TupleBuffer*>` (`RecordBuffer.hpp:74`) and leaks it via `getMemArea()`/`getReference()`.
  - `TupleBufferRef` (`Interface/BufferRef/TupleBufferRef.hpp`) — not a reference at all, it is the memory layout.

  ### Desired Structure

  ```mermaid
  classDiagram
      class BufferRef {
          +Sequence
          +Chunk
          +getTuple()
          +size()
      }
      class NautilusBuffer {
          +Copy
          +Move
          +Destroy
      }
      class MemoryLayout
      BufferRef *-- NautilusBuffer
      BufferRef *-- MemoryLayout
      MemoryLayout <|-- RowLayout
      MemoryLayout <|-- ColLayout
  ```
  
  - `BufferRef` manages a buffer, so it holds a `NautilusBuffer`; copy/move/destroy and child buffers stay there.
  - `BufferRef` holds a `MemoryLayout`, picked by the optimizer (see `MemoryLayoutTypeTrait`).
  - Code using a `BufferRef` knows neither the layout nor whether the buffer is owned or borrowed.
  - `TupleBuffer` to `RecordBuffer` is the same as `PagedVector` to `PagedVectorRef` — hence `BufferRef`.

  | today | new |
  | --- | --- |
  | `TupleBuffer` (`nes-memory/include/Runtime/TupleBuffer.hpp`) | `Buffer` |
  | `RecordBuffer` | `BufferRef` |
  | `TupleBufferRef` | `MemoryLayout` |
  | `NautilusBuffer` | stays — ownership wrapper, not a reference |
  
  ### Acceptance Criteria

  - [ ] `RecordBuffer` holds a `NautilusBuffer`, no raw `nautilus::val<TupleBuffer*>` member
  - [ ] `RecordBuffer` holds the memory layout; no call site constructs a `TupleBufferRef` itself
  - [ ] `RecordBuffer::getReference()` gone
  - [ ] Renames applied: `TupleBuffer` → `Buffer`, `RecordBuffer` → `BufferRef`, `TupleBufferRef` → `MemoryLayout`
  - [ ] All existing tests pass

  Overlaps with #1674, which moves `VariableSizedAccess` into `nes-nautilus` and touches both `TupleBufferRef` and `NautilusBuffer`. Branch this off the PR of #1674 instead of `main`.

keyseven123:
This PR renames our buffer wrappers. See #1831 for more information. 

## Verifying this change
This change is tested by our CI test.

This PR closes #1831 .

### DIFF
--- a/nes-executable/include/ExecutablePipelineStage.hpp
+++ b/nes-executable/include/ExecutablePipelineStage.hpp
@@ -16,7 +16,7 @@
 #include <cstdint>
 #include <memory>
 #include <ostream>
-#include <Runtime/TupleBuffer.hpp>
+#include <Runtime/Buffer.hpp>
 #include <PipelineExecutionContext.hpp>
 
 namespace NES
@@ -33,7 +33,7 @@ public:
 
     /// Executes the ExecutablePipelineStage with a readonly input tuple buffer.
     /// `execute` should never be called on a pipeline that has not previously been `started`.
-    virtual void execute(const TupleBuffer& inputTupleBuffer, PipelineExecutionContext& pipelineExecutionContext) = 0;
+    virtual void execute(const Buffer& inputTupleBuffer, PipelineExecutionContext& pipelineExecutionContext) = 0;
 
     /// Stops the ExecutablePipelineStage allowing it to flush left over state.
     /// `stop` should never be called on a pipeline that has not previously been `started`.
--- a/nes-executable/include/PipelineExecutionContext.hpp
+++ b/nes-executable/include/PipelineExecutionContext.hpp
@@ -20,15 +20,15 @@
 #include <vector>
 #include <Identifiers/Identifiers.hpp>
 #include <Runtime/AbstractBufferProvider.hpp>
+#include <Runtime/Buffer.hpp>
 #include <Runtime/Execution/OperatorHandler.hpp>
-#include <Runtime/TupleBuffer.hpp>
 
 namespace NES
 {
 class PipelineExecutionContext
 {
 public:
-    /// Policy whether an emitted TupleBuffer is able to process immediately or
+    /// Policy whether an emitted Buffer is able to process immediately or
     /// needs to be dispatched as a new Task.
     enum class ContinuationPolicy : uint8_t
     {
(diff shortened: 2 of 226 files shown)


ls-1801 (changes requested):
I think we should spend some more time establishing consistent names and responsibilities for these abstractions before committing to a repository-wide rename.

At the moment, I find the relationship between `Buffer`, `NautilusBuffer`, `TaskBufferRef`, `RecordView`, and `MemoryLayout` difficult to infer from their names. In particular:

- We usually use the `Ref` suffix for the Nautilus interface to an underlying runtime data structure, but there is no corresponding `TaskBuffer`, and `TaskBufferRef` contains no task-specific state or behavior.
- `RecordView` combines a `TaskBufferRef` with a `MemoryLayout`, making it closer to the schema-aware Nautilus interface described as `BufferRef` in #1831.
- The implementation differs from #1831, which proposes one facade holding both the `NautilusBuffer` and the `MemoryLayout`.
- Mechanical renames such as `getPagedVectorMemoryLayout`, despite that method returning a `Buffer*`, suggest that the terminology is not yet sufficiently clear.

One possible vocabulary could distinguish:

- `Buffer`: general-purpose managed byte storage.
- `NautilusBuffer`: the owned/borrowed Nautilus handle to that storage.
- `TupleBuffer`: a buffer whose opaque payload represents tuples and which carries tuple-stream metadata. This would broadly retain the meaning of the old `TupleBuffer`, but without using it for general-purpose storage.
- `MemoryLayout`: the physical representation used to interpret tuple memory.
- `TupleBufferRef` or a similar name: the Nautilus interface that combines tuple memory with its layout and provides access to individual tuples and fields.
- `Record`: the value-level representation used during operator execution, in contrast to a tuple’s encoded in-memory representation.

I do not necessarily want to prescribe these exact names in this review. My main concern is that we should first agree on the conceptual model—including the distinction between tuple memory and record values, the meaning of `Ref`, and the relationship between buffer ownership and memory layout—and then derive the names and interfaces from that model.

Could we spend some more time on this design and document the intended terminology before completing the rename?

Artraxon (changes requested):
I agree with @ls-1801 , if the naming in this PR deviates from what was described in #1831 , it should document it at least in this PR description.
(I also find carrying the issue number in the PR title quite odd)

keyseven123:
Closing this PR as this needs more discussion.

PR #1872, opened 2026-08-11
Is nan function

### LINKED ISSUE #1867: IsNan Function
We should add a function that checks if a floating point is a number or not `IsNaN`.

keyseven123:
This PR adds a `IsNaN` check as asked for in #1867 .

## Verifying this change
This change is tested by providing a new SLT.

## Documentation
- The change is reflected in the necessary documentation.

This PR closes #1867

### DIFF
--- a/docs/guide/query_api.md
+++ b/docs/guide/query_api.md
@@ -502,6 +502,26 @@ SELECT * FROM s WHERE optional_value IS NULL INTO sink
 SELECT * FROM s WHERE optional_value IS NOT NULL INTO sink
 ```
 
+`IS NaN` and `IS NOT NaN` test whether a numeric value is a floating-point *not a number*:
+
+```sql
+SELECT * FROM measurements WHERE reading IS NaN INTO sink
+```
+
+```sql
+SELECT * FROM measurements WHERE reading IS NOT NaN INTO sink
+```
+
+The operand must be numeric. Applying `IS NaN` to a `VARSIZED`, `BOOLEAN`, or `CHAR` field is
+rejected when the query is registered. Integer operands are accepted but are never NaN, so `IS NaN`
+is always false for them. Infinities are not NaN.
+
+💡 `IS NaN` follows SQL three-valued logic, see [null handling](../technical/null_handling.md) for
+more details.
+
+The equivalent function-call spelling `ISNAN(x)` is also available, which is what you need in a
+projection: `SELECT ISNAN(reading) AS invalid FROM measurements INTO sink`.
+
 #### Union
 
 Union combines two input streams with identical schema into one.
@@ -734,6 +754,8 @@ Functions are either unary (one input) or binary (two inputs).
 | Not In value list  | `SELECT * FROM s WHERE a NOT IN (1, 5) INTO sink`              |
 | Is Null            | `SELECT * FROM s WHERE a IS NULL INTO sink`                    |
 | Is Not Null        | `SELECT * FROM s WHERE a IS NOT NULL INTO sink`                |
+| Is NaN             | `SELECT * FROM s WHERE a IS NaN INTO sink`                     |
+| Is Not NaN         | `SELECT * FROM s WHERE a IS NOT NaN INTO sink`                 |
 
 #### **Other**
 
--- a/docs/technical/null_handling.md
+++ b/docs/technical/null_handling.md
@@ -322,6 +322,25 @@ SELECT ISNULL(CONCAT(a, b)) FROM stream;       -- true if the concat result is N
 
 `ISNULL()` is the **only way to reliably test for NULL**, since `value == NULL` would itself evaluate to NULL.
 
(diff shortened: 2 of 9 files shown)


keyseven123:
As discussed, we should not want to have an own physical function but rather want to convert a `IS NAN` into an equalslogicalfunction that compares the field against the `NaN` in double. If the field type is an integer, we simply add a `ConstantFunction`that returns true

nes-logical-operators/src/Functions/BooleanFunctions/IsNaNLogicalFunction.cpp
```diff
@@ -0,0 +1,129 @@
...
+    if (not inferredChild.getDataType().isNumeric())
+    {
+        throw CannotInferStamp("IS NaN expects a numeric input function but got {}", inferredChild.getDataType());
+    }
+    /// NaN is the only value that does not compare equal to itself, so IS NaN desugars into NOT(x = x) and needs no physical
+    /// function of its own. Comparing against a NaN constant would not work, as every comparison with NaN is false, NaN = NaN
+    /// included. Integers always compare equal to themselves and thus yield false, as required.
+    /// Unlike ISNULL, IS NaN follows the usual nullability propagation, i.e., NULL IS NaN is NULL. Both Equals and Negate
+    /// propagate NULL, so the desugared form keeps that behaviour.
+    return NegateLogicalFunction{EqualsLogicalFunction{inferredChild, inferredChild}}.withInferredDataType(schema);
```
yschroeder97, line 76: Note that this will duplicate the entire subtree of functions. Does nautilus eliminate this?

nes-sql-parser/AntlrSQL.g4
```diff
@@ -489,6 +494,7 @@ LIKE: 'LIKE';
 LIMIT: 'LIMIT' | 'limit';
 LIST: 'LIST';
 MERGE: 'MERGE' | 'merge';
+NANTOKEN: 'NAN' | 'nan' | 'NaN';
```
yschroeder97, line 497: Do we really want to reserve all three here? This means users can no longer name columns "nan".

yschroeder97 (approved):
Cool that we can have this syntactic sugar now. 

Please look at the comments, though.

docs/guide/query_api.md
```diff
@@ -502,6 +502,27 @@ SELECT * FROM s WHERE optional_value IS NULL INTO sink
...
+```
+
+The operand must be numeric. Applying `IS NaN` to a `VARSIZED`, `BOOLEAN`, or `CHAR` field is
+rejected when the query is registered. Integer operands are accepted but are never NaN, so `IS NaN`
+is always false for them. Infinities are not NaN.
+
+💡 `IS NaN` follows SQL three-valued logic: a NULL operand yields NULL rather than false, and a NULL
+predicate excludes the row. A NULL value is therefore returned by **neither** `IS NaN` nor
+`IS NOT NaN`. Combine with `IS NULL` to catch both: `WHERE reading IS NaN OR reading IS NULL`.
+
```
Artraxon:
```markdown
`IS NaN` follows SQL three-values logic, see technical/null_handling.md for more details.
```

docs/technical/null_handling.md
```diff
@@ -322,6 +322,28 @@ SELECT ISNULL(CONCAT(a, b)) FROM stream;       -- true if the concat result is N
 
 `ISNULL()` is the **only way to reliably test for NULL**, since `value == NULL` would itself evaluate to NULL.
 
+### ISNAN Function
+
+`ISNAN(expr)` — also spelled `expr IS NaN` / `expr IS NOT NaN` — tests whether a numeric expression
+is a floating-point *not a number*. The operand must be numeric; a `VARSIZED`, `BOOLEAN`, or `CHAR`
+input is rejected during type inference. Integer operands are accepted but never NaN.
+
```
Artraxon: This paragraph is redundant with query_api.md

Artraxon (changes requested):
looks fine, just keep the docs a bit more terse

nes-logical-operators/src/Functions/BooleanFunctions/IsNaNLogicalFunction.cpp
keyseven123: It should but I have changed it to a true equals function.

nes-sql-parser/AntlrSQL.g4
keyseven123: I would say so. I think it is not a good idea especially as usually fields are case-insensitive.

docs/technical/null_handling.md
keyseven123: Done

docs/guide/query_api.md
keyseven123: Done

Artraxon (approved):

PR #1876, opened 2026-08-11
feat(docs): add quick start guide

tekdogan:
## Summary

  Adds a practical Quick Start guide to the documentation website.

  The guide walks users through:

  - Starting a single NebulaStream worker with Docker
  - Creating a small topology that generates values, scales them, and writes CSV output
  - Registering the query with `nes-cli`
  - Inspecting the generated results
  - Running the same setup with Docker Compose
  - Checking query status and stopping the query

  ## Scope

  This PR changes only:

  - `docs/website/content/reference/GetStarted.md`

  ## Follow-up

  Nightly command validation: #1877 

## Issue Closed by this pull request:

This PR closes #1861

PR #1877, opened 2026-08-11
feat(CI): add nightly command check for docs

### LINKED ISSUE #1887: CI: add nightly validation for docs/quick-start
## Goal

  Add a nightly GitHub Actions workflow that validates the Docker commands documented in the Quick Start guide.

  ## Scope

  The check should exercise both documented setup paths:

  - Direct Docker flow:
    - Start `nebulastream/worker:main`
    - Submit the topology with `nebulastream/nes-cli:main`
    - Verify `output/results.csv` contains the expected scaled values

  - Docker Compose flow:
    - Start the worker and CLI services
    - Submit `compose-topology.yaml`
    - Verify the same CSV output

  ## Implementation

  Add:

  - `.github/workflows/nightly-quick-start.yml`
  - `docs/website/scripts/test-quick-start.sh`

  The workflow should run nightly and support manual execution with `workflow_dispatch`.

  ## Acceptance Criteria

  - The workflow runs on `ubuntu-latest`.
  - It cleans up containers and temporary files on success and failure.
  - It retries query registration while the worker starts.
  - It fails when the expected CSV output is not generated.
  - It validates both Docker and Docker Compose instructions.
  - The workflow can be triggered manually from GitHub Actions.

tekdogan:
## Summary

  Adds a nightly GitHub Actions check for the Docker commands documented in the Quick Start guide.

  The check validates both documented paths:

  - Starting a worker and submitting a query with `docker run`
  - Starting the same setup with Docker Compose
  - Waiting for `output/results.csv`
  - Verifying that the query produced scaled values

  ## Stack

  Stacked on #1876 , which adds the Quick Start documentation page.

  ## Files

  - `.github/workflows/nightly-quick-start.yml`
  - `docs/website/scripts/test-quick-start.sh`

  ## Trigger

  The workflow runs every night at `02:00 UTC` and can also be started manually through `workflow_dispatch`.

This PR closes #1887

### DIFF
--- a/.github/workflows/nightly-quick-start.yml
+++ b/.github/workflows/nightly-quick-start.yml
@@ -1,0 +1,23 @@
+name: Nightly Quick Start Check
+
+on:
+  workflow_call:
+    inputs:
+      head_sha:
+        required: false
+        type: string
+        default: main
+
+permissions:
+  contents: read
+
+jobs:
+  quick-start:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v5
+        with:
+          ref: ${{ inputs.head_sha }}
+      - name: Run Quick Start commands
+        run: bash docs/website/scripts/test-quick-start.sh
--- a/docs/website/scripts/test-quick-start.sh
+++ b/docs/website/scripts/test-quick-start.sh
@@ -1,0 +1,160 @@
+#!/usr/bin/env bash
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+
+#    https://www.apache.org/licenses/LICENSE-2.0
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -Eeuo pipefail
+
+script_dir="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)"
+quick_start_doc="$script_dir/../content/reference/GetStarted.md"
+
+work_dir="$(mktemp -d)"
+compose_file="$work_dir/compose.yaml"
+commands_dir="$work_dir/commands"
+
+cleanup() {
+  docker rm -f worker >/dev/null 2>&1 || true
+  if [[ -f "$compose_file" ]]; then
+    (cd "$work_dir" && docker compose down --volumes --remove-orphans) >/dev/null 2>&1 || true
+  fi
+  rm -rf "$work_dir"
+}
+trap cleanup EXIT
+
+mkdir -p "$work_dir/output" "$commands_dir"
+
+extract_code_block() {
+  local name="$1"
+  local language="$2"
+  local output="$3"
+  local start_marker="<!-- quick-start-${name}:start -->"
+  local end_marker="<!-- quick-start-${name}:end -->"
+  local opening_fence="\`\`\`$language"
+
+  if ! awk -v start_marker="$start_marker" -v end_marker="$end_marker" -v opening_fence="$opening_fence" '
+    $0 == start_marker { capture = 1; next }
+    capture && $0 == opening_fence { in_fence = 1; next }
+    in_fence && $0 == "```" { in_fence = 0; saw_fence = 1; next }
+    in_fence && $0 == end_marker { exit 1 }
+    capture && $0 == end_marker { complete = saw_fence; exit }
+    in_fence { print }
+    END { if (!complete) exit 1 }
+  ' "$quick_start_doc" >"$output"; then
+    echo "Could not extract the $name $language block from $quick_start_doc." >&2
+    return 1
+  fi
+
+  if [[ ! -s "$output" ]]; then
+    echo "The extracted $name $language block is empty." >&2
+    return 1
+  fi
+}
+
+run_documented_command() {
+  local name="$1"
+  local query_id="${2:-}"
+  local command
+
+  command="$(<"$commands_dir/$name.sh")"
+  if [[ -n "$query_id" ]]; then
+    command="${command//<query-id>/$query_id}"
+  fi
+  (cd "$work_dir" && bash -Eeuo pipefail -c "$command")
+}
+
+retry_documented_command() {
+  local name="$1"
+
+  for _ in {1..30}; do
+    if run_documented_command "$name"; then
+      return 0
+    fi
+    sleep 1
+  done
+
+  return 1
+}
+
+extract_code_block topology yaml "$work_dir/topology.yaml"
+extract_code_block compose yaml "$compose_file"
+for command in \
+  run-worker \
+  submit-query \
+  inspect-output \
+  stop-worker \
+  create-compose-topology \
+  start-compose \
+  list-compose \
+  submit-compose-query \
+  status-compose-query \
+  stop-compose-query \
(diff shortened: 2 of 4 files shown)


oczerwia-2010 (changes requested):
This is an important step towards keeping our documentation up to date, as requested via PN I would like to not only test the compose file but also the query to prevent syntax changes going unnoticed.

oczerwia-2010 (approved):

.github/workflows/nightly-quick-start.yml
```diff
@@ -0,0 +1,18 @@
+name: Nightly Quick Start Check
+
+on:
+  schedule:
+    - cron: "0 2 * * *"
```
keyseven123: Please add a workflow call in the existing `.github/workflows/nightly.yml`

.github/workflows/nightly-quick-start.yml
```diff
@@ -0,0 +1,18 @@
...
+
+permissions:
+  contents: read
+
+jobs:
+  quick-start:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
```
keyseven123: Update this to v5

keyseven123 (changes requested):

keyseven123 (changes requested):
A few things on `docs/website/scripts/test-quick-start.sh` before this goes in.

**`wait_for_output` assertion doesn't verify the scaling (L57)**

This assertion passes even if the scaling regressed. The generator emits
`SEQUENCE UINT64 0 250 1`, so the line `10` is already present in the unscaled
input — if `* UINT64(2)` broke, or the sink wrote `VALUE` instead of
`SCALED_VALUE`, this check is still green.

Suggest asserting a value only reachable after scaling, e.g. `^500$` (250 * 2).

**The documented commands are re-implemented rather than run (L73)**

Only the two YAML blocks come from the doc — every documented command is
re-implemented here in a different form: `--name "$worker_name"` vs
`--name worker`, `$work_dir` vs `$PWD`, `docker compose --project-name/--file`
vs plain `docker compose up -d`, `exec -T` vs `exec`. The doc's `sed`,
`docker compose ps`, `status`, `stop` and `docker compose down` lines are not
exercised at all.

That means a break in a documented command line goes undetected, which is the
thing this nightly is meant to catch. Could we extract the bash blocks with the
same marker mechanism (or at least the run/exec lines)?

**`docker rm -f` can abort the script mid-run (L96)**

The worker is started with `--rm`, so if it has already exited (crash after
producing output, for instance) Docker removed it and this fails with "No such
container". Under `set -Eeuo pipefail` that aborts the whole script right after
the docker-run flow *succeeded*, so the Compose flow is never checked and the
failure surfaces as a bare exit 1.

`|| true` here, same as in `cleanup`.

**Nit (L63)**

With `set -e`, `[[ -f "$results_file" ]] && cat "$results_file" >&2` terminates
the function when the file is missing, so the `return 1` below never runs.
Harmless today, but breaks silently if `wait_for_output` is ever used in a
conditional — `if [[ -f … ]]; then cat …; fi`.

docs/website/scripts/test-quick-start.sh
```diff
@@ -0,0 +1,116 @@
...
+}
+
+extract_yaml_block topology "$work_dir/topology.yaml"
+extract_yaml_block compose "$compose_file"
+
+wait_for_output() {
+  local results_file="$1"
+
+  for _ in {1..30}; do
+    if [[ -f "$results_file" ]] && grep -q '^10$' "$results_file"; then
```
keyseven123: This assertion passes even if the scaling regressed. The generator emits `SEQUENCE UINT64 0 250 1`, so the line `10` is already present in the *unscaled* input — if `* UINT64(2)` broke, or the sink wrote `VALUE` instead of `SCALED_VALUE`, this check is still green.

Suggest asserting a value only reachable after scaling, e.g. `^500$` (250 * 2).

docs/website/scripts/test-quick-start.sh
```diff
@@ -0,0 +1,116 @@
...
+    sleep 1
+  done
+
+  echo "Quick Start did not produce the expected CSV output." >&2
+  [[ -f "$results_file" ]] && cat "$results_file" >&2
+  return 1
+}
+
+echo "Checking the Docker run flow..."
+docker run -d --rm \
```
keyseven123: Only the two YAML blocks come from the doc — every documented command is re-implemented here in a different form: `--name "$worker_name"` vs `--name worker`, `$work_dir` vs `$PWD`, `docker compose --project-name/--file` vs plain `docker compose up -d`, `exec -T` vs `exec`. The doc's `sed`, `docker compose ps`, `status`, `stop` and `docker compose down` lines are not exercised at all.

That means a break in a documented command line goes undetected, which is the thing this nightly is meant to catch. Could we extract the bash blocks with the same marker mechanism (or at least the run/exec lines)?

docs/website/scripts/test-quick-start.sh
```diff
@@ -0,0 +1,116 @@
...
+    -t /catalog/topology.yaml start; then
+    registered=true
+    break
+  fi
+  sleep 1
+done
+[[ "$registered" == true ]] || { echo "Could not register the Docker run query." >&2; exit 1; }
+
+wait_for_output "$work_dir/output/results.csv"
+docker rm -f "$worker_name" >/dev/null
```
keyseven123: The worker is started with `--rm`, so if it has already exited (crash after producing output, for instance) Docker removed it and this fails with "No such container". Under `set -Eeuo pipefail` that aborts the whole script right after the docker-run flow *succeeded*, so the Compose flow is never checked and the failure surfaces as a bare exit 1.

`|| true` here, same as in `cleanup`.

docs/website/scripts/test-quick-start.sh
```diff
@@ -0,0 +1,116 @@
...
+
+  for _ in {1..30}; do
+    if [[ -f "$results_file" ]] && grep -q '^10$' "$results_file"; then
+      return 0
+    fi
+    sleep 1
+  done
+
+  echo "Quick Start did not produce the expected CSV output." >&2
+  [[ -f "$results_file" ]] && cat "$results_file" >&2
```
keyseven123: Nit: with `set -e`, this AND-list evaluates false when the file is missing and terminates the function right there, so the `return 1` below never runs. Harmless today, but it breaks silently if `wait_for_output` is ever used in a conditional — `if [[ -f … ]]; then cat …; fi`.

keyseven123 (changes requested):
A few things on `docs/website/scripts/test-quick-start.sh` before this goes in — details inline.

.github/workflows/nightly-quick-start.yml
tekdogan: done

.github/workflows/nightly-quick-start.yml
tekdogan: done

docs/website/scripts/test-quick-start.sh
tekdogan: done, now checking for `249*2` (as 250 is excluded)

docs/website/scripts/test-quick-start.sh
tekdogan: the nightly now extracts and runs the commands from the documentation

docs/website/scripts/test-quick-start.sh
tekdogan: worker cleanup now tolerates an already-removed container

docs/website/scripts/test-quick-start.sh
tekdogan: replaced it with an `if` block

keyseven123 (approved):

PR #1879, opened 2026-08-12
build: unify Rust integration with deduplicated umbrella crates

ls-1801:
## Summary

- introduce a repository-wide Cargo workspace with shared package metadata, dependencies, lockfile, target directory, and toolchain
- discover Rust crates from Cargo metadata and generate reusable umbrella static libraries for each transitive crate set, avoiding duplicate CXX runtime symbols
- add `add_cxx_bridge` and `target_link_rust_lib` as the public CMake integration helpers, including generated-code registration and CTest registration for workspace crates
- migrate the existing logger and network bindings to the unified workspace
- back the existing `CastToUnixTs` physical function with a Rust crate using Chrono while preserving its query API and supported timestamp behavior
- document the supported workspace and bridge-integration workflow

## Why

Linking multiple independent Rust `staticlib` crates into one C++ executable embeds duplicate copies of shared Rust dependencies, notably the CXX runtime. The umbrella layer builds each required Rust crate set as one static library, while the unified workspace keeps versions and dependency configuration consistent.

The timestamp function provides a concrete consumer of this integration: parsing now uses a maintained Rust date/time library instead of a handwritten C++ parser.

## Developer impact

New Rust crates are added to the root workspace, inherit package metadata and dependencies, create their bridge with `add_cxx_bridge`, and declare C++ target requirements with `target_link_rust_lib`. Rust unit tests are exposed through CTest as `rust-<crate>`.


Cargo.toml
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Unified Rust workspace for all NebulaStream Rust crates.
+# Add new crates here when introducing Rust code in any part of the project.
+# CMake auto-discovers members via `cargo metadata` (see cmake/EnableRust.cmake).
+[workspace]
+resolver = "2"
```
yschroeder97: Could we upgrade this to 3? Costs nothing and is the default in version 2024, which is used here.

nes-physical-operators/rust/cast-to-unix-timestamp/src/cast_to_unix_timestamp.rs
yschroeder97, line 1: Do we need this to prove that the umbrella staticlib generation works? I would not include this example. With the coordinator crates, an example that uses this will follow shortly. 

Furthermore, It also broadens what the parser accepts: `chrono::parse_from_rfc2822` takes inputs the old `date::from_stream` path rejected (arbitrary offsets and named zones, optional weekday, 2-digit years), so some malformed values that used to raise `FormattingError` now return a number. Existing queries over the same data can change results.

But if you want, we can also keep this in, but then it needs a test pinning the accepted format.

docs/rust-cmake-integration.md
```diff
@@ -0,0 +1,198 @@
...
+
+C++ library targets declare which Rust crates they need using a custom INTERFACE property:
+
+```cmake
+target_link_rust_lib(nes-sources nes_network_bindings)
+target_link_rust_lib(nes-common spdlog_bindings)
+target_link_rust_lib(nes-physical-operators nes_rust_timestamp)
+```
+
+This sets `INTERFACE_NES_RUST_LIBS` on the target. It's purely metadata — no actual linking happens. The finalizer recursively walks `target_link_libraries` relationships and collects the property from every reachable target.
```
yschroeder97:
```suggestion
```

Sounds like an implementation detail I don't really need to know here

docs/rust-cmake-integration.md
```diff
@@ -0,0 +1,198 @@
...
+
+At the end of CMake configuration, the build system discovers executables and
+collects their transitive Rust requirements. Each unique sorted crate set gets
+one generated umbrella static library, shared by every executable with the same
+requirements. Cargo therefore includes shared dependencies such as CXX only
+once. No per-executable Rust configuration is required.
+
+## Adding a New Rust Crate
+
+To add a new Rust crate that backs a C++ plugin (e.g. a function, source, or sink):
```
yschroeder97: What about a rust crate that is not linked with a plugin, but a regular internal lib?

Or do you talk about a plugin because this document is intended for people writing rust code in plugins specifically?

docs/rust-cmake-integration.md
```diff
@@ -0,0 +1,198 @@
...
+]
+```
+
+CMake auto-discovers it via `cargo metadata` — no changes to `EnableRust.cmake` needed. Add the crate to `[workspace.dependencies]` only when another workspace member consumes it.
+
+### 3. Create the CXX bridge CMakeLists.txt
+
+```cmake
+add_cxx_bridge(
+    my-func-bindings
```
yschroeder97:
```suggestion
    my-func-bindings # name of the C++ side of the bridge library target
```

cmake/EnableRust.cmake
```diff
@@ -63,38 +63,139 @@ endif ()
...
+    file(WRITE "${_NES_CARGO_CONFIG}" "[net]\noffline = true\n")
+else ()
+    file(REMOVE "${_NES_CARGO_CONFIG}")
 endif ()
 
 FetchContent_MakeAvailable(Corrosion)
 
+include(RustLibCollection)
+# Defer Rust link finalization using double-defer so it runs AFTER all other
+# deferred calls (e.g. generate_rust_tokio_source_registry which registers
```
yschroeder97: Please remove the examples to things that are not included in the PR.

cmake/RustLibCollection.cmake
```diff
@@ -0,0 +1,313 @@
...
+
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# RustLibCollection.cmake — Transitive Rust crate collection and per-executable umbrella generation.
+#
+# Multiple Rust staticlib crates (e.g. nes_source_bindings, nes_network_bindings) each embed
```
yschroeder97:
```suggestion
# Multiple Rust staticlib crates (e.g. nes_logger_bindings, nes_network_bindings) each embed
```

cmake/RustLibCollection.cmake
```diff
@@ -0,0 +1,313 @@
...
+# the CXX runtime, producing duplicate symbols when linked together. This module:
```
yschroeder97: More general, they may depend on the same crates.

cmake/RustLibCollection.cmake
```diff
@@ -0,0 +1,313 @@
...
+
+# Register a Rust crate's source directory so the umbrella generator can find it.
+function(register_rust_crate rust_crate manifest_dir)
+    set_property(GLOBAL PROPERTY NES_RUST_CRATE_${rust_crate}_PATH "${manifest_dir}")
+endfunction()
+
+# Add per-crate `cargo test` as ctest tests for a Rust workspace.
+#
+# Usage:
+#   add_rust_cargo_tests(nes-sources/rust nes_buffer_bindings nes_source_runtime ...)
```
yschroeder97: Again, use the example that is part of this PR

nes-common/tests/UnitTests/Util/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ add_nes_common_test(nes-common-tests
         "ReflectionTest.cpp"
         "VersionTest.cpp"
 )
-target_link_libraries(nes-common-tests nes_rust_bindings)
+target_link_libraries(nes-common-tests nes_network_bindings)
```
yschroeder97: These Util tests need the logger, not the network stack. nes-common already provides the logger through `target_link_rust_lib(nes-common spdlog_bindings)`, so this direct link to the raw nes_network_bindings staticlib looks like leftover from the rename. Linking the raw staticlib next to the spdlog umbrella also risks duplicate cxx runtime symbols.

nes-common/rust/spdlog-bindings/Cargo.toml
```diff
@@ -12,16 +12,16 @@
...
-tracing = "0.1"
-tracing-subscriber = "0.3"
-cxx = "1.0"
+tracing = { workspace = true }
+tracing-subscriber = { workspace = true }
+cxx = { workspace = true }
 
 [build-dependencies]
-cxx-build = "1.0"
+cxx-build = { workspace = true }
```
yschroeder97: This crate has no build.rs (codegen uses `add_cxx_bridge/corrosion`), so the `cxx-build` build-dependency is unused. It can be removed, along with the `[workspace.dependencies]` entry that the policy checker otherwise keeps alive for this one member.

yschroeder97 (changes requested):
Thanks, its a good step to make adding rust code easier in the future. 

- Some examples in the comments have leftovers from the io branch
- I am unsure whether we need the timestamp function FFI to prove the point
- other minor things.

Cargo.toml
ls-1801: Yes, changed it to resolver 3.

Commit: a499e99541

nes-physical-operators/rust/cast-to-unix-timestamp/src/cast_to_unix_timestamp.rs
ls-1801: We would like to keep the Rust implementation because it is much simpler than the old C++ version. I restricted it to the formats accepted before and added tests, so existing behavior should remain unchanged.

Commit: 8cfcad416a

docs/rust-cmake-integration.md
ls-1801: Agreed, removed.

Commit: f08d036a88

docs/rust-cmake-integration.md
ls-1801: They can be used anywhere like a normal C++ library. Functions, sources, and sinks are just places where we expect them frequently. I clarified this in the docs.

Commit: 0f395010f9

docs/rust-cmake-integration.md
ls-1801: Good point, added the clarification.

Commit: 11b28e3d40

cmake/EnableRust.cmake
ls-1801: Removed the examples that are not part of this PR.

Commit: 2e6da14ec7

cmake/RustLibCollection.cmake
ls-1801: Updated the example to use spdlog_bindings and nes_network_bindings, which are part of this PR.

Commit: 02b1944619

cmake/RustLibCollection.cmake
ls-1801: Right, this applies to shared dependencies in general. I updated the wording and kept CXX as an example.

Commit: 2f6f0901d7

cmake/RustLibCollection.cmake
ls-1801: Updated the example to match the root-workspace call used in this PR.

Commit: 53450950e7

nes-common/tests/UnitTests/Util/CMakeLists.txt
ls-1801: You are right, this was left over from the rename. I removed it and verified that nes-common-tests still builds and passes.

Commit: 70fcd4db86

nes-common/rust/spdlog-bindings/Cargo.toml
ls-1801: Removed it from the crate and workspace dependencies. Cargo check and tests still pass.

Commit: 9954b6e454

yschroeder97 (approved):

cmake/RustLibCollection.cmake
```diff
@@ -0,0 +1,314 @@
...
+    get_property(_rustc CACHE Rust_COMPILER_CACHED PROPERTY VALUE)
+    list(APPEND _test_env "RUSTC=${_rustc}")
+    if (NES_RUST_ENV_VARS)
+        list(APPEND _test_env ${NES_RUST_ENV_VARS})
+    endif ()
+
+    set(crate_names ${ARGN})
+    set(_prev_test "")
+    foreach (crate ${crate_names})
+        add_test(
```
keyseven123: These two crates have a C++ half that only CMake compiles, so `cargo test -p` can't link them.

cmake/EnableRust.cmake
```diff
@@ -63,38 +63,138 @@ endif ()
...
+        MANIFEST_PATH ${PROJECT_SOURCE_DIR}/Cargo.toml
+        CRATES ${_staticlib_crates}
+        IMPORTED_CRATES ALL_RUST_CRATES
         CRATE_TYPES staticlib
         FLAGS ${ADDITIONAL_CARGOFLAGS}
 )
 
-# Arguments are passed to cargo via an environment which we attach to the  nes_rust_bindings target and is loaded by corrosion
+# Add Rust unit tests to ctest (one test per workspace crate, run via `cargo test`)
+add_rust_cargo_tests("${PROJECT_SOURCE_DIR}" ${_all_crate_names})
```
keyseven123: `NES_RUST_ENV_VARS` is only set at :211, so on a fresh configure the Rust tests run unsanitized. Move this call below it.

nes-physical-operators/rust/cast-to-unix-timestamp/src/cast_to_unix_timestamp.rs
```diff
@@ -0,0 +1,120 @@
...
+    }
+}
+
+fn parse_timestamp_to_unix_milliseconds(timestamp: &str) -> Result<u64, String> {
+    let timestamp = timestamp.trim();
+    if timestamp.is_empty() {
+        return Err("cannot convert an empty timestamp".to_owned());
+    }
+
+    let offset_formats = ["%Y-%m-%dT%H:%M:%S%:z", "%Y-%m-%dT%H:%M:%S%.f%:z"];
```
keyseven123, line 30: chrono's `%:z` makes the colon optional, so `+0100` now parses — behavior change vs the old `%Ez`.

scripts/hash_dependencies.sh
```diff
@@ -22,18 +22,22 @@ set -euo pipefail
 # cd to root of git repo
 cd "$(git rev-parse --show-toplevel)"
 
+# Member manifests are deliberately excluded from the dependency image hash. Enforce the
+# invariants that make that safe: members can only inherit dependencies from the workspace,
+# and the workspace cannot keep dependencies that are not yet used (and therefore not vendored).
+python3 scripts/check_rust_dependency_policy.py
```
keyseven123, line 28: A non-zero exit here silently yields an empty `VCPKG_HASH` — the CMake callers don't check the status.

cmake/RustLibCollection.cmake
```diff
@@ -0,0 +1,314 @@
...
+    if (NOT "${NES_RUST_ENV_VARS}" STREQUAL "")
+        set_property(TARGET ${umbrella_name}
+                PROPERTY CORROSION_ENVIRONMENT_VARIABLES ${NES_RUST_ENV_VARS})
+    endif ()
+
+    _link_umbrella_with_bridges(${exe_target} ${umbrella_name} "${rust_libs}")
+endfunction()
+
+# Link the umbrella in a sandwich: umbrella → bridges → umbrella.
+# With the header/symbol split, C++ consumers only link -headers targets,
```
keyseven123: No `-headers` target exists — consumers link the bridges directly. Comment describes an invariant the build never had.

nes-common/rust/spdlog-bindings/CMakeLists.txt
```diff
@@ -13,14 +13,11 @@
...
-        FILES ../spdlog/lib.rs
-        # REGEN_TARGET builds only the generated headers, without cargo or the C++ sources.
-        REGEN_TARGET nes-logger-bindings-codegen
+        CRATE spdlog_bindings
+        FILES lib.rs
 )
-add_dependencies(nes-codegen nes-logger-bindings-codegen)
 target_sources(nes-logger-bindings PRIVATE src/LoggerBindings.cpp)
-target_link_libraries(nes-logger-bindings PUBLIC nes-common spdlog::spdlog magic_enum::magic_enum)
+target_link_libraries(nes-logger-bindings PRIVATE spdlog::spdlog magic_enum::magic_enum)
```
keyseven123: `LoggerBindings.hpp` is exported PUBLIC and includes <spdlog/logger.h>, so spdlog should stay PUBLIC. Works today only because nes-common links it too.

nes-physical-operators/src/Functions/CastToUnixTimestampPhysicalFunction.cpp
```diff
@@ -127,39 +40,35 @@ CastToUnixTimestampPhysicalFunction::CastToUnixTimestampPhysicalFunction(Physica
...
-            return VarVal{0, true, true}.castToType(outputType.type);
-        }
+        return VarVal{0, true, true}.castToType(outputType.type);
     }
 
     const auto var = value.getRawValueAs<VariableSizedData>();
-    const auto size = var.getSize();
-    const auto ptr = var.getContent();
-
     const auto parsedMilliSeconds = nautilus::invoke(
```
keyseven123, line 49: This proxy throws on malformed input, so it needs `nautilus::invokeGuarded` (and CPPTRACE_TRY/CATCH). Pre-existing, but this PR rewrites the call site.

flake.nix
```diff
@@ -4,15 +4,23 @@
...
     flake-utils.lib.eachSystem [ "x86_64-linux" "aarch64-linux" ] (
       system:
       let
-        pkgs = import nixpkgs { inherit system; };
+        pkgs = import nixpkgs {
+          inherit system;
+          overlays = [ (import rust-overlay) ];
+        };
         lib = pkgs.lib;
+        rustToolchain = pkgs.rust-bin.stable.latest.default;
```
keyseven123, line 23: Devshell uses stable, but `rust-toolchain.toml` pins nightly — needed for the `-Zbuild-std` sanitizer configs. Use `fromRustupToolchainFile ./rust-toolchain.toml`.

docs/rust-cmake-integration.md
```diff
@@ -0,0 +1,211 @@
...
+
+Instead of linking individual Rust staticlibs into C++ executables, the build system generates **umbrella crates** that bundle all needed Rust crates into a single staticlib. Executables that need the exact same set of crates share a single umbrella — the crate set is hashed to produce a stable key, so 37 executables needing the same crates produce only 1 Cargo build.
+
+## Unified Workspace
+
+All Rust crates live in a single Cargo workspace rooted at `/Cargo.toml`:
+
+```toml
+[workspace]
+resolver = "2"
```
keyseven123: Says `resolver = "2"`; root `Cargo.toml:17` uses `"3"`.

keyseven123 (changes requested):

cmake/RustLibCollection.cmake
ls-1801: Agreed. Cargo-test registration is now explicit through `add_rust_cargo_test`; only `nes_network` and `nes_rust_timestamp`, which link without CMake-built C++ objects, opt in.

cmake/EnableRust.cmake
ls-1801: Fixed as part of explicit registration. Cargo tests are now registered from the crate CMake files after `NES_RUST_ENV_VARS` has been configured.

nes-physical-operators/rust/cast-to-unix-timestamp/src/cast_to_unix_timestamp.rs
ls-1801: Is that a problem? `+01:00` is the ISO 8601 extended form and `+0100` is the basic form, both of which occur in real-world data. I added a test explicitly documenting support for `+0100`.

scripts/hash_dependencies.sh
ls-1801: Fixed. Both callers now use a shared `get_nes_dependency_hash` helper that captures stdout and stderr, stops configuration with the actual script error, and calculates the hash only once per configure run.

cmake/RustLibCollection.cmake
ls-1801: Agreed; that comment described an obsolete design and has been removed.

nes-common/rust/spdlog-bindings/CMakeLists.txt
ls-1801: Agreed. `spdlog::spdlog` is now PUBLIC because it is required by the exported header, while `magic_enum` remains PRIVATE.

nes-physical-operators/src/Functions/CastToUnixTimestampPhysicalFunction.cpp
ls-1801: `invokeGuarded` is not available in the Nautilus version used by this branch. I tested malformed Rust input through both interpreted and compiled `nautilus::invoke`; it safely propagates as an NES exception with the `FormattingError` code, so I left the call unchanged.

flake.nix
ls-1801: The Nix change here was only the minimum needed to provide Rust in the development shell. Aligning it with `rust-toolchain.toml` is left for @adrianmichalke as a follow-up if he ever needs  this.

docs/rust-cmake-integration.md
ls-1801: Fixed; the documentation now uses resolver 3, matching the root workspace.

keyseven123 (approved):

PR #1885, opened 2026-08-13
feat(functions): Add Conditional (CASE WHEN) function

### LINKED ISSUE #1370: [TASK] Create `CASE WHEN` function
### Description

Add the `CASE WHEN` conditional function: 
```sql
CASE
    WHEN condition1 THEN result1
    WHEN condition2 THEN result2
    WHEN conditionN THEN resultN
    ELSE result
END; 
```

It resembles an if-then-else chain, where the then expressions return the same type.
If no condition evaluates to true, the else expression is evaluated.

### Acceptance Criteria

- [ ] Add logical function
- [ ] Add physical function
- [ ] Add lowering
- [ ] Add E2E tests
- [ ] Add function to parser grammar

### LINKED ISSUE #1866: Case When Function
We should add a case when function, similar to the general SQL function.

yschroeder97:
## Summary

Implements the Conditional (CASE WHEN) function on top of current `main`, ported from the tutorial branch and adapted to how functions are written in current main.

## What's included

**Logical function** (`nes-logical-operators`)
- `ConditionalLogicalFunction` flat child list `[cond1, res1, …, condN, resN, default]`. Validates that conditions are BOOLEAN and results share the default's type.

**Physical function** (`nes-physical-operators`)
- `ConditionalPhysicalFunction` evaluates conditions in order, returns the first matching branch, else the default. Iterates `WhenThenPhysicalFunction` pairs.

## Validation

**Systests**: `nes-systests/function/Conditional.test`: priority ordering, numeric (INT32) results with boundary edges, deeply nested Conditionals, a Conditional used as another Conditional's condition, a nested arithmetic function inside a condition, and argument/type error cases (`ERROR 2002` / `ERROR 2006`).

**Unit tests**: limited to behaviour the systests cannot observe: value equality (`operator==`, used by the optimizer), reflection, and `explain()` form (EXPLAIN renders projections by field name, not by expression).

## Invoking it

The parser has **no dedicated `CASE WHEN` syntax yet**, so the function is called in function-call form:

```sql
Conditional(cond1, result1, cond2, result2, …, default)
```

`explain()` renders the same function-call form.

## Error handling

- Invalid **user input** is reported as query errors: wrong argument count → `CannotDeserialize`; non-BOOLEAN condition or mismatched result types → `DifferentFieldTypeExpected`.
- `INVARIANT` / `PRECONDITION` are used for structural/lowering bugs (they terminate), not user input.

## Follow-up work
- **Parser support for the real `CASE WHEN … THEN … ELSE … END` syntax**.

### DIFF
--- a/nes-plugins/Functions/Conditional/ConditionalLogicalFunction.cpp
+++ b/nes-plugins/Functions/Conditional/ConditionalLogicalFunction.cpp
@@ -1,0 +1,178 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <Functions/ConditionalLogicalFunction.hpp>
+
+#include <iterator>
+#include <ranges>
+#include <string>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+#include <DataTypes/DataType.hpp>
+#include <DataTypes/DataTypeProvider.hpp>
+#include <Functions/LogicalFunction.hpp>
+#include <Schema/Field.hpp>
+#include <Schema/Schema.hpp>
+#include <Schema/SchemaFwd.hpp>
+#include <Serialization/LogicalFunctionReflection.hpp>
+#include <Util/PlanRenderer.hpp>
+#include <Util/Reflection.hpp>
+#include <fmt/format.h>
+#include <fmt/ranges.h>
+#include <ErrorHandling.hpp>
+#include <LogicalFunctionRegistry.hpp>
+
+namespace NES
+{
+namespace
+{
+/// Groups the leading elements of the flat child list into (when, then) pairs, leaving out the
+/// trailing default result.
+std::vector<WhenThenLogicalFunction> groupWhenThens(const std::vector<LogicalFunction>& children)
+{
+    INVARIANT(
+        children.size() >= 3 and children.size() % 2 == 1,
+        "ConditionalLogicalFunction requires an odd number of children >= 3 (when/then pairs plus a default), but got {}",
+        children.size());
+
+    std::vector<WhenThenLogicalFunction> whenThens;
+    whenThens.reserve(children.size() / 2);
+    for (auto when = children.begin(); when != std::prev(children.end()); when += 2)
+    {
+        whenThens.push_back(WhenThenLogicalFunction{.when = *when, .then = *std::next(when)});
+    }
+    return whenThens;
+}
+}
+
+/// The result type stays untouched here; it is inferred once the input schema is known.
+ConditionalLogicalFunction::ConditionalLogicalFunction(std::vector<WhenThenLogicalFunction> whenThens, LogicalFunction elseCase)
+    : whenThens(std::move(whenThens)), elseCase(std::move(elseCase))
+{
+    INVARIANT(not this->whenThens.empty(), "ConditionalLogicalFunction requires at least one when/then pair");
+}
+
+ConditionalLogicalFunction::ConditionalLogicalFunction(const std::vector<LogicalFunction>& children)
+    : whenThens(groupWhenThens(children)), elseCase(children.back())
+{
+}
+
+DataType ConditionalLogicalFunction::getDataType() const
+{
+    return dataType;
+}
+
+std::vector<LogicalFunction> ConditionalLogicalFunction::getChildren() const
+{
+    std::vector<LogicalFunction> children;
+    children.reserve((whenThens.size() * 2) + 1);
+    for (const auto& [when, then] : whenThens)
+    {
+        children.push_back(when);
+        children.push_back(then);
+    }
+    children.push_back(elseCase);
+    return children;
+}
+
+ConditionalLogicalFunction ConditionalLogicalFunction::withChildren(const std::vector<LogicalFunction>& children) const
+{
+    auto copy = *this;
+    copy.whenThens = groupWhenThens(children);
+    copy.elseCase = children.back();
+    return copy;
+}
+
+std::string_view ConditionalLogicalFunction::getType() const
+{
+    return NAME;
+}
+
+bool ConditionalLogicalFunction::operator==(const ConditionalLogicalFunction& rhs) const
+{
+    return whenThens == rhs.whenThens and elseCase == rhs.elseCase;
+}
+
+std::string ConditionalLogicalFunction::explain(ExplainVerbosity verbosity) const
+{
+    /// Rendered in function-call form because the SQL parser does not support the CASE WHEN syntax yet.
+    const auto children = getChildren();
+    const auto explained = children | std::views::transform([verbosity](const LogicalFunction& child) { return child.explain(verbosity); })
+        | std::ranges::to<std::vector>();
+    return fmt::format("{}({})", NAME, fmt::join(explained, ", "));
+}
+
+LogicalFunction ConditionalLogicalFunction::withInferredDataType(const Schema<Field, Unordered>& schema) const
+{
+    /// The default result carries the type of the whole expression.
+    const auto inferredElseCase = elseCase.withInferredDataType(schema);
+    auto resultType = inferredElseCase.getDataType();
+
+    std::vector<WhenThenLogicalFunction> inferredWhenThens;
+    inferredWhenThens.reserve(whenThens.size());
+    for (const auto& [when, then] : whenThens)
+    {
+        auto inferredWhen = when.withInferredDataType(schema);
+        auto inferredThen = then.withInferredDataType(schema);
+        if (not inferredWhen.getDataType().isType(DataType::Type::BOOLEAN))
+        {
+            throw DifferentFieldTypeExpected("CASE WHEN condition must be BOOLEAN, but got {}", inferredWhen.getDataType());
+        }
+        if (not inferredThen.getDataType().isType(resultType.type))
+        {
+            throw DifferentFieldTypeExpected(
+                "CASE WHEN results must all share the default result's type {}, but got {}", resultType, inferredThen.getDataType());
+        }
+        /// A branch that can produce a null makes the whole expression nullable, as in SQL.
+        resultType = DataTypeProvider::provideDataType(resultType.type, resultType.joinNullable(inferredThen.getDataType()));
+        inferredWhenThens.push_back(WhenThenLogicalFunction{.when = std::move(inferredWhen), .then = std::move(inferredThen)});
+    }
+
+    auto inferred = ConditionalLogicalFunction{std::move(inferredWhenThens), inferredElseCase};
+    inferred.dataType = resultType;
+    return inferred;
+}
+
+Reflected
+Reflector<ConditionalLogicalFunction>::operator()(const ConditionalLogicalFunction& function, const ReflectionContext& context) const
+{
+    return context.reflect(detail::ReflectedConditionalLogicalFunction{.children = function.getChildren()});
+}
+
+ConditionalLogicalFunction
+Unreflector<ConditionalLogicalFunction>::operator()(const Reflected& reflected, const ReflectionContext& context) const
+{
+    const auto [children] = context.unreflect<detail::ReflectedConditionalLogicalFunction>(reflected);
+    /// The shape of a serialized plan is not trusted, so a bad child count is reported rather than asserted.
+    if (children.size() < 3 or children.size() % 2 == 0)
+    {
+        throw CannotDeserialize("ConditionalLogicalFunction requires an odd number of children >= 3, but got {}", children.size());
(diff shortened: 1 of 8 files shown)


nes-plugins/Functions/Conditional/include/Functions/ConditionalLogicalFunction.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include <Util/Logger/Formatter.hpp>
+#include <Util/PlanRenderer.hpp>
+#include <Util/ReflectionFwd.hpp>
+
+namespace NES
+{
+
+/// Represents a CASE WHEN expression. Children are stored as a flat list:
+/// [condition1, result1, condition2, result2, ..., conditionN, resultN, default]
+/// The list always holds an odd number of elements >= 3: at least one (condition, result) pair
```
Artraxon: why? Doesn't it make more sense to store directly the whenThen Pairs + else branch? In that case you know that each ConditionalLogicalFunction always has at least the correct number of children

nes-plugins/Functions/Conditional/include/Functions/ConditionalLogicalFunction.hpp
```diff
@@ -0,0 +1,86 @@
...
+    std::vector<LogicalFunction> children;
+
+    friend Reflector<ConditionalLogicalFunction>;
+};
+
+namespace detail
+{
+struct ReflectedConditionalLogicalFunction
+{
+    std::vector<std::optional<LogicalFunction>> children;
```
Artraxon: you don't need to wrap LogicalFunctions in std:optional anymore.

(Of course the comment about the whenThens + else also applies here)

nes-plugins/Functions/Conditional/ConditionalLogicalFunction.cpp
```diff
@@ -0,0 +1,177 @@
...
+}
+
+ConditionalLogicalFunction::ConditionalLogicalFunction(std::vector<LogicalFunction> children)
+    : dataType(DataTypeProvider::provideDataType(DataType::Type::UNDEFINED)), children(std::move(children))
+{
+    INVARIANT(
+        this->children.size() >= 3 and this->children.size() % 2 == 1,
+        "ConditionalLogicalFunction requires an odd number of children >= 3 (when/then pairs plus a default), but got {}",
+        this->children.size());
+    dataType = this->children.back().getDataType();
```
Artraxon: As a convention, we never set the datatype in the constructor (neither to undefined or to anything else), always only in the withInferredDataType().
Also, you are not checking here that the type of the last child is the same as of the other branches.

IMO if you do the size check here anyway I don't see any reason why you can't use better typed out children.

nes-plugins/Functions/Conditional/ConditionalLogicalFunction.cpp
```diff
@@ -0,0 +1,177 @@
...
+    return ConditionalLogicalFunction{std::move(children)};
+}
+
+LogicalFunctionRegistryReturnType
+LogicalFunctionGeneratedRegistrar::RegisterConditionalLogicalFunction(LogicalFunctionRegistryArguments arguments)
+{
+    /// The argument count comes straight from the user's query, so a bad count is reported rather than asserted.
+    if (arguments.children.size() < 3 or arguments.children.size() % 2 == 0)
+    {
+        throw CannotDeserialize(
```
Artraxon, line 172: You can ignore this, but I am surprised that we are always throwing CannotDeserialize and not something more apt

Artraxon (changes requested):
Just small changes. On a more general note, what is the reason on why we would want to put this into nes-plugins? This is quite a foundational function and it doesn't even carry any dependencies, IMO it should go into nes-logical-operators

nes-plugins/Functions/Conditional/ConditionalLogicalFunction.cpp
Artraxon: https://github.com/nebulastream/nebulastream/issues/1892

nes-plugins/Functions/Conditional/include/Functions/ConditionalLogicalFunction.hpp
yschroeder97: Done. The function now stores `std::vector<WhenThenLogicalFunction>` plus `elseCase`, like the physical function. The flat list only exists at the `getChildren` and `withChildren` boundary, so the shape is validated once, when it is grouped.

nes-plugins/Functions/Conditional/include/Functions/ConditionalLogicalFunction.hpp
yschroeder97: Removed. The reflected struct is now a plain `std::vector<LogicalFunction>`, which also drops the missing child check in the unreflector.

nes-plugins/Functions/Conditional/ConditionalLogicalFunction.cpp
yschroeder97: Fixed. No constructor touches `dataType` any more, it is assigned only in `withInferredDataType`, and the children are typed out as suggested.

The type check does happen, in `withInferredDataType`. `resultType` comes from the inferred default result and every branch result is compared against it. The constructor cannot check it because no schema is known there yet.

nes-plugins/Functions/Conditional/ConditionalLogicalFunction.cpp
yschroeder97: Agreed, thanks for filing #1892. I keep `CannotDeserialize` here so this function stays consistent with the other registrars. In the unreflector it stays anyway, that path really is deserialization.

yschroeder97:
Thanks for the review. All four inline comments are addressed in the amended commit.

The children are stored as `std::vector<WhenThenLogicalFunction>` plus `elseCase`, like in the physical function. The `std::optional` wrapper is gone from the reflected struct. `dataType` is set only in `withInferredDataType`. The registrar keeps `CannotDeserialize` for now, tracked in #1892.

On nes-plugins vs nes-logical-operators I would like to keep it where it is for this PR, because this function is the worked example of the guide in #1890.

nes-plugins/Functions/Conditional/ConditionalPhysicalFunction.cpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+}
+
+VarVal ConditionalPhysicalFunction::execute(const Record& record, ArenaRef& arena) const
+{
+    for (const auto& [when, then] : nautilus::static_iterable(whenThens))
+    {
+        if (when.execute(record, arena))
+        {
+            return then.execute(record, arena);
```
PhilippGrulich: please create a test with many cases (n=100) and check how this impacts compilation time. I think the multiple returns here could lead to an trace explosion.

nes-plugins/Functions/Conditional/ConditionalPhysicalFunction.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <ErrorHandling.hpp>
+#include <ExecutionContext.hpp>
+#include <PhysicalFunctionRegistry.hpp>
+#include <static.hpp>
+
+namespace NES
+{
+ConditionalPhysicalFunction::ConditionalPhysicalFunction(std::vector<PhysicalFunction> childFunctions) : elseCase{childFunctions.back()}
+{
+    /// The trailing child is the default result; the remaining children form (when, then) pairs.
```
keyseven123: I do not like this at all. Please pass it as a separate parameter.

nes-plugins/Functions/Conditional/ConditionalPhysicalFunction.cpp
keyseven123: @yschroeder97 you could do it via the position, so remember what position was correct and then call it after the loop.

keyseven123 (changes requested):

nes-plugins/Functions/Conditional/ConditionalPhysicalFunction.cpp
yschroeder97: Done. The constructor now takes the (when, then) pairs and the default result as separate parameters, plus the declared result type. The flat child list is only regrouped in the registrar, which already checks the shape.

nes-plugins/Functions/Conditional/ConditionalPhysicalFunction.cpp
yschroeder97: Measured it; there is no trace explosion: both grow linearly in the number of branches, even at n=100.

The multiple returns did cost a constant factor, though, so I applied the fix. Numbers from a Debug build on the MLIR backend, tracing the physical function directly and reading CompilationStatistics:

| branches | before: IR blocks | before: IR ops | before: compile | after: IR blocks | after: IR ops | after: compile |
|---|---|---|---|---|---|---|
| 1 | 10 | 39 | 5.2 ms | 4 | 22 | 4.6 ms |
| 10 | 82 | 318 | 10.5 ms | 31 | 148 | 7.8 ms |
| 25 | 202 | 783 | 25.0 ms | 76 | 358 | 13.9 ms |
| 50 | 402 | 1558 | 63.0 ms | 151 | 708 | 26.5 ms |
| 100 | 802 | 3108 | 138.4 ms | 301 | 1408 | 66.3 ms |

So the early return costs 8 IR blocks per branch; the single exit costs 3. At 100 branches, that is 2.7x less trace and about half the query compilation time.

Note: this is Nautilus query compilation time, not C++ build time.

yschroeder97: Applied suggestion of @keyseven123, with one deviation. Remembering the position and dispatching after the loop does not work directly because the matching position is a runtime value while the branch list is unrolled at trace time, so it cannot be used to index into the branches afterward. What works is remembering the result rather than the position: a flag records that a branch has already matched, the result accumulates in a single variable, and there is a single return.

The trace goes from 8 IR blocks per branch to 3, and query compilation for 100 branches goes from ~138 ms to ~66 ms.

The trade-off is that the conditions of the later branches are now evaluated even after a branch matches. Still, only the taken branch is computed.

yschroeder97:
Pushed an amended commit addressing the open review comments.

**Constructor** (@keyseven123): the physical function takes the (when, then) pairs and the default result as separate parameters now, with the flat list regrouped in the registrar.

**Trace size and compilation time** (@PhilippGrulich, @keyseven123): measured over 1 to 100 branches. There is no trace explosion, both the old and new form grow strictly linearly. The early return per branch did carry a constant factor of 8 IR blocks per branch against 3 for a single exit, so I switched to accumulating the result and returning once. At 100 branches that is 802 to 301 IR blocks and ~138 ms to ~66 ms of query compilation. Details in the thread.

**Null handling**: reviewing the above surfaced that the null semantics were written before nullability existed and were not SQL compliant. Two fixes:

- A branch that can produce a null now makes the whole expression nullable, joined across all results and the default. Previously every result had to match the default exactly including its nullable flag, so `Conditional(cond, nullable_field, INT32(-1))` was rejected outright rather than widened.
- A condition that is NULL is not true, so it does not select its branch. That already held via the boolean conversion on the value wrapper; it is now explicit, because the rewritten loop no longer goes through that conversion.

`nes-systests/function/Conditional.test` gained a case covering all three: a NULL condition falling through, results that disagree on nullability, and a NULL result flowing out. It fails without the type inference change.

Still open from your first review @Artraxon: whether this belongs in `nes-plugins` or `nes-logical-operators`. My preference is still to keep it here for this PR since it is the worked example for the guide in #1890, but happy to move it if you would rather have it settled first.

nes-plugins/Functions/Conditional/ConditionalPhysicalFunction.cpp
```diff
@@ -0,0 +1,97 @@
...
+{
+}
+
+/// The branch loop is unrolled during tracing, so a return inside it adds one exit block per
+/// branch. Accumulating into one variable keeps a single exit and roughly halves trace size and
+/// compilation time. The price is that every condition runs even after a branch matched, so a
+/// condition must be free of side effects. Results are still only evaluated for the branch taken.
+VarVal ConditionalPhysicalFunction::execute(const Record& record, ArenaRef& arena) const
+{
+    VarVal result = elseCase.execute(record, arena);
```
PhilippGrulich: In general I think the solution is fine, however it could be that it's not valid to execute the default case before the when cases.

CASE WHEN x==0 THEN 0 ELSE AS y/x

If possible you could initialize the result with a default value.

Artraxon (approved):

nes-plugins/Functions/Conditional/ConditionalPhysicalFunction.cpp
yschroeder97: Actually, I think the added complexity is not worth the gain, the plain loop has at most 2.7x less performance, and that only for n=100, which is far more than what the majority of queries containing CASE WHEN will actually look like. 

Therefore, I would like to stay with the plain loop.

yschroeder97: @keyseven123 @PhilippGrulich 

Settled now with the plain loop with two exits. Added a comment explaining the issue and that it might need to be revisited. 

This function is also used in the guide and the NES tutorial, so we should use it as an example to help you write simple code while leveraging query compilation.

keyseven123 (approved):
As I had some offline discussion with @yschroeder97 , I think this is fine for a tutorial function.

PR #1890, opened 2026-08-14
feat(docs): How to add a function plugin

### LINKED ISSUE #972: [DOC] Add function development guide
### What do you see as an issue?

_No response_

### Solving the problem

Add a guide on how to add a new function as a plugin.

### Code of Conduct

yschroeder97:
Stacked on top of #1885 (base branch `ys/case-when-fn`).

Adds `docs/guide/how_to_add_a_function.md`, a guide for adding a function as an optional plugin. It covers:

- the two parts of a function: the logical function used during planning and the physical function used during execution
- plugin registration for both parts, plus unreflection for the logical function, in one `CMakeLists.txt`
- the logical and physical interfaces and the registration function naming
- the `Conditional` function as a worked example

### DIFF
--- a/docs/guide/how_to_add_a_function.md
+++ b/docs/guide/how_to_add_a_function.md
@@ -1,0 +1,253 @@
+# How to add a `Function`
+
+## 1. What is a Function
+
+A function computes a single value from a single record and has no view of the stream.
+It reads field values and constants, applies some logic, and returns one value.
+The value is a single field, but that field may be of a composite type or hold variable-sized data such as text or an image.
+Examples are `a + b`, `CHAR_LENGTH(name)`, and the CASE WHEN expression used as the example in this guide.
+
+Functions appear inside operators.
+The predicate of a Selection is a function that returns a boolean.
+Each output column of a Projection is a function.
+A join or aggregation uses functions to express keys and aggregated values.
+Without functions an operator could only pass fields through unchanged, so functions are how a query expresses computation over the data.
+
+### Function versus Operator
+
+An operator and a function work at different levels.
+
+- An operator works at the level of the stream. It decides which records pass, and how they are grouped, joined, filtered, or combined. It consumes and produces streams of records, windows, or tables.
+- A function is evaluated once per record. Its input is one or more values from that record and its output is one value. It has no view of the stream.
+
+Operators contain functions, and functions nest.
+A Selection operator holds a predicate: a tree of functions that returns a boolean, for example an AND function whose children are two comparisons.
+Each output field of a Projection is computed by such a function tree.
+When you add arithmetic, a comparison, or a string operation to the system, you add a function, not an operator.
+
+```mermaid
+graph LR
+    IN["record stream<br/>(a=1, b=9)<br/>(a=7, b=8)"] --> SEL
+    subgraph SEL["Selection operator: which records pass"]
+        FN["predicate function: one value per record<br/>a + b &gt; 10"]
+    end
+    SEL --> OUT["record stream<br/>(a=7, b=8)"]
+```
+
+The operator sees the stream, the function sees one record and produces one value.
+
+## 2. Logical and Physical Functions
+
+A function has two representations, and you implement both.
+
+The logical function is used during query planning and optimization.
+It knows its child functions and its result data type.
+Given the schema of the input, it infers the result type of the expression, which is how the enclosing operator computes its own output schema.
+The logical function answers two questions: what is the type of this expression, and is the expression valid.
+
+The physical function is used during query compilation and execution.
+It defines the execution semantics: given a record at runtime, produce the result value.
+It is lowered from the logical function and compiled by nautilus into executable code.
+The physical function answers one question: how is the value computed.
+
+Two representations exist because planning and execution need different information.
+Planning works on types and structure and never touches record values.
+Execution works on record values and never reasons about schemas.
+Keeping them apart means the optimizer stays independent of the runtime, and the runtime stays free of planning concerns.
+The two connect by name: lowering looks up the physical function using the logical function's registry key, so the names must match.
+
+## 3. Overview
+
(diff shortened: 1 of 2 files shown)


oczerwia-2010 (approved):
LGTM

docs/guide/how_to_add_a_function.md
```diff
@@ -0,0 +1,318 @@
+# How to add a `Function`
+
+## 1. What is a Function
+
+A function computes a single value from a record.
```
alepping: what is a single value though?
If values can be complex, it could be a vector/list of values.

In my mind its more that it computes a single row/tuple, whereas an operator may produce an unbound stream of records or windows/tables (like CQL)

docs/guide/how_to_add_a_function.md
```diff
@@ -0,0 +1,318 @@
...
+The predicate of a Selection is a function that returns a boolean.
+Each output column of a Projection is a function.
+A join or aggregation uses functions to express keys and aggregated values.
+Without functions an operator could only pass fields through unchanged, so functions are how a query expresses computation over the data.
+
+### Function versus Operator
+
+An operator and a function work at different levels.
+
+- An operator transforms a stream of records. It decides which records pass, and how they are grouped, joined, filtered, or combined. Its input and output are records.
```
alepping: The output could also be windows/tables

docs/guide/how_to_add_a_function.md
```diff
@@ -0,0 +1,318 @@
...
+- A function transforms values inside a single record. Its input is one or more values and its output is one value. It has no view of the stream.
+
+Operators contain functions.
+A Selection operator holds one predicate function.
+A Projection operator holds one function per output field.
```
alepping: why only one? could also be multiple (nested)

docs/guide/how_to_add_a_function.md
```diff
@@ -0,0 +1,318 @@
...
+
+### Function versus Operator
+
+An operator and a function work at different levels.
+
+- An operator transforms a stream of records. It decides which records pass, and how they are grouped, joined, filtered, or combined. Its input and output are records.
+- A function transforms values inside a single record. Its input is one or more values and its output is one value. It has no view of the stream.
+
+Operators contain functions.
+A Selection operator holds one predicate function.
```
alepping: could be multiple

docs/guide/how_to_add_a_function.md
alepping: the general idea is that a value can be complex, e.g., VARSIZED, which can represent an image

docs/guide/how_to_add_a_function.md
```diff
@@ -0,0 +1,318 @@
...
+5) The activation of the function plugin
+
+The following is the commented `CMakeLists.txt` file for the Conditional function plugin.
+
+```cmake
+# Load Plugin Registry Utils
+include(${PROJECT_SOURCE_DIR}/cmake/PluginRegistrationUtil.cmake)
+include(${PROJECT_SOURCE_DIR}/cmake/UnreflectionRegistrationUtil.cmake)
+
+# 1) Register the logical function with the LogicalFunction registry owned by nes-logical-operators.
```
alepping: did you adapt this to the latest changes of @Artraxon ?

alepping: the syntax changed somewhat. also, there is no registrar anymore

alepping (changes requested):
I don't think this is up to date with the current plugin implementation, which changed significantly due to the changes by @Artraxon.

It also shows a general issue with documentation that is not live/tested code.
Imo, you can remove the concrete examples of how to write a plugin and simply point to the code.
If you feel fancy, you could also bootstrap the documentation from the live/tested code

docs/guide/how_to_add_a_function.md
yschroeder97: Adopted. The definition is now framed per record, and the value's type is qualified:

> A function computes a single value from a single record and has no view of the stream.
> It reads field values and constants, applies some logic, and returns one value.
> The value is a single field, but that field may be of a composite type or hold variable-sized data such as text or an image.

docs/guide/how_to_add_a_function.md
yschroeder97: Adopted. The comparison now reads:

> - An operator works at the level of the stream. It decides which records pass, and how they are grouped, joined, filtered, or combined. It consumes and produces streams of records, windows, or tables.
> - A function is evaluated once per record. Its input is one or more values from that record and its output is one value. It has no view of the stream.

docs/guide/how_to_add_a_function.md
yschroeder97: Adopted. The line now describes a function tree:

> A Selection operator holds a predicate: a tree of functions that returns a boolean, for example an AND function whose children are two comparisons.

docs/guide/how_to_add_a_function.md
yschroeder97: Adopted. Nesting is now explicit. The paragraph opens with "Operators contain functions, and functions nest." and the line reads:

> Each output field of a Projection is computed by such a function tree.

docs/guide/how_to_add_a_function.md
yschroeder97: Adopted. The commented CMakeLists listing and the registration function section are removed. Section 4 now summarizes how registration works (registry entries declared in CMake, registration at runtime) and points to `nes-plugins/Functions/Conditional/CMakeLists.txt`. Section 8 describes the static `create<PluginName>` factory pattern that replaced the generated registrars and points to the plugin sources. The branch is rebased onto main, so the referenced plugin is the current implementation.

yschroeder97:
The guide now references the plugin that is in the tree. The concrete registration examples are removed in favor of pointers to that code. The conceptual sections (interface, type inference, execution) keep their snippets; I checked that they match the current implementation.

alepping (approved):

PR #1894, opened 2026-08-14
fix(CI) Several bug fixes

keyseven123:
This PR adds three bug fixes:
1. Fixed a bug that would leave the network channel not respond, resulting in a deadlock of the query
2. Fixed a bug that would not stop local plans correctly
3. Fixes wrong page size allocation

## Verifying this change
This change is tested by newly added tests.

### DIFF
--- a/nes-frontend/src/QueryManager/QueryManager.cpp
+++ b/nes-frontend/src/QueryManager/QueryManager.cpp
@@ -32,6 +32,7 @@
 #include <Runtime/Execution/QueryStatus.hpp>
 #include <Util/Logger/Logger.hpp>
 #include <Util/Pointers.hpp>
+#include <cpptrace/from_current_macros.hpp>
 #include <fmt/ranges.h>
 #include <DistributedLogicalPlan.hpp>
 #include <DistributedQuery.hpp>
@@ -152,6 +153,29 @@ void QueryManager::QueryManagerBackends::rebuildBackendsIfNeeded() const
 
     if (!exceptions.empty())
     {
+        /// Local plans that already started are not tracked in `state.queries` on this path, so nobody would ever stop them. A network
+        /// sink whose downstream local plan never started finds no receiving end for its channel, so the receiver denies it. The
+        /// sender treats a denied channel as a transient failure and retries forever, which blocks worker shutdown. Thus, we stop the
+        /// partial deployment here.
+        for (const auto& [host, startedQueryIds] : localQueries)
+        {
+            for (const auto& localQueryId : startedQueryIds)
+            {
+                CPPTRACE_TRY
+                {
+                    if (const auto result = backends.at(host).stop(localQueryId); not result)
+                    {
+                        NES_WARNING(
+                            "Could not stop local query {} on {} while unwinding a failed start: {}", localQueryId, host, result.error());
+                    }
+                }
+                CPPTRACE_CATCH(const std::exception& exception)
+                {
+                    NES_WARNING(
+                        "Could not stop local query {} on {} while unwinding a failed start: {}", localQueryId, host, exception.what());
+                }
+            }
+        }
         return std::unexpected(exceptions);
     }
 
--- a/nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
+++ b/nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
@@ -32,6 +32,7 @@
 #include <DataTypes/VarVal.hpp>
 #include <DataTypes/VariableSizedData.hpp>
 #include <Interface/PagedVector/PagedVector.hpp>
+#include <Interface/PagedVector/PagedVectorRef.hpp>
 #include <Interface/Record.hpp>
 #include <Runtime/AbstractBufferProvider.hpp>
 #include <Runtime/Allocator/NesDefaultMemoryAllocator.hpp>
@@ -197,6 +198,17 @@ struct NoOversizedUnpooledBufferProvider : AbstractBufferProvider
 /// - the large 2 MiB buffer keeps everything on a single page and exercises the no-paging path.
 /// Schemas whose tuple size doesn't fit must be discarded via RC_PRE.
 constexpr std::array<uint64_t, 5> BUFFER_SIZE_POOL = {64, 128, 512, 4096, 2ULL * 1024 * 1024};
+
+/// We derive the page size as bufferSize / ratio instead of drawing a second entry from BUFFER_SIZE_POOL. As a result,
+/// page < buffer, i.e. the hash-join build's configuration, holds for every pool entry. In contrast, a second draw would
+/// collapse to page == buffer for wide schemas, which need 4096 bytes or more anyway as RC_PRE(pageBufferSize >= minimumPageSize(fieldTypes))
+/// throws away any run whose page is too small to hold one tuple. We stop the ratios at 8 because
+/// every page starts its own var-sized chain out of pooled buffers. Thus, tiny pages burn one buffer per few tuples and
+/// drain DirtyBufferProvider's pool of 16-32 buffers. This is a harness limit, not a PagedVector defect.
+constexpr std::array<uint64_t, 4> BUFFER_TO_PAGE_SIZE_RATIOS = {1, 2, 4, 8};
+
+/// Pages below the smallest pool entry are not a configuration the engine uses and only starve the harness' pool.
+constexpr uint64_t MIN_PAGE_SIZE = BUFFER_SIZE_POOL.front();
 
 /// Per-vector item-count range and max number of paged vectors used by the concat properties.
 constexpr uint64_t MAX_ITEMS_PER_CONCAT_VECTOR = 201;
@@ -393,9 +405,10 @@ rc::Gen<TestUtils::AnyVec> genAnyVec(std::vector<DataType> types, uint64_t buffe
         });
 }
 
-uint64_t estimateSchemaSize(const std::vector<DataType>& types)
+/// Smallest page size a PagedVector over these fields accepts. Asks the layout instead of doing the header math here.
+uint64_t minimumPageSize(const std::vector<DataType>& types)
 {
-    return getSizeInBytes(TestUtils::createSchemaFromDataTypes(types));
+    return DefaultPagedVectorTupleLayout{TestUtils::createSchemaFromDataTypes(types)}.getMinimumPageSize();
 }
 
 /// Reads pagedVector.at(idx) for a rapidcheck-drawn set of indices and asserts each record equals the reference.
@@ -419,21 +432,23 @@ void insertAndIterateProperty(TestUtils::EngineMode mode)
 {
     const auto fieldTypes = *TestUtils::genDataTypeSchema(TestUtils::ALL_VALUE_TYPES, 1, TestUtils::MAX_SCHEMA_FIELDS);
     const auto bufferSize = *rc::gen::elementOf(BUFFER_SIZE_POOL);
-    RC_PRE(PagedVector::Page::getHeaderSize() + estimateSchemaSize(fieldTypes) < bufferSize);
+    const auto pageBufferSize = bufferSize / *rc::gen::elementOf(BUFFER_TO_PAGE_SIZE_RATIOS);
+    RC_PRE(pageBufferSize >= MIN_PAGE_SIZE and pageBufferSize >= minimumPageSize(fieldTypes));
 
     const auto numberOfItems = *rc::gen::inRange<uint64_t>(0, TestUtils::MAX_ITEMS_PER_PROPERTY);
     const auto oversizedBudget = std::make_shared<int>(*rc::gen::inRange(0, MAX_OVERSIZED_DRAWS_PER_PROPERTY + 1));
 
     NES_INFO(
-        "Property insertAndIterate: fields={}, N={}, bufferSize={}, oversizedBudget={}, field_types={}",
+        "Property insertAndIterate: fields={}, N={}, bufferSize={}, pageBufferSize={}, oversizedBudget={}, field_types={}",
         fieldTypes.size(),
         numberOfItems,
         bufferSize,
+        pageBufferSize,
         *oversizedBudget,
         fmt::join(fieldTypes, ", "));
 
     auto bufferManager = DirtyBufferProvider::create(bufferSize, TestUtils::pooledBufferCountFor(bufferSize));
-    TestUtils::TestablePagedVector pagedVector{fieldTypes, *bufferManager, mode};
+    TestUtils::TestablePagedVector pagedVector{fieldTypes, *bufferManager, mode, pageBufferSize};
 
     std::vector<TestUtils::AnyVec> reference;
     reference.reserve(numberOfItems);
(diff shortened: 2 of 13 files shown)


nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalHashJoin.cpp
```diff
@@ -332,6 +334,24 @@ LoweringRuleResultSubgraph LowerToPhysicalHashJoin::apply(LogicalOperator logica
...
+                "header). Increase the page size.",
+                side,
+                tupleSize,
+                pageSize,
+                PagedVector::Page::getHeaderSize());
+        }
+    };
+    requireTupleFitsOnPage(leftHashMapOptions.pageSize, leftTupleLayout, "left");
+    requireTupleFitsOnPage(rightHashMapOptions.pageSize, rightTupleLayout, "right");
+
```
ls-1801, line 347: 1. I think, the exception type is way too specific
2. This logic does not belong in the lowering; we assume that the paged vector uses a simple row layout, so this check is now coupled into the paged vector implementation. This check should be done by the pagedvector layout.

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
```diff
@@ -779,12 +798,45 @@ TEST(PagedVectorTest, oversizedVarSizedThrowsWhenUnpooledUnavailable)
...
+    ASSERT_EQ(pagedVector.getPageBufferSize(), initPageSize);
+
+    pagedVector.appendPageIfFull(bufferManager.get());
+    ASSERT_EQ(pagedVector.getNumberOfPages(), 1);
+
+    /// Must be the size the capacity math used, not the provider's larger buffer size. Before the fix: providerBufferSize.
+    const auto page = mainBuffer.loadChildBuffer(ChildBufferIndex{0});
+    ASSERT_EQ(page.getBufferSize(), initPageSize);
+    ASSERT_GE(page.getBufferSize(), PagedVector::Page::getHeaderSize() + (pagedVector.getPageCapacity() * tupleSize));
+}
```
ls-1801: not a big fan of this test. We should make sure that it can be generated by the property-based tests above.

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
```diff
@@ -555,14 +571,17 @@ void concatCopyProperty(TestUtils::EngineMode mode)
 {
     const auto fieldTypes = *TestUtils::genDataTypeSchema(TestUtils::ALL_VALUE_TYPES, 1, TestUtils::MAX_SCHEMA_FIELDS);
     const auto bufferSize = *rc::gen::elementOf(BUFFER_SIZE_POOL);
-    RC_PRE(PagedVector::Page::getHeaderSize() + estimateSchemaSize(fieldTypes) < bufferSize);
+    const auto pageBufferSize = *rc::gen::elementOf(BUFFER_SIZE_POOL);
+    RC_PRE(pageBufferSize <= bufferSize and bufferSize <= pageBufferSize * MAX_BUFFER_TO_PAGE_SIZE_RATIO);
+    RC_PRE(PagedVector::Page::getHeaderSize() + estimateSchemaSize(fieldTypes) < pageBufferSize);
```
ls-1801: Maybe we should expose the PagedVectorLayout and have it answer questions like, "What is the minimum page size required for this schema?"

Instead of relying on the internals of the PagedVector.

ls-1801 (changes requested):
1. QueryManager change is fine
2. PagedVector change is mostly fine. I would prefer if the logic for checking what the minimal required page size is in the PagedVector(Ref/Layout).
3. The flush implementation effectively reimplements another layer of retry. I think the proper solution is to make the flush non-blocking and have the engines' repeat task handle the retry.
Here is a patch that should replace the second commit:
```diff
Subject: [PATCH] fix(NetworkChannel): Fixed a bug that would leave the network channel not respond, resulting in a deadlock of the query
---
Index: nes-network/network/src/sender/sender.rs
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/nes-network/network/src/sender/sender.rs b/nes-network/network/src/sender/sender.rs
--- a/nes-network/network/src/sender/sender.rs	(revision 3f3ce5dc0847a143b581c23ae3a02d57cf0c24b6)
+++ b/nes-network/network/src/sender/sender.rs	(revision aee948472b99776f4db26dbbdcbaa3bfbb207e6f)
@@ -11,16 +11,17 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 */
-
 use super::control::*;
 use crate::channel::Communication;
 use crate::protocol::*;
 use crate::sender::channel::{ChannelCommand, ChannelCommandQueue};
+use async_channel::TrySendError;
 use std::marker::PhantomData;
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 use tokio::runtime::Runtime;
 use tokio::sync::oneshot;
+use tokio::sync::oneshot::error::TryRecvError;
 use tracing::{Instrument, debug, info_span};
 
 /// Timeout for graceful tokio runtime shutdown
@@ -80,6 +81,7 @@
 /// ```
 pub struct SenderChannel {
     queue: ChannelCommandQueue,
+    pending_flush: Mutex<Option<oneshot::Receiver<bool>>>,
 }
 
 /// Result of a non-blocking send operation on a `SenderChannel`.
@@ -120,27 +122,50 @@
             _ => unreachable!(),
         }
     }
-    /// Flushes the network writer and checks if all data has been acknowledged.
+    /// Requests a flush and checks whether all queued data has been acknowledged.
     ///
-    /// This blocking operation flushes the underlying network stream to ensure
-    /// buffered data is sent, then checks the state of the channel queues.
+    /// This method is non-blocking with respect to network I/O. It submits at most
+    /// one flush request to the channel handler and polls that request on subsequent
+    /// calls. Callers should retry when `Ok(false)` is returned.
     ///
     /// # Returns
     ///
-    /// - `Ok(true)`: All data has been sent and acknowledged (both pending and in-flight queues are empty)
-    /// - `Ok(false)`: Flush completed but data is still pending or awaiting acknowledgment
-    /// - `Err(_)`: The network service or channel was closed
+    /// - `Ok(true)` if all pending data has been sent and acknowledged.
+    /// - `Ok(false)` if the flush request is still pending, the command queue is
+    ///   currently full, or data is still awaiting transmission or acknowledgment.
+    /// - `Err(_)` if the network service or channel has closed.
     ///
-    /// # Blocking
-    ///
-    /// This method blocks the calling thread. Do not call from async contexts without proper handling.
+    /// Concurrent calls are serialized internally to ensure that only one flush
+    /// request is outstanding at a time.
     pub fn flush(&self) -> Result<bool> {
-        let (tx, rx) = oneshot::channel();
-        self.queue
-            .send_blocking(ChannelCommand::Flush(tx))
-            .map_err(|_| "Network Service Closed")?;
-        rx.blocking_recv()
-            .map_err(|_| "Network Service Closed".into())
+        let mut pending_flush = self.pending_flush.lock().expect("BUG: IO-Thread Crashed");
+        if let Some(mut pending_flush_rx) = pending_flush.take() {
+            return match pending_flush_rx.try_recv() {
+                Ok(r) => Ok(r),
+                Err(TryRecvError::Empty) => {
+                    *pending_flush = Some(pending_flush_rx);
+                    Ok(false)
+                }
+                Err(TryRecvError::Closed) => Err("Network Service Closed".into()),
+            };
+        }
+
+        let (tx, mut rx) = oneshot::channel();
+        if let Err(e) = self.queue.try_send(ChannelCommand::Flush(tx)) {
+            return match e {
+                TrySendError::Full(_) => Ok(false),
+                TrySendError::Closed(_) => Err("Network Service Closed".into()),
+            };
+        }
+
+        match rx.try_recv() {
+            Ok(r) => Ok(r),
+            Err(TryRecvError::Empty) => {
+                *pending_flush = Some(rx);
+                Ok(false)
+            }
+            Err(TryRecvError::Closed) => Err("Network Service Closed".into()),
+        }
     }
 
     /// Closes this channel and propagates the close signal to the `ReceiverChannel`.
@@ -372,7 +397,10 @@
 
         // Receive the internal queue handle for the data channel
         let internal = rx.blocking_recv().map_err(|_| "Network Service Closed")?;
-        Ok(SenderChannel { queue: internal })
+        Ok(SenderChannel {
+            queue: internal,
+            pending_flush: Default::default(),
+        })
     }
 
     /// Shuts down the network service and all active channels.

```

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
keyseven123: Changed it.

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
keyseven123: Okay. Thought that we should test this always, but it is now part of the property-based tests.

nes-query-compiler/src/LoweringRules/LowerToPhysical/LowerToPhysicalHashJoin.cpp
keyseven123: 1. Changed it.
2. I disagree. I think this configuration error should be caught during query compilation so that in the future, the callee of the lowering can take adequat step.s

ls-1801: I was not just talking about *when* the check happens (it should definitely not happen at runtime).
I was referring more to the fact that the lowering should not know what schema is okay and what is too large; the logic should be part of the DataStructure (similar to https://github.com/nebulastream/nebulastream/pull/1894#discussion_r3796189386).

ls-1801 (approved):

nes-systests/operator/join/JoinPageSizeTooSmall.test
```diff
@@ -0,0 +1,25 @@
+# name: join/JoinPageSizeTooSmall.test
+# description: The hash join build keeps one PagedVector per distinct join key, and those pages are sized by the state
+#              page-size knob. A page that cannot hold a single tuple is rejected while lowering the join instead of
+#              tripping a precondition deep inside the build at runtime.
```
yschroeder97:
```suggestion
#              failing at a precondition inside the build at runtime.
```

We should try to avoid excessive "AI prose" in comments, where they tend to make sentences unnecessarily complex and inserting words that sound nice but confuse and interrupt the reading flow. 
I wrote a nes-comments skill that I can send you.

nes-sinks/src/NetworkSink.cpp
```diff
@@ -75,7 +75,14 @@ void NetworkSink::start(PipelineExecutionContext&)
 
 void NetworkSink::stop(PipelineExecutionContext& pec)
 {
-    PRECONDITION(channel, "Sender channel is not initialized");
+    /// A pipeline can be stopped without ever having been started, e.g. when the deployment of a query is unwound because another
```
yschroeder97:
```suggestion
    /// A pipeline can be stopped without ever having been started, e.g. when the deployment of a query is cancelled because another
```

"Unwound" is an overloaded term, should use a plainer verb.

nes-query-engine/QueryEngine.cpp
```diff
@@ -647,6 +647,14 @@ bool ThreadPool::WorkerThread::operator()(StopPipelineTask& stopPipelineTask) co
         },
         [&](const TupleBuffer&, std::chrono::milliseconds duration)
         {
+            if (terminating)
+            {
+                /// A stage that cannot finish its stop (e.g. a network sink whose peer never accepted the channel) would otherwise
+                /// re-arm this task forever, and the draining worker threads would never see an empty queue.
```
yschroeder97: Again, please instruct the agent to stop using fancy words that can not be understood. 

I don't know what "re-arming" means, and I also don't know what a "draining worker thread" is.

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
```diff
@@ -198,6 +199,18 @@ struct NoOversizedUnpooledBufferProvider : AbstractBufferProvider
 /// Schemas whose tuple size doesn't fit must be discarded via RC_PRE.
 constexpr std::array<uint64_t, 5> BUFFER_SIZE_POOL = {64, 128, 512, 4096, 2ULL * 1024 * 1024};
 
+/// The page size is derived from the provider's buffer size so that every buffer size covers pageBufferSize !=
```
yschroeder97: I don't understand this. What does "every buffer size covers pageBufferSize != providerBuffer size mean".
How can something "cover" the != operation? This reads really cryptic.

nes-frontend/src/QueryManager/QueryManager.cpp
```diff
@@ -152,6 +153,28 @@ void QueryManager::QueryManagerBackends::rebuildBackendsIfNeeded() const
 
     if (!exceptions.empty())
     {
+        /// Local plans that already started are not tracked in `state.queries` on this path, so nobody would ever stop them. A network
+        /// sink whose downstream node failed to register keeps retrying its rejected channels forever and blocks worker shutdown, so
```
yschroeder97: What does "failed to register" mean? We removed query registration a while ago.

yschroeder97 (changes requested):
Several issues: 

1. Neither the title nor the PR description precisely state what the exact issues fixed are. Fixing several unrelated things in one PR is fine for me, but then I would like to actually see an enumeration of what bugs, what caused them, and what the fix is. 
2. I don't like the agent-generated comments. They read weirdly complex and have a sentence structure that makes it sometimes harder/more confusing to understand what's going on. I'd suggest that you ask for another pass of the PR, rewriting the comments to state plain and concisely what/why the code is how it is. 

I also dislike the agent tendency of using fancy/misleading words that confuse rather than the normal technical terms. 
For example, just write "the type stores an id", instead of "the type carries an id" or "the id lives in the type".

nes-systests/operator/join/JoinPageSizeTooSmall.test
keyseven123: Done

nes-sinks/src/NetworkSink.cpp
keyseven123: Done

nes-query-engine/QueryEngine.cpp
keyseven123: Changed the commit message

nes-nautilus/tests/UnitTests/PagedVectorTest.cpp
keyseven123: Added more information to the comment

nes-frontend/src/QueryManager/QueryManager.cpp
keyseven123: Faield to register the network channel. I have added it ot the comment

keyseven123:
> Several issues:
> 
>     1. Neither the title nor the PR description precisely state what the exact issues fixed are. Fixing several unrelated things in one PR is fine for me, but then I would like to actually see an enumeration of what bugs, what caused them, and what the fix is.
> 

I thought that the commit messages would be enough... My bad. I have updated the pr description to contain what bugs we solve.


>     2. I don't like the agent-generated comments. They read weirdly complex and have a sentence structure that makes it sometimes harder/more confusing to understand what's going on. I'd suggest that you ask for another pass of the PR, rewriting the comments to state plain and concisely what/why the code is how it is.
> 
> 
> I also dislike the agent tendency of using fancy/misleading words that confuse rather than the normal technical terms. For example, just write "the type stores an id", instead of "the type carries an id" or "the id lives in the type".

I have addressed your comments and change the comments.

yschroeder97 (approved):

PR #1898, opened 2026-08-15
Fix #1855: Remove logical Schema dependency from the physical layer

### LINKED ISSUE #1855: [TASK] Remove Schema from the physical layer
### Current state
`Schema` is a logical-layer type, but it is currently reachable and used throughout the physical layer. We want to have a clear separation between a logical schema and a physical memory layout, see #1831 and #1579 .
We want to achive that the physical layer knows a memory layout, not a schema, and logical-to-physical lowering lives in one place. Thus, we can enable new logical data types without having to build a custom physical data type for it as well.

  ### Acceptance criteria

  - [ ] `nes-nautilus` does not link `nes-schema`, and no `nes-nautilus` public header includes `Schema/*.hpp`
  - [ ] `PhysicalOperatorWrapper` and `JoinSchema` hold no `Schema` members
  - [ ] No file under `nes-physical-operators/` includes `Schema/*.hpp`, including `nes-physical-operators/tests`
  - [ ] `nes-physical-operators` does not link `nes-schema`, directly or transitively
  - [ ] The `cmake/schema-firewall` scaffold is deleted
  - [ ] All existing tests pass

shubhrai23:
This PR removes the logical Schema dependency from the physical layer as requested in issue #1855.


ls-1801 (changes requested):
Requesting changes because this PR does not currently configure using the project’s documented Docker development workflow. CMake fails because required targets such as `nes-cli`, `nes-repl`, and `nes-repl-embedded` are no longer defined.

This PR removes substantial CMake configuration for several core components, including libraries, executables, registries, and tests. There are also no CI checks reported for the PR.

Please restore the unintentionally removed build configuration and verify that the project completes a clean configure and build and that all unit and system tests pass before requesting another review.

keyseven123 (changes requested):
@shubhrai23 please also use the pr description below:

```
**(for example:)** 

## Verifying this change
This change is tested by
*(for example:)*

*(for example:)*
- ExecutionEngine
- QueryCompiler
- QueryEngine
- RestAPI

## Documentation

```

PR #1900, opened 2026-08-15
Fix #1579: Remove passing Schema to TestTupleBuffer

### LINKED ISSUE #1579: Remove passing Schema to TestTupleBuffer
I do not like at all that we are now going back to using a schema directly to reason about the underlying order and datatypes of a schema. This PR is time-sensitive so it is fine, but we should have a physical layout though. Or at least use the LowerSchemaProvider class and use the current TupleBufferRef instead of passing a schema into the TupleBufferRef. I like to have an inefficient class for populating TupleBuffers but this here now ties a logical component to a physical one.

Or what does @Artraxon think about this? Can we maybe do it as one of your PRs? I have created an issue for this. If you @Artraxon tackle this as part of one of your PRs please assign yourself and the issue to the PR

_Originally posted by @keyseven123 in https://github.com/nebulastream/nebulastream/pull/1470#discussion_r3177744046_

shubhrai23:
Closes #1579. Refactored TestTupleBuffer and C++ tests to use physical types instead of logical schemas.

### DIFF
--- a/nes-nautilus/tests/TestTupleBufferTest.cpp
+++ b/nes-nautilus/tests/TestTupleBufferTest.cpp
@@ -1,3 +1,4 @@
+#include <Interface/BufferRef/LowerSchemaProvider.hpp>
 /*
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


keyseven123 (changes requested):
@shubhrai23 please use the pr description below and ensure that the project completes a clean configure and build and that all unit and system tests pass before requesting another review.

```
**(for example:)** 

## Verifying this change
This change is tested by
*(for example:)*

*(for example:)*
- ExecutionEngine
- QueryCompiler
- QueryEngine
- RestAPI

## Documentation

```

PR #1902, opened 2026-08-15
Fix #1731: Add ByteAmount TypeValidator / ScalarOption<Byte> for size configs

### LINKED ISSUE #1731: Add ByteAmount TypeValidator / ScalarOption<Byte> for size configs (support 4KiB, 4gb, etc.)

shubhrai23:
Fixes #1731. This PR introduces a strong-typedef Byte and extends YAML conversion / validation. This enables configuring size options with strings like '4KiB' instead of raw numbers. Options like operator_buffer_size, page_size, total_memory_in_bytes, and buffer_alignment_in_bytes have been migrated.

### DIFF
--- a/nes-configurations/include/Configurations/ScalarOption.hpp
+++ b/nes-configurations/include/Configurations/ScalarOption.hpp
@@ -20,6 +20,7 @@
 #include <Util/URI.hpp>
 #include <yaml-cpp/yaml.h>
 #include <ErrorHandling.hpp>
+#include <Configurations/Validation/ByteAmountValidation.hpp>
 
 namespace NES
 {
@@ -83,6 +84,10 @@ private:
         {
             /// Simple boolean conversion (true for "true", false otherwise)
             return strValue == "true";
+        }
+        else if constexpr (std::is_same_v<Type, NES::Byte>)
+        {
+            return NES::parseByteAmount(strValue);
         }
         else if constexpr (NESIdentifier<Type>)
         {
@@ -184,6 +189,7 @@ using StringOption = ScalarOption<std::string>;
 using FloatOption = ScalarOption<float>;
 using UIntOption = ScalarOption<uint64_t>;
 using BoolOption = ScalarOption<bool>;
+using ByteOption = ScalarOption<Byte>;
 
 }
 
--- a/nes-configurations/include/Configurations/Validation/ByteAmountValidation.hpp
+++ b/nes-configurations/include/Configurations/Validation/ByteAmountValidation.hpp
@@ -1,0 +1,81 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


ls-1801:
There is already a PR: https://github.com/nebulastream/nebulastream/pull/1775

PR #1903, opened 2026-08-15
Fix #1866: Implement CASE WHEN SQL function

### LINKED ISSUE #1866: Case When Function

shubhrai23:
This PR adds the CASE WHEN syntax and translates it to Nautilus IR.


yschroeder97:
Hey! Thanks for your contribution. 

Before opening any PRs, please look for already existing duplicates, which is the case here (#1885).

PR #1921, opened 2026-08-18
refactor(systest): split the systest lib into stage directories

### LINKED ISSUE #1912: Split the systest lib into stage directories

The systest library is a handful of large files with mixed responsibilities. Restructure it into per-stage directories (Config, Discovery, Parser, Runner, ResultChecker) with one file per concern, without behavior or type changes: split main() from the argument parsing, discovery from the state types, and logging from the executor, and introduce a test-file reader so the parser always receives text. The full corpus stays green; the diff is moves, renames and TU splits.

### LINKED ISSUE #1920: [EPIC] Systest refactoring: from control plane to test tool
### Description

systest has grown into a control plane of its own. It owns a source and sink catalog per test file, binds DDL itself, patches query plans mid-pipeline (splicing sink result paths, rewriting anonymous sources, writing source data at bind time), assigns query ids, and mixes discovery, parsing, execution and result checking across a few large files. That coupling makes the tool increasingly hard to maintain as features are added, and it exercises internals a test tool should not touch.

This epic refactors systest into a pipeline of stage-scoped types, in the order they run: configure, discover, read, parse, rewrite, run, check. Statements are made correct before submission instead of being patched as plans mid-run, so the tool interacts with the system the way a user would, through SQL text. The final step swaps the execution backend to the embedded Rust coordinator (#1577), reducing systest to its core: test discovery, DSL parsing, and result checking.

Beyond maintainability, the stage-scoped types and the SQL-only interaction are what make the system testable from the outside: a test states SQL in and expected rows out, and the tool holds none of the engine's internal state. Once that API is stable, the tests stop changing when internals do: a change to the optimizer, the sources or the sinks that preserves query results leaves the test files and the tool untouched, so the corpus keeps its value as a behavioral contract instead of tracking implementation details.

Each item below is one PR. Every intermediate state keeps the full systest corpus green on the current execution path; the coordinator only enters in the last step.

### Acceptance Criteria

- [x] https://github.com/nebulastream/nebulastream/issues/1912
- [x] https://github.com/nebulastream/nebulastream/issues/1913
- [x] https://github.com/nebulastream/nebulastream/issues/1914
- [x] https://github.com/nebulastream/nebulastream/issues/1928
- [x] https://github.com/nebulastream/nebulastream/issues/1927
- [ ] https://github.com/nebulastream/nebulastream/issues/1933
- [ ] https://github.com/nebulastream/nebulastream/issues/1934
- [ ] https://github.com/nebulastream/nebulastream/issues/1915
- [ ] https://github.com/nebulastream/nebulastream/issues/1916
- [ ] https://github.com/nebulastream/nebulastream/issues/1917
- [ ] https://github.com/nebulastream/nebulastream/issues/1918
- [ ] https://github.com/nebulastream/nebulastream/issues/1919

yschroeder97:

Part of the systest refactoring epic #1920. Restructuring of nes-systest-lib into per-stage directories, without behavior or type changes. The directories follow the stages of the systest pipeline in the order they run: configure, discover test files, read one, parse it, run it, and check the results.

- Parser/: the .test DSL parser and its tests.
- Config/: the configuration (`Config.hpp`, previously `SystestConfiguration.hpp`) and the argument parsing, split out of `SystestStarter` into `ConfigParser` with a public `parseConfig(argc, argv)`. `main()` shrinks to `src/Main.cpp`.
- Discovery/: test discovery (group and disable filters, recursive discovery, name shortening, the `--list` printer), split out of `SystestState`. A new `TestFileReader` is now the only place a test file is opened; the parser loses `loadFile` and always receives text via `loadString`.
- Runner/: `SystestRunner` and `QuerySubmitter`.
- ResultChecker/: `SystestResultCheck` and its tests.
- `Progress` and `Logging`: the progress tracker renamed, and the log setup split out of the executor.

## Verifying this change

This change is a refactoring with no behavior change, covered by the existing tests:
- slt-parser-test (41 tests) and slt-e2e-test (12 tests) pass.
- A systest directory run and `--list` behave as before.

- nes-systests only (the systest library and binary)

## Documentation

- The moved code keeps its in-code documentation; the new `ConfigParser`, `TestFileReader` and `Logging` entry points are documented in their headers.

This PR closes #1912


nes-systests/systest/include/Discovery/TestFileReader.hpp
```diff
@@ -0,0 +1,28 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#pragma once
+
+#include <filesystem>
+#include <string>
+
+namespace NES
```
ls-1801, line 20: the namespace is inconsistent

.github/workflows/clang_tidy_diff.yml
```diff
@@ -91,7 +91,11 @@ jobs:
         with:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |
-            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml -j $(nproc)
+            # Same driver as the pre-check above, so both steps select the same files and skip
+            # the headers that have no compile command to reconstruct.
```
ls-1801: remove these comments

scripts/tidy.sh
ls-1801, line 1: I would drop this change; you can create a dedicated .cpp file for your header to force clang-tidy to use it. I would prefer not to have two different sets of clang-tidy checks for headers and source files.

ls-1801 (changes requested):

nes-systests/systest/include/Discovery/TestFileReader.hpp
yschroeder97: Removed the namespace for all occurrences where it is not required to disambiguate between symbol names, as per the decision that we made in the past.

.github/workflows/clang_tidy_diff.yml
yschroeder97: Removed the change entirely.

scripts/tidy.sh
yschroeder97: Dropped the change.

ls-1801 (approved):

alepping (approved):

PR #1922, opened 2026-08-18
feat(systest): typed test-file model and parser stage

### LINKED ISSUE #1913: Introduce a typed test-file model and parser stage

Parsing currently fires callbacks that the binder consumes directly, so the parse result never exists as a value. Introduce a typed model of the parsed file (create, query, differential and explain statements with typed expectations) and a TestFileParser that produces it, and refactor the binder to consume the model. Verified by new parser unit tests and the unchanged corpus.

yschroeder97:

Part of the systest refactoring epic #1920, stacked on #1921.

### Why

Currently, the result of parsing a test file never exists as a value. 
The binder registers eight callbacks on the parser, and the relationships between the sections of a file are reconstructed through shared mutable state:

- A query and its expected answer come in two separate callbacks, so the binder keeps an id-keyed builder map and completes builders after the expected result (`plans.emplace(id, id).first->second.setExpectedResult(...)`). A builder is half-initialized between callbacks, and `build()` contains a (large) set of runtime preconditions that verify that every field was eventually set.
- Three variables hold state between callback invocations: the buffered local and global configuration lines, the last parsed query id, and the last merged overrides. A differential block gets its settings from the last merged overrides of the query parsed before it, which is correct only because of callback ordering, not by construction.
- The semantics of configuration lines (global accumulates, local is used up by the next query, lists multiply into alternatives) are implemented inside the binder, mixed with catalog mutation and plan rewriting, and cannot be tested without catalogs and an optimizer.

This change makes parsing a **stage** that produces a **value**. 
`parseTestFile` returns a `TestFile`: the statements of the file in order, each self-contained (a query has: its SQL, its number, its expectation, its settings and its sequential flag). 

The concrete advantages:
- The builder map, the three per-callback state variables and the binder's copy of the override merging are deleted; the binder iterates over a list.
- A malformed file fails at parse time with a clear error (a result block with no query above it, an `EXPLAIN` followed by an error expectation) instead of failing later as a half-built query.
- The parse stage is testable in isolation: the new `testfile-parser-test` covers query-answer pairing, settings combination and differential blocks in 9 tests that need no engine, no catalogs and no optimizer.
- The differential settings rule becomes structural: a block takes the settings of its own first half, which the builder passes over explicitly.
- Later steps of the epic consume exactly this value. The upcoming rewriter is a basically a function `TestFile -> RunnableTestFile`; without this stage it would have to read the callback state instead.

### Change log

- Added `Model/` with the parsed-form types: `TestFile` and the statement variants (`CreateStatement`, `QueryStatement`, `DifferentialStatement`, `ExplainStatement`), `Expectation` as a variant of expected rows, an expected error, or an expected plan, plus `ConfigurationOverride` and `SystestQueryId`. The duplicate definitions in `SystestState.hpp` are removed.
- Updated `SystestParser`: namespace `NES`, `loadString` returns void, and the schema structs nothing referenced are removed.
- `AttachedData` is a variant of `InlineRows` and `AttachedFile`, so a file reference is a path rather than the first element of a rows vector. The line parser still reports the pair shape, and `TestFileParser` converts at the boundary.
- Added `Parser/TestFileParser.hpp` with `parseTestFile(parser, path)`: registers the callbacks, forms the cartesian product of configuration alternatives (one `QueryStatement` per combination, sharing the query number), and pairs each query with the answer that follows it.
- Changed `SystestBinder` to visit the statement list in file order; the plan rewriting (sink and source patching) is unchanged and stays for a later step of the epic.
- `SystestQuery` carries an `Expectation` instead of a results-or-error variant plus explain fields; runner, result check and executor are adapted.

## Verifying this change

- New `testfile-parser-test` target (9 tests) covers the model construction, settings combination and query-answer pairing.
- `slt-parser-test` (41 tests), `slt-e2e-test` (12 tests) pass.
- A full run over the default discovery root passes 729 of 729 queries.
- Targeted runs over `differential/`, `explain/`, `parser/`, anonymous sources and a configuration-matrix file pass.

- nes-systests

## Documentation

- The model types and the parser stage entry point are documented in their headers.

This PR closes #1913


nes-systests/systest/private/Model/ConfigurationOverride.hpp
Artraxon, line 1: you can move the implementation to a .cpp. But, I would also not spend too much time on this since this is one of the few places where our PRs will conflict and I have to replace large parts of it anyway, mostly to adapt it to the new representations of configs. But I think I am following a similar approach, so it should be mostly coompatible with what you are doing here.

nes-systests/systest/include/Model/ConfigurationOverride.hpp
```diff
@@ -0,0 +1,72 @@
...
+        for (const auto& [key, value] : init)
+        {
+            overrideParameters.emplace(std::string{key}, std::string{value});
+        }
+    }
+
+    std::string& operator[](std::string_view key) { return overrideParameters[std::string{key}]; }
+
+    [[nodiscard]] const std::string& at(std::string_view key) const { return overrideParameters.at(std::string{key}); }
+
```
Artraxon: I would definitely indicate in a comment/docstring that it shares the semantics of an std::unordered_map

nes-systests/systest/include/Model/ConfigurationOverride.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <unordered_map>
+#include <utility>
+
+namespace NES
+{
+
+/// The worker settings that one configuration block of a test file asks for, as raw key value pairs.
+struct ConfigurationOverride
+{
+    std::unordered_map<std::string, std::string> overrideParameters;
```
Artraxon: why can't this field be private?

nes-systests/systest/include/Model/ConfigurationOverride.hpp
```diff
@@ -0,0 +1,72 @@
...
+        const auto mix = [](std::size_t currentSeed, std::size_t value) noexcept
+        { return currentSeed ^ (value + 0x9e3779b9 + (currentSeed << 6U) + (currentSeed >> 2U)); };
+
+        for (const auto& [key, value] : co.overrideParameters)
+        {
+            seed = mix(seed, hasher(key));
+            seed = mix(seed, hasher(value));
+        }
+        return seed;
+    }
```
Artraxon: you can just use the generic Hash object in Util/Hash.hpp

nes-systests/systest/private/Model/ParsedTestFile.hpp
```diff
@@ -0,0 +1,100 @@
...
+
+namespace Systest
+{
+/// One source or sink DDL statement, and the data that an `ATTACH` clause supplied for a physical source.
+struct CreateStatement
+{
+    std::string sql;
+    std::optional<AttachedData> attach;
+};
+
```
Artraxon, line 50: Could the systest parser already distinguish between a create sink and create source statement? Or in other words, isn't it always a CreateSinkStatement if there is data attached and we could split it up (and if we need a generic create statement make that a std::variant?)

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -1094,106 +1006,64 @@ struct SystestBinder::Impl
...
         {
-            parser.parse();
+            std::visit(
+                Overloaded{
+                    [&](Systest::CreateStatement& create)
+                    {
+                        std::optional<std::pair<TestDataIngestionType, std::vector<std::string>>> testData;
+                        if (create.attach.has_value())
+                        {
+                            /// This binder predates the attached-data variant and dies with the rewrite-upfront step,
```
Artraxon: ... and gets removed with the rewrite-upfront PR so the variant folds back into the current pair type

nes-systests/systest/include/Model/ConfigurationOverride.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <initializer_list>
+#include <string>
+#include <string_view>
+#include <unordered_map>
+#include <utility>
+
+namespace NES
+{
+
+/// The worker settings that one configuration block of a test file asks for, as raw key value pairs.
```
Artraxon: how can a configuration block of a test file ask for settings? It just overwrites them, right?

nes-systests/systest/private/Model/Expectation.hpp
Artraxon, line 1: IMO most of the comments here are not necessary. I also don't understand the comment about the json, but I am also not sure if it even is necessary? If there is something that will never produce json, why would it emit an object where it can be indicated that its json?

nes-systests/systest/include/Model/SystestQueryId.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+#include <cstdint>
+
+#include <Identifiers/NESStrongType.hpp>
+
+namespace NES
+{
+
+/// The number of a query within its test file, counted from one in file order.
+/// A test author selects a single query by this number, and the parser matches a query to its expected answer by it.
```
Artraxon: Second sentence can be removed IMO

nes-systests/systest/include/Model/TestFile.hpp
```diff
@@ -0,0 +1,100 @@
...
+{
+/// One source or sink DDL statement, and the data that an `ATTACH` clause supplied for a physical source.
+struct CreateStatement
+{
+    std::string sql;
+    std::optional<AttachedData> attach;
+};
+
+/// One query and its expected answer: the result rows, or an error it should raise.
+/// The settings are the ones that applied where the test wrote the query, and they select the worker that runs it.
```
Artraxon: > The settings are the ones that applied where the test wrote the query

I don't understand this sentence, its still unclear which settings are meant. And how can a test write a query?

nes-systests/systest/include/Model/TestFile.hpp
```diff
@@ -0,0 +1,100 @@
...
+    std::string sql;
+    SystestQueryId id;
+    Expectation expected;
+    ConfigurationOverride settings;
+    /// Whether the query above this one in the file has to finish first, which `SEQUENTIAL_EXECUTION` asks for.
+    bool sequential = false;
+};
+
+/// Two queries whose results have to agree, which is all a differential block asserts.
+/// Neither half states an expected answer, so this holds two statements rather than two expectations.
```
Artraxon: Can be simplified to "Two queries whose results must match"

nes-systests/systest/include/Model/TestFile.hpp
```diff
@@ -0,0 +1,100 @@
...
+{
+    std::filesystem::path path;
+};
+
+/// The data that an `ATTACH` clause supplies for a physical source.
+using AttachedData = std::variant<InlineRows, AttachedFile>;
+
+namespace Systest
+{
+/// One source or sink DDL statement, and the data that an `ATTACH` clause supplied for a physical source.
```
Artraxon: redundant

nes-systests/systest/src/Model/TestFile.cpp
```diff
@@ -0,0 +1,20 @@
...
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+/// The model is types rather than code, so nothing else compiles its headers on their own. This translation unit does,
+/// and the header it includes reaches every one of them, so a header that stops being self contained fails here rather
```
Artraxon: yes, but wouldn't any other TU that includes the header compile it?

nes-systests/systest/src/Parser/TestFileParser.cpp
```diff
@@ -0,0 +1,247 @@
...
+/// Whether a set of alternatives asks for nothing, which is what a file that wrote no configuration line has.
+/// Kept apart from an empty vector so combining with it yields the other side rather than nothing at all.
+bool asksForNothing(const std::vector<ConfigurationOverride>& alternatives)
+{
+    return alternatives.empty() or (alternatives.size() == 1 and alternatives.front().overrideParameters.empty());
+}
+
+/// Combines two sets of alternatives into every pairing of one from each, so a query asking for two settings that each
+/// list two values runs four times.
+/// A key both sides set takes the value from `overrides`, which is the line written closer to the query.
```
Artraxon: So a configuration value can be specified multiple times in a systest file but then gets overwritten silently?

Artraxon (changes requested):
Mostly good. There are some overlaps with my configuration changes but I can adjust mine. 

I would mostly like to see the comments reduced and I also don't like the silent overwrite of config values, I don't quite see the use case for it. IMO, either you specify an option for a file, or for a query, and if you try both its an error.

nes-systests/systest/private/Model/ConfigurationOverride.hpp
yschroeder97: Done, moved to Model/ConfigurationOverride.cpp.

nes-systests/systest/include/Model/ConfigurationOverride.hpp
yschroeder97: Done, class comment now states that lookup, iteration and equality behave as on the underlying std::unordered_map.

nes-systests/systest/include/Model/ConfigurationOverride.hpp
yschroeder97: It can, it is private now with map like accessors.

nes-systests/systest/include/Model/ConfigurationOverride.hpp
yschroeder97: Done. That also fixed a bug, the old mix depended on map iteration order, so two equal overrides could hash to different values (I think).

nes-systests/systest/private/Model/ParsedTestFile.hpp
yschroeder97: This stage parses no SQL, so splitting the type would mean already peeking keywords that the SQL parser in the binder does later. In the current implementation, a wrongly placed attach fails there. I would keep the generic statement, we can improve it at some point later :)

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: Removed the comment.

nes-systests/systest/include/Model/ConfigurationOverride.hpp
yschroeder97: Right, it overrides them. Reworded.

nes-systests/systest/private/Model/Expectation.hpp
yschroeder97: Reworked them. About the json flag, It's a leftover from a branch higher in the stack, to I removed it for now. Same for the qualifyingPrefix of ExpectedPlan.

nes-systests/systest/include/Model/SystestQueryId.hpp
yschroeder97: Removed.

nes-systests/systest/include/Model/TestFile.hpp
yschroeder97: Removed the sentence.

nes-systests/systest/include/Model/TestFile.hpp
yschroeder97: Done.

nes-systests/systest/include/Model/TestFile.hpp
yschroeder97: Removed.

nes-systests/systest/src/Model/TestFile.cpp
yschroeder97: Yes, but only behind its own earlier includes, which can supply a missing include and hide that the header is not self contained. This TU compiles the headers with nothing before them. The comment now says exactly that.

nes-systests/systest/src/Parser/TestFileParser.cpp
yschroeder97: Not anymore. A key set globally and again for a query is an error now, same for two lines setting the same key for one query. A later global line may still restate a key, TupleLargerThanBuffer switches buffer sizes mid file this way.

yschroeder97:
Reduced the comments throughout and made conflicting overrides an error, details in the single comments.

nes-systests/systest/private/Model/ParsedTestFile.hpp
Artraxon: sounds fine

nes-systests/systest/src/Parser/TestFileParser.cpp
Artraxon: Sounds kind of cursed, we just maybe also change that later one, but its fine for now

Artraxon (approved):

nes-systests/systest/src/Parser/TestFileBuilder.cpp
```diff
@@ -0,0 +1,197 @@
...
+                .firstSql = firstSql,
+                .firstId = firstId,
+                .secondSql = secondSql,
+                .secondId = secondId,
+                .settings = std::move(alternative),
+                .sequential = sequential});
+        }
+    }
+
+    [[nodiscard]] std::vector<Systest::Statement> take() && { return std::move(statements); }
```
adrianmichalke: Dont we have to check here if we have any pending queries?

nes-systests/systest/src/Model/ConfigurationOverride.cpp
```diff
@@ -0,0 +1,103 @@
...
+                if (onDuplicateKey == OnDuplicateKey::Reject and merged.contains(key))
+                {
+                    throw SLTUnexpectedToken("Configuration key '{}' is set more than once for the same query", key);
+                }
+                merged[key] = value;
+            }
+            combined.push_back(std::move(merged));
+        }
+    }
+    return combined;
```
adrianmichalke, line 75: The previous implementation did deduplicate the merged alternatives. Like this it can happen that the same query gets executed multiple times with the same configuration.
E.g.:
 ```text
   GlobalConfiguration thread_number: [1, 2]
   GlobalConfiguration thread_number: [1]
 ```

nes-systests/systest/src/Parser/TestFileBuilder.cpp
```diff
@@ -0,0 +1,197 @@
...
+
+    /// Reads the settings in scope now, rather than when the result completes the query,
+    /// because a configuration line that follows the query belongs to the next one.
+    void beginQuery(std::string sql, const bool sequential)
+    {
+        pendingQuery = PendingQuery{.sql = std::move(sql), .settings = settings.takeForNextQuery(), .sequential = sequential};
+    }
+
+    /// An EXPLAIN takes no settings, so a local line in scope stays for the next query.
+    void beginExplain(std::string sql) { pendingQuery = PendingQuery{.sql = std::move(sql), .settings = {}, .isExplain = true}; }
```
adrianmichalke: Its actually expected behaviour that the explain applies configurations

nes-systests/systest/private/Model/ParsedTestFile.hpp
```diff
@@ -0,0 +1,95 @@
...
+/// One query and its expected answer.
+struct QueryStatement
+{
+    std::string sql;
+    SystestQueryId id;
+    Expectation expected;
+    ConfigurationOverride settings;
```
adrianmichalke:
```suggestion
    ConfigurationOverride configurations;
```
Check other occurences as well

nes-systests/systest/private/Model/ParsedTestFile.hpp
```diff
@@ -0,0 +1,95 @@
...
+    /// Set by `SEQUENTIAL_EXECUTION`: the previous query has to finish before this block starts.
+    bool sequential = false;
+};
+
+/// One `EXPLAIN` and the plan text it expects.
+struct ExplainStatement
+{
+    std::string sql;
+    SystestQueryId id;
+    ExpectedResult expected;
```
adrianmichalke: nit: cant this be directly an ExpectedPlan?

nes-systests/systest/tests/Parser/SystestParserTests.cpp
```diff
@@ -260,8 +260,9 @@ TEST_F(SystestParserTest, testExplainCallbackWithVerbatimResultBlock)
     const std::string explainIn = "EXPLAIN (OPTIMIZED, FORMAT TEXT) SELECT id FROM stream WHERE value > 4 INTO sink;";
     const std::string queryIn = "SELECT id FROM stream INTO sink;";
 
-    /// The expected explain output contains lines that look like parser tokens: a line starting with `----`,
-    /// a section header, and an indented line starting with `SELECT`. All must be delivered verbatim.
+    /// The expected explain output contains lines that look like parser tokens: a line starting with `----`, a section header, and an
+    /// indented line starting with `SELECT`.
+    /// All must be delivered verbatim.
```
adrianmichalke: ?? :)

nes-systests/systest/src/Parser/TestFileBuilder.cpp
```diff
@@ -0,0 +1,197 @@
...
+{
+
+namespace
+{
+
+/// The configuration lines in scope for the next query.
+/// A global line holds until the file ends, and a later global line replaces the keys of an earlier one.
+/// A local line holds for one query, and any duplicate key with other local lines in scope is an error.
+/// A line may list several values, and the query runs once per pairing of values across the lines in scope.
+class CurrentSettings
```
adrianmichalke:
```suggestion
class CurrentConfiguration
```

adrianmichalke: As it not the full configuration but only the overrides on top of the default config it we might want to reflect that as well in the naming.

nes-systests/systest/src/Parser/TestFileBuilder.cpp
```diff
@@ -0,0 +1,197 @@
...
+                .sequential = sequential});
+        }
+    }
+
+    [[nodiscard]] std::vector<Systest::Statement> take() && { return std::move(statements); }
+
+private:
+    CurrentSettings settings;
+    std::optional<PendingQuery> pendingQuery;
+    std::vector<Systest::Statement> statements;
```
adrianmichalke: As you already started with rm the Systest namespaces we should rm them completely

nes-systests/systest/src/Parser/TestFileBuilder.cpp
```diff
@@ -0,0 +1,197 @@
...
+struct PendingQuery
+{
+    std::string sql;
+    std::vector<ConfigurationOverride> settings;
+    bool isExplain = false;
+    bool sequential = false;
+};
+
+/// Collects the statements of one test file from the parser.
+/// The parser reports a query and its expected result in two callbacks, so the query is marked pending until the answer completes it.
```
adrianmichalke: Nit:
```suggestion
/// The parser reports a query and its expected result in two callbacks, so the query is marked pending until the expected result completes it.
```

adrianmichalke (changes requested):

nes-systests/systest/src/Model/ConfigurationOverride.cpp
yschroeder97: Added deduplication and a comment to explain it. Also added a regression test to verify the behavior.

nes-systests/systest/src/Parser/TestFileBuilder.cpp
yschroeder97: Added a check that throws when the query is never completed with an expectation.

nes-systests/systest/src/Parser/TestFileBuilder.cpp
yschroeder97: No, look at this comment from current main: 

```
  /// EXPLAIN statements are not executed, so configuration overrides do not apply; 
  /// reset them so they do not leak into the next query
```
In main, the configuration is discarded, but definitely not "applied".
The optimizer is configured for the whole systest run with a config field, and the current implementation consumes the overrides above it.
It might be good to allow the test author configure the optimizer differently for EXPLAINs, but I don't see this to be in scope of this PR.

nes-systests/systest/private/Model/ParsedTestFile.hpp
yschroeder97: Agreed and changed.

nes-systests/systest/private/Model/ParsedTestFile.hpp
yschroeder97: Applied the suggestion, renamed the other occurrences as well.

nes-systests/systest/src/Parser/TestFileBuilder.cpp
yschroeder97: Renamed QueryStatement -> SelectStatement and Statement -> TestStatement, so the names do not collide with the binder/antlr parser. Then removed the Systest nested namespace completely.

nes-systests/systest/src/Parser/TestFileBuilder.cpp
yschroeder97: Applied.

nes-systests/systest/src/Parser/TestFileBuilder.cpp
yschroeder97: Applied.

nes-systests/systest/tests/Parser/SystestParserTests.cpp
yschroeder97: Restored.

adrianmichalke (approved):

PR #1924, opened 2026-08-18
refactor(worker): pass worker config options after `--`

Artraxon:
This PR changes how the SingleNodeWorker takes options for the configuration, making it the same as for the other frontends.
All config options must now also be passed behind -- .
In addition, one can now pass a .yaml file with configuration with -w .


ls-1801 (changes requested):
1. I guess some tests still need to be adjusted
2.  The README and documentation still instruct users to run the now-invalid old syntax, including README.md:134,
    docs/guide/nebulastream-frontend.md:117, and docs/development/systests.md:378.

Other than that, I like the change

ls-1801 (approved):

yschroeder97 (approved):

PR #1925, opened 2026-08-19
fix(clang-tidy): skip changed headers that have no compile command

yschroeder97:

`clang-tidy-diff` analyzes every changed file on its own, headers included. Its
default `-iregex` covers `h` and `hpp`. Only a source file has a compile command,
so clang-tidy adapts the command of the nearest source file in the compilation
database when the input is a header. A header in a directory tree that holds no
compiled source gets no include paths at all, and it fails on its first include:

```
Model/SystestQueryId.hpp:19:10: error: 'Identifiers/NESStrongType.hpp' file not found [clang-diagnostic-error]
Model/TestCaseId.hpp:20:10:    error: 'fmt/base.h' file not found         [clang-diagnostic-error]
Model/Expectation.hpp:22:10:   error: 'ErrorHandling.hpp' file not found  [clang-diagnostic-error]
```

That is a compiler error, so the job fails. The headers are correct and the build
compiles them. clang-tidy failed to reconstruct their flags.

A shared stem with a source file is not the criterion either.
`nes-common/include/Util/Pointers.hpp` has no `Pointers.cpp` and analyzes fine,
because the same tree holds compiled sources. Header-only directories break,
which is why this surfaced now.

`scripts/tidy.sh` also misreported the failure. It read any "file not found" as a
missing codegen output and told the reader to build `nes-codegen`, which does
nothing for this case. The change list is as follows:

- Changed `scripts/tidy.sh` to analyze sources and headers in two passes over the
  same diff, selected by `-iregex`.
- Added a check on the header pass. When the pass failed on nothing but missing
  includes, the script lists the skipped headers, and the sources decide the
  status of the run. Every other diagnostic, in either pass, still fails the run.
- Changed the workflow step that runs all checks to call the `tidy-diff` target,
  like the pre-check step above it, rather than repeat the invocation inline.
- Changed the hint about generated headers to fire only for sources, where a
  missing codegen output is what the symptom means.

A header that analyzes today still analyzes, and its findings still fail the job.
A skipped header produces no findings today, only a hard error.

## Verifying this change

This change is tested by running the driver in the dev container with
`readability-duplicate-include` as the check set, over a diff that carries the
three unanalyzable headers:

| case | expected | result |
|---|---|---|
| unanalyzable headers only | pass, name them | exit 0, three headers listed |
| real finding in an analyzable header | fail | exit 1 |
| unanalyzable headers plus a real header finding | fail | exit 1 |
| missing header in a source file | fail, keep the codegen hint | exit 1, hint printed |

This pull request changes no C++, so its own clang-tidy job reports nothing to
do. #1922 is the end to end check. It fails on those three headers without this
change, and it carries this change in its history now.

- CI: the `clang-tidy-diff` workflow and `scripts/tidy.sh`. Both diff targets,
  `tidy-diff` and `tidy-diff-fix`, and the local developer flow go through that
  script, so they change with it.
- No source code, no build configuration, no dependencies.

## Documentation

- The script carries the reasoning next to the two passes: why a header has no
  compile command of its own, and what a skip means.
- The header comment of the script already describes the modes and the
  environment variables, and needed no change.

This PR closes no issue. It came out of #1922, whose clang-tidy job failed on
the headers quoted above, and it unblocks #1577 for the same reason. #1922 has
since dropped one of those three headers as unused, so a rerun without this
change would report the two that remain.

### DIFF
--- a/.github/workflows/clang_tidy_diff.yml
+++ b/.github/workflows/clang_tidy_diff.yml
@@ -91,7 +91,11 @@ jobs:
         with:
           image_name: nebulastream/nes-ci:${{ inputs.dev_image_tag }}-${{ matrix.stdlib }}-${{matrix.sanitizer}}
           run: |
-            git diff -U0 ${{ inputs.base_sha }} -- ':!*.inc' ':!*nes-rust-bindings' ':!vcpkg/**' | clang-tidy-diff-19.py -clang-tidy-binary clang-tidy-19 -p1 -path build -export-fixes clang-tidy-result/fixes.yml -j $(nproc)
+            # Same driver as the pre-check above, so both steps select the same files and skip
+            # the headers that have no compile command to reconstruct.
+            export NES_TIDY_DIFF_BASE=${{ inputs.base_sha }}
+            export NES_TIDY_EXPORT_FIXES=$(pwd)/clang-tidy-result/fixes.yml
+            cmake --build build --target tidy-diff
       # Findings are printed by clang-tidy-diff in the steps above and exported here for download.
       # A failing step means clang-tidy flagged something in .clang-tidy's WarningsAsErrors list.
       - name: Upload Clang-Tidy Results
--- a/scripts/tidy.sh
+++ b/scripts/tidy.sh
@@ -111,6 +111,7 @@ if [ "$MODE" = "full" ]; then
         -header-filter="$HEADER_FILTER" "$SOURCE_FILTER" 2>/dev/null \
         | tee >("${DECOLOR[@]}" > "$REPORT_FILE")
     STATUS=${PIPESTATUS[0]}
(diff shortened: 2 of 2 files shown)


yschroeder97:
Closing in favour of the approach you suggested. `nes-systests/systest/src/Model/` now has a translation unit, so clang-tidy can reconstruct include paths for the headers in that directory, and CI needs no change.

One correction on the premise, for the record: the change here did not use two sets of checks. Both passes ran the same `.clang-tidy` and the same arguments, and only the file selection and the handling of a "file not found" error differed. That said, one compiled source in the directory is the smaller fix, and it also keeps the existing hint about generated headers accurate.

The measured detail behind it: clang-tidy borrows flags from a nearby compiled source rather than a same-stem one, which is why `nes-common/include/Util/Pointers.hpp` analyses fine without a `Pointers.cpp`. `Model/` was the only directory in the refactor with no source anywhere near it.

PR #1926, opened 2026-08-19
refactor(systest): discover test files as a list

### LINKED ISSUE #1914: Discover test files as a list

Discovery returns a list of discovered test files instead of a map keyed by path, and a discovered file no longer carries a source and a sink catalog. The binder builds one catalog pair per test file, because binding is the only stage that reads a catalog.

A test file's name is also a path: the result file of a query is named after it. Names are relative to the directory passed on the command line, so two directories given at once name their files against different roots, and two files of the same name collide. That run fails with `query already registered` and the two files write to one result file. Names should be relative to a fixed root instead, so a directory narrows the search rather than renaming what it finds.

yschroeder97:

Part of #1920. Stacked on #1922.

A test file's name is its path below `nes-systests/`, always: `operator/join/JoinNull`. The name is the report label and the result file path, so it says where the file is, and two files with the same file name cannot collide. Before, a file was keyed against whichever `-t` directory found it and shortened to its file name when unique, so `-t dirA dirB` with two `Dup.test` files failed with `query already registered`, and a report line did not say which directory a file was in.

- Discovery returns `std::vector<DiscoveredTestFile>` instead of `TestFileMap`, an `unordered_map<path, TestFile>`. No consumer looked a file up by its path: the binder, the group listing and the printer all iterated the values, so the key bought nothing and the unordered map dropped the discovery order. `TestFile` also carried a source and a sink catalog per file that only the binder read; the binder creates them itself now, and a discovered file holds only what discovery knows: path, name, groups, and the query filter from the command line.
- `TestName` and `TestGroup` are strong types instead of `std::string` aliases, so a name is not mistaken for a path or a group where it serves as a map key or a report label.
- `test_discover_root` is the directory that names are relative to (default `nes-systests/`); `-t` narrows the search below it and may be relative. A `-t` directory outside the root is named from its parent; if two files would still share a name, discovery stops with both paths.
- A malformed `# groups:` header is an error instead of being read best-effort.
- Behaviour change relative to main: names are no longer shortened to the bare file name. Benchmark metrics tagged by query name change once (`Nexmark` becomes `benchmark/Nexmark`).

## Verifying this change

This change is tested by
- 13 discovery unit tests in the new target `systest-test-discovery-test`: naming below the root, narrowing, directories outside the root, name collisions, command-line files with and without a query filter, malformed group headers
- `slt-parser-test` and `testfile-builder-test`
- by hand: `-t dirA dirB` with two same-named files, `-t nes-systests/explain` (relative path), `-t .../EXPLAIN.test:03`

- `nes-systests` only: discovery, the catalog construction in the binder, the test locations in the config parser.
- Configuration: `test_discover_root` is new; `test_discover_dirs` keeps its name and meaning.
- No dependency changes.

This PR closes #1914. Follow-up on the discovery root: #1969.

### DIFF
--- a/nes-systests/systest/private/Discovery/TestDiscovery.hpp
+++ b/nes-systests/systest/private/Discovery/TestDiscovery.hpp
@@ -14,16 +14,46 @@
 
 #pragma once
 
+#include <filesystem>
+#include <optional>
 #include <ostream>
-#include <Config/Config.hpp>
-#include <SystestState.hpp>
+#include <string>
+#include <unordered_set>
+#include <vector>
 
-namespace NES::Systest
+#include <Config/Config.hpp>
+#include <Identifiers/Identifiers.hpp>
+
+namespace NES
 {
 
-/// load test file map objects from files defined in systest config
-TestFileMap loadTestFileMap(const SystestConfiguration& config);
+/// A test file's name: its path relative the discovery root without extension.
+/// E.g., `operator/join/JoinNull` becomes `nes-systests/operator/join/JoinNull.test`.
+/// The name labels the file in the report and in the result file path.
+/// Strong types, so a name or a group is not confused with a path or with each other.
+using TestName = NESStrongStringType<struct TestName_, "INVALID">;
+using TestGroup = NESStrongStringType<struct TestGroup_, "INVALID">;
 
-std::ostream& operator<<(std::ostream& os, const TestFileMap& testMap);
+struct DiscoveredTestFile
+{
+    explicit DiscoveredTestFile(
+        const std::filesystem::path& file,
+        TestName testName,
+        std::optional<std::unordered_set<SystestQueryId>> enabledQueries = std::nullopt);
+    [[nodiscard]] std::string getLogFilePath() const;
+
+    [[nodiscard]] TestName name() const { return testName; }
+
+    std::filesystem::path file;
+    TestName testName;
+    /// The query numbers to run. Every query of the file runs when this is not set.
+    std::optional<std::unordered_set<SystestQueryId>> enabledQueries;
+    std::vector<TestGroup> groups;
+};
+
+std::ostream& operator<<(std::ostream& os, const std::vector<DiscoveredTestFile>& testFiles);
+
+/// Reads the configuration and searches the given directories, returning one invocation's test files.
+std::vector<DiscoveredTestFile> discoverTestFiles(const SystestConfiguration& config);
 
 }
--- a/nes-systests/systest/src/Config/ConfigParser.cpp
+++ b/nes-systests/systest/src/Config/ConfigParser.cpp
@@ -14,6 +14,7 @@
 
 #include <Config/ConfigParser.hpp>
 
+#include <algorithm>
 #include <cstddef>
 #include <cstdlib>
 #include <exception>
@@ -24,6 +25,7 @@
 #include <stdexcept>
 #include <string>
 #include <unordered_map>
+#include <utility>
 #include <vector>
 #include <Config/Config.hpp>
 #include <Configurations/Util.hpp>
@@ -217,16 +219,16 @@ void addTestQueryNumbers(NES::SystestConfiguration& config, const std::string& t
     }
 }
 
-std::filesystem::path parseTestLocationPath(const std::string& testFileDefinition, NES::SystestConfiguration& config)
+/// Splits `path` or `path:test_numbers` into the two.
+std::pair<std::filesystem::path, std::string> splitTestLocation(const std::string& testFileDefinition)
 {
     const size_t delimiterPos = testFileDefinition.find(':');
     if (delimiterPos == std::string::npos)
     {
-        return {testFileDefinition};
+        return {std::filesystem::path{testFileDefinition}, std::string{}};
     }
 
-    addTestQueryNumbers(config, testFileDefinition.substr(delimiterPos + 1));
-    return {testFileDefinition.substr(0, delimiterPos)};
+    return {std::filesystem::path{testFileDefinition.substr(0, delimiterPos)}, testFileDefinition.substr(delimiterPos + 1)};
 }
 
 std::vector<std::filesystem::path> findAllInTree(const std::filesystem::path& wanted, const std::filesystem::path& root)
@@ -243,15 +245,19 @@ std::vector<std::filesystem::path> findAllInTree(const std::filesystem::path& wa
     return hits;
 }
 
-void applyDiscoveredTestLocation(const std::filesystem::path& testFilePath, NES::SystestConfiguration& config)
+void applyDiscoveredTestLocation(
+    const std::filesystem::path& testFilePath, const std::vector<std::filesystem::path>& searchDirs, NES::SystestConfiguration& config)
 {
-    /// Search all discover directories for a matching file name
+    /// A bare file name is looked up in every directory that the run searches, so a test can be given without its folder.
     std::vector<std::filesystem::path> allMatches;
-    for (const auto& dir : config.testDiscoverDirs.getValues())
+    for (const auto& searchDir : searchDirs)
     {
-        auto matches = findAllInTree(testFilePath.filename(), dir.getValue());
+        auto matches = findAllInTree(testFilePath.filename(), searchDir);
         allMatches.insert(allMatches.end(), matches.begin(), matches.end());
     }
+    std::ranges::sort(allMatches);
+    const auto duplicates = std::ranges::unique(allMatches);
+    allMatches.erase(duplicates.begin(), duplicates.end());
 
     if (allMatches.empty())
     {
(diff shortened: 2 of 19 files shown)


nes-systests/systest/private/Discovery/TestDiscovery.hpp
```diff
@@ -14,16 +14,52 @@
...
+#include <vector>
+
 #include <Config/Config.hpp>
-#include <SystestState.hpp>
+#include <Identifiers/Identifiers.hpp>
+
+namespace NES
+{
+
+/// A test file's given name: its path below the discovery root, without the extension.
```
Artraxon: this is a bit unclear to me, do you mean that the path is relative to the discovery root? also, what has the path do to with the name? is the name the relative path to the discovery root?

nes-systests/systest/private/Discovery/TestDiscovery.hpp
```diff
@@ -14,16 +14,52 @@
...
+/// A strong type, to avoid confusing it with other labels.
+using TestName = NESStrongStringType<struct TestName_, "INVALID">;
+using TestGroup = std::string;
```
Artraxon: IMO you can just make it a strong type as well while your at it. Just aliasing simple types is IMO an antipattern and I didn't like the fact that we did it in the first place.

nes-systests/systest/private/Discovery/TestDiscovery.hpp
```diff
@@ -14,16 +14,52 @@
...
 
-namespace NES::Systest
+/// One selected test file.
+///
+/// The name is the file's path below the discovery root without its extension (i.e., '.test').
```
Artraxon: ok here its explained a bit more concise, but IMO its also only needed in one place

nes-systests/systest/private/Discovery/TestDiscovery.hpp
```diff
@@ -14,16 +14,52 @@
...
+/// This guarantees that two files that share a stem in different folders have unique names.
```
Artraxon: What is the stem in this case? I would have thought that a stem of is the path to the last common parent folder, but this kind of does't make sense with the rest of the comment

nes-systests/systest/private/Discovery/TestDiscovery.hpp
```diff
@@ -14,16 +14,52 @@
...
+/// - A name whose stem is unique across the run is shortened back to that stem once discovery has finished.
```
Artraxon: I assume a run is one systest invocation not one query? And what is 'that' stem to what the name is shortened back to? why does it need shortening?

nes-systests/systest/src/Config/ConfigParser.cpp
```diff
@@ -280,32 +286,51 @@ void applyTestLocations(const ArgumentParser& program, NES::SystestConfiguration
...
+        }
+        else
+        {
+            files.push_back(std::move(split));
+        }
+    }
 
-    const auto testLocations = program.get<std::vector<std::string>>("--testLocations");
-    for (const auto& location : testLocations)
+    /// A bare file name is looked up in the search directories, which are fixed before any lookup runs. Collecting them
```
Artraxon: ~~runs~~

Artraxon (changes requested):
Overall, I don't quite understand the stemming logic or at least the terms used. Is it even necessary? Why not just always use the full names? I think it would actually improve the ergonomics because you directly know where to look for a failed test instead of having to do a filesystem search.

nes-systests/systest/src/Discovery/TestDiscovery.cpp
```diff
@@ -198,33 +256,81 @@ TestFileMap discoverTestsRecursively(const std::filesystem::path& path, const st
...
+        if (const std::string entryExt = toLowerCopy(entry.path().extension().string()); !fileExtension || entryExt == desiredExtension)
         {
-            const TestFile testfile(
-                entry.path(),
-                testNameFromRelativePath(entry.path().lexically_relative(path)),
-                std::make_shared<SourceCatalog>(),
-                std::make_shared<SinkCatalog>());
-            testFiles.insert({testfile.file, testfile});
+            testFiles.emplace_back(entry.path(), testNameFromRelativePath(entry.path().lexically_relative(keyRoot)));
         }
```
adrianmichalke, line 218: This will not match correctly if the test dir path is given as relative path ie: `systest -t nes-systests/operator/map`. I guess would suffice to normalize it to an absolute path

nes-systests/systest/src/Discovery/TestDiscovery.cpp
```diff
@@ -198,33 +256,81 @@ TestFileMap discoverTestsRecursively(const std::filesystem::path& path, const st
...
     }
     return testFiles;
 }
 
+std::vector<TestGroup> readGroups(const DiscoveredTestFile& testfile)
```
adrianmichalke, line 223: as files are user input i would expect that it throws or print on malformated groups

adrianmichalke (changes requested):
In general im not a fan of AI generated PR comments as they list a lot of detailed changes but miss to convey the overall idea.

nes-systests/systest/private/Discovery/TestDiscovery.hpp
```diff
@@ -14,16 +14,52 @@
...
+    [[nodiscard]] TestName name() const { return testName; }
+
+    std::filesystem::path file;
+    TestName testName;
+    /// An empty set of query numbers runs every query of the file.
+    std::unordered_set<SystestQueryId> onlyEnableQueriesWithTestQueryNumber;
+    std::vector<TestGroup> groups;
+};
+
+using DiscoveredTestFiles = std::vector<DiscoveredTestFile>;
```
ls-1801: Not a big fan of this typedef. Its just one additional indirection. I don't think std::vector<DiscoveredTestFile> is too verbose. But having to lookup what the DiscoveredTestFiles is, is one extra step. Also it could be annoying to debug a compiler error because a function expects DiscoveredTestFiles but received DiscoveredTestFile

nes-systests/systest/private/Discovery/TestDiscovery.hpp
```diff
@@ -14,16 +14,52 @@
...
-namespace NES::Systest
+/// One selected test file.
+///
+/// The name is the file's path below the discovery root without its extension (i.e., '.test').
+/// This guarantees that two files that share a stem in different folders have unique names.
+/// - A name whose stem is unique across the run is shortened back to that stem once discovery has finished.
+/// - A file given explicitly on the command line keeps its stem.
+struct DiscoveredTestFile
 {
+    explicit DiscoveredTestFile(const std::filesystem::path& file);
```
ls-1801: Could these overloads be reduced and their intended call cases documented? In particular, when is a `DiscoveredTestFile` created without an explicit `TestName`? Using an optional query filter, with a name such as `enabledTestQueryFilter`, could make the difference between “run every query” and “run only these queries” explicit instead of encoding it as an empty set.

nes-systests/systest/src/Discovery/TestDiscovery.cpp
```diff
@@ -148,45 +148,103 @@ std::optional<std::string> getSkipReason(const NES::Systest::TestFile& testFile,
...
+    {
+        return std::nullopt;
+    }
+    return directory.filename().string();
 }
 
-namespace NES::Systest
+/// Two files from different directories could have the same name.
+/// Grow each of them by the directory above until they are unique.
+void separateDuplicateTestNames(NES::DiscoveredTestFiles& testFiles)
```
ls-1801: Could we derive each minimal unique name by reversing the file path components, lexicographically sorting them, and comparing adjacent suffixes? That directly computes the minimum distinguishing suffix instead of first shortening names and then repeatedly growing duplicates. It should also scale predictably as the nightly suite continues to grow.

ls-1801 (changes requested):
It doesn’t need to happen in this particular pr.

However, I find the whole test discovery process rather strange, particularly the hardcoding of the cmake root directory. This is a poor design.

I would prefer if we restricted the system test to a single discovery directory. This directory can either be explicitly set or defaults to the current working directory.

Additionally, a second filter parameter could be added to select individual test:2.

nes-systests/systest/private/Discovery/TestDiscovery.hpp
yschroeder97: Made it into a strong type.

nes-systests/systest/private/Discovery/TestDiscovery.hpp
yschroeder97: Removed the alias. I agree that it is confusing.

nes-systests/systest/src/Discovery/TestDiscovery.cpp
yschroeder97: I removed the shortening logic, as it removed ~100 LoC and didn't really provide much benefit (beyond being a bit nicer on the eye).

nes-systests/systest/private/Discovery/TestDiscovery.hpp
yschroeder97: Reduced the overloads to two constructors and an `optional<unordered_set>`. 

Added a factory `DiscoveredTestFile::fromCommandLine(path, optionalQueries)`, which derives the stem name.

nes-systests/systest/src/Discovery/TestDiscovery.cpp
yschroeder97: Changed it so that an invalid group declaration throws a helpful error message.

nes-systests/systest/private/Discovery/TestDiscovery.hpp
yschroeder97: The stem of `function/ConditionalFunction.test` is `ConditionalFunction`, a.k.a the name of only the file without its extension.

yschroeder97: Added an example comment.

nes-systests/systest/private/Discovery/TestDiscovery.hpp
yschroeder97: Tried to make the comment more understandable.

nes-systests/systest/src/Config/ConfigParser.cpp
yschroeder97: Done.

nes-systests/systest/private/Discovery/TestDiscovery.hpp
yschroeder97: Yes, a run is one invocation of the SLT. Both can be used interchangeably; both are kind of overloaded terms, but I don't have a better name.

Removed the shortening logic as per your suggestion, so the rest of the comment no longer applies and was removed.

ls-1801 (approved):

Artraxon (approved):

nes-systests/systest/src/Discovery/TestDiscovery.cpp
adrianmichalke: Have you looked at this problem?

adrianmichalke (changes requested):

adrianmichalke:
Codex review:
 High — direct selection violates fixed-root test identity
    nes-systests/systest/src/Discovery/TestDiscovery.cpp:253-255 always names a directly selected file by its stem. Thus the same file is
    operator/join/JoinNull during directory discovery but JoinNull with -t .../JoinNull.test:3. Result paths, report labels, distributed IDs,
    and benchmark labels therefore depend on how the file was selected, contrary to #1914 and the PR’s “always relative to the root” premise.

adrianmichalke:
Codex review: Medium — overlapping external search roots make names argument-order dependent
    nes-systests/systest/src/Discovery/TestDiscovery.cpp:326-338 derives names per search root and keeps the first canonical file seen.
    Searching /ext/a/nested /ext/a names the same file nested/F, while reversing them names it a/nested/F. Collapse overlapping roots or
    choose the naming base independently of traversal order.

nes-systests/systest/src/Discovery/TestDiscovery.cpp
yschroeder97: Fixed. Directories from testLocations are normalized to absolute paths before matching, so relative arguments like nes-systests/operator/map resolve correctly. Directly specified files get the same normalization. Two tests cover the relative directory and the relative file case.

nes-systests/systest/src/Discovery/TestDiscovery.cpp
yschroeder97: Fixed. A malformed groups line now throws with the file and the offending line. Empty group names are rejected as well. RejectsAGroupsLineWithoutBrackets and RejectsAnEmptyGroupName cover both.

adrianmichalke (approved):

PR #1929, opened 2026-08-19
feat(systest): lock the working directory of a run

### LINKED ISSUE #1928: Lock the working directory of a run

Every systest invocation defaults to the same working directory and clears it at startup. A second run therefore deletes the data and result files that the first one is still reading, and the first run fails on a missing or empty result file. That failure reads as a broken query rather than as two runs colliding.

A run should claim the directory before it clears it, and a second run should report the conflict and name the option that gives it a directory of its own.

yschroeder97:

Every systest invocation defaults to the same working directory,
`<build>/nes-systests/working-dir`, and clears it at startup. The directory
holds the data and result files, and both are tagged from the test file's name/path and the query number, so they repeat across invocations. 
A second run, therefore, could delete the files the first run was still reading, and the first run failed due to a missing or empty result file. 

- Added `WorkingDirectoryGuard`. It creates the working directory, opens it
  with `O_DIRECTORY`, takes an exclusive `flock` on the directory itself, and
  empties it afterward.
- Changed `SystestExecutor` to build the guard where it used to remove and
  recreate the directory.
- Added the `private/` include directory to the systest library, for headers
  that the library uses but does not publish.

The lock lives on the directory's inode, so there is no lock file anywhere.
Emptying keeps the directory and removes every direct child, which preserves
the locked inode for the whole run. The kernel releases a `flock` however the
process ends, so a crashed run leaves nothing to remove by hand, and the lock
also holds across containers, which matters because each CI run is its own
container. Two containers that mount the same host directory open the same
inode, so the conflict is detected across mounts as well, and nothing is
written outside the working directory.

The guard reports a filesystem error with the path in the message, and it
creates nested working directories, which the previous single-level call did
not.

Two details from review:

- An earlier revision locked a sibling file, `<workingDir>.lock`, because a
  lock file inside the directory would be deleted when the guard empties it.
  The sibling left an artifact next to the directory, required write permission
  outside it, and needed the trailing separator of `--workingDir=/dir/` to be
  normalized away so that `/dir/` and `/dir` shared one lock. Locking the
  directory itself removes all three concerns, since `open` resolves both
  spellings to the same inode.
- The descriptor of the locked directory is owned by a member, not a bare
  `int`. A constructor that throws runs no destructor of its own, so a failure
  after the directory was opened (lock taken by another run, directory not
  clearable) would have leaked the descriptor and, with it, kept the lock for
  the rest of the process.

## Verifying this change

This change is tested by
- four unit tests in `slt-parser-test`: the guard empties a prepopulated
  directory and leaves no artifact inside or next to it, a second guard on the
  same directory is refused with or without a trailing separator, the lock is
  released when the guard goes out of scope, and a refused guard leaves no
  open descriptor behind
- starting a run against a working directory that another run holds. The second
  run stops with `another run holds the working directory <path>, and each would
  clear it under the other. Wait for that run, or pass --workingDir to use a
  different working directory`.
- the full corpus in the dev container, which passes with the guard in place
- a single test file run, which clears and repopulates its working directory as
  before

- nes-systests.
- No dependency changes, and no test target changes.

This PR closes #1928.

### DIFF
--- a/nes-systests/systest/src/WorkingDirectoryGuard.cpp
+++ b/nes-systests/systest/src/WorkingDirectoryGuard.cpp
@@ -1,0 +1,89 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <WorkingDirectoryGuard.hpp>
+
+#include <filesystem>
+#include <system_error>
+
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/file.h>
+
+#include <Util/Files.hpp>
+#include <ErrorHandling.hpp>
+
+namespace NES
+{
+namespace
+{
+
+/// Empties the working directory of one systest invocation.
+/// Data and result file names are derived from test file names, so they repeat across invocations.
+/// For example, the result checker could read a leftover file as the result of a query that wrote nothing.
+void clearWorkingDirectory(const std::filesystem::path& workingDir)
+{
+    std::error_code errorCode;
+    for (const auto& entry : std::filesystem::directory_iterator{workingDir, errorCode})
+    {
+        std::filesystem::remove_all(entry.path(), errorCode);
+        if (errorCode)
+        {
+            break;
+        }
+    }
+    if (errorCode)
+    {
+        throw TestException("could not clear the working directory {}: {}", workingDir.string(), errorCode.message());
+    }
+}
+
+}
+
+WorkingDirectoryGuard::WorkingDirectoryGuard(const std::filesystem::path& directory)
+{
+    std::error_code errorCode;
+    std::filesystem::create_directories(directory, errorCode);
+    if (errorCode)
+    {
+        throw TestException("could not create the working directory {}: {}", directory.string(), errorCode.message());
+    }
+
+    /// NOLINTNEXTLINE(cppcoreguidelines-pro-type-vararg) - POSIX API requires varargs
+    lockedDirectory.descriptor = ::open(directory.c_str(), O_DIRECTORY | O_RDONLY | O_CLOEXEC);
+    if (lockedDirectory.descriptor < 0)
+    {
+        throw TestException("could not open the working directory {}: {}", directory.string(), getErrorMessageFromERRNO());
+    }
+    /// The kernel releases a flock however the process ends, so a crashed run leaves no lock to remove by hand.
+    /// An flock also works across containers.
+    if (::flock(lockedDirectory.descriptor, LOCK_EX | LOCK_NB) != 0)
+    {
+        throw TestException(
+            "another run holds the working directory {}, and each would clear it under the other. "
+            "Wait for that run, or pass --workingDir to use a different working directory",
+            directory.string());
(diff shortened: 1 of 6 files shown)


nes-systests/systest/src/WorkingDirectoryGuard.cpp
```diff
@@ -0,0 +1,86 @@
...
+        throw TestException("could not create the working directory {}: {}", workingDir.string(), errorCode.message());
+    }
+}
+
+}
+
+WorkingDirectoryGuard::WorkingDirectoryGuard(const std::filesystem::path& directory)
+{
+    /// The lock file is a sibling of the directory, because emptying the directory would delete a lock file inside it.
+    const auto lockFile = std::filesystem::path{directory.string() + ".lock"};
```
adrianmichalke: Will this work with a working dir like `--workingDir=/dir/` with a ´/´ at the end?

nes-systests/systest/src/WorkingDirectoryGuard.cpp
```diff
@@ -0,0 +1,86 @@
...
+    if (errorCode)
+    {
+        throw TestException("could not clear the working directory {}: {}", workingDir.string(), errorCode.message());
+    }
+
+    std::filesystem::create_directories(workingDir, errorCode);
+    if (errorCode)
+    {
+        throw TestException("could not create the working directory {}: {}", workingDir.string(), errorCode.message());
+    }
```
adrianmichalke: What happens with the lock if one of these exceptions throws?

adrianmichalke (changes requested):

nes-systests/systest/src/WorkingDirectoryGuard.cpp
yschroeder97: Good catch. Normalized using `lexically_normal` so that both variants resolve to the same lockfile.

nes-systests/systest/src/WorkingDirectoryGuard.cpp
yschroeder97: Then the destructor would not be called on the partially initialized object, which means the lock would only be closed/released when the process terminates rather than when the guard goes out of scope. 

Introduced a small `LockFile` RAII wrapper for a file descriptor to make the throwing constructor safe.

nes-systests/systest/src/WorkingDirectoryGuard.cpp
```diff
@@ -0,0 +1,99 @@
...
+    const auto normal = directory.lexically_normal();
+    return normal.has_filename() ? normal : normal.parent_path();
+}
+
+}
+
+WorkingDirectoryGuard::WorkingDirectoryGuard(const std::filesystem::path& directory)
+{
+    const auto workingDir = withoutTrailingSeparator(directory);
+    /// The lock file is a sibling of the directory, because emptying the directory would delete a lock file inside it.
```
ls-1801: Could the lock live at <workingDir>/.nes-systest.lock? We could create the directory, acquire that lock, and then remove every direct child except the lock file. This avoids the sibling artifact and trailing-separator normalization while preserving the locked inode for the whole run.

ls-1801 (changes requested):
I would prefer to keep the lock in the working directory. I am thinking about scenarios where we might not have permissions outside the working directory or use a docker mount

adrianmichalke (approved):

nes-systests/systest/src/WorkingDirectoryGuard.cpp
yschroeder97: The guard now opens the working directory itself with `O_DIRECTORY` and takes the flock on that descriptor, so the lock is taken on the dir inode. Clearing works as you proposed, the directory stays and every direct child is removed, which keeps the locked inode alive for the whole run. 

Compared to `.nes-systest.lock` this leaves no artifact inside the directory and nothing to skip when emptying it.

ls-1801 (approved):

PR #1931, opened 2026-08-19
refactor(systest): interpret the run-related options into a run policy

### LINKED ISSUE #1927: Interpret the run options into a run plan

The executor decides what a run does from single options as it goes: whether to shuffle the queries, how many to submit at once, whether to keep issuing them, and whether to record timings. Options that only mean something together are read in several places.

One interpretation step should read them once, into an ordering, a concurrency, a repetition and, for a measuring run, the report it writes. The executor then follows the plan rather than the options, and a test can build a plan without starting a run.

yschroeder97:

Stacked on #1929.

The executor decided what a run does from individual options of `SystestConfiguration`: whether to
shuffle the queries, how many to submit at once, whether to keep issuing them,
and whether to record timings.
These options are logically related and belong together, since they define how to run a test battery.

- Added `Config/RunPolicy`, holding an ordering (`RunInFileOrder` or
  `RunInShuffledOrder`), a concurrency, a repetition (`SubmitOnce` or
  `SubmitUntilStopped`) and, for a measuring run, the report that it writes.
  `RunPolicy::create` is the only way to build one (private constructor), so a
  measuring run cannot be handed more than one query at a time.
- Changed `SystestExecutor` to follow the policy rather than the options.
- Bugfix: gave the endless mode option the correct key. It had
  `query_compiler_config`, the key of the query compiler config option, so a
  config file could set neither of the two on its own.

## Verifying this change

This change is tested by
- two unit tests in `systest-run-policy-test`: a measuring run submits one
  query at a time (the rule the parser's `-b`/`-n` check and the executor rely
  on), and its report is `<workingDir>/BenchmarkResults.json` (the name
  `scripts/benchmarking/benchmark.py` reads)
- format and clang-tidy on the changed lines

- nes-systests only: the run options, and how the executor reads them.
- Configuration: the endless mode option answers to `endless_mode` rather than
  to the key of another option.
- One new test target, `systest-run-policy-test`. No dependency changes.

## Documentation

- `RunPolicy` has in-code documentation for each policy, and for what a
  measuring run does.

This PR closes #1927.

### DIFF
--- a/nes-systests/systest/private/Config/RunPolicy.hpp
+++ b/nes-systests/systest/private/Config/RunPolicy.hpp
@@ -1,0 +1,71 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#pragma once
+
+#include <cstddef>
+#include <filesystem>
+#include <optional>
+#include <variant>
+
+namespace NES
+{
+
+/// Runs the queries in file order (default).
+struct RunInFileOrder
+{
+};
+
+/// Runs the queries in a random order, which is how tests that depend on each other are found.
+/// Every query in a run is submitted to the same worker, so a test can pass on state that an earlier query left behind.
+struct RunInShuffledOrder
+{
+};
+
+using OrderingPolicy = std::variant<RunInFileOrder, RunInShuffledOrder>;
+
+/// Submits every query once (default).
+struct SubmitOnce
+{
+};
+
+/// Submits the queries round after round, until something outside the run stops it.
+struct SubmitUntilStopped
+{
+};
+
+using RepetitionPolicy = std::variant<SubmitOnce, SubmitUntilStopped>;
+
+class SystestConfiguration;
+
+/// One invocation's policy, read once from the systest config.
+struct RunPolicy
+{
+    RunPolicy() = delete;
+
+    /// Reads the options into the policy that the run follows.
+    [[nodiscard]] static RunPolicy create(const SystestConfiguration& config);
+
+    OrderingPolicy ordering;
+    /// How many queries the run submits at once.
+    size_t concurrency;
+    RepetitionPolicy repetition;
+    /// `std::nullopt` means the run measures nothing.
+    std::optional<std::filesystem::path> measureReport;
+
+private:
+    RunPolicy(OrderingPolicy ordering, size_t concurrency, RepetitionPolicy repetition, std::optional<std::filesystem::path> measureReport);
+};
+
+}
--- a/nes-systests/systest/src/Config/ConfigParser.cpp
+++ b/nes-systests/systest/src/Config/ConfigParser.cpp
@@ -159,16 +159,19 @@ void applyBenchmarkMode(const ArgumentParser& program, NES::SystestConfiguration
     }
 
     config.benchmark = true;
+    /// A measuring run is limited to one query at a time, so -b and -n above 1 contradict each other.
     if ((program.is_used("-n") || program.is_used("--numberConcurrentQueries")) && program.get<int>("--numberConcurrentQueries") > 1)
     {
-        NES_ERROR("Cannot run systest in Benchmarking mode with concurrency enabled!");
-        std::cout << "Cannot run systest in benchmarking mode with concurrency enabled!\n";
+        const auto message = fmt::format(
+            "Cannot combine -b with -n {}: a measuring run submits one query at a time. Drop one of the two.",
+            program.get<int>("--numberConcurrentQueries"));
+        NES_ERROR("{}", message);
+        std::cout << message << '\n';
         std::exit(-1); ///NOLINT(concurrency-mt-unsafe)
     }
 
     std::cout << "Running systests in benchmarking mode. Only one query is run at a time!\n";
     std::cout << "Any included differential queries and queries expecting an error will be skipped.\n";
-    config.numberConcurrentQueries = 1;
 }
 
 void applyDebugMode(const ArgumentParser& program)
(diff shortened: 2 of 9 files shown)


nes-systests/systest/private/Config/RunPolicy.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <cstddef>
+#include <filesystem>
+#include <optional>
+#include <variant>
+
+namespace NES
+{
+
+/// Runs the queries in file order (default).
+struct SourceOrder
```
Artraxon: I would make the typenames here a bit more explicit. Like here RunInFileOrder

nes-systests/systest/private/Config/RunPolicy.hpp
```diff
@@ -0,0 +1,67 @@
...
+{
+};
+
+/// Runs the queries in a random order, which is how tests that depend on each other are found.
+/// Every query in a run is submitted to the same worker, so a test can pass on state that an earlier query left behind.
+struct Shuffled
```
Artraxon: RunInShuffledOrder

nes-systests/systest/private/Config/RunPolicy.hpp
```diff
@@ -0,0 +1,67 @@
...
+{
+};
+
+using OrderingPolicy = std::variant<SourceOrder, Shuffled>;
+
+/// Submits every query once (default).
+struct Once
```
Artraxon: SubmitOnce

nes-systests/systest/private/Config/RunPolicy.hpp
```diff
@@ -0,0 +1,67 @@
...
+{
+};
+
+/// Submits the queries round after round, until something outside the run stops it.
+struct UntilStopped
```
Artraxon: SubmitUntilStopped

nes-systests/systest/src/Config/ConfigParser.cpp
```diff
@@ -159,16 +159,20 @@ void applyBenchmarkMode(const ArgumentParser& program, NES::SystestConfiguration
     }
 
     config.benchmark = true;
+    /// A measuring run is limited to one query at a time when the options are interpreted, so this only reports
+    /// the conflict. It stays in the parser because only argparse knows whether -n was passed.
     if ((program.is_used("-n") || program.is_used("--numberConcurrentQueries")) && program.get<int>("--numberConcurrentQueries") > 1)
```
Artraxon: I would remove everything after "to one query at a time", because I don't really understand it and don't think it helps understand anything here because the logic is straight forward

Artraxon (changes requested):
Looks mostly good to me

nes-systests/systest/private/Config/RunPolicy.hpp
```diff
@@ -0,0 +1,67 @@
...
+
+class SystestConfiguration;
+
+/// One invocation's policy, read once from the systest config.
+struct RunPolicy
+{
+    OrderingPolicy ordering;
+    /// A measuring run always submits one query at a time.
+    /// Queries that run together compete for the worker, so their measurements are not meaningful.
+    size_t concurrency;
```
adrianmichalke: I do not fully understand this comment. Is it related to concurrency? if so how?

nes-systests/systest/private/Config/RunPolicy.hpp
```diff
@@ -0,0 +1,67 @@
...
+    RepetitionPolicy repetition;
+    /// `std::nullopt` means the run measures nothing.
+    std::optional<std::filesystem::path> measureReport;
+
+    /// Reads the options into the policy that the run follows.
+    [[nodiscard]] static RunPolicy create(const SystestConfiguration& config);
```
adrianmichalke: Should the be the only way to construct it? In this case a common pattern is to make the default constructor private

nes-systests/systest/tests/Config/RunPolicyTests.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+TEST_F(RunPolicyTest, AMeasuringRunSubmitsOneQueryAtATime)
+{
+    SystestConfiguration config;
+    config.benchmark = true;
+    config.numberConcurrentQueries = 4;
+
+    EXPECT_EQ(RunPolicy::create(config).concurrency, 1);
+}
```
adrianmichalke, line 57: Not sure if these tests add anything, but it up to you if you want to keep them

adrianmichalke (changes requested):
only minor comments

nes-systests/systest/private/Config/RunPolicy.hpp
yschroeder97: Done.

nes-systests/systest/private/Config/RunPolicy.hpp
yschroeder97: Done.

nes-systests/systest/private/Config/RunPolicy.hpp
yschroeder97: Done.

nes-systests/systest/private/Config/RunPolicy.hpp
yschroeder97: Done.

nes-systests/systest/src/Config/ConfigParser.cpp
yschroeder97: Applied as proposed.

nes-systests/systest/private/Config/RunPolicy.hpp
yschroeder97: The field controls how many queries can be submitted and run concurrently. 

A systest invocation that "measures", a.k.a benchmarking, needs concurrency = 1, because otherwise measurements are not meaningful, as the different queries compete for the worker's resources. 

It makes sense to split and just state here what concurrency means, moving the reason to the actual construction in the cpp file.

nes-systests/systest/private/Config/RunPolicy.hpp
yschroeder97: Made the ctor private, which loses the designated initializer in the factory method, which is fine for me. 

Now, only `create` can construct a `RunPolicy`.

nes-systests/systest/tests/Config/RunPolicyTests.cpp
yschroeder97: Removed the first one that just restates the defaults. 

Kept the others; although they are not the most important, they do add value: 
- Assert that the run creates the benchmarks in the location that `benchmark.py` expects
- Assert that benchmark queries do not run concurrently

Artraxon (approved):

ls-1801 (approved):

greenBene (approved):
I really appreciate this change :) It makes it easier to reason about the options!

adrianmichalke (approved):

Issue #1823, opened 2026-07-28
[DOC] Update NES reference documentation


oczerwia-2010:
### What do you see as an issue?

The current documentation reference is outdated and refers to the "old" NES version making the out-of-the-box use of NES more difficult than it has to be.

### Solving the problem

Replace / update the current references for NES with all currently implemented capabilities and structure it accordingly so that it can be hosted on the current documentation webpage.

### Are you willing to submit a PR?

- [x] Yes I am willing to submit a PR!

### Code of Conduct

- [x] I agree to follow this project's [Code of Conduct](https://github.com/nebulastream/nebulastream/blob/main/CODE_OF_CONDUCT.md)

PR #1932, opened 2026-08-19
docs(restructuring): make the docs website reflect the docs directory

oczerwia-2010:

This PR reduces redundancy both within the documentation itself as well as between the different components. 
List of changes:
- Added `functions.md` to centralize all scalar / aggregation functions with their syntax, input parameters as well as simple examples.
- added `operators.md` for the same purpose.
- added `sources-and-sinks.md` for the same purpose.
- Removed all contents relating to functions, operators sources and sinks from the `query_api.md` file.
- Removed the Nebuli branding
- Renamed general concepts to stream processing theory to create space where we can discuss key components of the system in more detail.
- Adjusted the CI jobs `nightly-quick-start.yml` and `test-quick-start.sh` to work with the new directory layout

Structural changes:
- removed the hugo metadata from each markdown into a seperate content adapter file (_content.gotmpl)
- adjusted the docs/ directory structure into a more flat file structure
- the website now reflects the flat directory structure of the docs directory
- Introduced a third header called "Getting started" which now contains all end to end tutorials / instructions on how to make the system run (does not include implementation!)

## Verifying this change
The webpage can be build and inspected locally by running:
```
git clone https://github.com/nebulastream/nebulastream.git && 
  cd nebulastream &&
  git checkout docs/sql-function-operator-reference &&
  cd docs/website &&
  npm install &&
  npm run start

```

- Documentation

This PR is part of issue #1823


docs/guide/functions.md
```diff
@@ -0,0 +1,107 @@
...
+case-insensitively, so `Abs(x)`, `ABS(x)`, and `abs(x)` are equivalent.
+
+## Field access, constants, and casts
+
+| Signature | Description | Example |
+|---|---|---|
+| `<field>` | References a record attribute by name. Every bare identifier in an expression is this. | `SELECT x FROM s INTO sink` |
+| `<literal>` | An inline literal value (number, string, bool). | `SELECT 42 FROM s INTO sink` |
+| `<expr> AS <alias>` | Binds the result of `<expr>` to a new field name in the output schema. | `SELECT x AS x1 FROM s INTO sink` |
+| `CAST(x AS <type>)` | Converts `x` to `<type>` using standard SQL cast syntax. Takes exactly one argument (`x`); the target type is a type annotation, not a second function argument. See [Casting rules](#casting-rules). | `SELECT CAST(x AS FLOAT64) FROM s INTO sink` |
```
tekdogan: `CAST(x AS <type>)` does not parse according to [current parser](https://github.com/nebulastream/nebulastream/blob/0d53176cf412f0b72ae79beeb7cf05a0703d64a5/nes-sql-parser/AntlrSQL.g4#L464).
You can replace it wtih `CASTTOTYPE(x AS <type>)`.

docs/guide/operators.md
```diff
@@ -0,0 +1,238 @@
...
+
+Join is a stateful, binary operator that combines tuples from two input streams within the same window
+that satisfy the join predicate. The join's timestamp field currently must have the same name in both
+input streams.
+
+- `ON <predicate>` is required; an equality predicate over fields from both sides.
+- `WINDOW` is required, see [Windows](#windows).
+
+```sql
+SELECT * FROM s INNER JOIN (SELECT * FROM t) ON sid = tid WINDOW TUMBLING(ts, SIZE 1 MINUTE) INTO sink
```
tekdogan, line 182: I am not sure if this join gets parsed without error. It requires two timestamps for joins. [Reference](https://github.com/nebulastream/nebulastream/blob/0d53176cf412f0b72ae79beeb7cf05a0703d64a5/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp#L1112-L1118)

docs/guide/operators.md
```diff
@@ -0,0 +1,238 @@
...
+WINDOW TUMBLING(ts, SIZE 1 MINUTE)
+HAVING MAX(price) > 100.0 AND COUNT(*) >= 10 INTO sink
+```
+
+## Join
+
+**Syntax:** `SELECT ... FROM <left> [INNER] JOIN (<right>) ON <predicate> <window> INTO <sink>`
+
+Join is a stateful, binary operator that combines tuples from two input streams within the same window
+that satisfy the join predicate. The join's timestamp field currently must have the same name in both
```
tekdogan: Does it really require the same timestamp name?

[Here](https://github.com/nebulastream/nebulastream/blob/0d53176cf412f0b72ae79beeb7cf05a0703d64a5/nes-frontend/tests/DistributedPlanningTest.cpp#L360-L368) is a test with different timestamp names, and [this](https://github.com/nebulastream/nebulastream/blob/0d53176cf412f0b72ae79beeb7cf05a0703d64a5/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp#L902-L917) is the implementation reference.

docs/guide/operators.md
```diff
@@ -0,0 +1,238 @@
...
+| [Union](#union) | Stateless | Combine two streams with the same underlying schema |
+| [Aggregation](#aggregation) | Stateful | Accumulate windows of a single stream |
+| [Join](#join) | Stateful | Combine two streams in windows based on a predicate |
+| [Model Inference](#model-inference) | Stateless | Run an ONNX ML model on each tuple in a stream |
+
+---
+
+## Projection
+
+**Syntax:** `SELECT <field | function | subquery> [AS alias], ... FROM <source> INTO <sink>`
```
tekdogan: [Currently](https://github.com/nebulastream/nebulastream/blob/0d53176cf412f0b72ae79beeb7cf05a0703d64a5/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp#L726), subqueries seem to be only supported in `FROM`.

docs/guide/functions.md
```diff
@@ -0,0 +1,107 @@
...
+| `a <= b` | Less-than-or-equal comparison. | `SELECT * FROM s WHERE a <= b INTO sink` |
+| `a BETWEEN lo AND hi` | Inclusive range check. Supports `NOT BETWEEN`. | `SELECT * FROM t WHERE amount BETWEEN 100.0 AND 1000.0 INTO sink` |
+| `a IN (v1, v2, ...)` | Membership in an explicit, non-empty value list. Supports `NOT IN`. | `SELECT * FROM s WHERE status IN ('queued', 'running') INTO sink` |
+| `a IS NULL` | Tests whether `a` is null. Supports `IS NOT NULL`. | `SELECT * FROM s WHERE optional_value IS NULL INTO sink` |
+| `IsNull(a)` | Function-call spelling of `IS NULL`, usable in a projection (not just `WHERE`). | `SELECT IsNull(optional_value) AS is_missing FROM s INTO sink` |
+| `a IS NaN` | Tests whether a numeric value is floating-point *not a number*. Supports `IS NOT NaN`. The operand must be numeric; integers are accepted but are never NaN. | `SELECT * FROM measurements WHERE reading IS NaN INTO sink` |
+| `ISNAN(x)` | Function-call spelling of `IS NaN`, usable in a projection. | `SELECT ISNAN(reading) AS invalid FROM measurements INTO sink` |
+
+`IS NaN` follows SQL three-valued logic, see [null handling](../technical/null_handling.md) for details.
```
tekdogan: Currently there is no such page for null handling. Can you create it?

docs/guide/sources-and-sinks.md
```diff
@@ -0,0 +1,159 @@
...
+  'lrb.json' as "SOURCE".FILE_PATH,
+  'JSON' as INPUT_FORMATTER."TYPE"
+);
+```
+
+One source reads CSV-formatted data from a TCP socket, while the other reads JSON-formatted data from a
+file. Both produce tuples that conform to the `lrb` schema. The CSV file might look like this:
+
+```
+creationTS,vehicle,speed,highway,lane,direction,position
```
tekdogan: The shown CSV contains a header row, but the CSV input formatter has no header-skipping option. Could you remove the header?

tekdogan (changes requested):
Great changes! It looks much cleaner now.

I found a few inconsistencies between the documentation and the current implementations, which I've noted below.

docs/guide/functions.md
```diff
@@ -0,0 +1,107 @@
+# Functions
+
+This page is the reference for every scalar and aggregate function NebulaStream exposes in SQL: the
+operations you can apply to fields inside a `SELECT` list, `WHERE` clause, or aggregation. For the
+operators that functions are used inside of (`SELECT`, `WHERE`, `WINDOW`, `JOIN`, …), see
+[Operators](operators.md).
```
ThanGeo: The last sentence here could be improved in terms of phrasing. I don't really understand what it means. I suppose it tries to differentiate between functions and operators but its confusing right now

docs/guide/sources-and-sinks.md
```diff
@@ -0,0 +1,159 @@
...
+```sql
+CREATE PHYSICAL SOURCE FOR source_name TYPE TCP SET(
+  'CSV' as INPUT_FORMATTER."TYPE",
+  '\n' as INPUT_FORMATTER.TUPLE_DELIMITER,
+  ',' as INPUT_FORMATTER.FIELD_DELIMITER,
+  ...
+);
+```
+
+NebulaStream supports CSV and JSON input formats.
```
ThanGeo, line 153: You should add here that there are certain structures in JSON that are not supported yet such as dynamically-sized JSON arrays. We recently added nested JSON support if I am not mistaken. Maybe the input/output formats details would justify a separate .md file? So that we can list existing/not-existing features for each one.

ThanGeo (changes requested):
Big fan of this PR and thanks for using concise human language and not incomprehensible AI-generated filler walls of text.

Left a couple of minor comments

docs/guide/functions.md
oczerwia-2010: The technical directory is currently explicitly not included in the website build, I will do another pass in the next PR to carefully rework the technical directory as their concepts are a bit harder to update. For now I have commented out this sentence

docs/guide/operators.md
oczerwia-2010: Changed

docs/guide/operators.md
oczerwia-2010: Good catch, changed the statement

docs/guide/operators.md
oczerwia-2010: True, changed

docs/guide/functions.md
```diff
@@ -0,0 +1,107 @@
...
+
+A function is either unary (one argument) or binary (two arguments). Passing the wrong number of
+arguments throws `InvalidQuerySyntax` at query-compile time. A function name is matched
```
keyseven123: Instead of `query-compile time`, I would use `during query submission`.

docs/website/static/scripts/run_keystrokes_scenario.sh
```diff
@@ -1,50 +0,0 @@
-#!/bin/bash
-
```
keyseven123, line 2: Why are you removing all files in this folder? I thought we need them for the guide?

docs/guide/operators.md
keyseven123: i still see it in the `SELECT` but this might be just a cached version...

docs/guide/operators.md
```diff
@@ -0,0 +1,238 @@
...
+
+```sql
+WINDOW TUMBLING(ts, SIZE 1 SEC) INTO sink
+```
+
+```sql
+WINDOW SLIDING(ts, SIZE 1 SEC, ADVANCE BY 100 MS) INTO sink
+```
+
+💡 A sliding window's timestamp field must be `UINT64`, with millisecond resolution.
```
keyseven123: It does not have to be. NebulaStream assumes it is.

docs/guide/sources-and-sinks.md
keyseven123: We support nested. Maybe we can have one markdoown file per format?

keyseven123 (changes requested):
How are we ensuring that a change in the query syntax does not invalidate the docs?

On a side note: This PR could have benefitted from having multiple commits with each commit providing some more information about why a change is done. Also, it would help with the review as one can review one commit after the other.

docs/guide/operators.md
oczerwia-2010: changed

oczerwia-2010:
> How are we ensuring that a change in the query syntax does not invalidate the docs?

Good point, at this point I have no robust solution. I would either make a change in syntax / parsing files trigger a required review by a documentation team member.

Another approach that is still tested is the LLM based analysis.

Third option might be potentially introducing versioning of the documentation and pointing towards last working commits.

Still open for discussion though

docs/guide/functions.md
oczerwia-2010: changed

docs/website/static/scripts/run_keystrokes_scenario.sh
oczerwia-2010: These are from the old quick start guide and are no longer referenced

docs/guide/functions.md
```diff
@@ -0,0 +1,107 @@
+# Functions
+
+This page is the reference for every scalar and aggregate function NebulaStream exposes in SQL. Functions
+can be applied to [fields](sources-and-sinks.md#logical-sources) within the `SELECT` list, `WHERE`
+clause, or aggregation. See [Operators](operators.md) for the operators these clauses belong to.
```
keyseven123: I do not think that we support agg functions in the where clause

docs/guide/functions.md
```diff
@@ -0,0 +1,107 @@
...
+| `a <= b` | Less-than-or-equal comparison. | `SELECT * FROM s WHERE a <= b INTO sink` |
+| `a BETWEEN lo AND hi` | Inclusive range check. Supports `NOT BETWEEN`. | `SELECT * FROM t WHERE amount BETWEEN 100.0 AND 1000.0 INTO sink` |
+| `a IN (v1, v2, ...)` | Membership in an explicit, non-empty value list. Supports `NOT IN`. | `SELECT * FROM s WHERE status IN ('queued', 'running') INTO sink` |
+| `a IS NULL` | Tests whether `a` is null. Supports `IS NOT NULL`. | `SELECT * FROM s WHERE optional_value IS NULL INTO sink` |
+| `IsNull(a)` | Function-call spelling of `IS NULL`, usable in a projection (not just `WHERE`). | `SELECT IsNull(optional_value) AS is_missing FROM s INTO sink` |
+| `a IS NaN` | Tests whether a numeric value is floating-point *not a number*. Supports `IS NOT NaN`. The operand must be numeric; integers are accepted but are never NaN. | `SELECT * FROM measurements WHERE reading IS NaN INTO sink` |
+| `ISNAN(x)` | Function-call spelling of `IS NaN`, usable in a projection. | `SELECT ISNAN(reading) AS invalid FROM measurements INTO sink` |
+
+<!-- `IS NaN` follows SQL three-valued logic, see [null handling](../technical/null_handling.md) for details. -->
```
keyseven123: Why is this commented out?

docs/guide/operators.md
```diff
@@ -0,0 +1,244 @@
...
+
+### Window measures
+
+| Measure | Description |
+|---|---|
+| Event time | Uses a timestamp field defined in the tuples themselves to assign them to windows. Supply the field as the first `WINDOW` argument, e.g. `WINDOW TUMBLING(ts, SIZE 1 SEC)`. |
+| Ingestion time | Assigns tuples to windows based on when they were ingested into the system. Omit the timestamp field: `WINDOW TUMBLING(SIZE 1 MINUTE)`. |
+
+💡 For binary windowed operators (joins), the event-time timestamp fields of the two input streams can
+have different names: supply both as `WINDOW TUMBLING(<left_ts>, <right_ts>, SIZE <size> <unit>)`.
```
keyseven123, line 131: Is this true? I do not think so but I might be wrong

keyseven123 (changes requested):

docs/guide/functions.md
oczerwia-2010: Fair catch — reworded to be precise: aggregate functions only apply post-window in the `SELECT`/`HAVING` list, not in `WHERE`. Fixed in 8244d858f4.

docs/guide/functions.md
oczerwia-2010: It shouldn't be — the null_handling.md page exists and covers this (ISNAN section) accurately. Uncommented in 8244d858f4.

docs/guide/operators.md
oczerwia-2010: Verified against the grammar and query-plan-creator code (AntlrSQL.g4 timestampParameter, exitTimestampParameter/exitTumblingWindow) plus existing systests (e.g. JoinMultipleKeys.test, guide11.test) — joins do support two differently-named timestamp fields, one per input side. This is accurate as written, no change needed.

ls-1801 (approved):

docs/guide/sources-and-sinks.md
```diff
@@ -0,0 +1,300 @@
...
+  'localhost:8080' AS "SINK"."HOST",
+  '<path>' as "SINK".FILE_PATH,
+  'CSV' as "SINK".OUTPUT_FORMAT,
+  FALSE as "SINK".APPEND
+);
+```
+
+The sink name (`csv_sink`) must match the name used in the query's `INTO` clause.
+
+Available sink types:
```
ThanGeo, line 213: The phrasing here is not very consistent with how the supported physical sources are listed above. I would recommend being consistent across this document wrt sinks/sources. Maybe even an overview table at the very top that briefly outlines/maintains all supported types would be a nice starting point.

Also, probably the KafkaSink will be merged before this document so keep an eye out for that to include it here.

ThanGeo (changes requested):
Minor comment on the sources/sink document and its formatting

PR #1935, opened 2026-08-19
fix(config): Fix scalar SequenceOption CLI parsing

### LINKED ISSUE #1908: [BUG] Scalar SequenceOption elements cannot be parsed from command-line input
### Version / Commit

main at commit 5ecb0de0088c9137ac921b967b5ba7fa726f4583

### What happened?

A scalar element cannot be added to a `SequenceOption` through the generic command-line configuration path using the sequence’s declared option key.

For example, passing `{{"--test_groups", "Join"}}` to `BaseConfiguration::overwriteConfigWithCommandLineInput()` for a configuration containing `SequenceOption<StringOption>` throws:

```text
invalid config parameter; Identifier test_groups is not known.
```

Conversely, the malformed key `--test_groups.` is accepted. This happens because the scalar child is default-constructed without a name and `ScalarOption::parseFromString()` looks up that empty name instead of the nested identifier supplied by `SequenceOption`.

The current systest executable handles its sequence arguments separately through argparse; this report concerns the generic configuration API.

- `--test_groups=Join` should add the string value `"Join"`.
- `--test_query_numbers=42` should add the unsigned integer value `42`.
- A malformed trailing-dot key such as `--test_groups.` should be rejected
  without modifying the sequence.
- A value such as `"Join,Aggregation"` should remain one string value; this
  fix should not introduce comma-separated or repeated-key semantics.
- YAML parsing and programmatic `.add()` behavior should remain unchanged.

### How to reproduce

The current `systest` executable parses `--groups` separately with `argparse` and applies values using `.add()`, so it bypasses the faulty generic configuration path. The following minimal configuration exercises that path directly:

```cpp
class ScalarSequenceConfiguration final : public BaseConfiguration
{
public:
    SequenceOption<StringOption> testGroups{
        "test_groups", "Test groups to run"};
    SequenceOption<UIntOption> testQueryNumbers{
        "test_query_numbers", "Test query numbers to run"};

protected:
    std::vector<BaseOption*> getOptions() override
    {
        return {&testGroups, &testQueryNumbers};
    }
};
```

Exercise these inputs in a configuration unit test:

```cpp
config.overwriteConfigWithCommandLineInput(
    {{"--test_groups", "Join"}});

config.overwriteConfigWithCommandLineInput(
    {{"--test_query_numbers", "42"}});

config.overwriteConfigWithCommandLineInput(
    {{"--test_groups.", "Join"}});

config.overwriteConfigWithCommandLineInput(
    {{"--test_groups", "Join,Aggregation"}});
```

The focused regression was built and run in the GCC/libstdc++ development
container:

```bash
cmake -S . -B cmake-build-debug -G Ninja -DCMAKE_BUILD_TYPE=Debug
cmake --build cmake-build-debug \
  --target configuration-help-message-tests -j
ctest --test-dir cmake-build-debug \
  -R 'SequenceOptionCliParsingTest' \
  --output-on-failure
```

Observed deterministically (added unit tests):

```text
ParsesStringValueFromNormalKey:
  Identifier test_groups is not known.

ParsesUnsignedIntegerValueFromNormalKey:
  Identifier test_query_numbers is not known.

RejectsTrailingDotKey:
  Expected an exception; actual: throws nothing.
  The sequence is modified.

DoesNotSplitCommaSeparatedStringValue:
  Identifier test_groups is not known.

PreservesYamlSequenceParsing: Passed
PreservesProgrammaticAdd: Passed

33% tests passed, 4 tests failed out of 6
```

### Operating System

Ubuntu 22.04.3 LTS

### Deployment details

Local WSL2 checkout built in nebulastream/nes-development:local, prepared using the GCC/libstdc++ option of scripts/install-local-docker-environment.sh. The repository was bind-mounted into the container. No distributed
deployment was involved.

### Anything else

This reproduces on every focused test run.

**Suspected cause**  
`SequenceOption<T>::parseFromString()` default-constructs `T()`. A scalar
created this way has an empty option name. `ScalarOption<T>::parseFromString()`
then looks up `this->getName()` rather than its supplied `identifier`.
Consequently, a normal map keyed by `test_groups` is not found, while the
malformed trailing-dot input is grouped under an empty child key and accepted.

Originally encountered in a thesis remote-systest extension. The remote systest workflow generated worker configuration arguments such as `--worker.link_capacities=sink-node:9090:1000000` and `--worker.link_latencies=sink-node:9090:250`, both backed by `SequenceOption<StringOption>`. Those options are thesis-specific and are not part of current upstream main; the minimal configuration test above reproduces the same generic parsing defect on current main.

### Code of Conduct

maximilian-breuer:

Fix scalar `SequenceOption` command-line parsing.

`SequenceOption` passes the option identifier to `ScalarOption::parseFromString()`, but the parser previously looked up `this->getName()`, which is empty for the temporary scalar option. As a result, valid arguments such as `--test_groups=Join` failed to parse.

This change makes `ScalarOption::parseFromString()` use the supplied identifier and adds regression coverage.

## Verifying this change

- Added string `SequenceOption` CLI parsing coverage.
- Added unsigned-integer `SequenceOption` CLI parsing coverage.
- Verified malformed trailing-dot keys are rejected.
- Verified comma-separated scalar values are not split.
- Verified YAML sequence parsing remains unchanged.
- Verified programmatic `.add()` behavior remains unchanged.
- Full Debug build passed.
- Formatting check passed.
- All 921 enabled tests passed.
- One existing test remains intentionally disabled.

- `nes-configurations`
- Scalar and sequence configuration option parsing

## Documentation

No documentation changes are required.

The configuration rework may replace this implementation later, but the regression tests should preserve the expected behavior.

## Issue Closed by this pull request

This PR closes #1908

### DIFF
--- a/nes-configurations/include/Configurations/ScalarOption.hpp
+++ b/nes-configurations/include/Configurations/ScalarOption.hpp
@@ -159,11 +159,11 @@ void ScalarOption<T>::parseFromYAMLNode(YAML::Node node)
 template <class T>
 void ScalarOption<T>::parseFromString(std::string identifier, std::unordered_map<std::string, std::string>& inputParams)
 {
-    if (!inputParams.contains(this->getName()))
+    if (!inputParams.contains(identifier))
     {
         throw InvalidConfigParameter("Identifier {} is not known.", identifier);
     }
-    std::string value = inputParams[this->getName()];
+    std::string value = inputParams[identifier];
     if (value.empty())
     {
         throw InvalidConfigParameter("Identifier {} is not known.", identifier);
--- a/nes-configurations/tests/CMakeLists.txt
+++ b/nes-configurations/tests/CMakeLists.txt
@@ -13,5 +13,10 @@
 # Google Testing Framework ----------------------------------------------------
 include(ExternalProject)
 ### Config Tests ###
-add_nes_unit_test(configuration-help-message-tests "UnitTests/Configurations/ConfigurationHelpMessageTests.cpp" "UnitTests/Validation/EndpointValidationTest.cpp" "UnitTests/Configurations/ExplicitlySetTrackingTests.cpp")
+add_nes_unit_test(
+    configuration-help-message-tests
+    "UnitTests/Configurations/ConfigurationHelpMessageTests.cpp"
+    "UnitTests/Configurations/ExplicitlySetTrackingTests.cpp"
+    "UnitTests/Configurations/SequenceOptionTests.cpp"
+    "UnitTests/Validation/EndpointValidationTest.cpp")
 set_tests_properties(${Tests} PROPERTIES TIMEOUT 35)
--- a/nes-configurations/tests/UnitTests/Configurations/SequenceOptionTests.cpp
+++ b/nes-configurations/tests/UnitTests/Configurations/SequenceOptionTests.cpp
@@ -1,0 +1,133 @@
+/*
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        https://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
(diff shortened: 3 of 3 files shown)


ls-1801 (approved):

Artraxon (approved):

maximilian-breuer:
@ls-1801 rebased onto current `main` and additional empty-value regression test included (codecov)

PR #1936, opened 2026-08-20
refactor(systest): verdict model and typed result checkers

### LINKED ISSUE #1915: Introduce a verdict model and pure result checkers

Replace the monolithic result check with an outcome model (a verdict per test case, reported as checked queries keyed by file, variant and query number) and a family of pure checkers for result rows, explain plans and differential agreement that operate on files and text only. The old comparison engine is deleted in the same change, so there is never a second checker in the tree.

yschroeder97:

Stacked on #1931.

Result checking was one engine (`SystestResultCheck`, 1,238 lines) that mixed
loading result files, comparing, and reporting, keyed on the runner's types.
This PR replaces it with a typed verdict model and one value type per kind of
check. The change list is as follows:

- Added `Model/Verdict.hpp`. A `Verdict` is `std::expected<Success, Mismatch>`: a
  passed check holds `Success`, a failed one the text that describes how actual
  and expected differ. A running query holds an `optional<Verdict>`, empty until
  it was checked.
- Added the checker family. `QueryResultCheck` compares a query's result file
  against the schema and the tuples that the test declared, `DifferentialCheck`
  compares the two result files of a differential block tuple by tuple, and
  `ExplainLinesCheck` / `ExplainRegexCheck` compare a printed plan against
  expected lines or against regex assertions; which of the two applies is
  decided where the check is built, from the tags on the expected lines. Each
  is a plain struct holding its inputs, with a `check()` member that returns a
  `Verdict`.
- Added `Model/TestCaseId.hpp`. A `TestCaseId` identifies one case in a run's
  reports: the test file's discovered name, the query's number within that
  file, and the configuration overrides of that run of it. It prints through
  `operator<<` and `FMT_OSTREAM` as `file:N [key=value]`, the number unpadded.
- Added `ResultChecker/Check.hpp`. The `Check` concept states the shape,
  `AnyCheck` is the variant of the four, a `static_assert` reads the alternatives
  from the variant and checks each against the concept, and `runCheck` visits
  it. The runner reaches every checker through `runCheck`.
- Moved result file reading into `QueryResult` and the schema- and type-aware
  comparison into `ResultComparison`, both private units behind the checkers.
  Reading returns `expected<QueryResult, string>` with one message per cause
  (file not written, empty, empty header, malformed header), so a bad result
  file fails that query with a verdict instead of stopping the run; both
  checkers guard the same way.
- The verdict for a comparison takes a `ComparisonOrigin` rather than a bool,
  so a failure inside a differential block still reports that it happened
  there, and the two call sites say which case they are.
- Deleted `SystestResultCheck.{hpp,cpp}`.

Why this shape:

- A check outcome is a value rather than a bool plus log output. The checkers
  do not print; the caller decides how to report a `Mismatch`.
- A check is a value too. A test builds an `ExpectedOutputCheck` from a file
  and its expected tuples and reads the `Verdict`, without any runner state.
- The set of checks is closed and compiler checked. A new check becomes usable
  by being added to `AnyCheck`, which is also where the `Check` concept is
  enforced on it, and a dispatch site that handles the alternatives by name
  stops compiling when one is added.
- The checkers are declared in namespace `NES`. `Systest` stays reserved for
  names that would clash in a library that links `nes-systest-lib`.

Failure messages no longer start with two blank lines, so a report reads
`Error: Result Mismatch` on one line instead of leaving `Error:` alone above a
gap. The parts of a comparison report are joined with a blank line between them
rather than each part prefixing its own.

## Verifying this change

This change is tested by
- `slt-parser-test` in the dev container: 51 tests passing, of which 16 cover
  the checkers and 2 the case id. Among them: dispatch through `runCheck` over
  every `AnyCheck` alternative, a differential block reporting which of its
  queries wrote no output, empty file vs empty header vs malformed header as
  verdicts, and two golden tests that pin the text of a schema and of a result
  mismatch report.
- the whole `nes-systests` corpus in the dev container: 760 queries, all
  passing, which covers the result, differential, explain, and error cases
- `clang-tidy-diff` on the changed lines, with no findings in the files this PR
  touches
- the format gate (`scripts/format.sh`), clean on the branch

- nes-systests only: how a query's output is checked and how a failure is
  reported. No dependency changes. The checker tests joined `slt-parser-test`,
  so there is no new test target.

## Documentation

- Each checker struct documents what its check compares, and `Check.hpp`
  documents the concept, the variant, and the dispatch.

This PR closes #1915, part of the systest refactoring epic #1920. Follow-up on
reporting the phase of a failed query: #1974.


nes-systests/systest/private/Model/TestCaseId.hpp
```diff
@@ -0,0 +1,54 @@
...
+        if (id.queryIdInFile != NES::INVALID<NES::SystestQueryId>)
+        {
+            /// Two digits, so the numbers of one file line up in the report.
+            /// A file with more than 99 queries widens the column.
+            out = fmt::format_to(out, ":{:02}", id.queryIdInFile.getRawValue());
+        }
+        return out;
+    }
+};
+
```
Artraxon: even though I prefer specializing the formatter struct, our usual approach is to implement the operator<< for ostream and then use the FMT_OSTREAM macro to add the fmt specialization. Of course you can still use fmt internally to implement the operator<< . Also, all this can be in a .cpp

nes-systests/systest/private/Model/Verdict.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+/// Why a check failed, as text describing how the actual output differs from the expected one.
+/// A struct, so it can later store more structured mismatch information without touching call sites.
+struct Mismatch
+{
+    std::string detail;
+};
+
+/// The outcome of one check: the empty value when it passed, and a mismatch describing the failure otherwise.
+using Verdict = std::expected<void, Mismatch>;
```
Artraxon: IMO its much better to use std::monostate instead of void here, because void is kind of a weird type in c++ . Its both the trivially constructible, no-information unit-type (like std::monostate), but its also the impossible to construct type at the same time. Only downside of std::monostate is that you bring in another header. Or, alternatively just define your own empty Success struct with no fields, will result in the same thing.

nes-systests/systest/private/ResultChecker/Check.hpp
```diff
@@ -0,0 +1,56 @@
...
+/// A new check becomes usable by being added here.
+using AnyCheck = std::variant<ExpectedOutputCheck, DifferentialCheck, ExplainCheck>;
+
+namespace detail
+{
+template <typename... Ts>
+consteval bool allAlternativesAreChecks(std::variant<Ts...>*)
+{
+    return (Check<Ts> and ...);
+}
```
Artraxon:
```c++
template <typename... Ts>
concept AllAlternativesAreChecks = (Check<Ts> and ...);
```

nes-systests/systest/private/ResultChecker/ExplainChecker.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <string>
+
+#include <Model/Expectation.hpp>
+#include <Model/Verdict.hpp>
+
+namespace NES
+{
+
+/// One explain check.
+/// A `<REGEX>` or `<!REGEX>` tag on the expected lines makes the checker match each as a regex instead of comparing it directly.
```
Artraxon: would it make sense to encode this in the type?

nes-systests/systest/private/ResultChecker/ResultFile.hpp
```diff
@@ -0,0 +1,39 @@
...
+{
+
+struct TestCaseOutput
+{
+    Schema<UnqualifiedUnboundField, Ordered> outputSchema;
+    std::vector<std::string> outputTuples;
+};
+
+/// Reads a result file, taking the first line as the schema header and the rest as data rows.
+/// Returns nullopt when the file cannot be opened or holds nothing.
```
Artraxon: what is when the expected output is supposed to be nothing?

nes-systests/systest/private/SystestState.hpp
```diff
@@ -159,7 +160,8 @@ struct RunningQuery
     std::optional<DistributedQueryStatusSnapshot> queryStatus;
     std::optional<uint64_t> bytesProcessed{0};
     std::optional<uint64_t> tuplesProcessed{0};
-    bool passed = false;
+    /// What the check said about this query, which the runner fills in once the query has been checked.
+    Verdict verdict;
```
Artraxon: Is this initialized correctly? IMO probably makes sense to at least make it an optional so its clear when it has not been run yet.

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
```diff
@@ -0,0 +1,679 @@
...
+
+template <typename T, typename Tag>
+class ResultCheckStrongType
+{
+public:
+    explicit constexpr ResultCheckStrongType(T value) : value(std::move(value)) { }
+
+    using Underlying = T;
+    using TypeTag = Tag;
+
```
Artraxon: What are these type aliases used for?

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
```diff
@@ -0,0 +1,679 @@
...
+using ExpectedResultField = ResultCheckStrongType<std::string, struct ExpectedResultFields_>;
+using ActualResultField = ResultCheckStrongType<std::string, struct ActualResultFields_>;
+
+template <typename FieldType, typename Tag>
+class ResultTuple
+{
+public:
+    explicit ResultTuple(std::string tuple) : tuple(std::move(tuple)) { }
+
+    using TupleType = Tag;
```
Artraxon: what is this type alias used for?

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
```diff
@@ -0,0 +1,679 @@
...
+        }
+        const auto& expectedIdentifier = *expectedField.getFullyQualifiedName().begin();
+        const auto& actualIdentifier = *actualField.getFullyQualifiedName().begin();
+        if (expectedIdentifier.isCaseSensitive())
+        {
+            return expectedIdentifier == actualIdentifier;
+        }
+        return expectedIdentifier.asCanonicalString() == toUpperCase(actualIdentifier.asCanonicalString());
+    };
+
```
Artraxon, line 365: why not just use the operator== of the UnqualifiedUnboundField? And btw the Identifiers and QualifiedIdentifiers also have it implemented

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
```diff
@@ -0,0 +1,679 @@
...
+        for (auto it = begin; it != end; ++it)
+        {
+            const auto idx = static_cast<size_t>(std::distance(begin, it));
+            if (fieldsMatch(expectedField, *it) and not matchedActualResultFields.contains(idx))
+            {
+                matchingFieldIt = it;
+                break;
+            }
+        }
+        if (matchingFieldIt != end)
```
Artraxon: At least do a for each loop over the schema with views::enumerate since manual loops are always a bit fishy. you might be able to just use std::ranges::find as well

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
```diff
@@ -0,0 +1,679 @@
...
+            expectedToActualFieldMap.expectedResultsFieldSortIdx.emplace_back(expectedFieldIdx);
+            expectedToActualFieldMap.actualResultsFieldSortIdx.emplace_back(offset);
+        }
+        else
+        {
+            expectedToActualFieldMap.schemaError += fmt::format("\n- '{}' is missing from actual result schema.", expectedField);
+            expectedToActualFieldMap.expectedToActualFieldMap.emplace_back(expectedField.getDataType(), std::nullopt);
+        }
+    }
+    for (size_t fieldIdx = 0; fieldIdx < std::ranges::size(actualResultSchema.getRawValue()); ++fieldIdx)
```
Artraxon: In general I don't quite understand what the goal of this big loop is, what is the expected output here?

Why can't it be solved by just using asserting the same size of the schemas and using std::views:zip to combine the schemas, and then filter for where the schemas don't match?

You can also maybe take a look at SinkLogicalOperators inferSchema method where we also check for the schema sameness and then try to build a good error string

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
```diff
@@ -0,0 +1,679 @@
...
+    }
+    for (const auto& additionalIdx : expectedToActualFieldMap.additionalActualFields)
+    {
+        expectedLine += "_ ";
+        actualLine += fmt::format("{} ", splitActualResult.at(additionalIdx));
+    }
+    resultError += fmt::format("\n{} | {}", expectedLine, actualLine);
+    lineIdxIt.advanceExpected();
+    lineIdxIt.advanceActual();
+}
```
Artraxon: building strings iteratively is always not so nice, I'd either use a stringstream, or even better, just use fmt with fmt::join for the iterator parts

nes-systests/systest/src/ResultChecker/ResultFile.cpp
```diff
@@ -0,0 +1,107 @@
...
+namespace
+{
+UnqualifiedUnboundField parseField(const std::string_view field)
+{
+    std::vector<std::string_view> parts;
+    for (const auto subrange : std::ranges::split_view(field, ':'))
+    {
+        parts.emplace_back(trimWhiteSpaces(std::string_view(subrange)));
+    }
+    INVARIANT(parts.size() == 3, "a header field is written as name:TYPE:NULLABILITY, but was: {}", field);
```
Artraxon: are you sure an invariant is right here? if user input (i.e. a systest file) can cause this, an exception should be thrown instead

nes-systests/systest/src/Runner/SystestRunner.cpp
```diff
@@ -85,16 +92,120 @@ void reportResult(
...
+    }
+
+    if (query.differentialQueryPlan.has_value())
+    {
+        return runCheck(
+            DifferentialCheck{.firstOutputFile = query.resultFile(), .secondOutputFile = query.resultFileForDifferentialQuery()});
+    }
+
+    const auto* expectedRows = std::get_if<ExpectedRows>(&query.expectation);
+    INVARIANT(expectedRows != nullptr, "Systest was expected to have an expected result");
```
Artraxon: you can use the NES::get in Variant.hpp for this check

Artraxon (changes requested):
Looks mostly good to me, its a nice improvement to the state before

nes-systests/systest/private/ResultChecker/ExplainChecker.hpp
adrianmichalke: agree, REGEX- and VERBATIM-explain should be seperate checks

nes-systests/systest/private/Model/TestCaseId.hpp
```diff
@@ -0,0 +1,54 @@
...
+namespace NES
+{
+
+/// Identifies one case in a run's reports: the test file that it came from, and the query's number within that file.
+/// A report line carries this rather than the file path, because one run reports many cases from the same file.
+struct TestCaseId
+{
+    std::string originFile;
+    SystestQueryId queryIdInFile = INVALID<SystestQueryId>;
```
adrianmichalke, line 33: Isnt it also defined by the configuration variant?

nes-systests/systest/src/ResultChecker/DifferentialChecker.cpp
```diff
@@ -0,0 +1,67 @@
...
+std::expected<TestCaseOutput, std::string> outputToCompare(const std::filesystem::path& file)
+{
+    auto output = loadTestCaseOutput(file);
+    if (not output)
+    {
+        return std::unexpected(fmt::format("wrote no output to {}", file.string()));
+    }
+    if (std::ranges::size(output->outputSchema) == 0)
+    {
+        return std::unexpected(fmt::format("wrote a header that declares no fields to {}", file.string()));
```
adrianmichalke:
```suggestion
   return std::unexpected(
       fmt::format("result file has an empty schema header: {}", file));
```

adrianmichalke: add the same expectedOutputChecker

nes-systests/systest/src/ResultChecker/QueryResultChecker.cpp
adrianmichalke, line 1: IMO this should be the QueryResultChecker as output is a rather broad term

nes-systests/systest/src/ResultChecker/QueryResultChecker.cpp
```diff
@@ -0,0 +1,40 @@
...
+namespace NES
+{
+
+Verdict ExpectedOutputCheck::check() const
+{
+    const auto loaded = loadTestCaseOutput(outputFile);
+    if (not loaded)
+    {
+        return std::unexpected(Mismatch{fmt::format("result file was not written: {}", outputFile.string())});
+    }
```
adrianmichalke, line 32: why not check here for the empty header as well?

nes-systests/systest/src/ResultChecker/ResultFile.cpp
adrianmichalke: aggree, an invariant should be used here to track bugs within the systest tool. 
In this case the query did produce a wrong header which should not terminate the systest tool

nes-systests/systest/src/ResultChecker/ResultFile.cpp
```diff
@@ -0,0 +1,107 @@
...
+        NES_ERROR("Failed to open result file: {}", outputFilePath);
+        return std::nullopt;
+    }
+
+    TestCaseOutput result;
+    std::string firstLine;
+    if (!std::getline(resultFile, firstLine))
+    {
+        NES_ERROR("Result file '{}' is empty", outputFilePath);
+        return std::nullopt;
```
adrianmichalke: nit: as we will report the verdict anyways error prints/logs are redundant

nes-systests/systest/src/ResultChecker/ResultFile.cpp
```diff
@@ -0,0 +1,107 @@
...
+    if (not nullable)
+    {
+        throw SLTUnexpectedToken("Unknown nullable: {}", parts.at(2));
+    }
+
+    const auto type = magic_enum::enum_cast<DataType::Type>(parts.at(1));
+    if (not type and toLowerCase(parts.at(1)) != "varsized")
+    {
+        throw SLTUnexpectedToken("Unknown basic type: {}", parts.at(1));
+    }
```
adrianmichalke: All the error cases here shoud just create a failed verdict. I would perfer to use another control flow mechanism for this (expected, optional) rather then using exeptions.

nes-systests/systest/private/ResultChecker/QueryResultChecker.hpp
adrianmichalke, line 1: Why not name it TestCaseOutput? It does not include any ResultFile type

nes-systests/systest/src/Runner/SystestRunner.cpp
```diff
@@ -105,45 +216,8 @@ void processQueryWithError(
...
-                }
-
-                return std::string{};
-            }
-
-            return fmt::format("Query Failed with unexpected error: {}", *runningQuery->exception);
-        },
-        performanceMessageBuilder);
+    auto verdict = checkFailedQuery(runningQuery->exception, runningQuery->systestQuery.expectation);
+    reportResult(runningQuery, progressTracker, failed, std::move(verdict), performanceMessageBuilder);
```
adrianmichalke, line 233: Previoulsly we differenciated between between errors during i) parsing ii) registration/compilation/startup and iii) runtime. The step of the error got added in the log print. IMO this differenciation is useful and should be preserved. 

I guess it got removed with e3c7e60fa835b767fa24bd8eb0b31ae718226ada

nes-systests/systest/private/ResultChecker/ResultFile.hpp
adrianmichalke: Then it only includes the header.

nes-systests/systest/private/ResultChecker/ResultFile.hpp
```diff
@@ -0,0 +1,39 @@
...
+#include <Schema/SchemaFwd.hpp>
+
+namespace NES
+{
+
+struct TestCaseOutput
+{
+    Schema<UnqualifiedUnboundField, Ordered> outputSchema;
+    std::vector<std::string> outputTuples;
+};
```
adrianmichalke: To avoid convsions we should call it Observed- or ActualTestCaseOutput in contrast to the ExpectedTestCaseOutput

nes-systests/systest/src/Runner/SystestRunner.cpp
```diff
@@ -85,16 +92,120 @@ void reportResult(
...
+    if (std::holds_alternative<ExpectedError>(query.expectation))
+    {
+        return std::unexpected(
+            Mismatch{fmt::format("expected error {} but query succeeded", std::get<ExpectedError>(query.expectation).code)});
+    }
+
+    if (writesIntoDiscardingSink(query))
+    {
+        NES_INFO("Skipping result check for {}:{} because it writes to a Void sink.", query.testName, query.queryIdInFile);
+        return {};
```
adrianmichalke: I would prefer if the Verdict would have a dedicated Success type. Modelling it as an empty set here for a success is rather confusing

nes-systests/systest/tests/ResultChecker/ResultCheckerTests.cpp
adrianmichalke, line 1: I would like to see some tests here for checking that the result mismatch reports still work as before

adrianmichalke (changes requested):

nes-systests/systest/private/Model/TestCaseId.hpp
```diff
@@ -0,0 +1,54 @@
...
+    static constexpr auto parse(const format_parse_context& ctx) -> decltype(ctx.begin()) { return ctx.begin(); }
+
+    static auto format(const NES::TestCaseId& id, const format_context& ctx) -> decltype(ctx.out())
+    {
+        auto out = fmt::format_to(ctx.out(), "{}", id.originFile);
+        if (id.queryIdInFile != NES::INVALID<NES::SystestQueryId>)
+        {
+            /// Two digits, so the numbers of one file line up in the report.
+            /// A file with more than 99 queries widens the column.
+            out = fmt::format_to(out, ":{:02}", id.queryIdInFile.getRawValue());
```
keyseven123: I mean 99 queries per file are not idea, but why not give us here 100 or even 1k queries per file?

nes-systests/systest/private/ResultChecker/DifferentialChecker.hpp
```diff
@@ -0,0 +1,33 @@
...
+namespace NES
+{
+
+/// One differential check: the two result files have to agree.
+struct DifferentialCheck
+{
+    std::filesystem::path firstOutputFile;
+    std::filesystem::path secondOutputFile;
+
+    [[nodiscard]] Verdict check() const;
```
keyseven123, line 32: Maybe add here as a comment or for the struct that the check happens on a tuple base. So we do not check if the file is byte-identical

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
```diff
@@ -0,0 +1,679 @@
...
+    [[nodiscard]] bool hasNext() const
+    {
+        return (expectedResultTupleIdx.getRawValue() + actualResultTupleIdx.getRawValue()) < totalResultLinesSize;
+    }
+
+    [[nodiscard]] ExpectedResultIndex getExpected() const { return expectedResultTupleIdx; }
+
+    [[nodiscard]] ActualResultIndex getActual() const { return actualResultTupleIdx; }
+
+    void advanceExpected() { this->expectedResultTupleIdx = ExpectedResultIndex(this->expectedResultTupleIdx.getRawValue() + 1); }
```
keyseven123: I like to use `{}` for constructor calls instead of `()`, as it is easier to spot and also disallows narrowing conversions. But this is not binding.

nes-systests/systest/src/Runner/SystestRunner.cpp
keyseven123: I agree with @adrianmichalke. There should be a clear distinction between a success and an empty Verdict.

keyseven123 (changes requested):
I really like this `Verdict` class. Cool idea.

nes-systests/systest/private/Model/Verdict.hpp
yschroeder97: Agreed, decided for the dedicated `Success` struct.

nes-systests/systest/src/ResultChecker/ResultFile.cpp
yschroeder97: Changed, see my other replies.

nes-systests/systest/src/ResultChecker/ResultFile.cpp
yschroeder97: Rewrote loadTestCaseOutput to return std::expected<TestCaseOutput, std::string>, with the checker converting the output into a Verdict, and the string into aMismatch.
This way, we have a unified error handling strategy in all places where errors should be a failed test case.

nes-systests/systest/src/ResultChecker/ResultFile.cpp
yschroeder97: Fixed.

nes-systests/systest/src/ResultChecker/QueryResultChecker.cpp
yschroeder97: Done.

nes-systests/systest/src/ResultChecker/DifferentialChecker.cpp
yschroeder97: Fixed.

nes-systests/systest/private/ResultChecker/ResultFile.hpp
yschroeder97: Clarified in a comment.

nes-systests/systest/private/ResultChecker/Check.hpp
yschroeder97: A pack concept cannot look inside a variant, so the assertion would have to repeat the four alternatives below. 

The specialization reads them from AnyCheck, so adding a check stays one edit with this: 
```c++
template <typename>
constexpr bool AllAlternativesAreChecks = false;
template <typename... Ts>
constexpr bool AllAlternativesAreChecks<std::variant<Ts...>> = (Check<Ts> and ...);

static_assert(AllAlternativesAreChecks<AnyCheck>);
```

nes-systests/systest/private/ResultChecker/ExplainChecker.hpp
yschroeder97: Agree, split this into distinct types, which leads to the branching happening in the runner, and each type can check its specific output type.

nes-systests/systest/private/ResultChecker/ResultFile.hpp
yschroeder97: Renamed to `QueryResult` and `QueryResultCheck`

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
yschroeder97: Removed.

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
yschroeder97: Removed.

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
yschroeder97: Agree, changed.

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
yschroeder97: Changed to a lambda with `std::ranges::find_if`.

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
yschroeder97: Identifier::operator== compares canonical forms. The expected side comes from the test's SQL, where an unquoted id canonicalizes to ID.
The header from the result file is parsed case-preserving, so the report shows what the sink wrote. Under ==, those compare unequal, so every unquoted field name would fail. The lambda keeps the case rule; I added a comment saying why == cannot do it

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
yschroeder97: The loop matches fields by name and type, not by position, so the sink may write the columns in a different order than the test declares. Unmatched fields on either side are reported. zip would require the same order and count, which is stricter than what the checker accepts today. The logic is in main; moved. I added a comment above the loop that says this.

nes-systests/systest/src/Runner/SystestRunner.cpp
yschroeder97: Changed to return `Sucess{}`.

nes-systests/systest/src/Runner/SystestRunner.cpp
yschroeder97: Changed.

nes-systests/systest/src/Runner/SystestRunner.cpp
yschroeder97: You may be right, but I don't see it as my task for this PR. We can tackle this later, IMO.

nes-systests/systest/private/Model/TestCaseId.hpp
yschroeder97: Changed.

nes-systests/systest/private/Model/TestCaseId.hpp
yschroeder97: Removed the constraint.

nes-systests/systest/private/ResultChecker/DifferentialChecker.hpp
yschroeder97: Added a comment.

keyseven123 (approved):

nes-systests/systest/tests/ResultChecker/ResultCheckerTests.cpp
yschroeder97: Done, added two tests to confirm this.

nes-systests/systest/private/Model/TestCaseId.hpp
yschroeder97: Yes, added this to distinguish cases with different configurations (for the report).

nes-systests/systest/private/ResultChecker/QueryResultChecker.hpp
yschroeder97: Renamed to `QueryResultCheck` for consistency.

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
Artraxon: That sounds pretty cursed by I guess its fine for now

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
Artraxon: Why should we allow some arbitrary order here? This is part of the semantics of a query. I also don't think that any query requires this, since I made the column order deterministic in the schema inference PR

Artraxon (dismissed):
I only left one more comment in one of the old threads, I kept it unresolved. Everything else looks fine

> Why should we allow some arbitrary order here? This is part of the semantics of a query. I also don't think that any query requires this, since I made the column order deterministic in the schema inference PR

nes-systests/systest/src/ResultChecker/ResultComparison.cpp
Artraxon: If its easy to change and doesn't break any tests IMO it can be just done here, otherwise It'd warrant an issue

adrianmichalke (approved):

yschroeder97:
@Artraxon 

Fields now pair up by position, and the row values compare positionally too. 
A column at the wrong position is reported as missing on one side and unexpected on the other. 
The overall verdict is unchanged because the schema equality check before was already positional, so no test needed to change.

Demo:
```
# name: bug/SchemaOrderDemo.test
# description: demo of the positional schema mismatch report
# groups: [Demo]

CREATE LOGICAL SOURCE input(id UINT64 NOT NULL, value UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR input TYPE File;
ATTACH INLINE
1,10
2,20

SELECT id, value FROM input INTO File();
====
SELECT value, id FROM input INTO File();

Error: Schema Mismatch
---------------
(ID UINT64), (VALUE UINT64) != (VALUE UINT64), (ID UINT64)
- 'ID UINT64' is missing from actual result schema.
- 'VALUE UINT64' is missing from actual result schema.
+ 'VALUE UINT64' is unexpected field in actual result schema.
+ 'ID UINT64' is unexpected field in actual result schema.

Result Mismatch
Expected Results(Sorted) | Actual Results(Sorted)
-------------------------------------------------
1 10 _ _ | _ _ 10 1
2 20 _ _ | _ _ 20 2
```

PR #1937, opened 2026-08-20
feat(systest): SQL rewriter

### LINKED ISSUE #1916: Add a SQL rewriter that makes statements correct upfront

Add a rewriter that transforms a parsed test file into runnable SQL: qualification of every catalog-visible name with the test-file key, physical-source option splicing (file path, host, input formatter), inline data staged to files, sink inlining with a per-query result file, and a query-name stamp as the correlation key. A pure library (ANTLR lexer/parser only) checked by its own golden tests; nothing calls it yet.

yschroeder97:

Stacked on #1936.

The rewriter turns a parsed test file into statements that we can then submit unchanged, so any mid-pipeline plan patching of the binder can be deleted in the next step. 
Nothing calls it yet, except its unit tests. 

- Added `Model/RunnableTest.hpp`, the rewriter's output and the runner's future input: `RunnableTest` holds the setup DDL, the cases to run and check, and the data to stage (`InlineData` for a planned CSV, `ServedData` for a TCP source).
- Added the rewriter as three phases behind one umbrella class,`SqlRewriter::rewrite(const TestFile&) const`, each phase a file pair in `Rewriter/`:
  - `PreparedStatement`: parses and classifies every CREATE statement.
  - `Declarations`: registers every catalog-visible name and captures the declared sinks, sealed (i.e., consumed) before any rewriting.
  - `Emitter`: emits the setup DDL and rewrites the queries, in file order.
    An instance exists for a single test file, so no state carries over.
- Qualification prefixes every catalog-visible name with the test file key,
  token by token through the lexer (`NameQualifier`), so literals and
  whitespace stay as the test wrote them, and every test file of one
  invocation can share a single catalog.
- Physical sources get their file path, host and input formatter injected;
  inline data becomes a planned CSV file and socket data a served payload
  (`SourceRewriting`, `DataStaging`).
- The sink of each query is inlined with its schema and a result file of its
  own (`SinkRewriting`), and a `SET` clause puts the query name as the
  correlation key. A statement that does not parse is still qualified,
  so a syntax error test will still be caught later.

## Verifying this change

This change is tested by
- two new unit test targets: `systest-sql-rewriter-test`
  (24 tests: qualification, source splicing, sink inlining, EXPLAIN,
  differential blocks, staged data) and `systest-name-qualifier-test`
  (14 tests: key derivation, sealing, unqualification)
- the parser and runner suites, unchanged: `testfile-parser-test` (9) and
  `slt-parser-test` (40)
- the full corpus in the dev container

- nes-systests only. The library now has a `Rewriter/` stage and links
  `tcp_source_plugin_library`; no other dependency changes.
- Two new test targets.

## Documentation

- Every phase type and every rewriting rule has extensive in-code docs.

This PR closes #1916, part of the systest refactoring epic #1920.

### DIFF
--- a/nes-systests/systest/CMakeLists.txt
+++ b/nes-systests/systest/CMakeLists.txt
@@ -95,7 +95,7 @@ find_package(reflectcpp CONFIG REQUIRED)
 find_package(argparse CONFIG REQUIRED)
 
 add_library(nes-systest-lib STATIC ${SLT_SOURCE_FILES})
-target_link_libraries(nes-systest-lib PUBLIC nes-schema nes-frontend-embedded-lib nes-single-node-worker-lib nes-sql-parser nes-query-optimizer reflectcpp::reflectcpp PRIVATE nes-input-formatters argparse::argparse)
+target_link_libraries(nes-systest-lib PUBLIC nes-schema nes-frontend-embedded-lib nes-single-node-worker-lib nes-sql-parser nes-query-optimizer reflectcpp::reflectcpp PRIVATE nes-input-formatters argparse::argparse tcp_source_plugin_library)
 target_include_directories(nes-systest-lib PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/private>)
 target_compile_definitions(nes-systest-lib PUBLIC
         PATH_TO_BINARY_DIR="${CMAKE_BINARY_DIR}"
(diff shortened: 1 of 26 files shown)


nes-systests/systest/src/CMakeLists.txt
```diff
@@ -11,11 +11,12 @@
 # limitations under the License.
 
 # The stages of the systest pipeline, in the order they run:
-# configure, discover test files, read one, parse it, run it and check the results.
+# configure, discover test files, read one, parse it, rewrite its statements, run it and check the results.
```
adrianmichalke: nit: would rm the comment as we have to maitain it

yschroeder97:
Closed by a branch rename (`ys/systest-rewriter` -> `ys/systest/rewriter`). Continues in #1993.

PR #1940, opened 2026-08-21
fix(worker): Unblocked termination signals on early error paths

keyseven123:
`main` blocks `SIGINT`/`SIGTERM` before starting the worker, and only the shutdown hook consumes them — but that hook does not exist until the gRPC server is running. If `main` leaves via the early error paths before that point, a `SIGTERM` stays pending forever and the process can only be killed with `SIGKILL`.

Restores the default disposition on those paths so the process terminates promptly.


Artraxon:
I have created another PR that also deals with early exists because of invariants and direct returns correctly https://github.com/nebulastream/nebulastream/pull/1961

Artraxon (changes requested):
IMO you can close this PR since we have a more extensive fix in another one

keyseven123:
> IMO you can close this PR since we have a more extensive fix in another one

Closed this pr

Issue #1798, opened 2026-07-18
[TASK] Uniform Configuration Interface


Artraxon:
### Description

Currently, we have two configuration approaches, both with a set of problems:

#### DescriptorConfig
The descriptors where created to receive config parameters from a frontend, validate them, store them and serialize and deserialize them.
While they where fine as an intermediate solution, combining all of these responsibilities in one class comes with big restrictions on their functionality and ergonomics.

#### CLI-Options
The CLI-Option system is used to pass config values for the system (not catalog objects) via both the CLI and YAML.
It tries to combine the definition of options, the validation and parsing, with the holding the values. Its more of a YAML-Binder that happens to also bind some simpler structures to CLI-Options.

#### Issues of both systems
1. Validation and parsing are separate, requiring duplicate logic (see the generator source)
2. Errors are either not forwarded at all (most of the DescriptorConfig) which means users don't understand why their arguments where rejected and have to look into the log, or via exceptions, which is potentially unsafe, depending on the context in which the validation is run.

#### Of DescriptorConfig

1. When it was conceived, we didn't have the reflection system yet, so it can only store elements of a fixed variant. This has two downsides:
1.1. If you want to store more complicated data structures (such as the schema of the generator source), you have to store it as a string in some kind of format and reinterpret it every time you want to access it (i.e. instantiate a source).
1.2 You cannot encode the validity of arguments (for the consuming code) in types, such as strong types or variants.

In other words, despite their being validation functions, you never know whether the config value are actually correct or whether you need to validate again.

#### Of CLI Options

1. All parsing is done through YAML, which is not frontend agnostic.
2. At the same time, it also tries to work with explicitly nested config structures, and allows the sequencing of it, which of course is something which neither makes sense for the CLI-Options or for any configuration of catalog objects.
3. The structure of the type holding the validated information needs to match the structure of the input 1:1, you cannot combine multiple values or verify the validity of the combination of multiple values (as in, either just this option is set, or just the other option).
4. Querying whether something was explicitly set is IMO an antipattern, and is rather an indicator that its to difficult to just make an options return type something like std::optional<size_t> and set the default to std::nullopt, or reason about the combination of set options. 
5. Quite complex code in general, for example collecting all options requires a visitor pattern since its a nested data structure.

To a certain degree, the problem of the CLI Frontend is that it tries to do too much. 2) should be moved to frontend specific logic, and 4) can be replaced with better ergonomics around the more "complex" config types and the whole configuation flow in general.

[...]

PR #1948, opened 2026-08-23
Migrate Sources and Input Formatter to Uniform Config System

Artraxon:
Migrates the sources and input formatters to the new uniform config system as part of #1798 .
Most of it is pretty straightforward, the biggest change is that I also slightly refactored the SourceCatalog and the flow to get a SourceDescriptor.
You can now ask the SourceCatalog to give you a `SourceConfigSchema` for a given source type and input formatter type, and which you can use to resolve your config literals.
It will directly give you a tuple of a `GeneralSourceConfig` (containing things like maxInflightBuffers), a `PluginSourceConfiguration` (which is a pair of the source type and its type erased config struct), an `InputFormatterDescriptor` (a pair of the same things) and an optional `Schema<UnqualifiedUnboundField, Ordered>`.

Creating a physical source for a logical source and creating an anonymous source now share more common functions and model the absence/presence of schemas better, since the untyped descriptor configs are not passed around and stored for as long as they were previously.

I had to leave the old `DescriptorConfig`-like interface in there in addition, because the binders depend on it, they will be refactored  by the last PR in the stack.

Sources and InputFormatters


nes-input-formatters/include/InputFormatterProvider/InputFormatterValidationProvider.hpp
```diff
@@ -13,14 +13,9 @@
...
 #pragma once
 
-#include <optional>
-#include <string>
+#include <expected>
 #include <string_view>
-#include <unordered_map>
-#include <Configurations/Descriptor.hpp>
 
 namespace NES::InputFormatterValidationProvider
```
keyseven123: Remove this file and its .cpp

nes-input-formatters/src/InputFormatterValidationProvider.cpp
```diff
@@ -14,27 +14,13 @@
...
 #include <string_view>
-#include <unordered_map>
 #include <utility>
-#include <Configurations/Descriptor.hpp>
+
 #include <InputFormatterDescriptor.hpp>
-#include <InputFormatterValidationRegistry.hpp>
 
 namespace NES::InputFormatterValidationProvider
 {
```
keyseven123, line 34: Do we need this empty .cpp file now?

nes-plugins/Sources/GeneratorSource/GeneratorSource.cpp
```diff
@@ -15,55 +15,327 @@
...
+    std::vector<std::string_view> params;
+
+    for (const auto& param : splitOnMultipleDelimiters(configString, {'\n', ','}))
+    {
+        params.emplace_back(trimWhiteSpaces(param));
+    }
+
+    if (params.size() == 2)
+    {
+        const auto amplitudeParams = splitWithStringDelimiter<std::string_view>(params[0], " ");
```
keyseven123, line 90: We should check of the return of `splitWithStringDelimiter` is actually a size of 2. `GENERATOR_RATE_CONFIG = 'amplitude, frequency 2'` would throw

nes-plugins/Sources/GeneratorSource/CMakeLists.txt
```diff
@@ -13,11 +13,13 @@
 include(${PROJECT_SOURCE_DIR}/cmake/RuntimeRegistrationUtil.cmake)
 
 # One implementation library serving all registries this plugin participates in.
-add_library(generator_source_plugin_library STATIC GeneratorSource.cpp Generator.cpp GeneratorFields.cpp FixedGeneratorRate.cpp SinusGeneratorRate.cpp)
+add_library(generator_source_plugin_library STATIC
```
keyseven123, line 16: You removed the files but did not delete them

keyseven123 (changes requested):

nes-input-formatters/include/InputFormatterProvider/InputFormatterValidationProvider.hpp
Artraxon: done

nes-input-formatters/src/InputFormatterValidationProvider.cpp
Artraxon: nope, removed

nes-plugins/Sources/GeneratorSource/GeneratorSource.cpp
Artraxon: added checks

nes-plugins/Sources/GeneratorSource/CMakeLists.txt
Artraxon: done

keyseven123 (approved):

nes-sources/src/SourceCatalog.cpp
```diff
@@ -285,4 +379,170 @@ std::unordered_map<LogicalSource, std::unordered_set<SourceDescriptor>> SourceCa
...
+    const std::unordered_map<Identifier, std::string>& descriptorConfig,
+    const std::unordered_map<Identifier, std::string>& parserConfig)
+{
+    const auto& typeKey = InputFormatterDescriptor::TYPE_FIELD.getName();
+    const auto typeIter = std::ranges::find_if(parserConfig, [&](const auto& entry) { return entry.first == typeKey; });
+    if (typeIter == parserConfig.end())
+    {
+        return std::unexpected{InvalidConfigParameter("Source config does not contain input formatter type")};
+    }
+    const auto inputFormatterType = Identifier::parse(typeIter->second);
```
ls-1801, line 411: There is no early rejection of an empty input formatter type. Passing `'' AS INPUT_FORMATTER."TYPE"` reaches `Identifier::parse`, trips its nonempty precondition, and terminates the process instead of returning a configuration error. Use `Identifier::tryParse` and propagate the error.

Reproduced with this systest:

```sql
SELECT ID, VALUE, TIMESTAMP
FROM File(
    'small/stream8.csv' AS "SOURCE".FILE_PATH,
    '' AS INPUT_FORMATTER."TYPE",
    SCHEMA(id UINT64 NOT NULL, value UINT64 NOT NULL, timestamp UINT64 NOT NULL) AS "SOURCE"."SCHEMA")
INTO File();
----
ERROR 1000 "Could not create an anonymous source descriptor because of invalid config parameters"
```

Running `systest -t nes-systests/regression/EmptyInputFormatterType.test:01` aborts with:

```text
Precondition violated: (!value.empty()): Invalid identifier, cannot be empty
terminate called without an active exception
```

nes-sources/src/SourceCatalog.cpp
```diff
@@ -285,4 +379,170 @@ std::unordered_map<LogicalSource, std::unordered_set<SourceDescriptor>> SourceCa
...
+    {
+        /// The formatter type selects the schema; it is only a resolvable field for registered
+        /// formatter plugins (the NATIVE format has none).
+        if (key == typeKey)
+        {
+            continue;
+        }
+        /// Transitional contract of the map API: whatever does not parse as a typed literal stays
+        /// a raw string (delimiters like "\n" trim to empty and must survive as strings).
+        literals.emplace_back(QualifiedIdentifier::create(key), parseConfigLiteral(value).value_or(ConfigLiteral{value}));
```
ls-1801, line 432: We already discussed this in the initial PR, but I still don’t think eager type conversion is what we want here. Numeric-looking strings should remain usable in fields that expect strings, especially when explicitly quoted in SQL.

This might be a slightly contrived example, but this systest fails because we cannot use `'0'` as the delimiter:

```sql
CREATE LOGICAL SOURCE digitDelimited(leftValue UINT64 NOT NULL, rightValue UINT64 NOT NULL);

# The legacy adapter receives the string "0" after SQL binding. Reinterpreting it
# as an integer makes the formatter's string-only delimiter validation fail.
CREATE PHYSICAL SOURCE FOR digitDelimited TYPE File SET(
    'CSV' AS INPUT_FORMATTER."TYPE",
    '0' AS INPUT_FORMATTER.FIELD_DELIMITER);
ATTACH INLINE
102
304
506

SELECT leftValue, rightValue FROM digitDelimited INTO File();
----
1,2
3,4
5,6
```

Could we defer conversion until the expected field type is known?

nes-plugins/Sources/GeneratorSource/GeneratorSource.cpp
```diff
@@ -15,55 +15,336 @@
...
+         && !std::holds_alternative<FixedGeneratorRateConfig>(config.get(GENERATOR_RATE_CONFIG)))
+        || (config.get(GENERATOR_RATE_TYPE) == GeneratorRate::Type::SINUS
+            && !std::holds_alternative<SinusGeneratorRateConfig>(config.get(GENERATOR_RATE_CONFIG))))
+    {
+        return std::unexpected{InvalidConfigParameter(
+            "Generator rate config must match specified rate type {}", magic_enum::enum_name(config.get(GENERATOR_RATE_TYPE)))};
+    }
+    return GeneratorSourceConfig{
+        .seed = static_cast<uint32_t>(config.get(SEED)),
+        .maxRuntime = static_cast<int32_t>(config.get(MAX_RUNTIME_MS)),
```
ls-1801, line 337: Both `MAX_RUNTIME_MS` and `SEED` are declared as `ConfigField<int64_t>`, but are stored as 32-bit values in `GeneratorSourceConfig` (`int32_t maxRuntime` and `uint32_t seed`). These casts narrow the accepted 64-bit values without a range check.

nes-plugins/Sources/MQTTSource/MQTTSource.cpp
```diff
@@ -57,24 +63,81 @@ std::string generateClientId(std::string clientId)
...
+const ConfigField<std::string> CLIENT_ID{
+    Identifier::parse("CLIENT_ID"), "The MQTT client ID (leave default to auto-generate)", std::string{GENERATE_CLIENT_ID_TOKEN}};
+
+const ConfigField<std::string> TOPIC{Identifier::parse("TOPIC"), "The MQTT topic to subscribe to"};
+
+const ConfigField<int64_t> FLUSH_INTERVAL_MS{
+    Identifier::parse("FLUSH_INTERVAL_MS"),
+    "Interval in ms between tuple buffer flushes; 0 disables periodic flushing",
+    [](const ConfigLiteral& literal) { return tryGetOr<int64_t>(literal, expectedType<int64_t>()); },
+    int64_t{0}};
```
ls-1801, line 87: The flush interval should not allow negative values. Could we use an unsigned config type and validate the range when converting from the signed integer literal? An enabled interval should be positive, while `0` should remain valid to disable periodic flushing.

ls-1801 (changes requested):

ls-1801 (changes requested):
Rerequest once comments are addressed

PR #1949, opened 2026-08-24
Expose query engine task events as a queryable stream

nikla44:

The query engine already publishes task level events through `QueryEngineStatisticListener`, but the only consumer was `GoogleEventTracePrinter`, which writes a Chrome trace file to be inspected after the fact. This pull request makes the same events an ordinary stream, so that a NES query can select, filter and aggregate the engine's own statistics while it runs.

Sources are constructed per query from nothing but a `SourceDescriptor`, so nothing the worker owns can be injected into one. The bridge is therefore a process global feed keyed by a name string, following what `NetworkSource` does with the global receiver service:

```
TaskStatisticListener  --CSV rows-->  InProcessFeed  --rows-->  InProcessSource
```

- Added `InProcessFeed` and `InProcessFeedRegistry` to `nes-sources`, a bounded MPMC row queue looked up by name. `tryPush()`/`tryPop()` never block, and `droppedRows()` counts what did not fit.
- Added an `InProcess` source plugin (`nes-plugins/Sources/InProcessSource`) that reads the feed of the configured name and emits its rows, registered through `add_registry_entry` for both the `Source` and `SourceValidation` registries.
- Added `TaskStatisticListener` to `nes-single-node-worker`, a sibling of `GoogleEventTracePrinter` that reuses its skeleton: `onEvent` runs on the worker threads and only filters and enqueues, a dedicated thread formats the rows, and both stages drop rather than block.
- Handled `TaskExecutionStart`, `TaskEmit`, `TaskExecutionComplete` and `TaskExpired`, rendered as the row schema `(event_type, ts_us, thread_id, query_id, pipeline_id, task_id, tuples)`. Columns an event does not carry are zeroed.
- Excluded queries that read an `InProcess` source from the statistics themselves. Such a query would otherwise observe the events its own execution produces, which never settles: every buffer it emits creates tasks, which create the events that fill the next buffer. The query is recognised in its submitted plan, which arrives before it can emit anything.
- Added the `--enable_task_statistics` (default `false`) and `--task_statistics_feed` (default `engine_events`) worker options, and wired the listener up in `SingleNodeWorker` before the engine is built, since `CompositeStatisticListener` does not synchronize its listener list against the worker threads that read it.
- Added `scripts/run-engine-stats-demo.sh`, an end-to-end demo of the whole path.

## Verifying this change

This change is tested by

- *Added `InProcessSourceTest` unit tests in `nes-sources/tests`, covering that the feed drops rows once it is full, that the registry hands out the same feed for the same name, and that a source resolved through `SourceProvider` emits exactly what was pushed into its feed.*
- *Added `scripts/run-engine-stats-demo.sh`, which builds the embedded REPL, starts it with `--enable_task_statistics`, runs a Generator query for load plus a second query reading the engine's own task events, and asserts the invariants that hold: statistics rows were produced, every started task also completed, the statistics query excluded itself, and no events were dropped.*

This deliberately stands in for a systest. The statistics source never reaches end-of-stream, so a systest would wait on a query that never finishes, and the number of events an engine produces is not reproducible, so there is no stable expected-result block to compare against.

- **QueryEngine** — adds a second consumer of `QueryEngineStatisticListener` events. The hot path only filters and enqueues; when statistics are disabled no listener is registered at all.
- **Sources** — new `InProcessFeed`/`InProcessFeedRegistry` in `nes-sources` and a new `InProcess` source plugin. Existing sources are untouched.
- **SingleNodeWorker** — two new configuration options and the listener registration.
- **Build** — one new plugin directory registered via `add_plugin`.
- **Dependencies** — none added or upgraded.

Not affected: QueryCompiler, RestAPI, ExecutionEngine operators.

## Documentation

  - In-code documentation is present: the row schema, the non-blocking drop behaviour, the self-exclusion rule and its brittleness, and the listener registration ordering constraint are all documented at their definitions. No separate design document is added; `scripts/run-engine-stats-demo.sh` doubles as executable usage documentation.


Artraxon (changes requested):
IMO InProcess is a bit of a confusing name, and I also don't get why the engine events and the buffer manager events should be in the same source type if they have different schemas anyway.
These should be different source types altogether, e.g. one `EngineEventsSource` and one `BufferManagerEventsSource` or something along the lines.

Of course, they can share internal infrastructure.

nes-sources/src/InProcessFeed.cpp
```diff
@@ -0,0 +1,87 @@
...
+{
+
+InProcessFeed::InProcessFeed(const size_t capacity) : queue(capacity)
+{
+    PRECONDITION(capacity > 0, "An InProcessFeed needs a capacity of at least one row");
+}
+
+bool InProcessFeed::tryPush(std::string row)
+{
+    if (queue.writeIfNotFull(std::move(row)))
```
ls-1801: The current full-queue behavior keeps the oldest rows and discards each newly arriving row. Could we instead retain the most recent `capacity()` rows by removing one row and retrying `writeIfNotFull` in a loop until the new row is accepted?

nes-sources/src/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_source_files(nes-sources
         SourceCatalog.cpp
         FileSource.cpp
         NetworkSource.cpp
+        InProcessFeed.cpp
```
ls-1801: `InProcessFeedRegistry` is implemented in `nes-sources`, so every producer that publishes to it must depend on `nes-sources` even when it is not a source. Could the feed and registry move to a neutral shared library such as `nes-common`, or to a dedicated library, with `nes-sources` and producers depending on that instead?

nes-sources/src/InProcessFeed.cpp
ls-1801: Since this path already permits dropping events, could we use `MPMCQueue::write` instead of `writeIfNotFull`? `write` returns immediately during a concurrent read, while `writeIfNotFull` may wait for that read to finish; the tradeoff is that `write` can reject an event that `writeIfNotFull` would eventually accept.

nes-plugins/Sources/InProcessSource/InProcessSource.cpp
```diff
@@ -0,0 +1,132 @@
...
+{
+    const auto available = tupleBuffer.getAvailableMemoryArea<char>();
+    const auto deadline = std::chrono::steady_clock::now() + flushInterval;
+    size_t written = 0;
+
+    while (not stopToken.stop_requested())
+    {
+        if (not pendingRow.has_value())
+        {
+            pendingRow = feed->tryPop(POP_TIMEOUT);
```
ls-1801: `tryPop(POP_TIMEOUT)` can wait 10 ms even after a shorter `flushInterval` deadline has expired. With `FLUSH_INTERVAL_MS=1` and a single queued row, the source does not return that partial buffer until the next fixed-duration pop times out; could the pop deadline be capped at the remaining flush time once `written > 0`?

nes-single-node-worker/src/TaskStatisticListener.cpp
```diff
@@ -0,0 +1,198 @@
...
+    if (queued.has_value())
+    {
+        warnOnOverflow(!events.writeIfNotFull(std::move(queued.value())));
+    }
+}
+
+void TaskStatisticListener::onEvent(SystemEvent event)
+{
+    const auto* submit = std::get_if<SubmitQuerySystemEvent>(&event);
+    if (submit == nullptr || toUpperCase(submit->query).find(IN_PROCESS_SOURCE_MARKER) == std::string::npos)
```
ls-1801: This exclusion does not work for anonymous `InProcess` sources: they render as `ANONYMOUS_SOURCE(... type: INPROCESS ...)`, so the `SOURCETYPE: INPROCESS` search misses them. The statistics query can then publish its own task events back into the feed. Could this use source descriptors instead of matching one debug-plan spelling?

scripts/run-engine-stats-demo.sh
```diff
@@ -0,0 +1,195 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# End-to-end demo of the query engine task statistics stream.
+#
+# Starts an embedded worker with --enable_task_statistics, runs a Generator query to give the engine
+# something to do, and runs a second query that reads the engine's own task events through an
+# InProcess source. Everything happens in one process, so no ports or containers to coordinate.
+#
+# This exists instead of a systest: the statistics source never reaches end-of-stream (an idle engine
```
ls-1801: Could this end-to-end scenario be integrated into the existing Bats suite instead of living as a standalone demo script? The embedded REPL harness already supplies `NES_REPL_EMBEDDED` and an isolated `TMP_DIR`, while the invariant checks here do not require a fixed event count. Keeping it in the normal CI suite would ensure this path runs continuously.

nes-plugins/Sources/InProcessSource/InProcessSource.cpp
```diff
@@ -0,0 +1,132 @@
...
+InProcessSource::InProcessSource(const SourceDescriptor& sourceDescriptor)
+    : feedName(sourceDescriptor.getFromConfig(ConfigParametersInProcess::FEED_NAME))
+    , capacity(sourceDescriptor.getFromConfig(ConfigParametersInProcess::CAPACITY))
+    , flushInterval(std::chrono::milliseconds{sourceDescriptor.getFromConfig(ConfigParametersInProcess::FLUSH_INTERVAL_MS)})
+{
+}
+
+void InProcessSource::open(std::shared_ptr<AbstractBufferProvider>)
+{
+    feed = InProcessFeedRegistry::instance().getOrCreate(feedName, capacity);
```
ls-1801: Could registration be producer-driven instead? The producer can choose capacity from its expected event volume, while `InProcessSource` should only acquire an existing feed by name; this would also remove `CAPACITY` from the source configuration.

nes-single-node-worker/include/TaskStatisticListener.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+    /// Only the events that end up as rows travel through the queue, plus the exclusion markers. Routing
+    /// the markers through the same queue keeps them ordered against the events they suppress, and leaves
+    /// the exclusion set owned exclusively by the formatting thread.
+    using QueuedEvent = std::variant<TaskExecutionStart, TaskEmit, TaskExecutionComplete, TaskExpired, ExcludeQuery>;
+
+    void threadRoutine(const std::stop_token& token);
+
+    std::string feedName;
+    std::shared_ptr<InProcessFeed> feed;
```
ls-1801: Could we avoid routing these events through formatted strings? The feed carries only a handful of event types, so a binary representation could let `InProcessSource` skip both formatting and `InputFormatter` parsing, and could remove the dedicated formatting thread. We should also consider producing events only while a consumer is attached, so no feed work happens when no query is listening.

nes-single-node-worker/include/TaskStatisticListener.hpp
```diff
@@ -0,0 +1,80 @@
...
+    void onEvent(SystemEvent event) override;
+
+    /// Starts the thread that formats queued events. Must be called after construction.
+    void start();
+
+private:
+    static constexpr size_t QUEUE_LENGTH = 4096;
+
+    /// Tells the formatting thread to ignore everything belonging to a query, see 'onEvent(SystemEvent)'.
+    struct ExcludeQuery
```
ls-1801: Could we avoid excluding the entire statistics query? If feed production can disable selected event types, we could prevent the self-sustaining path without hiding every task caused by internal statistics queries, keeping the reported task count accurate. The exclusion policy could then be expressed per event type rather than per `QueryId`.

nes-sources/src/InProcessFeed.cpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+        return true;
+    }
+    dropped.fetch_add(1, std::memory_order_relaxed);
```
ls-1801: The current full-queue behavior keeps the oldest rows and discards each newly arriving row. Could we instead retain the most recent `capacity()` rows by removing one row and retrying the write until the new row is accepted?

nes-sources/src/InProcessFeed.cpp
```diff
@@ -0,0 +1,87 @@
...
+    static InProcessFeedRegistry registry;
+    return registry;
+}
+
+std::shared_ptr<InProcessFeed> InProcessFeedRegistry::getOrCreate(const std::string& name, const size_t capacity)
+{
+    const std::lock_guard lock(mutex);
+    if (const auto existing = feeds.find(name); existing != feeds.end())
+    {
+        return existing->second;
```
ls-1801: Could the registry hand out an exclusive consumer handle and release it when `InProcessSource::close()` runs? Returning the same queue to multiple sources means concurrent queries divide the rows and observe different subsets instead of each seeing the feed.

nes-plugins/Sources/EngineEventsSource/EngineEventsSource.cpp
```diff
@@ -0,0 +1,132 @@
...
+    feed.reset();
+}
+
+Source::FillTupleBufferResult InProcessSource::fillTupleBuffer(TupleBuffer& tupleBuffer, const std::stop_token& stopToken)
+{
+    const auto available = tupleBuffer.getAvailableMemoryArea<char>();
+    const auto deadline = std::chrono::steady_clock::now() + flushInterval;
+    size_t written = 0;
+
+    while (not stopToken.stop_requested())
```
ls-1801, line 83: Could stopping wake the reader with a poison marker instead of polling `stop_requested()` around timed pops? A `std::stop_callback` could enqueue `std::variant<PoisonMarker, std::string>`, leaving the pop deadline responsible only for the flush interval.

nes-single-node-worker/src/TaskStatisticListener.cpp
```diff
@@ -0,0 +1,198 @@
...
+        }
+
+        std::visit(
+            Overloaded{
+                [&](const ExcludeQuery& exclude) { excludedQueries.emplace(exclude.queryId); },
+                [&](const TaskExecutionStart& taskStart)
+                {
+                    if (!excludedQueries.contains(taskStart.queryId))
+                    {
+                        feed->tryPush(formatRow(
```
ls-1801, line 161: Could this feed carry a typed binary event representation instead of formatted strings? Only a small set of events is published, so `InProcessSource` could avoid `InputFormatter` parsing and the dedicated formatting thread could go away.

nes-single-node-worker/src/TaskStatisticListener.cpp
```diff
@@ -0,0 +1,198 @@
...
+            [](const TaskExecutionStart& taskStart) { return std::optional<QueuedEvent>{taskStart}; },
+            [](const TaskEmit& taskEmit) { return std::optional<QueuedEvent>{taskEmit}; },
+            [](const TaskExecutionComplete& taskComplete) { return std::optional<QueuedEvent>{taskComplete}; },
+            [](const TaskExpired& taskExpired) { return std::optional<QueuedEvent>{taskExpired}; },
+            [](const auto&) { return std::optional<QueuedEvent>{}; }},
+        std::move(event));
+
+    if (queued.has_value())
+    {
+        warnOnOverflow(!events.writeIfNotFull(std::move(queued.value())));
```
ls-1801, line 122: Could `onEvent` avoid enqueuing events while the feed has no attached consumer? Otherwise task events are queued, formatted, and written even when no query is listening.

nes-single-node-worker/src/TaskStatisticListener.cpp
```diff
@@ -0,0 +1,198 @@
...
+    const auto* submit = std::get_if<SubmitQuerySystemEvent>(&event);
+    if (submit == nullptr || toUpperCase(submit->query).find(IN_PROCESS_SOURCE_MARKER) == std::string::npos)
+    {
+        return;
+    }
+
+    NES_INFO("Excluding query {} from task statistics, it reads an InProcess source", submit->queryId);
+    /// Submission is rare and happens before the query produces any event, so blocking here costs nothing
+    /// and guarantees the marker is not dropped when the queue happens to be full.
+    events.blockingWrite(QueuedEvent{ExcludeQuery{submit->queryId}});
```
ls-1801, line 137: Could we avoid excluding the entire statistics query here? If feed production can disable selected event types, we could break the self-observation cycle without hiding tasks caused by internal statistics queries, keeping the reported task count accurate.

ls-1801 (changes requested):
There is also a follow up that needs to happen:

The registry is process-global, so multiple workers in one process would share feeds; worker-local ownership may be needed.

nes-sources/src/InProcessFeed.cpp
nikla44: You are correct that current implementation is wrong, I am just wondering if an exclusive handle is the right approach. What do you think about fan-out as this is in line with how sources behave generally?

nes-sources/src/InProcessFeed.cpp
nikla44: This was inherited from `GoogleEventTracePrinter.cpp:357` but I can change it if you prefer it

scripts/run-engine-stats-demo.sh
nikla44: Absolutely. Somehow, I just missed this before creating the PR

nes-sources/src/InProcessFeed.cpp
ls-1801: There is a pipe source/sink pr from me that can multiplex a stream to multiple sources. I would prefer if we keep them separate, if you need to process the same events in multiple streams, than I think you should use the pipe source/sink abstraction (if it gets ever merged) or rely on the multi sink queries that Felix is working on.

The problem is that the fan out gets complicated, when some sources apply backpressure and thus not read from the queue.
Technically we could just drop the events in that case. I would prefer to have a single responsibility tho and rely on composition.

I would keep it simple. If you can come up with a simple famout solution that works reliable I am fine, otherwise I would stick to the single consumer approach and fail sources that try to attach to a feed that is already used.

nes-sources/src/InProcessFeed.cpp
nikla44: I think just switching to evict-oldest strategy is not trivial: the producer would have to pop to evict,
  and that pop races the consumer's pop, so it can take a row the consumer was about to receive. In my opinion, this would force me to to implement a custom overwriting SPSC ring rather than using `folly::MPMCQueue`. Just want to make sure this is what you had in mind

nes-single-node-worker/src/TaskStatisticListener.cpp
nikla44: I think hiding only selected events is still problematic, as it may suppress `TaskEmit` but `TaskExecutionStart`/`TaskExecutionComplete` still become rows. Thus, I would suggest to drop exclusion entirely but this means that the observer's own cost always appears in the data.

greenBene (changes requested):
I really appreciate this change :) This allows us to get easier insights into a running system! I also appreciate the on-point inline documentation and the really clean code. 

The one thing I am missing is a test for a distributed setup, similar to the ` embedded.bats`  one. Could we add one 1) to validate the source works in distributed settings and 2) to have a working example of how one would set up such a query?

Artraxon (commented):
I didn't look at the code in detail so I don't want to give an approve-approve, but my concern was addressed. If you require a full code-review text me

nes-single-node-worker/src/TaskStatisticListener.cpp
ls-1801: I think that's fine; the stats should honestly report what's going on in the engine, if later analysis is not interested in them, then they can be filtered out. If we apply some batching, we could also avoid the cycle of every stat task producing more stat tasks.

PR #1961, opened 2026-08-25
fix(SingleNodeWorker): terminate correctly during startup, even when exceptions or invariants are triggered

Artraxon:
Even after https://github.com/nebulastream/nebulastream/pull/1889 , terminating from the outside doesn't work correctly and fails often in the CI when it comes before the signal handler was registered, for example because an exception was thrown.

This pr moves the registration of the system handler towards the beginning of the startup process, and instead of passing a reference to the grpc server, we later store it in an atomic when it was started up.

This implementation also handles triggered preconditions or direct returns correctly in contrast to https://github.com/nebulastream/nebulastream/pull/1940 , which only fixes the issue for exceptions.

EDIT: I also added a validation check for the buffer amount, since you cannot initialize the MPMC queue with a size of one, the precondition never gets reached.
Also, I split up the logging of exceptions and invariants into two, so that first the invariant/exception message is logged, and then the trace is gathered, because if an exception is thrown/invariant triggered during startup, collecting it takes 13 seconds or so which causes the tests to fail.

### DIFF
--- a/nes-common/include/ErrorHandling.hpp
+++ b/nes-common/include/ErrorHandling.hpp
@@ -130,13 +130,15 @@ inline std::string formatStacktrace(const cpptrace::stacktrace& stacktrace, cons
     /// This documents (and checks) requirements for calling a function. If violated, the function was called incorrectly.
     /// @param condition must be true to correctly call function guarded by precondition
     /// @param formatString can contain `{}` to reference varargs. Must not contain positional reference like `{0}`.
+    /// Log the marker before generating the (potentially slow) symbolized stacktrace so it reaches
+    /// the log even if the process is killed mid-symbolization (e.g. a `timeout` SIGKILL at startup).
     #define PRECONDITION(condition, formatString, ...) \
         do \
         { \
             if (!(condition)) \
             { \
-                auto trace = NES::formatStacktrace(cpptrace::generate_trace(), true); \
-                NES_ERROR("Precondition violated: ({}): " formatString "\u001B[0m\n\n{}", #condition __VA_OPT__(, ) __VA_ARGS__, trace); \
+                NES_ERROR("Precondition violated: ({}): " formatString, #condition __VA_OPT__(, ) __VA_ARGS__); \
+                NES_ERROR("\u001B[0m\n\n{}", NES::formatStacktrace(cpptrace::generate_trace(), true)); \
                 if (auto logger = ::NES::Logger::getInstance()) \
                 { \
                     if (!logger->isConsoleLoggingEnabled()) \
@@ -152,13 +154,15 @@ inline std::string formatStacktrace(const cpptrace::stacktrace& stacktrace, cons
     /// This documents what is assumed to be true at this particular point in a program. If violated, there is a misunderstanding and maybe a bug.
     /// @param condition is assumed to be true
     /// @param formatString can contain `{}` to reference varargs. Must not contain positional referencen like `{0}`.
+    /// Log the marker before generating the (potentially slow) symbolized stacktrace so it reaches
+    /// the log even if the process is killed mid-symbolization (e.g. a `timeout` SIGKILL at startup)
     #define INVARIANT(condition, formatString, ...) \
         do \
         { \
             if (!(condition)) \
             { \
-                auto trace = NES::formatStacktrace(cpptrace::generate_trace(), true); \
-                NES_ERROR("Invariant violated: ({}): " formatString "\u001B[0m\n\n{}", #condition __VA_OPT__(, ) __VA_ARGS__, trace); \
+                NES_ERROR("Invariant violated: ({}): " formatString, #condition __VA_OPT__(, ) __VA_ARGS__); \
+                NES_ERROR("\u001B[0m\n\n{}", NES::formatStacktrace(cpptrace::generate_trace(), true)); \
                 if (auto logger = ::NES::Logger::getInstance()) \
                 { \
                     if (!logger->isConsoleLoggingEnabled()) \
--- a/nes-common/src/ErrorHandling.cpp
+++ b/nes-common/src/ErrorHandling.cpp
@@ -135,26 +135,37 @@ std::string formatLogMessage(const std::exception& e)
 
 void tryLogCurrentException()
 {
+    /// Log the concise error headline (code + message) BEFORE resolving the stacktrace. Symbolizing a trace against a
+    /// large, statically linked binary can take many seconds; emitting the headline first guarantees the error marker
+    /// reaches the log even if the process is terminated (e.g. a `timeout` SIGKILL) part-way through symbolization.
     try
     {
         throw;
     }
     catch (const Exception& e)
     {
-        auto msg = formatLogMessage(e);
-        NES_ERROR("{}", msg)
+        const auto headline = fmt::format("failed to process with error code ({}) : {}", e.code(), e.what());
+        NES_ERROR("{}", headline)
         if (auto logger = Logger::getInstance(); logger && !logger->isConsoleLoggingEnabled())
         {
-            fmt::print(stderr, "{}\n", msg);
+            fmt::print(stderr, "{}\n", headline);
+        }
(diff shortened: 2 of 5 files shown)


alepping (approved):

ls-1801 (changes requested):
This improves the shutdown hook in several ways:

  - Replaces periodic sigtimedwait polling with a blocking sigwait. A std::stop_callback sends a thread-directed signal to wake the shutdown thread during normal teardown.
  - Removes std::_Exit, allowing normal stack unwinding and destructors to run. _Exit skips all cleanup and could leave buffered output, persistent resources, or external state uncleaned.
  - Avoids masking startup deadlocks or stuck initialization. If startup can hang, that should be fixed at the responsible initialization step instead of relying on _Exit to terminate the process.
  - Preserves the termination reason using conventional exit codes: 130 for SIGINT and 143 for SIGTERM.

```diff
Subject: [PATCH] improve termination handler
---
Index: nes-single-node-worker/src/SingleNodeWorkerStarter.cpp
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/nes-single-node-worker/src/SingleNodeWorkerStarter.cpp b/nes-single-node-worker/src/SingleNodeWorkerStarter.cpp
--- a/nes-single-node-worker/src/SingleNodeWorkerStarter.cpp	(revision ffc4f4fdbe5d9b95632ab05b79618c7f127b93da)
+++ b/nes-single-node-worker/src/SingleNodeWorkerStarter.cpp	(date 1787746001820)
@@ -12,27 +12,29 @@
     limitations under the License.
 */
 
-/// The POSIX signal APIs used below are not provided by the C++ <csignal> header.
-/// NOLINTNEXTLINE(modernize-deprecated-headers)
-#include <atomic>
 #include <cstdlib>
-#include <ctime>
 #include <exception>
 #include <iostream>
+#include <optional>
 #include <sstream>
 #include <stop_token>
 #include <string>
 #include <vector>
+#include <pthread.h>
+/// The POSIX signal APIs used below are not provided by the C++ <csignal> header.
+/// NOLINTNEXTLINE(modernize-deprecated-headers)
 #include <signal.h>
 #include <Configurations/Util.hpp>
 #include <Identifiers/Identifiers.hpp>
 #include <Plugins/BuiltinPlugins.hpp>
+#include <Util/Files.hpp>
 #include <Util/Logger/LogLevel.hpp>
 #include <Util/Logger/Logger.hpp>
 #include <Util/Logger/impl/NesLogger.hpp>
 #include <Util/Signal.hpp>
 #include <argparse/argparse.hpp>
 #include <cpptrace/from_current.hpp>
+#include <folly/Synchronized.h>
 #include <grpcpp/security/server_credentials.h>
 #include <grpcpp/server_builder.h>
 #include <ErrorHandling.hpp>
@@ -41,13 +43,57 @@
 #include <SingleNodeWorkerConfiguration.hpp>
 #include <Thread.hpp>
 #include <Version.hpp>
+#include <scope_guard.hpp>
 
 namespace
 {
-/// Published once the gRPC server is running so the termination thread can shut it down gracefully. While it is null
-/// (during startup, before the server exists) a termination signal has no server to unwind and the process exits directly.
+
+class Cleanup
+{
+    static constexpr int SIGNAL_EXIT_CODE_OFFSET = 128;
+
+    struct Internal
+    {
+        std::optional<int> terminationSignal;
+        grpc::Server* server = nullptr;
+    };
+
+    folly::Synchronized<Internal> cleanupMutex;
+
+public:
+    std::optional<int> clearServerAndGetTerminationSignal()
+    {
+        auto cleanup = cleanupMutex.wlock();
+        cleanup->server = nullptr;
+        auto result = cleanup->terminationSignal;
+        cleanup->terminationSignal.reset();
+        return result;
+    }
+
+    std::optional<int> setServerIfNotTerminated(grpc::Server* server)
+    {
+        auto cleanup = cleanupMutex.wlock();
+        if (cleanup->terminationSignal)
+        {
+            return cleanup->terminationSignal;
+        }
+        cleanup->server = server;
+        return std::nullopt;
+    }
+
+    void requestTermination(int signal)
+    {
+        auto cleanup = cleanupMutex.wlock();
+        cleanup->terminationSignal = signal + SIGNAL_EXIT_CODE_OFFSET;
+        if (cleanup->server != nullptr)
+        {
+            cleanup->server->Shutdown();
+        }
+    }
+};
+
 /// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
-std::atomic<grpc::Server*> activeServer{nullptr};
+Cleanup cleanup;
 
 /// Block termination signals before any worker threads are created. All subsequently created threads inherit this mask, allowing
 /// a dedicated thread to synchronously wait for the signals without doing non-async-signal-safe work in a signal handler.
@@ -68,27 +114,31 @@
         "shutdown-hook",
         [terminationSignals](const std::stop_token& stopToken) mutable
         {
-            /// Wait with a bounded timeout so that a clean shutdown without signal can stop this thread via its stop_token
-            /// NOLINTNEXTLINE(misc-include-cleaner) timespec/siginfo_t come from <ctime>/<signal.h>
-            const timespec pollInterval{.tv_sec = 0, .tv_nsec = 100'000'000};
-            while (!stopToken.stop_requested())
+            /// Wake sigwait with a thread-directed signal when the thread is asked to stop during a clean shutdown.
+            const auto waitingThread = pthread_self();
+            const std::stop_callback stopCallback(
+                stopToken,
+                [waitingThread]
+                {
+                    if (const auto error = pthread_kill(waitingThread, SIGINT); error != 0)
+                    {
+                        NES_ERROR("Failed to wake the termination thread: {} ({})", NES::getErrorMessage(error), error);
+                    }
+                });
+
+            int signal{};
+            const auto error = sigwait(&terminationSignals, &signal);
+            if (error != 0)
             {
-                siginfo_t info{};
-                const int signal = sigtimedwait(&terminationSignals, &info, &pollInterval);
-                if (signal < 0)
-                {
-                    continue;
-                }
-                NES_INFO("Received signal {}. Shutting down.", signal);
-                if (auto* server = activeServer.load())
-                {
-                    server->Shutdown();
-                    return;
-                }
-                /// The signal arrived before the gRPC server existed. There is nothing to unwind gracefully, so terminate
-                /// directly. std::_Exit skips destructors and atexit functions.
-                std::_Exit(130);
-            }
+                NES_ERROR("Failed to wait for a termination signal: {} ({})", NES::getErrorMessage(error), error);
+                return;
+            }
+            if (stopToken.stop_requested())
+            {
+                return;
+            }
+            NES_INFO("Received signal {}. Shutting down.", signal);
+            cleanup.requestTermination(signal);
         }};
 }
 }
@@ -107,6 +157,13 @@
         return 1;
     }
     NES::Logger::setupLogging("singleNodeWorker.log", NES::LogLevel::LOG_DEBUG);
+    SCOPE_EXIT
+    {
+        if (const auto logger = NES::Logger::getInstance())
+        {
+            logger->forceFlush();
+        }
+    };
 
     /// Register termination handler right now so that the process can be terminated before the rest of the system has started up.
     const auto terminationHandler = terminationThread(terminationSignals);
@@ -178,14 +235,18 @@
                 return 1;
             }
 
-            /// Publish the running server so the termination thread shuts it down gracefully on SIGINT/SIGTERM.
-            activeServer.store(server.get());
+            if (auto terminationSignal = cleanup.setServerIfNotTerminated(server.get()))
+            {
+                return *terminationSignal;
+            }
             NES_INFO("Server listening on {}", configuration->grpcAddressUri.getValue());
             server->Wait();
-            activeServer.store(nullptr);
+            if (auto terminationSignal = cleanup.clearServerAndGetTerminationSignal())
+            {
+                return *terminationSignal;
+            }
             NES_INFO("GRPC Server was shutdown. Terminating the SingleNodeWorker");
         }
-        NES::Logger::getInstance()->forceFlush();
         return 0;
     }
     CPPTRACE_CATCH(...)

```

ls-1801 (approved):

PR #1963, opened 2026-08-26
refactor(systest): rewrite statements upfront

### LINKED ISSUE #1917: Rewrite statements upfront instead of patching plans mid-pipeline

Use the rewriter on the live path so submitted statements are correct in the first place, and delete the mid-pipeline plan patching from the binder: the sink factory that re-registers sinks with spliced result paths, the anonymous-source rewriting, and the source-data writing at bind time. The test tool then interacts with the system the way a user would, through SQL text.

yschroeder97:
Sixth commit of the systest interpolation (EPIC #1920). 
Makes statements correct before submission instead of patching plans mid-pipeline.
Therefore, removes most of `SystestBinder.cpp`.

The pipeline gets a rewrite stage: parse -> partition by settings ->
rewrite each part -> stage the source data -> bind the resulting CREATE statements into one shared catalog set per invocation. 
What remains of the binder is a plan compiler; it still runs entirely on main's control plane (not the coordinator), so this commit changes how statements are prepared, not who executes them.

Deleted: the sink factory and the whole sink filepath patching, the anonymous-source path/host/parser-config patching, and the inline-file writing. 

Closes #1917.

## Verification

- build, 139/139 unit tests
- smoke: `sources/TCP.test`, `sources/Mixed.test`, `formatter/JSON_OUTPUT`

Adds the partition tests to `testfile-builder-test` and the new
`systest-binder-test` target for the query selection.

## Current limitations

- `SourceDataProvider` in nes-sources is now unused (about 114 lines). It can go in a follow up.
- `getTestFileKey` rejects a test file outside the discovery root even though discovery can find one there.
- The rewrite hoists every CREATE above the queries, so a CREATE written between two queries takes effect earlier than the file order suggests.
- An EXPLAIN inside a `SEQUENTIAL_EXECUTION` chain is answered at bind time and does not wait for the query above it.
- A benchmark run that selects query numbers can abort when a selected query depends on an unselected one.
- A differential block counts its input files for both halves, so throughput accounting counts the shared inputs twice.


yschroeder97:
Closed by a branch rename (`ys/systest-rewrite-upfront` -> `ys/systest/rewrite-upfront`). Continues in #1994.

PR #1971, opened 2026-08-31
fix: Make the error messages for seven failing queries actionable (#1957)

### LINKED ISSUE #1957: [TASK] usabilty improvements for ai agents and first time human users
## Why

Callers who don't know NES internals — first-time users and LLM agents — fix a failed
query by reading the error. Today the error often names an identifier they never typed,
points at a clause that is correct, or reports a symptom several rewrite stages
downstream of the mistake.

The queries below *should* fail. The ask is only that the message be enough to fix them.

We measured this with an LLM agent as a surrogate first-time user: given the schema, no
NES knowledge, allowed to retry. Across 176 logged errors it recovered from ~72% of
failures on NES's own message, at a median of two failed submissions each.

## Setup

Reproduced by hand, one query at a time, no agent involved. Each case is minimal: it
differs from an accepted query by one construct. Errors are quoted in full, C++ stack
trace omitted.

Topology submitted with every query (ODBC credentials from the environment):

```yaml
sinks: []

logical:

  - name: LAB_VALUES
    schema:
      - { name: LABVAL_ID, type: INT32 }
      - { name: MLIFE_FALLNR, type: INT32 }
      - { name: ZEITPUNKT, type: UINT64 }
      - { name: GRUPPENNR, type: INT32 }
      - { name: WERT, type: FLOAT64 }
      - { name: TSLOG, type: UINT64 }
      - { name: LAB_ID, type: INT32 }
      - { name: GRUPPE, type: VARSIZED }
      - { name: BEZ, type: VARSIZED }
      - { name: EINHEIT, type: VARSIZED }
      - { name: WERTEBEREICH, type: VARSIZED }
      - { name: PATIENTDETAILS, type: VARSIZED }

  - name: VITAL_VALUES
    schema:
      - { name: MLIFE_FALLNR, type: INT32 }
      - { name: ZEITPUNKT, type: UINT64 }
      - { name: VALUE, type: FLOAT64 }
      - { name: TYPENAME, type: VARSIZED }
      - { name: PATIENTDETAILS, type: VARSIZED }

physical:
  - logical: LAB_VALUES
    type: ODBC
    parser_config:
      type: Native
    source_config:
      db_host: "${MLIFE_DB_HOST}"
      db_port: "1433"
      database: "mlife_custom"
      username: "${MLIFE_DB_USER}"
      password: "${MLIFE_DB_PASSWORD}"
      driver: "ODBC Driver 18 for SQL Server"
      poll_interval_ms: "60000"
      trust_server_certificate: "true"
      timezone_offset_hours: "2"
      # Incremental poll: the trailing ? is bound to the last-seen timestamp.
      query: |-
        SELECT LabVal_ID, va.mlife_FallNr, DATEADD(minute, 90, Zeitpunkt) AS Zeitpunkt, GruppenNr, TRY_CAST(Wert AS FLOAT) AS Wert, vl.TSLog, Lab_ID, Gruppe, Bez, Einheit, Wertebereich, COALESCE(CONCAT(LEFT(vp.Vorname,1),'.', LEFT(vp.Nachname,1), '. geb.', YEAR(vp.GebDat), ' (Station/Bett ',(Bett COLLATE Latin1_General_CI_AS), ') um ',CONVERT(char(5), vl.Zeitpunkt, 108),' Uhr am ',CONVERT(char(10), vl.Zeitpunkt, 104)),'NA') AS PatientDetails FROM mlife_custom.dbo.v_Labor_null as vl INNER JOIN mlife_custom.dbo.v_Aufenthaltsort as va ON va.mlife_FallNr = vl.mlife_FallNr INNER JOIN mlife_custom.dbo.v_Patienten as vp ON va.mlife_PID = vp.mlife_PID WHERE TRY_CAST(Wert AS FLOAT) IS NOT NULL AND Einheit IS NOT NULL AND va.Station = 'WD1I' AND Bett IS NOT NULL AND va.BIS IS NULL AND DATEADD(minute, 90, Zeitpunkt) > ?

  - logical: VITAL_VALUES
    type: ODBC
    parser_config:
      type: Native
    source_config:
      db_host: "${MLIFE_DB_HOST}"
      db_port: "1433"
      database: "mlife_custom"
      username: "${MLIFE_DB_USER}"
      password: "${MLIFE_DB_PASSWORD}"
      driver: "ODBC Driver 18 for SQL Server"
      poll_interval_ms: "60000"
      trust_server_certificate: "true"
      timezone_offset_hours: "2"
      # Incremental poll: the trailing ? is bound to the last-seen timestamp.
      query: |-
        SELECT vv.mlife_FallNr, Zeitpunkt,TRY_CAST(Value AS FLOAT) AS Value, TypeName, COALESCE(CONCAT(LEFT(vp.Vorname,1),'.', LEFT(vp.Nachname,1), '. geb.', YEAR(vp.GebDat), ' (Station/Bett ',(Bett COLLATE Latin1_General_CI_AS), ') um ',CONVERT(char(5), vv.Zeitpunkt, 108),' Uhr am ',CONVERT(char(10), Zeitpunkt, 104)),'NA') AS PatientDetails FROM mlife_custom.ikim.v_Vitalwerte_needmi_long as vv INNER JOIN mlife_custom.dbo.v_Aufenthaltsort as va ON va.mlife_FallNr = vv.mlife_FallNr INNER JOIN mlife_custom.dbo.v_Patienten as vp ON va.mlife_PID = vp.mlife_PID WHERE va.Station = 'WD1I' AND Bett IS NOT NULL AND va.BIS IS NULL AND Zeitpunkt > ?
```

```sh
nes-cli -s <worker>:8080 -t topology.yaml start "<query>"
# exit 1; stdout and stderr are EMPTY; the message lands in ./nes-cli.log
```

All queries end in the sink

```
INTO HTTP('JSON' as `SINK`.INPUT_FORMAT, 'nebulastream-agent' AS `SINK`.`IP_ADDRESS`,
     'nesNotification' AS `SINK`.ENDPOINT, '3000' AS `SINK`.PORT,
     '/data/probe.log' AS `SINK`.LOG_FILE_PATH);
```

which is omitted below (case 7 uses a `File` sink, shown in full).

--

## 1. `HAVING` on the aggregate expression — 2003, names a field that does not exist

```sql
SELECT TYPENAME, COUNT(VALUE) AS N FROM VITAL_VALUES
WINDOW TUMBLING(ZEITPUNKT, SIZE 60 SEC) GROUP BY TYPENAME
HAVING COUNT(VALUE) > INT32(0)
```
```
failed to process with error code (2003) : cannot infer schema;
field VALUE_COUNT is not part of the schema Schema(fields(
  Field(name: VITAL_VALUES$TYPENAME, DataType: DataType(type: VARSIZED)),
  Field(name: VITAL_VALUES$N, DataType: DataType(type: UINT64))), size in bytes: 24)
```

`VALUE_COUNT` appears nowhere in the query — it is NES's internal rewrite name. The rule
being enforced is "`HAVING` must use the projected alias": `HAVING N > INT32(0)` is
accepted. **Should say so.** This message shape is ~110 of our 176 errors.

## 2. Windowing drops a column — 2003, reported as if it never existed

```sql
SELECT TYPENAME, COUNT(VALUE) AS N FROM VITAL_VALUES
WHERE TYPENAME = VARSIZED('SPO2') WINDOW TUMBLING(ZEITPUNKT, SIZE 60 SEC)
```
```
failed to process with error code (2003) : cannot infer schema;
field TYPENAME is not part of the schema Schema(fields(
  Field(name: VITAL_VALUES$START, DataType: DataType(type: UINT64)),
  Field(name: VITAL_VALUES$END, DataType: DataType(type: UINT64)),
  Field(name: VITAL_VALUES$N, DataType: DataType(type: UINT64))), size in bytes: 24)
```

`TYPENAME` is a real source column, filtered on in the same query. It was removed by
windowing. Adding `GROUP BY TYPENAME` fixes it. **Should say which stage dropped it.**

## 3. Aggregate over text — 2002, right message under the wrong name

```sql
SELECT MAX(GRUPPE) AS G FROM LAB_VALUES WINDOW TUMBLING(ZEITPUNKT, SIZE 60 SEC)
```
```
failed to process with error code (2002) : cannot deserialize;
aggregations on non numeric fields is not supported, but got DataType(type: VARSIZED)
```

The body is correct and actionable. Nothing is being deserialized. **Rename the code.**
Lowest recovery rate in our study (33%) purely because the name misdirects.

## 4. `LIKE` / `IN` — 2000, names an internal builder variable

```sql
SELECT BEZ FROM LAB_VALUES WHERE BEZ LIKE VARSIZED('KAL%')
SELECT BEZ FROM LAB_VALUES WHERE BEZ IN (VARSIZED('KALIUM'), VARSIZED('NATRIUM'))
```
```
failed to process with error code (2000) : invalid query syntax;
There were more than 1 functions in the functionBuilder in exitWhereClause.
```

Not about predicate count: `WHERE WERT > FLOAT64(1) AND WERT < FLOAT64(9)` is accepted.
`LIKE` and `IN` are simply unsupported. **Should say "LIKE is not supported".**

## 5. Scalar function — 2000, reported as an unknown *aggregation*, with a token id

```sql
SELECT UPPER(BEZ) AS B FROM LAB_VALUES
```
```
failed to process with error code (2000) : invalid query syntax;
Unknown (aggregation) function: UPPER, resolved to token type: 154
```

`UPPER` is scalar, not an aggregation. "Token type 154" is meaningless outside the
grammar. `STDDEV` and `PERCENTILE` produce the same text. **Should name the supported set.**

## 6. `COUNT(*)` — 2000, correct but withholds the fix

```sql
SELECT COUNT(*) AS N FROM VITAL_VALUES WINDOW TUMBLING(ZEITPUNKT, SIZE 60 SEC)
```
```
failed to process with error code (2000) : invalid query syntax;
Aggregation requires argument at COUNT(*)
```

`COUNT(VALUE)` works. The message never says that. 15 occurrences in our run.

## 7. Sink parameter — 1000, echoes a name the caller did not write

```sql
SELECT WERT FROM LAB_VALUES
INTO File('/data/probe.csv' AS `SINK`.FILEPATH, 'CSV' AS `SINK`.INPUT_FORMAT);
```
```
failed to process with error code (1000) : invalid config parameter;
Unknown configuration parameter: filepath.
```

The caller wrote `FILEPATH`; the error says `filepath`, so it cannot be matched against
the query text, and no list of accepted `File` sink parameters is offered.

---

## Reference

For the same class of error — a column that cannot be resolved — PostgreSQL answers:

```
ERROR:  column "value" does not exist
LINE 1: select id, sum(VALUE) from runs;
                       ^
```

It names the identifier the user typed, points at the position, and fits in one
sentence. When the problem is semantic rather than a typo it states the rule instead:
`column "runs.id" must appear in the GROUP BY clause or be used in an aggregate function`
— the shape asked for in cases 1 and 4.

NES already does this in places; 2030 and 2006 need no change at all:

```
error code (2030) : unknown source name; Logical source not registered. Source Name: MEDIKATION
error code (2006) : different field type expected; TimeBasedWindow should use a uint for time field VALUE
```

The ask is to bring 2000, 2002 and 2003 up to that standard.

### Acceptance Criteria

- [ ] **1.** `HAVING COUNT(VALUE) > INT32(0)` after a window reports that `HAVING` must
      use the projected alias, and names it — not `field VALUE_COUNT is not part of the
      schema`. No error names an internally generated identifier; where one must appear,
      the user expression it came from is named alongside it.
- [ ] **2.** When windowing removes a column the query references, the message says the
      column was dropped by aggregation and names the stage — not that it does not exist.
- [ ] **3.** Aggregate-over-non-numeric no longer reports under `cannot deserialize`:
      the code is renamed, or the error moves to a code whose name matches the cause.
- [ ] **4.** `LIKE` and `IN` are refused by name ("LIKE is not supported"). No internal
      builder or parser-callback identifier appears in any message.
- [ ] **5.** Scalar functions are not reported as unknown *aggregations*; no grammar
      token id appears in the message; the supported function set is named or linked.
- [ ] **6.** `COUNT(*)` / `COUNT(1)` state the supported form `COUNT(<column>)`.
- [ ] **7.** Sink parameter errors echo the parameter as the caller wrote it, and list
      the accepted parameters for that sink type.
- [ ] Each of the seven queries above has been re-run against the same two-source
      topology and its new message recorded in this issue.
- [ ] The controls still succeed: `HAVING N > INT32(0)`, `WHERE WERT > FLOAT64(1) AND
      WERT < FLOAT64(9)`, `COUNT(VALUE)`, `COUNT(EINHEIT) ... GROUP BY EINHEIT`.

keyseven123:
The seven queries in #1957 fail for good reasons, but their errors name internal identifiers, the wrong category, or a clause that is correct. One commit per query, each pairing a test case with the change that produces the message. The change list is as follows:
- Changed the SQL parser to refuse an aggregate repeated in `HAVING` by naming the SELECT alias, instead of failing on the generated name `VALUE_COUNT`.
- Changed projection schema inference to name the stage that dropped a column, instead of reporting the column as non-existent.
- Changed all five aggregations to reject non-numeric input as `UnsupportedQuery`, naming the field and its type.
- Changed the parser to name an unsupported predicate operator (`LIKE`) and to report an unknown function without aggregation wording or a grammar token id.
- Changed descriptor validation to list the accepted parameters for the sink or source type.
- Added `COUNT(1)` coverage; `COUNT(*)` and `COUNT(1)` are already accepted, so that commit is tests only.

## Verifying this change
This change is tested by
- Added a test case per query, each asserting the error code and the load-bearing phrase of its message.
- Added the controls from the issue's acceptance criteria (`HAVING` on the alias, `IN`, `COUNT`, `GROUP BY`) where not already covered.
- Full local `ctest`, plus the six touched `.test` files run directly.

- SQLParser
- LogicalOperators
- Configurations
- Systest

## Documentation

This PR closes #1957.

Three of its reports are stale against `main` and needed no message change: `IN` and `COUNT(*)`/`COUNT(1)` are already supported, and the non-numeric aggregation had already moved off `cannot deserialize` to `cannot infer stamp` — still a code whose name does not match the cause, hence the move to `UnsupportedQuery`.

https://claude.ai/code/session_01Eg3WxK1EqcW8VFzNnbV7dQ

### DIFF
--- a/nes-logical-operators/src/Operators/Windows/Aggregations/SumAggregationLogicalFunction.cpp
+++ b/nes-logical-operators/src/Operators/Windows/Aggregations/SumAggregationLogicalFunction.cpp
@@ -26,23 +26,40 @@
 #include <DataTypes/DataTypeProvider.hpp>
 #include <Functions/FieldAccessLogicalFunction.hpp>
 #include <Functions/LogicalFunction.hpp>
+#include <Functions/UnboundFieldAccessLogicalFunction.hpp>
+#include <Identifiers/Identifier.hpp>
 #include <Operators/Windows/Aggregations/WindowAggregationLogicalFunction.hpp>
 #include <Schema/Field.hpp>
 #include <Schema/Schema.hpp>
 #include <Schema/SchemaFwd.hpp>
 #include <Serialization/LogicalFunctionReflection.hpp>
+#include <Util/Overloaded.hpp>
 #include <Util/PlanRenderer.hpp>
 #include <Util/Reflection.hpp>
 #include <fmt/format.h>
 #include <folly/hash/Hash.h>
+#include <magic_enum/magic_enum.hpp>
 #include <AggregationLogicalFunctionRegistry.hpp>
 #include <ErrorHandling.hpp>
 
 namespace NES
 {
+namespace
+{
+Identifier fieldNameOf(const AggregationFieldAccess& inputFunction)
+{
+    return std::visit(
+        Overloaded{
+            [](const TypedLogicalFunction<UnboundFieldAccessLogicalFunction>& unbound) { return unbound->getFieldName(); },
+            [](const TypedLogicalFunction<FieldAccessLogicalFunction>& bound) { return bound->getField().getLastName(); }},
+        inputFunction);
+}
+}
+
 SumAggregationLogicalFunction::SumAggregationLogicalFunction(AggregationFieldAccess inputFunction)
     : inputFunction(inputFunction)
-    , aggregateType(inferFromInput(std::visit([](const auto& input) { return input->getDataType(); }, inputFunction)))
+    , aggregateType(
+          inferFromInput(std::visit([](const auto& input) { return input->getDataType(); }, inputFunction), fieldNameOf(inputFunction)))
 {
 }
 
@@ -66,7 +83,7 @@ std::string_view SumAggregationLogicalFunction::getName() const noexcept
 /// PostgreSQL: integers are promoted to the widest available integer type (sum(int2/int4) -> int8), while floats keep their type
 /// (sum(real) -> real) since they saturate to infinity instead of wrapping around. Sums exceeding the 64-bit integer range still
 /// wrap around silently.
-DataType SumAggregationLogicalFunction::inferFromInput(const DataType inputType)
+DataType SumAggregationLogicalFunction::inferFromInput(const DataType inputType, const Identifier& inputFieldName)
 {
     if (inputType.type == DataType::Type::UNDEFINED)
     {
@@ -87,13 +104,14 @@ DataType SumAggregationLogicalFunction::inferFromInput(const DataType inputType)
     {
         return inputType;
     }
-    throw CannotInferStamp("aggregations on non numeric fields is not supported (got {})", inputType);
+    throw UnsupportedQuery(
+        "{} is only supported on numeric fields, but field {} has type {}", NAME, inputFieldName, magic_enum::enum_name(inputType.type));
 }
 
 SumAggregationLogicalFunction SumAggregationLogicalFunction::withInferredType(const Schema<Field, Unordered>& schema) const
 {
     const auto newInputFunction = inferFieldAccess(inputFunction, schema);
-    const DataType outputType = inferFromInput(newInputFunction->getDataType());
+    const DataType outputType = inferFromInput(newInputFunction->getDataType(), newInputFunction->getField().getLastName());
     return SumAggregationLogicalFunction{newInputFunction, outputType};
 }
 
--- a/nes-sql-parser/private/AntlrSQLParser/AntlrSQLHelper.hpp
+++ b/nes-sql-parser/private/AntlrSQLParser/AntlrSQLHelper.hpp
@@ -39,7 +39,10 @@ namespace NES::Parsers
 
 class AntlrSQLHelper
 {
+public:
     using Projection = std::pair<std::optional<Identifier>, LogicalFunction>;
+
+private:
     std::vector<LogicalFunction> whereClauses; ///where and having clauses need to be accessed in reverse
     std::vector<LogicalFunction> havingClauses;
     std::optional<Identifier> source;
@@ -76,6 +79,7 @@ public:
 
     std::optional<Windowing::TimeBasedWindowType> windowType;
     std::vector<std::pair<WindowAggregationLogicalFunction, std::optional<Identifier>>> windowAggs;
+    std::vector<Identifier> userAggregationAliases;
     std::vector<SinkDescriptor> sinkDescriptor;
     std::vector<std::string> constantBuilder;
     std::vector<LogicalFunction> functionBuilder;
@@ -101,6 +105,10 @@ public:
     /// E.g., AVG(i + UINT64(1)) becomes: Projection(*, i + UINT64(1) AS _AGG_INPUT_0) → AVG(_AGG_INPUT_0).
     std::vector<Projection> preAggregationProjections;
     size_t aggExprCounter = 0;
+
+    /// Number of aggregations collected before the HAVING clause was entered, so that aggregations written
+    /// inside HAVING can be distinguished from the ones the SELECT list already projects.
+    size_t aggCountBeforeHaving = 0;
 
     /// Flag set while parsing a MODEL_INFERENCE TVF source to suppress identifier capture as FROM source
     bool isModelInference = false;
--- a/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
+++ b/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
@@ -1405,7 +1405,13 @@ void AntlrSQLQueryPlanCreator::exitFunctionCall(AntlrSQLParser::FunctionCallCont
                 }
                 else
                 {
-                    throw InvalidQuerySyntax("Unknown (aggregation) function: {}, resolved to token type: {}", funcName, tokenType);
+                    auto registeredNamesAggregation
+                        = AggregationLogicalFunctionRegistry::instance().getRegisteredNames() | std::ranges::to<std::vector>();
+                    std::ranges::sort(registeredNamesAggregation);
+                    throw InvalidQuerySyntax(
+                        "Unknown function: {}. Supported aggregation functions are {}.",
+                        funcName,
+                        fmt::join(registeredNamesAggregation, ", "));
                 }
             }
     }
--- a/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
+++ b/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
@@ -56,6 +56,7 @@
 #include <Functions/LogicalFunctionProvider.hpp>
 #include <Functions/UnboundFieldAccessLogicalFunction.hpp>
 #include <Identifiers/Identifier.hpp>
+#include <Iterators/BFSIterator.hpp>
 #include <Operators/ProjectionLogicalOperator.hpp>
 #include <Operators/Windows/Aggregations/AvgAggregationLogicalFunction.hpp>
 #include <Operators/Windows/Aggregations/CountAggregationLogicalFunction.hpp>
@@ -287,6 +288,31 @@ void validateSourceQualifiers(const AntlrSQLHelper& helper)
         if (qualifier != expectedQualifier.value())
         {
             throw InvalidQuerySyntax("Unknown source qualifier {}, expected {}", qualifier, expectedQualifier.value());
+        }
+    }
+}
+
+void validateAggregatedSelectList(AntlrSQLHelper& helper)
+{
+    const auto isAvailableAfterAggregation = [&helper](const Identifier& field)
+    {
+        return field == Identifier::parse("start") or field == Identifier::parse("end")
+            or std::ranges::any_of(helper.groupByFields, [&field](const auto& key) { return key.getFieldName() == field; })
+            or std::ranges::any_of(
+                   helper.windowAggs, [&field](const auto& agg) { return agg.second.has_value() and agg.second.value() == field; });
+    };
+
+    for (const auto& projection : helper.getProjections() | std::views::values)
+    {
+        for (const auto& function : BFSRange{projection})
+        {
+            if (const auto access = function.tryGetAs<UnboundFieldAccessLogicalFunction>();
+                access.has_value() and not isAvailableAfterAggregation(access.value()->getFieldName()))
+            {
+                throw InvalidQuerySyntax(
+                    R"(column "{}" must appear in the GROUP BY clause or be used in an aggregate function)",
+                    access.value()->getFieldName().getOriginalString());
+            }
         }
     }
 }
@@ -823,6 +849,7 @@ void AntlrSQLQueryPlanCreator::exitPrimaryQuery(AntlrSQLParser::PrimaryQueryCont
                 currentWindowTimestampOpt.has_value() ? "two" : "none");
         }
         auto characteristic = std::get<Windowing::UnboundTimeCharacteristic>(currentWindowTimestampOpt.value());
+        validateAggregatedSelectList(helpers.top());
         auto aggregations = helpers.top().windowAggs
             | std::views::transform(
                                 [](auto& agg)
--- a/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
+++ b/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
@@ -507,7 +507,13 @@ void AntlrSQLQueryPlanCreator::exitBoolComparison(AntlrSQLParser::BoolComparison
     }
     else
     {
-        throw UnsupportedQuery("Predicate is currently not supported: {}", comparison->getText());
+        /// `kind` stores the operator token (LIKE, RLIKE, DISTINCT, ...) to include in the error output.
+        INVARIANT(comparison->kind != nullptr, "Predicates without a kind token are handled above, but got {}", comparison->getText());
+        const auto keyword = toUpperCase(comparison->kind->getText());
+        const auto* const isPrefix = comparison->IS() != nullptr ? "IS " : "";
+        const auto* const negation = comparison->NOT() != nullptr ? "NOT " : "";
+        const auto* const distinctSuffix = comparison->DISTINCT() != nullptr ? " FROM" : "";
+        throw UnsupportedQuery("{}{}{}{} is not supported: {}", isPrefix, negation, keyword, distinctSuffix, comparison->getText());
     }
 
     functions.push_back(std::move(function));
--- a/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
+++ b/nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
@@ -14,6 +14,7 @@
 
 #include <AntlrSQLParser/AntlrSQLQueryPlanCreator.hpp>
 
+#include <algorithm>
 #include <array>
 #include <cctype>
 #include <cstddef>
@@ -69,6 +70,7 @@
 #include <Plans/LogicalPlanBuilder.hpp>
 #include <Util/Overloaded.hpp>
 #include <Util/PlanRenderer.hpp>
+#include <Util/Strings.hpp>
 #include <WindowTypes/Measures/TimeCharacteristic.hpp>
 #include <WindowTypes/Measures/TimeMeasure.hpp>
 #include <WindowTypes/Types/SlidingWindow.hpp>
@@ -81,6 +83,32 @@
 
 namespace NES::Parsers
 {
+
+namespace
+{
+/// Renders an aggregation back as SQL (COUNT(VALUE)) so error messages quote the caller's own expression.
+/// An argument that was desugared into a pre-aggregation projection is rendered as the expression the caller wrote,
+/// not as the generated _AGG_INPUT_N field.
+std::string
+describeAggregation(const WindowAggregationLogicalFunction& aggregation, const std::vector<AntlrSQLHelper::Projection>& desugaredArguments)
+{
+    const auto input = aggregation.getInputFunction();
+    const auto field = [&]
+    {
+        if (const auto* unbound = std::get_if<TypedLogicalFunction<UnboundFieldAccessLogicalFunction>>(&input))
+        {
+            const auto desugared = std::ranges::find_if(
+                desugaredArguments, [&unbound](const auto& projection) { return projection.first == (*unbound)->getFieldName(); });
+            if (desugared != std::ranges::end(desugaredArguments))
+            {
+                return desugared->second.explain(ExplainVerbosity::Short);
+            }
+        }
+        return std::visit([](const auto& access) { return access->explain(ExplainVerbosity::Short); }, input);
+    }();
+    return fmt::format("{}({})", toUpperCase(aggregation.getName()), field);
+}
+}
 
 LogicalPlan AntlrSQLQueryPlanCreator::getQueryPlan() const
 {
@@ -701,6 +729,7 @@ void AntlrSQLQueryPlanCreator::enterIdentifier(AntlrSQLParser::IdentifierContext
             helpers.top().windowAggs.pop_back();
             aggFunc.second = bindIdentifier(context);
             helpers.top().windowAggs.push_back(aggFunc);
+            helpers.top().userAggregationAliases.push_back(aggFunc.second.value());
             helpers.top().addProjection(aggFunc.second, UnboundFieldAccessLogicalFunction{aggFunc.second.value()});
         }
         else
@@ -994,12 +1023,41 @@ void AntlrSQLQueryPlanCreator::exitCastExpression(AntlrSQLParser::CastExpression
 void AntlrSQLQueryPlanCreator::enterHavingClause(AntlrSQLParser::HavingClauseContext* context)
 {
     helpers.top().isWhereOrHaving = true;
+    helpers.top().aggCountBeforeHaving = helpers.top().windowAggs.size();
     AntlrSQLBaseListener::enterHavingClause(context);
 }
 
 void AntlrSQLQueryPlanCreator::exitHavingClause(AntlrSQLParser::HavingClauseContext* context)
 {
     helpers.top().isWhereOrHaving = false;
+    /// An aggregation written inside HAVING is a second aggregation, not a reference to the projected one.
+    /// Refuse it and report the alias that already exists.
+    const auto& aggregations = helpers.top().windowAggs;
+    if (aggregations.size() > helpers.top().aggCountBeforeHaving)
+    {
+        const auto& havingAggregation = aggregations[helpers.top().aggCountBeforeHaving].first;
+        const auto expression = describeAggregation(havingAggregation, helpers.top().preAggregationProjections);
+        const auto projectedAlias = [&]() -> std::optional<Identifier>
+        {
+            for (const auto& [aggregation, alias] : aggregations | std::views::take(helpers.top().aggCountBeforeHaving))
+            {
+                if (aggregation == havingAggregation and alias.has_value()
+                    and std::ranges::contains(helpers.top().userAggregationAliases, alias.value()))
+                {
+                    return alias;
+                }
+            }
+            return std::nullopt;
+        }();
+        if (projectedAlias.has_value())
+        {
+            throw InvalidQuerySyntax(
+                "HAVING must use the projected alias: write {} instead of {}", projectedAlias.value().getOriginalString(), expression);
+        }
+        throw InvalidQuerySyntax(
+            "HAVING can only use fields the SELECT list projects: add {} AS <alias> to the SELECT list and use <alias> in HAVING",
+            expression);
+    }
     if (helpers.top().functionBuilder.size() != 1)
     {
         throw InvalidQuerySyntax("There was more than one function in the functionBuilder in exitHavingClause.");
--- a/nes-systests/operator/aggregation/WindowAggregationProjection.test
+++ b/nes-systests/operator/aggregation/WindowAggregationProjection.test
@@ -143,3 +143,21 @@ INTO sinkStreamNested;
 1,4
 1,2
 
+# A column that is neither grouped nor aggregated is rejected while the query is still being parsed.
+SELECT id, COUNT(id) AS count_id
+FROM stream
+WHERE id > UINT64(15)
+WINDOW TUMBLING(timestamp, size 1 sec)
+INTO sinkStreamNested;
+----
+ERROR 2000 "column "id" must appear in the GROUP BY clause or be used in an aggregate function"
+
+# Control for the case above: with the column in GROUP BY the same query should be accepted.
+SELECT id, COUNT(id) AS count_id
+FROM stream
+WHERE id > UINT64(15)
+GROUP BY id
+WINDOW TUMBLING(timestamp, size 1 sec)
+INTO sinkStreamNested;
+----
+16,1
--- a/nes-systests/operator/aggregation/WindowAggregationVarSizedMinimal.test
+++ b/nes-systests/operator/aggregation/WindowAggregationVarSizedMinimal.test
@@ -17,4 +17,3 @@ FROM stream GROUP BY (key) WINDOW TUMBLING(ts, size 100 ms) INTO FILE ();
 0,100,apple,1
 0,100,book,1
 0,100,cat,1
-
--- a/nes-systests/parser/VarSizedSupport.test
+++ b/nes-systests/parser/VarSizedSupport.test
@@ -48,3 +48,8 @@ WHERE text1 IN (VARSIZED('test3')) INTO sinkStreamWithText;
 ----
 11,2,test3,test2
 11,3,test3,test1
+
+# An unknown scalar function should be reported and specify the set of supported functions.
+SELECT UPPER(text1) AS t FROM streamWithText INTO sinkStreamWithText;
+----
+ERROR 2000 "Unknown function: UPPER. Supported aggregation functions are AVG, COUNT, MAX, MEDIAN, MIN, SUM."
--- a/nes-systests/parser/VarSizedSupport.test
+++ b/nes-systests/parser/VarSizedSupport.test
@@ -36,3 +36,15 @@ SELECT * FROM streamWithText INTO sinkStreamWithText;
 2,3,test1,test1
 6,4,test1,test2
 8,5,test1,test2
+
+# LIKE should be refused by name.
+SELECT text1 FROM streamWithText WHERE text1 LIKE VARSIZED('test%') INTO sinkStreamWithText;
+----
+ERROR 2008 "LIKE is not supported"
+
+# Control for the case above: IN over the same column should be supported.
+SELECT id, value, text1, text2 FROM streamWithText
+WHERE text1 IN (VARSIZED('test3')) INTO sinkStreamWithText;
+----
+11,2,test3,test2
+11,3,test3,test1
--- a/nes-systests/parser/WindowAggregationParser.test
+++ b/nes-systests/parser/WindowAggregationParser.test
@@ -59,6 +59,50 @@ INTO sinkStream;
 2000,3000,16,2
 1500,2500,16,2
 
+# HAVING that repeats the aggregate expression should be refused.
+# The error message should hint to use an alias from the SELECT list.
+SELECT start, end, id, SUM(value) as sumValue
+FROM stream
+WHERE id > 15
+WINDOW SLIDING(timestamp, size 1 sec, advance by 500 ms)
+GROUP BY id
+HAVING SUM(value) > UINT64(0)
+INTO sinkStream;
+----
+ERROR 2000 "HAVING must use the projected alias: write sumValue instead of SUM(VALUE)"
+
+SELECT start, end, id, SUM(value)
+FROM stream
+WHERE id > 15
+WINDOW SLIDING(timestamp, size 1 sec, advance by 500 ms)
+GROUP BY id
+HAVING SUM(value) > UINT64(0)
+INTO sinkStream;
+----
+ERROR 2000 "add SUM(VALUE) AS <alias> to the SELECT list"
+
+SELECT start, end, id, SUM(value + UINT64(1)) AS sumValue
+FROM stream
+WHERE id > 15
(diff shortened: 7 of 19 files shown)


nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -994,12 +1018,32 @@ void AntlrSQLQueryPlanCreator::exitCastExpression(AntlrSQLParser::CastExpression
...
+    /// the caller never wrote, so refuse it here and point at the alias the SELECT list already gives it.
+    const auto& aggregations = helpers.top().windowAggs;
+    if (aggregations.size() > helpers.top().aggCountBeforeHaving)
+    {
+        const auto& [havingAggregation, _] = aggregations[helpers.top().aggCountBeforeHaving];
+        const auto expression = describeAggregation(havingAggregation);
+        const auto projected = std::ranges::find_if(
+            aggregations | std::views::take(helpers.top().aggCountBeforeHaving),
+            [&](const auto& entry) { return entry.first == havingAggregation and entry.second.has_value(); });
+        if (projected != (aggregations | std::views::take(helpers.top().aggCountBeforeHaving)).end())
```
tim-nick: CI-blocking: `clang-tidy-diff` reports `clang-analyzer-core.CallAndMessage: 1st function call argument is an uninitialized value` here (treated as an error under `-warnings-as-errors`). Looks like a false positive — the analyzer loses track of a reference into a structured binding (`havingAggregation`) once it's captured by the `find_if` lambda across `std::ranges`/`std::invoke`. Verified locally that rebinding it as a plain named reference clears the diagnostic:

```suggestion
        const auto& havingAggregation = aggregations[helpers.top().aggCountBeforeHaving].first;
        const auto expression = describeAggregation(havingAggregation);
        const auto projectedAggregations = aggregations | std::views::take(helpers.top().aggCountBeforeHaving);
        const auto projected = std::ranges::find_if(
            projectedAggregations,
            [&](const auto& entry) { return entry.first == havingAggregation and entry.second.has_value(); });
        if (projected != std::ranges::end(projectedAggregations))
```

Also moves the repeated `aggregations | std::views::take(...)` expression into a named variable instead of building it twice — not required to fix the CI failure, just avoids computing the same thing twice.

tim-nick (commented):
Went through this locally: checked out the branch, built with the dev container, and ran all six touched systest files individually (all 27 queries pass). Great work overall.

One thing is currently blocking CI (inline comment below); everything else I found is either already-non-blocking clang-tidy hygiene or a minor wording nit, not worth holding up the merge for.

nes-systests/sinks/FileSinkConfig.test
yschroeder97, line 1: I am a bit skeptical this needs to be a systest, unit tests are fast and you can test the exact same thing using one IMO.

`new-configurations/tests/UnitTests/Configurations` already holds comparable tests. 
This has so much overhead, you have to write out the setup statements and even attach data just to verify that the error message contains the accepted config parameters?

But I also see the benefit to view this from the "users" angle.

nes-systests/sinks/FileSinkConfig.test
```diff
@@ -0,0 +1,21 @@
...
+CREATE PHYSICAL SOURCE FOR input TYPE File;
+ATTACH INLINE
+1,1,1000
+2,2,2000
+
+# An unknown sink parameter should be echoed as the caller wrote it and the accepted parameters for the
+# sink type should be listed.
+SELECT value FROM input INTO File('/tmp/probe.csv' AS "SINK".FILEPATH);
+----
+ERROR 1000 "Unknown configuration parameter: FILEPATH. Accepted parameters for File are:"
```
yschroeder97: Why does the assertion stop at "are:"?
If the list is empty, then this test will pass.

nes-systests/sinks/FileSinkConfig.test
```diff
@@ -0,0 +1,21 @@
...
+
+# Control for the case above: the accepted spelling FILE_PATH should be a valid parameter.
+SELECT value FROM input INTO File();
```
yschroeder97: Comment says this checks that `FILE_PATH` is a valid parameter, but then doesn't use it at all in the config

nes-systests/operator/aggregation/WindowAggregationProjection.test
```diff
@@ -143,3 +143,22 @@ INTO sinkStreamNested;
 1,4
 1,2
 
+# A column the window aggregation dropped should be reported as dropped, naming the stage that dropped
+# it.
```
yschroeder97:
```suggestion
# A column the window aggregation dropped should be reported as dropped, reporting the stage that dropped it.
```

nes-logical-operators/src/Operators/ProjectionLogicalOperator.cpp
```diff
@@ -200,14 +201,39 @@ void ProjectionLogicalOperator::inferLocalSchema()
...
+              " add the field to GROUP BY to keep it."
+            : "";
+        /// Rethrown as a plain Exception so the original message keeps its single "cannot infer schema; " prefix.
+        auto message = std::string{exception.what()};
+        while (not message.empty() and (message.back() == '\n' or message.back() == ' '))
+        {
+            message.pop_back();
+        }
+        throw Exception{
+            fmt::format("{}; the SELECT list is applied to the output of {}.{}\n", message, child->getName(), hint),
```
yschroeder97: This leads to a plan-internal operator name leaking into a user-facing msg, which is not ideal (`child->getName()` returns e.g., `WindowAggregationOperator`). 

Isn't what we really want here the SQL message: `column "id" must appear in the GROUP BY clause or be used in an aggregate function`? 
This should actually be caught by the binder and not here, since the binder has the SELECT list and the GROUP BY clause available and could check for that before any (wrong) operator could even be instantiated.

Also, `ProjectionPushdownRule` and `DecideFieldOrder` construct new projections that are not user-written, and if they have any internal bug, the users SELECT list would be blamed.

nes-logical-operators/src/Operators/ProjectionLogicalOperator.cpp
```diff
@@ -200,14 +201,39 @@ void ProjectionLogicalOperator::inferLocalSchema()
...
+                          {
+                              auto inferredFunction = projection.second.withInferredDataType(inputSchema);
+                              return UnboundProjection{projection.first, inferredFunction};
+                          })
+            | std::ranges::to<std::vector>();
+    }
+    CPPTRACE_CATCH(const Exception& exception)
+    {
+        /// A SELECT list only sees what its input operator emits, so a column that operator dropped should be
+        /// reported as dropped by that operator, not as a column that never existed.
```
yschroeder97: This comment reads as if it belongs to the if condition that rethrows, I would move above the catch, or to the msg construction.

nes-logical-operators/src/Operators/ProjectionLogicalOperator.cpp
```diff
@@ -200,14 +201,39 @@ void ProjectionLogicalOperator::inferLocalSchema()
...
+            : "";
+        /// Rethrown as a plain Exception so the original message keeps its single "cannot infer schema; " prefix.
+        auto message = std::string{exception.what()};
+        while (not message.empty() and (message.back() == '\n' or message.back() == ' '))
+        {
+            message.pop_back();
+        }
+        throw Exception{
+            fmt::format("{}; the SELECT list is applied to the output of {}.{}\n", message, child->getName(), hint),
+            static_cast<uint64_t>(ErrorCode::CannotInferSchema)};
```
yschroeder97:
```suggestion
        auto& message = exception.what();
        /// Every exception message ends in a newline, so the clause goes in front of it.
        message.resize(message.find_last_not_of(" \n") + 1);
        message += fmt::format("; the SELECT list is applied to the output of {}.{}\n", child->getName(), hint);
        throw;
```

nes-systests/operator/aggregation/WindowAggregationProjection.test
```diff
@@ -143,3 +143,22 @@ INTO sinkStreamNested;
...
+SELECT id, COUNT(id) AS count_id
+FROM stream
+WHERE id > UINT64(15)
+WINDOW TUMBLING(timestamp, size 1 sec)
+INTO sinkStreamNested;
+----
+ERROR 2003 "add the field to GROUP BY"
```
yschroeder97: This only asserts the generic hint - is this intended?

nes-systests/operator/aggregation/WindowAggregationVarSizedMinimal.test
```diff
@@ -18,3 +18,8 @@ FROM stream GROUP BY (key) WINDOW TUMBLING(ts, size 100 ms) INTO FILE ();
 0,100,book,1
 0,100,cat,1
 
+# Aggregating a VARSIZED field should be refused as an unsupported query naming the field and its type.
```
yschroeder97:
```suggestion
# Aggregating a VARSIZED field should be refused as an unsupported query reporting the field and its type.
```

nes-systests/operator/aggregation/WindowAggregationVarSizedMinimal.test
```diff
@@ -18,3 +18,8 @@ FROM stream GROUP BY (key) WINDOW TUMBLING(ts, size 100 ms) INTO FILE ();
...
+SELECT start, end, MAX(key) as key_out
+FROM stream WINDOW TUMBLING(ts, size 100 ms) INTO File();
+----
+ERROR 2008 "MAX is only supported on numeric fields, but field KEY has type VARSIZED"
```
yschroeder97: This is exactly the problem I see with having E2E tests for these very narrow cases. 

Now we have it only tested for `MAX`, the others not, which is inconsistent. 
Should we now write an E2E test for every agg function, checking the error message?

nes-logical-operators/src/Operators/Windows/Aggregations/SumAggregationLogicalFunction.cpp
```diff
@@ -87,7 +88,7 @@ DataType SumAggregationLogicalFunction::inferFromInput(const DataType inputType)
     {
         return inputType;
     }
-    throw CannotInferStamp("aggregations on non numeric fields is not supported (got {})", inputType);
+    throw UnsupportedQuery("SUM is only supported on numeric fields, but got type {}", magic_enum::enum_name(inputType.type));
```
yschroeder97: Field name is missing here, which is included in the other msgs of the agg functions.

nes-systests/operator/aggregation/WindowAggregationVarSizedMinimal.test
yschroeder97: If you insist on having these as systests, add the other cases as well. 

Positive: here you actually included the specific details (field name and type) in the message.

nes-sql-parser/private/AntlrSQLParser/AntlrSQLHelper.hpp
```diff
@@ -102,6 +102,10 @@ class AntlrSQLHelper
     std::vector<Projection> preAggregationProjections;
     size_t aggExprCounter = 0;
 
+    /// Number of aggregations collected before the HAVING clause was entered, so that aggregations written
+    /// inside HAVING can be told apart from the ones the SELECT list already projects.
```
yschroeder97: Dislike this LLM sentence construction "can be told apart": 

```suggestion
    /// inside HAVING can be distinguished from the ones the SELECT list already projects.
```

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -994,12 +1018,32 @@ void AntlrSQLQueryPlanCreator::exitCastExpression(AntlrSQLParser::CastExpression
...
+    helpers.top().aggCountBeforeHaving = helpers.top().windowAggs.size();
     AntlrSQLBaseListener::enterHavingClause(context);
 }
 
 void AntlrSQLQueryPlanCreator::exitHavingClause(AntlrSQLParser::HavingClauseContext* context)
 {
     helpers.top().isWhereOrHaving = false;
+    /// HAVING is applied after the window operator, which only emits the fields the SELECT list named. An
+    /// aggregation spelled out again in HAVING therefore resolves against a generated name (VALUE_COUNT) that
+    /// the caller never wrote, so refuse it here and point at the alias the SELECT list already gives it.
```
yschroeder97:
```suggestion
    /// HAVING is applied after the window operator, which only emits the fields from the SELECT list. 
    /// An aggregation written in HAVING therefore resolves against a generated name (VALUE_COUNT) that
    /// the caller never wrote, so refuse it and point at the alias in the SELECT list.
```

Try to use plain verbs in these comments instead of these verbose ones like "names", "spells", etc.

nes-systests/parser/WindowAggregationParser.test
```diff
@@ -59,6 +59,29 @@ INTO sinkStream;
 2000,3000,16,2
 1500,2500,16,2
 
+# HAVING that repeats the aggregate expression should be refused by naming the alias the SELECT list
+# gives it.
```
yschroeder97:
```suggestion
# HAVING that repeats the aggregate expression should be refused.
# The error message should hint to use an alias from the SELECT list.
```

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -77,11 +78,23 @@
...
 {
 
+namespace
+{
+/// Renders an aggregation back as SQL (COUNT(VALUE)) so error messages quote the caller's own expression.
+std::string describeAggregation(const WindowAggregationLogicalFunction& aggregation)
+{
+    const auto field
+        = std::visit([](const auto& access) { return access->explain(ExplainVerbosity::Short); }, aggregation.getInputFunction());
+    return fmt::format("{}({})", toUpperCase(std::string{aggregation.getName()}), field);
```
yschroeder97: `HAVING SUM(value + 1)` would render as `SUM(_AGG_INPUT_1)`, because `ensureFieldAccessArgument` desugars non-field arguments into `_agg_input_N`. 
The msg in L1044 would tell the user to write this into the SELECT list.

Nit: `toUpperCase` takes a `string_view`, so you can spare yourself the temp string.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -994,12 +1018,32 @@ void AntlrSQLQueryPlanCreator::exitCastExpression(AntlrSQLParser::CastExpression
...
+    if (aggregations.size() > helpers.top().aggCountBeforeHaving)
+    {
+        const auto& [havingAggregation, _] = aggregations[helpers.top().aggCountBeforeHaving];
+        const auto expression = describeAggregation(havingAggregation);
+        const auto projected = std::ranges::find_if(
+            aggregations | std::views::take(helpers.top().aggCountBeforeHaving),
+            [&](const auto& entry) { return entry.first == havingAggregation and entry.second.has_value(); });
+        if (projected != (aggregations | std::views::take(helpers.top().aggCountBeforeHaving)).end())
+        {
+            throw InvalidQuerySyntax("HAVING must use the projected alias: write {} instead of {}", projected->second.value(), expression);
```
yschroeder97: You assume that `entry.second` is an alias that the user chose in the query, but e.g., `exitFunctionCall:1391` stores `{field}_{FUNC}` for every aggregation, and only an explicit `AS` will overwrite that. 

`SELECT SUM(value) FROM ... HAVING SUM(value) > 0` yields `write VALUE_SUM instead of SUM(VALUE)`, which tells the user it needs to know about and use an internally-generated name.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
yschroeder97: In general, the comment is hard to understand. Alternative: 

```
/// An aggregation written inside HAVING is a second aggregation, not a reference to the projected one. 
/// Refuse it and report the alias that already exists.
```

nes-systests/parser/WindowAggregationParser.test
```diff
@@ -59,6 +59,29 @@ INTO sinkStream;
...
+SELECT start, end, id, SUM(value) as sumValue
+FROM stream
+WINDOW SLIDING(timestamp, size 1 sec, advance by 500 ms)
+GROUP BY id
+HAVING SUM(value) > UINT64(0)
```
yschroeder97, line 79: Control below adds `WHERE id > 15` which is omitted here.
Both should either use it or drop it.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -453,7 +466,18 @@ void AntlrSQLQueryPlanCreator::exitBoolComparison(AntlrSQLParser::BoolComparison
     }
     else
     {
-        throw UnsupportedQuery("Predicate is currently not supported: {}", comparison->getText());
+        /// `kind` carries the operator token (LIKE, RLIKE, DISTINCT, ...). The message should name that token so
+        /// the caller knows which construct to drop.
```
yschroeder97:
```suggestion
        /// `kind` stores the operator token (LIKE, RLIKE, DISTINCT, ...) to include in the error output.
```

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -1314,7 +1358,24 @@ void AntlrSQLQueryPlanCreator::exitFunctionCall(AntlrSQLParser::FunctionCallCont
                 }
                 else
                 {
-                    throw InvalidQuerySyntax("Unknown (aggregation) function: {}, resolved to token type: {}", funcName, tokenType);
+                    /// The registry also holds plan-internal entries that are not callable from SQL; listing them
+                    /// would send the caller after names that cannot be written in a query.
```
yschroeder97:
```suggestion
                    /// The registry also holds plan-internal entries that are not callable from SQL
```

What does "send the caller after names" mean?

nes-systests/parser/VarSizedSupport.test
```diff
@@ -36,3 +36,21 @@ SELECT * FROM streamWithText INTO sinkStreamWithText;
...
+
+# Control for the case above: IN over the same column should be supported.
+SELECT id, value, text1, text2 FROM streamWithText
+WHERE text1 IN (VARSIZED('test3')) INTO sinkStreamWithText;
+----
+11,2,test3,test2
+11,3,test3,test1
+
+# An unknown scalar function should be reported as an unknown function and should name the supported
+# function set.
```
yschroeder97:
```suggestion
# An unknown scalar function should be reported and specify the set of supported functions.
```

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -1314,7 +1358,24 @@ void AntlrSQLQueryPlanCreator::exitFunctionCall(AntlrSQLParser::FunctionCallCont
...
+                    static constexpr std::array PlanInternalFunctions{"FIELDACCESS", "UNBOUNDFIELDACCESS", "CONSTANTVALUE"};
```
yschroeder97: This denylist will go stale very soon, and some of these cannot be written as function calls, for example:
- ADD
- DIV
- EQUALS
- ISNAN is IS NAN

The functions that are printed here as pointers what to use actually are not callable from SQL.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -1314,7 +1358,24 @@ void AntlrSQLQueryPlanCreator::exitFunctionCall(AntlrSQLParser::FunctionCallCont
...
+                    auto registeredNamesFunctions = LogicalFunctionRegistry::instance().getRegisteredNames()
+                        | std::views::filter([](const auto& name) { return not std::ranges::contains(PlanInternalFunctions, name); })
+                        | std::ranges::to<std::vector>();
+                    auto registeredNamesAggregation = AggregationLogicalFunctionRegistry::instance().getRegisteredNames()
```
yschroeder97: This is a no-op, because the aggregation functions cannot contain any of the above.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -1314,7 +1358,24 @@ void AntlrSQLQueryPlanCreator::exitFunctionCall(AntlrSQLParser::FunctionCallCont
...
+                        | std::views::filter([](const auto& name) { return not std::ranges::contains(PlanInternalFunctions, name); })
+                        | std::ranges::to<std::vector>();
+
+                    std::ranges::sort(registeredNamesFunctions);
+                    std::ranges::sort(registeredNamesAggregation);
+                    throw InvalidQuerySyntax(
+                        "Unknown function: {}. Supported aggregation functions {}. "
```
yschroeder97:
```suggestion
                        "Unknown function: {}. Supported aggregation functions are {}. "
```

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -453,7 +466,18 @@ void AntlrSQLQueryPlanCreator::exitBoolComparison(AntlrSQLParser::BoolComparison
...
+        /// `kind` carries the operator token (LIKE, RLIKE, DISTINCT, ...). The message should name that token so
+        /// the caller knows which construct to drop.
+        const auto predicateName = [&]() -> std::string
+        {
+            if (comparison->kind == nullptr)
+            {
+                return "This predicate";
+            }
+            const auto keyword = toUpperCase(comparison->kind->getText());
+            return comparison->IS() != nullptr ? fmt::format("IS {}", keyword) : keyword;
```
yschroeder97: `IS NOT? kind=(TRUE|FALSE|UNKNOWN)` and `IS NOT? kind=DISTINCT FROM right=...` both lose information here. 
`x IS NOT TRUE` reports "IS TRUE is not supported", and `IS DISTINCT FROM` reports "IS DISTINCT is not supported" with the FROM dropped

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
```diff
@@ -453,7 +466,18 @@ void AntlrSQLQueryPlanCreator::exitBoolComparison(AntlrSQLParser::BoolComparison
...
     {
-        throw UnsupportedQuery("Predicate is currently not supported: {}", comparison->getText());
+        /// `kind` carries the operator token (LIKE, RLIKE, DISTINCT, ...). The message should name that token so
+        /// the caller knows which construct to drop.
+        const auto predicateName = [&]() -> std::string
+        {
+            if (comparison->kind == nullptr)
+            {
+                return "This predicate";
+            }
```
yschroeder97: `kind` cannot be null here, the kind-less alternatives are handled by the branches above. 
"This predicate" does not give any information, should at least include `comparison->getText()`. 

Also, dropping `getText()` from the message means a query with multiple predicate does not specify which one failed.

nes-systests/parser/VarSizedSupport.test
```diff
@@ -36,3 +36,21 @@ SELECT * FROM streamWithText INTO sinkStreamWithText;
...
+SELECT UPPER(text1) AS t FROM streamWithText INTO sinkStreamWithText;
+----
+ERROR 2000 "Unknown function: UPPER"
```
yschroeder97: Where is the supported function set?

yschroeder97 (changes requested):
Thanks for the effort in improving usability. 

The main problem to solve is that generated names still leak in the cases the issue did not cover. 
VALUE_SUM when the SELECT aggregate has no AS, _agg_input_N when a HAVING aggregate wraps an expression, WindowedAggregation for a dropped column., and ADD or NEGATE in the unknown function list. 

Furthermore, the comments should be streamlined and made more concise.

ThanGeo (changes requested):
Agreed with Yannik's comments. Will accept after they are addressed and @arcuellar88 and @tim-nick have given feedback as well.

nes-systests/operator/aggregation/WindowAggregationProjection.test
keyseven123: Done

nes-systests/sinks/FileSinkConfig.test
keyseven123: Fixed.

nes-systests/sinks/FileSinkConfig.test
keyseven123: I have added it, as we also have a `File_Config.test` for the source. I have converted it into a unit test, though.

nes-systests/operator/aggregation/WindowAggregationVarSizedMinimal.test
keyseven123: I like naming more here.

nes-logical-operators/src/Operators/Windows/Aggregations/SumAggregationLogicalFunction.cpp
keyseven123: Added

nes-systests/parser/WindowAggregationParser.test
keyseven123: Done

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: Took it.

nes-sql-parser/private/AntlrSQLParser/AntlrSQLHelper.hpp
keyseven123: Taken

nes-systests/parser/WindowAggregationParser.test
keyseven123: Done

nes-systests/operator/aggregation/WindowAggregationProjection.test
keyseven123: Nope, added the field name to the expected error message.

nes-logical-operators/src/Operators/ProjectionLogicalOperator.cpp
keyseven123: Taken.

nes-logical-operators/src/Operators/ProjectionLogicalOperator.cpp
keyseven123: Done

nes-systests/operator/aggregation/WindowAggregationVarSizedMinimal.test
keyseven123: I have moved this into a unit test. We are now also testing all registered aggregation types.

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: Done

nes-logical-operators/src/Operators/ProjectionLogicalOperator.cpp
keyseven123: Done

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: Taken

nes-systests/parser/VarSizedSupport.test
keyseven123: Done

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: Removed

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: Removed the denylist.

nes-systests/parser/VarSizedSupport.test
keyseven123: Added the function set

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: Good catch. Changed

nes-sql-parser/src/AntlrSQLQueryPlanCreator.cpp
keyseven123: Done.

yschroeder97 (approved):

ThanGeo (approved):

PR #1973, opened 2026-08-31
build(CI): Fix dependency-image build failures and Node 20 action deprecations

keyseven123:
@oczerwia-2010 please ignore the last commit. It was merged accidentially and has already been approved in #1598 .

Fixes the failures hit when the dependency images are actually rebuilt, i.e. on-demand image runs and any change to the dependency hash. Regular PRs reuse a cached image, so none of this surfaces there. The change list is as follows:

- Set `AWS_RETRY_MODE`/`AWS_MAX_ATTEMPTS` for the vcpkg binary cache, so throttled S3 reads back off instead of being treated as a cache miss and rebuilt from source.
- Fetch ClangBuildAnalyzer as a checksummed release tarball instead of cloning it, since GitHub answers throttled anonymous git traffic with a 401 that makes `git` prompt for a username and fail.
- Bumped `docker/login-action`, `docker/setup-buildx-action` and `docker/build-push-action` to their Node 24 majors, and dropped the `install` input the buildx major removes.
- Resolve `head_sha` to a commit sha once in `create_executable_images.yml`. Callers pass a mutable ref (the nightly passes `main`), and each job re-resolved it independently, so a ref moving mid-run made the dependency hash and the build disagree.

## Verifying this change
This change is covered by runnign the CI.

- CI (`create_executable_images.yml`, `get_dev_images.yml`, `pr-merge.yml`)
- Dependencies (dependency/development image build; merging forces a one-time rebuild of all dependency images)


docker/dependency/Dependency.dockerfile
```diff
@@ -49,6 +49,15 @@ ENV VCPKG_FORCE_SYSTEM_BINARIES=1
 #   1. S3 credentials provided: authenticated readwrite cache
 #   2. Only VCPKG_CACHE_PUBLIC_URL provided: read-only HTTP cache (public bucket)
 #   3. Nothing provided: no cloud cache, local build only
+#
+# The 12 dependency-image matrix jobs share host-triplet cache objects, so the bucket throttles the
+# authenticated tier with 429/ServiceUnavailable. vcpkg shells out to the aws CLI per package and
+# treats a non-zero exit as a cache miss, so AWS_RETRY_MODE/AWS_MAX_ATTEMPTS let the CLI back off
+# instead of silently rebuilding the package from source.
```
ls-1801: I would shorten this comment: Saying that we enable retry to no rebuild because of rate limits

ls-1801 (approved):

oczerwia-2010 (approved):

PR #1975, opened 2026-08-31
fix(Network): send child buffers at their used size, not their capacity

nmlt:
Child buffers do not get their size truncated if they fit into their anticipated allocated memory. This fixes it. Should reduce network usage in cases with many small child buffers.

## Verifying this change
- Tests
- I measured network throughput in [this PR](https://github.com/ESPAT-Orga/nebulastream-espat/pull/43/commits) with and without this fix and it reduced it a lot ;)

Network

## Documentation
n/a

n/a


ls-1801 (approved):
Do we even set the number of tuples anywhere for child buffers, except for the network source receive path?

nmlt:
> Do we even set the number of tuples anywhere for child buffers, except for the network source receive path?

There was another path that uses the field that way. So this is just a fix without really addressing the confusing name of the field.

keyseven123 (approved):

PR #1976, opened 2026-09-01
test(CMake): Serialize the DockerCompose test suites

keyseven123:
The three suites driving nes-cli derive the same image prefix and network label, and each setup_file force-removes both, tearing down a concurrently running sibling. Each suite now derives its own identity, serializing the label guards the rest, and bats_lib.bats fails if two ever share one again.

## Verifying this change
This change is tested by existing test but also the newly added `distributed_bats_lib.bash`.

- Bats tests


greenBene (approved):

scripts/testing/tests/bats_lib.bats
ls-1801, line 1: I wouldn't test the testing infrastructure here.

ls-1801 (approved):

PR #1978, opened 2026-09-01
fix(engine): synchronize delayed task shutdown

ls-1801:
## Summary

The DelayedTaskSubmitter could previously insert a task after the worker
threads have drained the queue. This caused the dtor of the queue to destroy
task's target pipelines would trip over the invariant that all pipelines
need to be terminated.

The new implementation initializes the shutdown on the DelayedTaskSubmitter,
which will fail all pending tasks, and reject and future delayed task submissions.

This ensures that the queue is truly empty after it is drained.

This would have manifested in flakey tests:

```
  systest: /home/runner/work/nebulastream/nebulastream/nes-query-engine/RunningQueryPlan.cpp:118: NES::RunningQueryPlanNode::~RunningQueryPlanNode(): Assertion `!requiresTermination && "Node was destroyed without termination. This should not happen"' failed.
  ```
  
  https://github.com/nebulastream/nebulastream/actions/runs/33396979743/job/99507861717

### DIFF
--- a/nes-query-engine/DelayedTaskSubmitter.hpp
+++ b/nes-query-engine/DelayedTaskSubmitter.hpp
@@ -52,38 +52,61 @@ private:
         bool operator()(const ScheduledTask& left, const ScheduledTask& right) const { return left.deadline > right.deadline; }
     };
 
+    using TaskQueue = std::priority_queue<ScheduledTask, std::vector<ScheduledTask>, TaskComparator>;
+
+    struct State
+    {
+        TaskQueue taskQueue;
+        bool acceptingTasks = true;
+    };
+
     SubmitFn submitFn;
 
     std::condition_variable_any cv;
-    folly::Synchronized<std::priority_queue<ScheduledTask, std::vector<ScheduledTask>, TaskComparator>, std::mutex> taskQueueMtx;
+    folly::Synchronized<State, std::mutex> stateMtx;
 
     /// The DelayedTaskSubmitter is implemented as its own dedicated thread. Most of the time is spent blocking on an empty task queue
     /// or waiting until the deadline has passed to submit the next task.
+    std::once_flag shutdownOnce;
     Thread workerThread;
 
     void workerLoop(const std::stop_token& stop);
+    static void failTaskDuringShutdown(Task& task);
 
 public:
     explicit DelayedTaskSubmitter(SubmitFn submitFn);
 
     /// Template function to accept any std::chrono::duration type
+    /// The task will be submitted via the submit function after a delay.
+    /// If the DelayedTaskSubmitter is shutdown the task will trigger onFailure and onComplete.
     template <typename Rep, typename Period>
     void submitTaskIn(Task task, std::chrono::duration<Rep, Period> delay)
     {
         auto deadline = ClockType::now() + delay;
 
-        auto taskQueue = taskQueueMtx.lock();
-        const bool isEarliest = taskQueue->empty() || deadline < taskQueue->top().deadline;
-        taskQueue->emplace(ScheduledTask{std::move(task), deadline});
+        auto state = stateMtx.lock();
+        if (!state->acceptingTasks)
+        {
+            state.unlock();
+            failTaskDuringShutdown(task);
+            return;
+        }
+
+        const bool isEarliest = state->taskQueue.empty() || deadline < state->taskQueue.top().deadline;
+        state->taskQueue.emplace(ScheduledTask{std::move(task), deadline});
 
         /// Wake up the worker thread if this task has an earlier deadline or if the taskqueue was empty.
         /// This is necessary, as we need to set the cv_wait_until to ensure no task being missed.
         if (isEarliest)
         {
-            taskQueue.unlock();
+            state.unlock();
             cv.notify_one();
         }
     }
+
+    /// Stops the submitter thread and fails all outstanding tasks. Concurrent calls block until the single shutdown operation has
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: nes-query-engine/DelayedTaskSubmitter.hpp [hunk 1]
```diff
@@ -124,40 +148,71 @@ void DelayedTaskSubmitter<CT>::workerLoop(const std::stop_token& stop)
...
+        shutdownOnce,
+        [this]
+        {
+            {
+                auto state = stateMtx.lock();
+                state->acceptingTasks = false;
+            }
+
+            /// Joining guarantees that no task submission into the engine task queue is still in flight when shutdown returns.
+            workerThread = {};
```
Artraxon, line 191: How does assigning a a thread to a new one join the current thread with the thread saved in the variable? Does that really work?

nes-query-engine/DelayedTaskSubmitter.hpp
```diff
@@ -124,40 +148,71 @@ void DelayedTaskSubmitter<CT>::workerLoop(const std::stop_token& stop)
...
+            {
+                auto state = stateMtx.lock();
+                if (state->taskQueue.empty())
+                {
+                    break;
+                }
+
+                /// Priority queue's API only returns a const reference. We have exclusive access and immediately remove the element.
+                /// NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+                auto task = std::move(const_cast<ScheduledTask&>(state->taskQueue.top()).task);
```
Artraxon, line 205: Why is a move needed here? no rvalue ctor is used (or any other function that requires rvalues), all the functions just take normal mutable references.

Artraxon (changes requested):

nes-query-engine/DelayedTaskSubmitter.hpp
ls-1801: The task object is move-only, so we are using the move assignment here. `top()` returns an (const-)lvalue reference that we can move out of, leaving a moved-from object in the queue, which is destroyed by `pop`.

nes-query-engine/DelayedTaskSubmitter.hpp
ls-1801: Yes, it does. jthread joins on destruction, which is triggered by assigning a new default-constructed thread (which does not do anything)

nes-query-engine/DelayedTaskSubmitter.hpp
```diff
@@ -52,45 +52,69 @@ class DelayedTaskSubmitter
...
 
     void workerLoop(const std::stop_token& stop);
+    static void failTaskDuringShutdown(Task& task);
 
 public:
     explicit DelayedTaskSubmitter(SubmitFn submitFn);
 
     /// Template function to accept any std::chrono::duration type
+    /// The task will be submitted via the submit function after a delay.
+    /// If the DelayedTaskSubmitter is shutdowm the task will fail.
```
keyseven123:
```c++
    /// If the DelayedTaskSubmitter is shut dowm the task will fail.
```
and also maybe mention that the onComplete runs and not only the onFailure

keyseven123 (changes requested):

nes-query-engine/QueryEngine.cpp
```diff
@@ -647,17 +658,6 @@ bool ThreadPool::WorkerThread::operator()(StopPipelineTask& stopPipelineTask) co
...
-            {
-                /// A stage that cannot finish its stop (e.g. a network sink whose peer never accepted the channel) would otherwise
-                /// add this task into the queue forever, and the worker threads would never see an empty queue.
-                ENGINE_LOG_WARNING("Dropping pipeline stop retry during query engine termination");
-                /// Dropping the task destroys the node it still owns. Clear the termination flag first so the node satisfies the
-                /// ~RunningQueryPlanNode invariant instead of aborting on the `!requiresTermination` assertion.
-                stopPipelineTask.pipeline->requiresTermination = false;
-                return;
-            }
-
```
keyseven123, line 660: The repeat handler moves the owning `unique_ptr<RunningQueryPlanNode>` into `submitTaskIn`, which takes the Task by value and now destroys it         
                           on the shutdown-rejection path — so the stage is destroyed while its own `stop()` is still on the stack.

keyseven123 (changes requested):

Artraxon (approved):

nes-query-engine/QueryEngine.cpp
```diff
@@ -647,33 +669,29 @@ bool ThreadPool::WorkerThread::operator()(StopPipelineTask& stopPipelineTask) co
...
     auto pipelineId = stopPipelineTask.pipeline->id;
     auto queryId = stopPipelineTask.queryId;
     stopPipelineTask.pipeline->stage->stop(pec);
+
+    if (requiresTaskRepetition)
+    {
+        const auto duration = requiresTaskRepetition.value();
+        Task repeatedTask
+            = StopPipelineTask(stopPipelineTask.queryId, std::move(stopPipelineTask.pipeline), std::move(stopPipelineTask.callback));
+        if (duration.count() > 0)
```
keyseven123:
```c++
        StopPipelineTask repeatedTask{stopPipelineTask.queryId, std::move(stopPipelineTask.pipeline), std::move(stopPipelineTask.callback)};
```

keyseven123 (approved):
One thing you can pick, if you want.

PR #1983, opened 2026-09-01
fix(SystestResultCheck): Fix parsing errors for quoted field identifiers containing commas in parseFieldNames

### LINKED ISSUE #1981: [BUG] SystestResultCheck fails to parse field-identifiers containing ','
### Version / Commit

965964636a3838132d2d9cd948451b7363b6297b

### What happened?

We allow to quote field identifiers to keep them case-sensitive and include special characters. However, the CSV header created by SchemaFormatter.cpp lacks the quotes for these quoted identifiers.  This is a problem, as the SystestResultCheck splits this header on ',' to obtain the identifier-datatype pairs. Without quotes, there is no way to differentiate between delimiting comma and commas included in a quoted field-identifier here. This causes systest-queries like these to trigger an invariant in parseFieldNames in SystestResultCheck.cpp:

```sql
CREATE LOGICAL SOURCE "measurements,stream" ("sensor,value" UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR "measurements,stream" TYPE File;
ATTACH INLINE
42

SELECT "sensor,value" FROM "measurements,stream" INTO FILE();
----
42
```

Special characters like ',' in quoted identifiers are ignored during parsing. The mentioned query should pass without errors.

### How to reproduce

Create a systest with the mentioned query.

### Operating System

Does not matter

### Deployment details

Does not matter

### Anything else

_No response_

### Code of Conduct

LeoRaasch:
Currently, `parseFieldNames` in `SystestResultCheck.cpp` will fail to correctly parse field identifiers containing special characters such as "," or ":", even if these identifiers were quoted in the schema definition. This is because `SchemaFormatter::getFormattedSchema()` omits the quotes of these identifiers. As `SystestResultCheck` obtains the Schema from this function, it currently has no way of differentiating between field-delimiting commas and commas that are part of a field identifier. Systest-Queries like the following therefore trigger an invariant instead of passing:

```sql
CREATE LOGICAL SOURCE "measurements,stream" ("sensor,value" UINT64 NOT NULL);
CREATE PHYSICAL SOURCE FOR "measurements,stream" TYPE File;
ATTACH INLINE
42

SELECT "sensor,value" FROM "measurements,stream" INTO FILE();
----
42
```
This pull request fixes this issue by enforcing that `SchemaFormatter::getFormattedSchema()` displays case-sensitive field identifiers (ergo they appeared quoted in the schema definition) quoted instead of unquoted. Additionally, the pull request introduces the `std::vector<std::string_view>
splitOnMultipleDelimiters(std::string_view input, const std::vector<char>& delimiters, const std::vector<char>& encapsulators);` function to `Strings.hpp`, which ignores all delimiters encapsulated within an identical pair of chars out of the encapsulators vector. `SystestResultCheck` uses this function to ignore special characters within quoted field identifiers.

### Brief change log

- Adds splitOnMultipleDelimiters with encapsulators parameter to Strings.hpp
- SchemaFormatter::getFormattedSchema() displays case-sensitive field identifiers quoted instead of unquoted
- parseFieldNames in SystestResultCheck.cpp ignores special chars in quoted field identifiers during parse

## Verifying this change
A new test-case containing the aforementioned query was added to `QuotedIdentifiers.test`.

- nes-sinks: The CSV output-format uses the output of SchemaFormatter::getFormattedSchema() as file header. These headers are affected by the changes to this function (displaying quoted field identifiers as quoted instead of removing the quotes).
- SystestResultCheck
- nes-common (Strings.hpp)

## Documentation
- Change to SchemaFormatter was explained via comment
- Some comments in SystestResultCheck were adjusted to match the new behavior

This PR closes #1981


keyseven123 (approved):
Maybe @yschroeder97 knows a pr that you can stack this pr on top of, as he is currently refactoring the systest.

nes-sinks/src/SinksParsing/SchemaFormatter.cpp
```diff
@@ -34,9 +35,14 @@ std::string SchemaFormatter::getFormattedSchema()
                 | std::views::transform(
                     [](const auto& field)
                     {
+                        const Identifier& identifier = field.getFullyQualifiedName();
+                        /// SystestResultCheck currently splits the output of this function on ',' to gain the identifier-datatype pairs.
+                        /// Without quotes, SystestResultCheck has no way to differentiate between the delimiting ',' and ',' that might appear within the quoted field identifier.
+                        /// Therefore, we reapply the quotes of case-sensitive identifiers here.
```
yschroeder97: There is no `SystestResultCheck` anymore; please update this.

yschroeder97 (approved):
Please update the outdated comment :)

Thanks for the fix!

nes-sinks/src/SinksParsing/SchemaFormatter.cpp
LeoRaasch: Updated it

PR #1985, opened 2026-09-02
refactor(SinkDescriptor): Move backpressure threshold config parameters to Network Sink configuration

### LINKED ISSUE #1964: [TASK] Move Backpressure Config Options to Network Sink
### Description

They are Network Sink specific and should live there.

### Acceptance Criteria

They are moved.

LeoRaasch:
The sink config parameters `BACKPRESSURE_UPPER_THRESHOLD` and `BACKPRESSURE_LOWER_THRESHOLD` are currently `SinkDescriptor` configuration parameters, even though they are only relevant for the `NetworkSink` and the `MQTTSink` plugin. This pull request removes both config parameters from the sink descriptor and adds them to the parameters of the `NetworkSink` and `MQTTSink`.
## Verifying this change
This change is tested by
- Setting these parameters still works exactly the same. Adjusted repl .bats test by removing the parameters from the expected configurations of the FILE sink

- nes-sinks
- MQTTSink plugin

This PR closes #1964


Artraxon (approved):

ls-1801 (changes requested):
Is there a specific reason you need this change? I explicitly moved them from the network sink to be a global sink option so other sinks can use them in a PR a while back.

The intention is that this logic can be used by any kind of sink; for example, the file sink could be implemented in a non-blocking way, which can signal backpressure.

LeoRaasch:
> Is there a specific reason you need this change? I explicitly moved them from the network sink to be a global sink option so other sinks can use them in a PR a while back.
> 
> The intention is that this logic can be used by any kind of sink; for example, the file sink could be implemented in a non-blocking way, which can signal backpressure.

I did not open the issue for this PR, but I can see why it was opened: I thought that the config parameters offered by the sink descriptor should ideally be used by every sink implementation. In other instances, like the InputFormatterDescriptor, we currently handle things in a similar manner: `ALLOW_COMMAS_IN_STRINGS` might be applicable to formats besides CSV but is located in `ConfigParametersCSVInputFormatIndexer` as the CSVInputFormatter is currently the only input formatter using it. If the user sees config parameters, which do not actually affect the sink, popping up when creating it, they might assume that the parameters actually impact the sink in some way.
Maybe @Artraxon can back me up on this, as he opened the issue.

PR #1987, opened 2026-09-03
fix(docs): Sync docs to new plugin registry

Artraxon:
Small fixups of the documentation where the old plugin registry was referenced.
Should we also remove the DD of the old plugin registry completely? IMO there are multiple things in there that I would consider highly inaccurate/wrong, some large reasoning jumps and questionable design constraints.

The approach that we now have is kind of the standard c++ way of doing it, and how one uses it is already well documented.


docs/design/20240701_static_plugin_registry.md
```diff
@@ -1,3 +1,5 @@
+> **Note (2026-09):** The configure-time `GeneratedRegistrar`/`add_plugin_library` design described here has been superseded by runtime registries (`create_runtime_registry`/`add_registry_entry`, see `cmake/RuntimeRegistrationUtil.cmake` and [guide/extensibility.md](../guide/extensibility.md)). This document is kept for historical context.
```
ThanGeo, line 1: Maybe in such cases we can use a keyword label such as [OUTDATED] or something?

docs/guide/how_to_add_an_optimizer_rule.md
```diff
@@ -36,31 +36,32 @@ Registering a rule plugin consists of:
...
+# Load runtime registry utils
+include(${PROJECT_SOURCE_DIR}/cmake/RuntimeRegistrationUtil.cmake)
 
-# 1) Register Plugin 
-add_plugin_as_library(
-        RedundantUnionRemoval               # Plugin Name 
-        PlanRule                            # Registry Name
-        RedundantUnionRemovalRulePlugin     # Plugin library Name
-        RedundantUnionRemovalRule.cpp)      # List of source files
+# 1) Build the plugin implementation library
```
ThanGeo, line 42: Just a question for my comprehension, why do we use plugins for new rules? Is it just for flexibility?

ThanGeo (approved):
Just a few minor comments. On the obsolete/outdated design doc I think there should be an explicit tag/header that clearly states that the design doc is either outdated, partly outdated or up-to-date with the current implementation. Maybe we annotate design documents with flags in each part based on their freshness level

docs/guide/how_to_add_an_optimizer_rule.md
greenBene: It's to allow users to easily define their own rules, similar to functions, operators, etc :)

PR #1990, opened 2026-09-04
refactor(catalogs): resolve source/sink configs in the statement binder

Artraxon:
The previous PRs in the stack changed already how descriptors internally store their data, that is that they can now hold some arbitrary configuration type, and how config options are validated and used to instantiate those configuration types.

But, the PR needed to keep the old map-based descriptor interface since all the binders are working with it.

This PR removes the map-based descriptors fully, and ports over the the anonymous source/sink operators, the binding rules and all binders to the new interface.

Options are now also scoped by the plugin name, i.e. the SEED option for the generator source is now scoped under GENERATOR_SOURCE.SEED . It is also possible to omit the qualifier if the config option is otherwise unambiguous.


nes-frontend/apps/cli/CLIStarter.cpp
```diff
@@ -554,16 +557,46 @@ std::vector<std::string> loadQueries(
     return queries;
 }
 
-std::unordered_map<NES::Identifier, std::string> bindConfig(const std::unordered_map<std::string, std::string>& config)
+NES::Schema<NES::LiteralConfigValue, NES::Ordered> bindSourceConfig(const std::unordered_map<std::string, std::string>& config)
```
greenBene, line 560: As discussed before, while I see that `NES::Schema` might provide the functionality you need, I find it highly confusing for this use case. Since we are in a database group, if one reads"schema", the first and most logical connection is, IMO, a database schema, which does not map to this use case since not only is the ' schema ' of the config included in the NES:Schema object, but in the FieldType is the value encoded, if I understand correctly?

I am not sure what the right solution is here: Maybe rename NES::Schema to the name of the data structure that it best describes and have a new `NES::Schema` and a `NES:Config` inherit from it?

nes-logical-operators/src/Operators/Sources/AnonymousSourceLogicalOperator.cpp
```diff
@@ -50,39 +52,25 @@ AnonymousSourceLogicalOperator AnonymousSourceLogicalOperator::withInferredSchem
...
 
 std::string AnonymousSourceLogicalOperator::explain(ExplainVerbosity verbosity, OperatorId id) const
 {
     if (verbosity == ExplainVerbosity::Debug)
     {
-        return fmt::format("ANONYMOUS_SOURCE(opId: {}, type: {} traitSet: {})", id, getSourceType(), traitSet.explain(verbosity));
+        return fmt::format("INLINE_SOURCE(opId: {}, type: {} traitSet: {})", id, pluginSourceConfig.getType(), traitSet.explain(verbosity));
     }
-    return fmt::format("ANONYMOUS_SOURCE({})", getSourceType());
+    return fmt::format("INLINE_SOURCE({})", pluginSourceConfig.getType());
```
greenBene, line 73: We on purpose renamed anonymous source to inline source. Why are you changing it back?

nes-systests/systest/src/SystestExecutor.cpp
```diff
@@ -179,9 +191,76 @@ void exitOnFailureIfNeeded(const std::vector<Systest::RunningQuery>& failedQueri
...
+        ConfigFieldDefault{QualifiedIdentifier::parse("TCP_SOURCE.OVERWRITEABLE_HOST_AND_PORT"), [] { return true; }},
+        /// Sink defaults only make `CREATE SINK` statements without options bindable; systest
+        /// replaces every sink's config with a per-query result file anyway (see SLTSinkFactory).
+        ConfigFieldDefault{QualifiedIdentifier::parse("OUTPUT_FORMATTER.TYPE"), [] { return "CSV"; }},
+        ConfigFieldDefault{QualifiedIdentifier::parse("FILE_SINK.FILE_PATH"), [] { return "/tmp/systest_file_source_placeholder.txt"; }},
+        ConfigFieldDefault{
+            QualifiedIdentifier::parse("CHECKSUM_SINK.FILE_PATH"), [] { return "/tmp/systest_checksum_sink_placeholder.txt"; }},
+        ConfigFieldDefault{
+            QualifiedIdentifier::parse("SINK.HOST"), [&] { return config.clusterConfig.allowSinkPlacement.at(0).getRawValue(); }},
+        ConfigFieldDefault{QualifiedIdentifier::parse("FILE_SINK.PATH"), [] { return "/tmp/systest_file_sink_placeholder.txt"; }}};
```
greenBene, line 232: Basically, this is an ordered map, but it requires quite a lot of fluff to write it down. Can we streamline this, so it looks sth like: 
```
{{"TCP.SOURCE.SOCKET_HOST", std::monostate{}},{"TCP_SOURCE.SOCKET_PORT", std::monostate{}}, ...}
```

nes-sql-parser/src/CommonParserFunctions.cpp
```diff
@@ -75,6 +87,81 @@ Identifier bindIdentifier(AntlrSQLParser::StrictIdentifierContext* strictIdentif
...
+                                      {
+                                          return LiteralConfigValue{
+                                              defaultInputFormatter.getFullyQualifiedName(), defaultInputFormatter.get()};
+                                      });
+                          }),
+                  [] { return InvalidQuerySyntax("INPUT_FORMATTER.TYPE must be specified"); })
+                  .and_then(
+                      [](const LiteralConfigValue& value)
+                      { return tryGetOr<std::string>(value.getValue(), expectedType<std::string>()).and_then(Identifier::tryParse); });
+        return unwrapOrThrow(std::move(expected));
```
greenBene, line 121: I think this is really difficult to reason about. Could we simplify it by using normal if-clauses and more than one temporary variable?

greenBene (changes requested):
While I am really happy with the idea of this PR (stack of PRs), I have one major issue: The reliance on `NES::Schema` for the configuration. As I argue below, I believe using the term Schema for the config is really confusing and misleading. I think you built a quite interesting data structure, but it deserves a more fitting name :)

PR #1993, opened 2026-09-08
feat(systest): SQL rewriter

### LINKED ISSUE #1916: Add a SQL rewriter that makes statements correct upfront

yschroeder97:

Stacked on #1999.

The rewriter turns a parsed test file into statements we can submit unchanged, so we can delete any mid-pipeline plan patching of the systest binder in the next step. 
Nothing calls it yet, except its unit tests. 

- Added `Model/RunnableTestFile.hpp`, the rewriter's output and the runner's future input: `RunnableTestFile` holds the setup DDL, the cases to run and check, and the data to stage (`InlineData` for a planned CSV, `ServedData` for a TCP source).
- Added the rewriter as three phases, each producing a single value as output, behind one umbrella class,`SqlRewriter::rewrite(const ParsedTestFile&) const`:
  - `ClassifiedStatement`: parses and classifies every CREATE statement.
  - `Declarations`: registers every catalog-visible name and captures the declared sinks, sealed (i.e., consumed) before any rewriting.
  - `Emitter`: emits the setup DDL and rewrites the queries, in file order.
    An instance exists for a single test file.
- Qualification prefixes every catalog-visible name with the test file key,
  through the lexer (`NameQualifier`), and every test file of one
  invocation can share a single catalog.
- Physical sources get their file path, host, etc. injected;
  inline data becomes a planned CSV file and socket data a served payload
  (`SourceRewriting`, `DataStaging`).
- The sink of each query is inlined with its schema and a result file (`SinkRewriting`).

## Verifying this change

This change is tested by
- two new unit test targets: `systest-sql-rewriter-test`
  (41 tests: qualification, source splicing, sink inlining, EXPLAIN,
  differential blocks, staged data) and `systest-name-qualifier-test`
  (16 tests: key derivation, sealing, unqualification)
- the parser and runner suites, unchanged: `testfile-builder-test` (14) and
  `slt-parser-test` (49)
- the full corpus in the dev container

- nes-systests only. The library now has a `Rewriter/` stage and links
  `tcp_source_plugin_library`; no other dependency changes.
- Two new test targets.

## Documentation

- Every phase type and every rewriting rule has extensive in-code docs.

This PR closes #1916, part of the systest refactoring epic #1920.


nes-systests/systest/private/Model/RewrittenTest.hpp
```diff
@@ -0,0 +1,150 @@
...
+{
+    if (const auto* query = std::get_if<RewrittenQuery>(&testCase.action))
+    {
+        return query->id;
+    }
+    if (const auto* explain = std::get_if<RewrittenExplain>(&testCase.action))
+    {
+        return explain->id;
+    }
+    return std::get<RewrittenDifferential>(testCase.action).firstId;
```
ls-1801: I would prefer a `std::visit(Overloaded{...}, testCase);` as it would fail to compile if the variant is extended.

nes-systests/systest/private/Model/RewrittenTest.hpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/// One runnable test file.
+/// The rewriter produces this and the runner consumes it.
+struct RewrittenTest
```
ls-1801: NIT:

I am not too happy with the `Rewritten` prefix, as it does not really imply anything. Why was it rewritten? What was it rewritten to?

The term rewritten also does not exclude the possibility of it being rewritten again.

I think you already proposed a better name for it: RunnableTestFile.

And I would align our naming with other testing frameworks; I think they would call it a TestSuite with many TestCases. Just calling it Test makes it non-obvious, that these are multiple tests.

nes-systests/systest/src/Rewriter/NameQualifier.cpp
```diff
@@ -0,0 +1,170 @@
...
+        if (const auto type = token->getType(); type == AntlrSQLLexer::IDENTIFIER || type == AntlrSQLLexer::BACKQUOTED_IDENTIFIER)
+        {
+            if (const auto qualified = names.qualified(token->getText()))
+            {
+                rewritten.append(sql.substr(cursor, token->getStartIndex() - cursor));
+                rewritten.append(*qualified);
+                cursor = token->getStopIndex() + 1;
+            }
+        }
+    }
```
ls-1801: I think we cannot use rewrite every identifier.

```sql
  CREATE LOGICAL SOURCE File (v UINT64);
  CREATE PHYSICAL SOURCE FOR File TYPE File;
```

Would be rewritten to:
```sql
  CREATE LOGICAL SOURCE prefix_File (v UINT64);
  CREATE PHYSICAL SOURCE FOR prefix_File TYPE prefix_File;
```

and creates a weird error: `prefix_File` is not a valid source.

ls-1801:
```c++
/// A logical source can share its name with a plugin, but qualifying the source must preserve the plugin type.
TEST_F(NameQualifierTest, PreservesPluginTypeWhenItMatchesLogicalSourceName)
{
    NameRegistry registry{"TESTKEY"};
    registry.declare("File");
    const auto names = std::move(registry).seal();

    EXPECT_EQ(
        rewriteIdentifiers("CREATE PHYSICAL SOURCE FOR File TYPE File;", names),
        "CREATE PHYSICAL SOURCE FOR TESTKEY_FILE TYPE File;");
}
```

nes-systests/systest/src/Rewriter/NameQualifier.cpp
```diff
@@ -0,0 +1,170 @@
...
+    std::string rewritten;
+    size_t cursor = 0;
+    for (const auto& token : lexer.getAllTokens())
+    {
+        if (const auto type = token->getType(); type == AntlrSQLLexer::IDENTIFIER || type == AntlrSQLLexer::BACKQUOTED_IDENTIFIER)
+        {
+            if (const auto qualified = names.qualified(token->getText()))
+            {
+                rewritten.append(sql.substr(cursor, token->getStartIndex() - cursor));
+                rewritten.append(*qualified);
```
ls-1801: ANTLR indexing and C++ indexing don't match. ANTLR indexes will account for Unicode characters, whereas C++ will not. This will manifest as a corrupted test file:

```SQL
SELECT 'é' FROM stream INTO File();
```
should be rewritten to:

```SQL
SELECT 'é' FROM KEY_STREAM INTO File();
```
 but probably is going to end up as:

```
```SQL
SELECT 'é' FROMKEY_STREAMm INTO File();
```
As the antlr index offset is different than the actual byte offset. (as 'é' is represented as 2 bytes).

I think antlr has a TokenStreamRewriter that can handle this for us

nes-systests/systest/src/Rewriter/NameQualifier.cpp
```diff
@@ -0,0 +1,170 @@
...
+    if (character == '/')
+    {
+        return "_D_";
+    }
+    return fmt::format("_{:02X}_", static_cast<unsigned char>(character));
+}
+
+}
+
+std::string getTestFileKey(const std::filesystem::path& testFile, const std::filesystem::path& discoveryRoot)
```
ls-1801: This issue may be a bit esoteric, but I don't think prefixing alone would solve name conflicts.

For example, you could have a file called A.test and A_.test 
File A defines a source called _B
File A_ defines a source called B

Both would produce the same A__B prefixed source name and would clash. 

I think we should solve namespaces on a different level. Or provide different views into the catalog. 

Alternatively, you could make the prefix unique (I think) if you also include the size of the prefix:

So it would become A1_B and A_2_B

nes-systests/systest/src/Rewriter/SinkRewriting.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+std::string SinkRewriter::declaredSinkStatement(SqlParse& parse, const AntlrSQLParser::CreateSinkDefinitionContext* definition) const
+{
+    const std::vector options{
+        Sql::option(Sql::Sink, Sql::Host, target.sinkHost.view()),
+        Sql::option(Sql::Sink, Sql::FilePath, target.resultFile(definition->sinkName->getText()).string()),
+        Sql::option(Sql::Sink, Sql::OutputFormat, Sql::Csv)};
+
+    antlr4::TokenStreamRewriter rewriter{&parse.tokenStream()};
+    rewriter.insertAfter(definition->getStop(), fmt::format(" {}", Sql::setClause(Sql::optionList(options))));
```
ls-1801: Does this always append a new SET clause at the end of the query? What if that sink already had a SET clause?

  CREATE SINK out(id UINT64) TYPE File
  SET ('CSV' AS "SINK"."OUTPUT_FORMAT");

would become:

  CREATE SINK KEY_OUT(id UINT64) TYPE File
  SET ('CSV' AS "SINK"."OUTPUT_FORMAT")
  SET ('localhost:8080' AS "SINK"."HOST", ...);

Sinks for normal queries are replaced with inline sinks, so this is only used for explain queries?

nes-systests/systest/src/Rewriter/SinkRewriting.cpp
```diff
@@ -0,0 +1,157 @@
...
+{
+namespace
+{
+
+/// Returns whether a sink of this type writes a result file.
+/// The File and Checksum sinks do, and they take a path and an output format.
+/// The sink that discards its input takes neither and leaves the query with nothing to check.
+bool writesResultFile(const std::string_view sinkType)
+{
+    return not Sql::sameName(sinkType, Sql::Void);
```
ls-1801: i would invert the check, and make it explicitly only true for File and Checksum. As most other sinks will not work, e.g., print sink

ls-1801 (changes requested):

nes-systests/systest/private/Model/RewrittenTest.hpp
yschroeder97: Applied.

nes-systests/systest/private/Model/RewrittenTest.hpp
yschroeder97: Changed to `RunnableTestFile`, and also changed `ClassifiedTest` --> `ClassifiedTestFile`. 

This gives us consistent naming for the types each stage produces: `ParsedTestFile` -> `ClassifiedTestFile` -> `RunnableTestFile`. 

Regarding alignment with other frameworks, I'd prefer `File` over `Suite` because that's what it is: it belongs to a file that's the logical group where each case is defined. 

But I agree that just `Test` would be too ambiguous. Are you fine with the `TestFile` suffix and `TestCase` to differentiate? 

The case-level types are still `RewrittenCase`, `RewrittenQuery`, `RewrittenDifferential`, `RewrittenExplain`.

nes-systests/systest/src/Rewriter/NameQualifier.cpp
yschroeder97: Confirmed and fixed. Thanks for the test.

It turned out to be bigger than TYPE. 

The grammar writes a plugin type as a plain identifier in four places, and two of them have no TYPE keyword in front: a source or a sink written directly into a query. So `SELECT x FROM File(...) INTO File(...)` broke the same way in a file that declares a source named File.

The fix: parse the statement first, ask the tree which words sit in a type slot, and skip those when renaming. I removed `rewriteIdentifiers`, and the renaming now goes through `TokenStreamRewriter`.

One choice worth explaining: I kept renaming every matching word and only subtracted the type slots, rather than renaming only the words the parse tree calls "names". 
The second version sounds safer but is not: it means listing every kind of name position, and forgetting one leaves it unrenamed, which only shows up later as a missing catalog entry. 
Subtracting the type slots removes exactly the cases you found and changes nothing else.
 
I added your test, adapted to the current API, plus another for the two anonymous positions.

Two side effects:
- A file may now mix an EXPLAIN with a query writing into the same declared sink. That used to be rejected because the old code renamed the text before parsing it, so the sink lookup no longer matched. Parsing the original text fixes it. The test that pinned the limitation now asserts the working behavior.
- A query that does not parse is passed through with its original names, since there is no tree to work from. It fails at parse time before any name is looked up, so nothing changes for the test author.

nes-systests/systest/src/Rewriter/NameQualifier.cpp
yschroeder97: Fixed by the same change as above.

Qualifying now happens via `TokenStreamRewriter`, which the rest of the rewriter already uses, so ANTLR does the text replacement instead of us.

nes-systests/systest/src/Rewriter/SinkRewriting.cpp
yschroeder97: Both are right; fixed.

The append was unconditional, and `createSinkDefinition` allows one optional `SET` clause, not a list, so a sink that wrote its own would have produced two and failed to parse at the worker. 
`spliceSetClause` and `declaredOptions` are now templates in `SqlParse.hpp` shared by both, since the grammar spells optionsClause? identically for a physical source and a sink.

There was a second bug in the same three lines: the three generated options were added unconditionally, so a sink declaring its own `OUTPUT_FORMAT` would have got the key twice even once the clause was spliced correctly. 
They now go through `declaresOption` guards, and the test's own options are appended last.

`FILE_PATH` is the exception and now throws, because the checker reads the file the rewriter chose. 

And yes, EXPLAIN only. Every other sink is inlined into the query that writes to it. 
Nothing in the corpus declares a sink with a SET clause, so both bugs were latent; we now have tests for the merge and the rejection.

nes-systests/systest/src/Rewriter/SinkRewriting.cpp
yschroeder97: Inverted: only File and Checksum write a result file now.

They are also the only two that take a file path, so Print and MQTT were being handed a `SINK.FILE_PATH` that `Descriptor` rejects with `"Unknown configuration parameter"`, naming something the test never wrote.

I went one step further and threw for them rather than treating them like Void. 

Tests added for the rejection and for Void. Nothing in the corpus uses Print or MQTT.

nes-systests/systest/src/Rewriter/NameQualifier.cpp
yschroeder97: You are right, though that example does not create a collision. `encodeKeyCharacter` escapes _ as __, so A_ keys as A__, and _B is not a legal unquoted identifier, so it stays quoted:

A.test  declaring "_B"  ->  "A__B"
A_.test declaring B     ->  A___B

But the claim holds, and this one does collide:

A/B.test declaring C  ->  A_D_B_C 
A.test   declaring D_B_C  ->  A_D_B_C

Your length idea fixes it, and works cleanly if the length goes first, since 5_A_D_B_C is decodable by reading digits up to the first underscore. That changes every qualified name in the corpus and in the expected-plan tests, so I would rather not fold it into this PR for a case nothing currently hits. A

Agreed, the real answer is namespaces at a different level, which would be, IMO, the catalog.
However, I see the issue as quite tiny and would like to defer it until the namespaces are implemented in the catalog.

nes-systests/systest/private/Model/RewrittenTest.hpp
ls-1801: test file is fine

nes-systests/systest/src/Rewriter/NameQualifier.cpp
ls-1801: fine

ls-1801 (approved):

nes-systests/systest/private/Rewriter/NameQualifier.hpp
```diff
@@ -0,0 +1,126 @@
...
+/// Removes the qualifying prefix wherever it occurs in text, so a plan reads as the test wrote it.
+std::string unqualified(std::string_view text, std::string_view qualifyingPrefix);
+
+/// Replaces every identifier that names something the test file declared with its qualified spelling.
+/// Only registered names change, so column names, aliases, keywords and string literals stay as the test wrote them.
+/// An identifier naming a plugin type keeps its spelling, because a type is not a name that any test file declares, and the
+/// grammar admits the same word in both positions: `CREATE PHYSICAL SOURCE FOR File TYPE File` names a source and a plugin.
+/// The replacements join the given rewriter, so a caller combines them with its own edits in one pass, and an edit that
+/// already covers a name wins over the replacement of that name.
+void qualifyNames(SqlParse& parse, antlr4::TokenStreamRewriter& rewriter, const QualifiedNames& names);
```
adrianmichalke: nit: how about using the term 'namespace' instead of 'qualified name'?

nes-systests/systest/private/Rewriter/RewriteTarget.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+namespace NES
+{
+
+/// Context of one test file's rewrite.
+/// Directories are absolute paths on the host running this process, so different workers can resolve them
+/// independently from the location where their process was started.
+/// The source host and the sink host are separate fields, because a run on a cluster may place sources and sinks
+/// on different workers, respectively.
+struct RewriteTarget
```
adrianmichalke: 1) Why dont call it RewriteContext then?
2) I find all comments in this file rather confusing. IMHO, we should rm them

nes-systests/systest/private/Rewriter/SinkRewriting.hpp
```diff
@@ -0,0 +1,85 @@
...
+/// The result file is absent when the sink writes none (e.g., Checksum, Void).
+struct RewrittenSink
+{
+    std::string sql;
+    std::optional<std::filesystem::path> resultFile;
+};
+
+/// Returns the query sink, and rejects a query that writes into none or more than one.
+/// A query needs exactly one, because the runner reads back a single result file per query.
+AntlrSQLParser::SinkContext* requireSingleSink(const SqlParse& parse, const std::string& sql);
```
adrianmichalke, line 50: Do we event support that yet in SQL?

nes-systests/systest/private/Rewriter/SinkRewriting.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+private:
+    [[nodiscard]] RewrittenSink inlined(
+        const std::string& type,
+        AntlrSQLParser::NamedConfigExpressionSeqContext* declared,
+        const std::string& trailingOptions,
+        const std::filesystem::path& candidateResultFile) const;
+
+    /// A non-owning view for the duration of one rewrite; the owner outlives every sub-rewriter it hands these to.
+
```
adrianmichalke:
```suggestion
```

nes-systests/systest/private/Rewriter/Emitter.hpp
```diff
@@ -0,0 +1,88 @@
...
+    /// A non-owning view for the duration of one rewrite; the owner outlives every sub-rewriter it hands these to.
+    const RewriteTarget& target; /// NOLINT(cppcoreguidelines-avoid-const-or-ref-data-members)
+    Declarations declarations;
+    SourceRewriter sourceRewriter;
+    SinkRewriter sinkRewriter;
+
+    /// The data files that each logical source reads, filled by the setup pass and read by the case pass.
+    std::unordered_map<Identifier, std::vector<std::filesystem::path>> inputFilesBySource;
+    RunnableTestFile runnable;
+};
```
adrianmichalke: One high level comment at the top of the class would help way more than the detailed comments per method/member

nes-systests/systest/private/Rewriter/RewriteTarget.hpp
```diff
@@ -0,0 +1,65 @@
...
+{
+    /// Used to qualify the catalog-visible names
+    TestFileKey testFileKey;
+    /// This file's display name in failures and progress lines, which discovery chose so that two files that
+    /// share a stem in different folders do not report the same.
+    std::string displayName;
```
adrianmichalke:
```suggestion
    std::string name;
```
The only time we have the notion of a display name here

nes-systests/systest/src/Rewriter/Emitter.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+void Emitter::emitSetup(std::vector<ClassifiedCreate> setup, const bool submitsDeclaredSinks)
+{
+    for (auto& create : setup)
+    {
+        emitCreate(std::move(create), submitsDeclaredSinks);
+    }
+}
+
+void Emitter::emitCases(std::vector<CaseStatement> cases)
```
adrianmichalke:
```suggestion
void Emitter::emitTestCases(std::vector<TestCaseStatement> testcases)
```
To not confuse with sql case statement

nes-systests/systest/src/Rewriter/Emitter.cpp
adrianmichalke, line 1: The emitters' code path does not seem to be very robust in the face of changes. We should replace this sooner rather than later by implementing support for SQL namespaces in the coordinator.

adrianmichalke (changes requested):
LGTM, some comments on naming and comments.

1) The code path looks legit, but in general, it would be cool if we could replace it rather sooner than later with the SQL namespacing idea we had discussed. With the current solution we are basically maintaining another SQL parser and specifically the emitter looks a bit brittle. Let me know if you need help you with it.

2) I would remove most of the AI-generated comments and replace them with higher-level comments that reflect the actual intent.

nes-systests/systest/private/Rewriter/NameQualifier.hpp
yschroeder97: I would rather not do that, because we then mix something we want to introduce the actual namespace feature with something we don't have yet. 

It's actually only a prefix currently, not a real namespace.

nes-systests/systest/private/Rewriter/RewriteTarget.hpp
yschroeder97: Renamed to `RewriteContext`. 

Stripped the comments to the bare minimum, trying to keep the really useful ones.

nes-systests/systest/private/Rewriter/SinkRewriting.hpp
yschroeder97: Yes, the grammar allows it: `sinkClause: INTO sink (',' sink)*` in `nes-sql-parser/AntlrSQL.g4`. 

The rewriter rejects it only because we read only one result file per query.

nes-systests/systest/private/Rewriter/SinkRewriting.hpp
yschroeder97: Done.

nes-systests/systest/private/Rewriter/Emitter.hpp
yschroeder97: Removed the low-level comments.

nes-systests/systest/private/Rewriter/RewriteTarget.hpp
yschroeder97: Renamed.

nes-systests/systest/src/Rewriter/Emitter.cpp
yschroeder97: Done.

nes-systests/systest/src/Rewriter/Emitter.cpp
yschroeder97: Namespaces in one of the very first two additions to the coordinator after it is initially merged. 

The rewriter itself even after that, though: the runner still has to put the generated data and result file paths into the physical sources and sinks before it submits a statement. 

With namespaces, we can strip the qualification, and the emitter shrinks to path substitution and sink inlining.

nes-systests/systest/private/Rewriter/NameQualifier.hpp
adrianmichalke: I found qualifying very confusing. It made it difficult to understand the code bc it can mean a lot of things. What about 'prefix'?

nes-systests/systest/src/Rewriter/Emitter.cpp
adrianmichalke: Can you change the other occurences as well?

adrianmichalke (approved):
Approved if the last two comments are addressed

nes-systests/systest/src/Rewriter/Emitter.cpp
yschroeder97: Applied.

nes-systests/systest/private/Rewriter/NameQualifier.hpp
yschroeder97: Taken.

PR #1994, opened 2026-09-08
refactor(systest): rewrite statements upfront

### LINKED ISSUE #1917: Rewrite statements upfront instead of patching plans mid-pipeline

yschroeder97:
Part of the systest refactoring (#1920).
Statements are made correct before submission instead of patching plans in the middle of the pipeline, so systest interacts with the system through SQL text only.
- A rewrite stage after parsing: partition a file by its settings, rewrite each part, stage the source data, bind the CREATE statements into one catalog set per invocation.
- The sink path patching, the anonymous source patching and the inline-file writing are removed with most of `SystestBinder`.
- The systest data adaptors leave the source plugins.
  The rewriter emits a physical source with its file path or socket options, and the runner writes the CSV or starts the TCP server before submitting.
- `TCPDataServer` no longer aborts on a clean close after a failed connection; it logs a warning when the close itself fails.

## Verifying this change
- Partition tests in `testfile-builder-test`, a new `systest-binder-test` for the query selection, the existing systest corpus.

- Systest, the File and TCP source plugins.

## Documentation
- In-code documentation on the rewrite stage.

This PR closes #1917.


nes-systests/systest/private/Parser/TestFilePartition.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+#include <Model/ConfigurationOverride.hpp>
+#include <Model/ParsedTestFile.hpp>
+
+namespace NES
+{
+
+/// One configuration override and the corresponding statements of a test file.
+/// Each part runs on its own worker, because a worker takes its configuration at startup and cannot change it later.
+struct TestFilePart
```
greenBene, line 27: Please either rename the file to ` TestFilePart` or rename the struct to ` TestFilePartition`

greenBene: I'd lean towards the second, but it's up to you :)

nes-systests/systest/src/Parser/TestFilePartition.cpp
```diff
@@ -0,0 +1,66 @@
...
+                    [](const SelectStatement& query) { return query.overrides; },
+                    [](const DifferentialStatement& block) { return block.overrides; },
+                    /// An EXPLAIN does not use overrides thus far, so it joins the default partition.
+                    [](const ExplainStatement&) { return ConfigurationOverride{}; }},
+                statement))
+        {
+            auto partition = std::ranges::find(partitions, *overrides, &TestFilePart::overrides);
+            if (partition == partitions.end())
+            {
+                partition = partitions.emplace(partitions.end(), *overrides, ParsedTestFile{.path = testFile.path, .statements = creates});
```
greenBene, line 54: Why not a simple `.emplace_back()` ?

greenBene (changes requested):
Just minor nits. I'm a huge fan of this simplification of the systest code! :)

nes-systests/explain/EXPLAIN.test
```diff
@@ -77,7 +77,7 @@ FROM (
 INTO File();
 ----
 == Initial Logical Plan ==
-SINK(FILE)
+ANONYMOUS_SINK(AnonymousSink)
```
ls-1801, line 80: I am not sure how helpful the inner AnonymousSink is. This should probably show the config

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +206,274 @@ struct SystestBinder::Impl
...
     {
-        sltSinkProvider.registerSink(statement.sinkType, statement.name, statement.schema, statement.sinkConfig);
+        std::visit(
+            Overloaded{
+                [](const PlainStatement&) {},
+                [](const StatementWithInlineData& withInline) { writeInlineData(withInline.data); },
+                [&](StatementWithServedData& withServed)
+                {
+                    auto [thread, options] = serve(std::move(withServed.data));
+                    withServed.sql = addSourceOptions(withServed.sql, options);
```
ls-1801, line 343: Could we make `addSourceOptions` merge options by configuration key, allowing explicitly declared values to override defaults? It currently concatenates the generated options with the existing SET clause. For example, `TYPE TCP SET (200 AS "SOURCE".FLUSH_INTERVAL_MS)` with attached data gets both the generated `100` and the explicit `200`, so the SQL binder rejects the duplicate key and the systest fails during setup. The default flush interval should only be inserted when the test has not specified one.

nes-systests/systest/src/Runner/SystestRunner.cpp
```diff
@@ -271,22 +259,28 @@ std::vector<RunningQuery> runQueries(
...
-                    active.emplace(*regDiff, std::make_shared<RunningQuery>(nextQuery, *regDiff, *reg));
+                    active.emplace(*first, std::make_shared<RunningQuery>(nextQuery, *first, *second));
+                    active.emplace(*second, std::make_shared<RunningQuery>(nextQuery, *second, *first));
                 }
                 else
                 {
+                    /// The first half started but its partner did not, so its terminal state only removes it.
+                    if (first)
+                    {
+                        discardedQueries.insert(*first);
```
ls-1801, line 276: If either half of a differential test fails, could we stop the other query and wait for its termination? These tests compare successful results; continuing the surviving query cannot produce a useful comparison. In particular, if the first submission succeeds and the second fails here, the first query is only added to `discardedQueries`, never to `active`. The main loop checks only `active` and `pending`, so it can exit without stopping or waiting for that query. I reproduced this with the real remote systest runner: the second query failed worker compilation while `nes-cli status` still showed the first as Running after systest exited. Please also cancel the partner when a failure is observed after both queries have started.

ls-1801 (changes requested):

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +206,274 @@ struct SystestBinder::Impl
...
+            /// The rewrite consumes the part's statements. Only its overrides are read afterwards.
+            auto runnable = rewriteTestFile(
+                std::move(file),
+                RewriteTarget{
+                    .testFileKey = partKey,
+                    .displayName = testfile.name().getRawValue(),
+                    .workingDir = workingDir,
+                    .testDataDir = testDataDir,
+                    .sourceHost = clusterConfiguration.allowSourcePlacement.at(0),
+                    .sinkHost = clusterConfiguration.allowSinkPlacement.at(0)});
```
adrianmichalke, line 285: Currently, qualification/namespacing also rewrites functions, e.g., a source abs and leads to rewriting ABS()

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +206,274 @@ struct SystestBinder::Impl
...
-            case TestDataIngestionType::FILE: {
-                if (testData.second.size() != 1)
+        std::vector<SystestQuery> queries;
+        std::unordered_set<SystestQueryId> foundQueries;
+        for (size_t index = 0; index < parts.size(); ++index)
+        {
+            auto& [overrides, file] = parts.at(index);
+            const auto partKey = discoveryRoot.keyOf(testfile.file, index, parts.size());
+            /// The rewrite consumes the part's statements. Only its overrides are read afterwards.
+            auto runnable = rewriteTestFile(
```
adrianmichalke, line 277: Shouldnt `keepSelectedCases` happen before? Here, we can throw on statements that are actually excluded later

nes-systests/systest/src/Runner/SystestRunner.cpp
```diff
@@ -163,7 +148,7 @@ Verdict checkExplainedQuery(const SystestQuery& query)
 
     INVARIANT(query.actualExplainOutput.has_value(), "checking an EXPLAIN requires a computed explain output");
     const auto expectedLines = NES::get<ExpectedPlan>(query.expectation).lines;
-    const auto& actual = query.actualExplainOutput.value();
+    const auto actual = unqualified(query.actualExplainOutput.value(), query.qualifyingPrefix);
```
adrianmichalke: What about the case: orders.test with source named orders_input. Then we get:
 - Original: ORDERS_INPUT
 - Qualified: ORDERS_ORDERS_INPUT
 - Normalized: INPUT

nes-systests/systest/src/Parser/TestFilePartition.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+#include <Model/ConfigurationOverride.hpp>
+#include <Model/ParsedTestFile.hpp>
+#include <Util/Overloaded.hpp>
+
+namespace NES
+{
+
+std::vector<TestFilePart> partitionByOverrides(const ParsedTestFile& testFile)
+{
```
adrianmichalke: We should only parition if we actually need a new worker ie the configuration would require a worker restart. Most configuration do not require that we launch a worker for it, we can just change the config during runtime. The original implmenentation did this

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -489,8 +191,9 @@ struct SystestBinder::Impl
                 }
                 ++loadedFiles;
             }
-            catch (const Exception& exception)
+            catch (const std::exception& exception)
```
adrianmichalke, line 200: nit: why rm our Exception type?

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +206,274 @@ struct SystestBinder::Impl
...
+        }();
 
-        return createUniqueFile(fmt::format("{}/input", sourceDir), ".csv").second;
-    }
+        PRECONDITION(
+            not clusterConfiguration.allowSourcePlacement.empty(),
+            "Topology must list at least one worker in allow_source_placement to assign a default source host");
+        PRECONDITION(
+            not clusterConfiguration.allowSinkPlacement.empty(),
+            "Topology must list at least one worker in allow_sink_placement to assign a default sink host");
```
adrianmichalke, line 252: Is this user input? if so use exceptions

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +206,274 @@ struct SystestBinder::Impl
...
+        return queries;
+    }
 
-        const auto logicalSource = sourceCatalog->getLogicalSource(statement.attachedTo);
-        if (not logicalSource.has_value())
+private:
+    /// Compiles one rewritten test file part: its setup goes into the catalogs, and each case becomes a query to run.
+    /// The part key goes into each query's distributed id, because the coordinator rejects two plans with one id and
+    /// the parts of a file repeat its query numbers.
+    [[nodiscard]] std::vector<SystestQuery> compile(
```
adrianmichalke: misleading naming as no query compilation happens

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +206,274 @@ struct SystestBinder::Impl
...
+        return std::move(binding).value();
     }
 
-    LogicalOperator setAnonymousSink(
-        const std::string_view& testFileName,
-        SLTSinkFactory& sltSinkProvider,
-        const SystestQueryId& currentQueryNumberInTest,
-        const TypedLogicalOperator<AnonymousSinkLogicalOperator>& sinkOperator) const
+    /// Parses one rewritten statement into a plan under the given distributed id, and optimizes it.
+    [[nodiscard]] DistributedLogicalPlan compilePlan(const std::string& sql, const std::string& distributedId) const
```
adrianmichalke: same here, no compilation happens

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +206,274 @@ struct SystestBinder::Impl
...
     }
 
-    LogicalOperator setNamedSink(
-        SystestQueryBuilder& currentBuilder,
-        const std::string_view& testFileName,
-        SLTSinkFactory& sltSinkProvider,
-        const SystestQueryId& currentQueryNumberInTest,
-        const TypedLogicalOperator<SinkLogicalOperator>& sinkOperator) const
+    [[nodiscard]] SystestQuery
+    compileQuery(const RewrittenQuery& runnableQuery, const DiscoveredTestFile& testfile, const std::string& partKey) const
```
adrianmichalke: same here

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +206,274 @@ struct SystestBinder::Impl
...
+        return query;
     }
 
-    [[nodiscard]] SystestQueryBuilder bindExplainStatement(
-        const StatementBinder& binder,
-        const std::string_view& testFileName,
-        SLTSinkFactory& sltSinkProvider,
-        const ExplainStatement& statement) const
+    [[nodiscard]] SystestQuery
+    compileDifferential(const RewrittenDifferential& block, const DiscoveredTestFile& testfile, const std::string& partKey) const
```
adrianmichalke: and here

adrianmichalke (changes requested):

nes-systests/systest/private/Parser/TestFilePartition.hpp
yschroeder97: Done. Renamed to `TestFilePartition`.

nes-systests/systest/src/Parser/TestFilePartition.cpp
yschroeder97: Using `emplace_back()` now.

nes-systests/explain/EXPLAIN.test
yschroeder97: The anonymous sink now prints its type, `ANONYMOUS_SINK(FILE)`. 

I would avoid including the full config, since it would put the run's result path into the expectation, so it prints only the type.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: `addSourceOptions` now takes `{group, key, value}` and drops an option already set.

With this, a `FLUSH_INTERVAL_MS` that the test wrote is overwritten. 
A `SOCKET_HOST` or `SOCKET_PORT` on a source with attached data is rejected, as is 'FILE_PATH' with an ATTACH, because the source has to connect to the server that sends the data. 

Added tests for both.

nes-systests/systest/src/Runner/SystestRunner.cpp
yschroeder97: Added a function: remove a partner that has already finished; stop and discard one that is still running when its status arrives. 
A first half whose partner didn't start joins the active queries, so the run waits for its terminal state before exiting. 

Added runner test to validate.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: Function name tokens are now excluded from prefixing, as plugin types are.
Added a test (`SELECT ABS(x) FROM abs`) to validate. 

A field or alias that shares a source's name is still renamed. 

I wouldn't want to spend a lot of time now searching for all possible (very unlikely to hit) edge cases for something SQL namespaces will replace in the foreseeable future.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: The selection now happens before the rewrite: the **parsed** statements are filtered by query number, so a malformed query that the run doesn't select can't cause the whole file to fail. 
A part with only CREATEs left is skipped before the rewrite.

nes-systests/systest/src/Runner/SystestRunner.cpp
yschroeder97: Confirmed, and it also breaks names that were never prefixed.

**Before**: the check deleted the prefix string wherever it appeared in the plan (replaceAll(plan, "ORDERS_", "")). With a source orders_input in orders.test, the plan prints ORDERS_ORDERS_INPUT and both occurrences were removed: SOURCE(INPUT). 
A field orders_total was cut the same way, although fields are never prefixed.

**Now**: the rewriter records, for every name it prefixes, the spelling the test declared (ORDERS_ORDERS_INPUT -> ORDERS_INPUT). 
The EXPLAIN check replaces exactly those whole tokens, in one pass over the plan, and touches nothing else.

Result: `SOURCE(ORDERS_ORDERS_INPUT)` reads back as `SOURCE(ORDERS_INPUT); PROJECTION(fields: [ORDERS_TOTAL])` stays. 
The prefix string no longer exists in the model, only this mapping.

nes-systests/systest/src/Parser/TestFilePartition.cpp
yschroeder97: Partitioning does not start workers. 

The executor groups the queries by override and starts one worker per distinct override, configured at startup (`SystestExecutor.cpp`: `queriesByOverride`, `configCopy`, `runQueriesAtLocalWorker`), as current main does in the same file. 
The partition gives each alternative its own key (`KEY_C1`, `KEY_C2`), so that runs of one query number don't share catalog names or result files. 

AFAIK, the worker has no runtime reconfiguration, so I'm not sure what you mean here.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: `Exception` inherits from `std::exception`, so the catch block doesn't lose information. 

Loading a file can also throw standard exceptions (`std::stoul` on a malformed error code in an expectation, for one), so we need to catch those as well if we don't want to fail the whole invocation. 

Added a comment to explain.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: It is user input from the topology YAML. 

Added a check that runs when the file is read: 
The config parser rejects an empty placement list with a message before loading any test file. 

The preconditions in the binder stay, because after that they hold by construction. 

Added some tests to validate.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: Renamed the family to `bind...` because the class is the binder. The plan helper is `optimizedPlan`.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: Renamed, see above.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: Renamed, see above.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: Renamed, see above.

greenBene (approved):

nes-systests/systest/src/Parser/TestFilePartition.cpp
ls-1801: Yeah, I think we should do this in a follow-up. A lot of the current configuration should not be worker configuration, but should be moved to the optimizer instead. Then we would need to split different configs and only group the worker configs.

ls-1801 (approved):

nes-systests/systest/src/Parser/TestFilePartition.cpp
adrianmichalke: @yschroeder97 To clarify: We have worker configurations, such as the buffer size, which requires a worker restart, and others that can simply be changed at runtime (e.g., optimizer configurations). Since the number of worker partitions can increase significantly, we did not start a new worker for every partition, but only for those that would require a restart. For all others, we simply change the configuration on-the-fly at runtime between individual test cases.

I can create a new issue for that

nes-systests/systest/src/SystestBinder.cpp
```diff
@@ -503,577 +214,271 @@ struct SystestBinder::Impl
...
+                foundQueries.insert(numbers.begin(), numbers.end());
+            }
+            keepSelectedStatements(file.statements, selected);
+            /// When the selection drops every test case of a part, nothing runs, so nothing is rewritten or staged.
+            if (std::ranges::all_of(
+                    file.statements, [](const auto& statement) { return std::holds_alternative<CreateStatement>(statement); }))
+            {
+                continue;
+            }
+            const auto partKey = discoveryRoot.keyOf(testfile.file, index, parts.size());
```
adrianmichalke, line 275: Now something like --testLocations /tmp/example.test does not work anymore bc of namespace generation

nes-systests/systest/src/SystestBinder.cpp
adrianmichalke, line 1: Now source D_B_S in a.test and source S in a/b.test both become A_D_B_S

nes-systests/systest/src/Rewriter/NamePrefixer.cpp
adrianmichalke, line 1:
```
 CREATE LOGICAL SOURCE s(s UINT64 NOT NULL);
 ATTACH INLINE
 1

 CREATE SINK out(s UINT64 NOT NULL) TYPE File;

 SELECT s FROM s INTO out;
 ----
 1
```
The inline sink does not get renamed and we get: 
```
 ERROR 2003: cannot infer schema
 Expected field: S
 Unexpected field: FILE_S
```

adrianmichalke: It also broke EXPLAIN with the VISUAL format:
```
 CREATE LOGICAL SOURCE s(id UINT64 NOT NULL);
 ATTACH INLINE
 1

 CREATE SINK out(id UINT64 NOT NULL) TYPE File;

 EXPLAIN (LOGICAL) FORMAT VISUAL SELECT id FROM s INTO out;
 ----
 == Initial Logical Plan ==
        SINK(OUT)
             │
 PROJECTION(fields: [ID])
             │
        SOURCE(S)
```

adrianmichalke (changes requested):
There are a few (edge?) cases that dont currently work. 
Since a implementation of the coordinator namespace feature is planned, I wouldnt require that all edge cases get fixed, but maybe you could add proactive invariants to show that we are aware of them.

nes-systests/systest/src/Rewriter/NamePrefixer.cpp
yschroeder97: First one: fixed. The cause was the column, though, not the sink. `s` was prefixed everywhere it appeared, so the projection produced `<KEY>_S` while the sink expected `S`. Prefixing now rewrites only the grammar positions that hold a source, sink, or model name; everything else is untouched.  

Second: the issue is that the formatting depends on the width of the operator labels, which is computed from the prefixed names. As a result, restoring the names in the printed text cannot recover the layout. 
An EXPLAIN check with VISUAL now fails and is not supported. 

Comparable suites do the same.
CockroachDB and Materialize assert on the indented text format, and DuckDB never compares the visual rendering either, only regex or JSON.
I don't think it is a good idea at all to assert all the visual intricacies of the explain operator in an E2E testing framework.
IMO, it should compare results, not whether bars or boxes align.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: Fixed. A test file outside the discovery root uses its absolute path (/tmp/example.test → _D_TMP_D_EXAMPLE) instead of being rejected. 
An absolute path starts with the separator token, and a path under the root never does, so the two kinds of keys cannot collide. 
The same applies to --testDiscoverDirs pointing outside the root.

nes-systests/systest/src/SystestBinder.cpp
yschroeder97: Correct, and it isn't limited to subdirectories: the key encoding itself is injective (no two distinct inputs collide on the same result), but the join isn't injective, as you noted. 

Making the join injective would lengthen every prefixed name wherever it shows up (error output, sink file names), and the coordinator namespaces will replace prefixing anyway, so I left the encoding alone and added a check instead: the run now fails at load time with an error pointing at both files and the collision, and both declared names, instead of throwing a "source already exists" in the catalog.

adrianmichalke (approved):

PR #1997, opened 2026-09-09
Added Kafka Sink support as a plugin

ThanGeo:
This pull request adds support for Kafka Sinks in NebulaStream as a plugin plus rdkafka dependencies.

## Verifying this change
This change is tested by preliminary distributed bats testing (similar to MQTT #1781)

## Legacy
This feature was developed in the a demo fork and migrated/updated to main in this PR for the first time.

Demo Fork Link: https://github.com/nebulastream/nes-demo-process-intelligence/


ls-1801 (approved):

nes-plugins/Sinks/KafkaSink/tests/util/create_compose.sh
```diff
@@ -0,0 +1,197 @@
...
+  echo "Error: yq is required. Install with: sudo snap install yq"
+  exit 1
+fi
+
+WORKERS_FILE=$1
+
+if [ ! -f "$WORKERS_FILE" ]; then
+  echo "$WORKERS_FILE does not exist"
+  exit 1
+fi
```
greenBene, line 55: The 52-55 are rendundant to lines 39-42. I'd suggest removing lines 39-45

greenBene (changes requested):
Just a minor redundancy in one of the helper scripts :) The code looks good, and I'm really excited to get Kafka support into main! 🥳

nes-plugins/Sinks/KafkaSink/tests/util/create_compose.sh
ThanGeo: Done, thanks

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -0,0 +1,123 @@
...
+/// to the configured topic.
+class KafkaSink final : public Sink
+{
+public:
+    static constexpr std::string_view NAME = "Kafka";
+    static constexpr auto BACKPRESSURE_RETRY_INTERVAL = std::chrono::milliseconds(10);
+    /// Bounds how long start() waits for broker metadata and how long a queued message may
+    /// be retried before librdkafka reports it as failed via the delivery report callback.
+    /// Without a bound, an unreachable broker would hang start() indefinitely and a downed
+    /// broker mid-query would sit in librdkafka's internal retry queue rather than failing.
```
yschroeder97:
```suggestion
    /// Without a bound, start and execute may block indefinitely.
```

"Internal retry queue" is an unnecessary impl detail that we don't really need to know about here.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -0,0 +1,123 @@
...
+    void stop(PipelineExecutionContext&) override;
+    void execute(const TupleBuffer& inputTupleBuffer, PipelineExecutionContext&) override;
+
+    static DescriptorConfig::Config validateAndFormat(std::unordered_map<std::string, std::string> config);
+
+protected:
+    std::ostream& toString(std::ostream& os) const override;
+
+private:
+    /// Outcome of a single publish attempt. Mirrors NetworkSink's `SendResult`:
```
yschroeder97: If it is mirrored, we should think about factoring it out into a common, reusable type in `nes-sinks`.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -0,0 +1,123 @@
...
+    ///   Ok     - librdkafka accepted the message into its outbound queue.
+    ///   Full   - outbound queue is at capacity; caller should buffer & retry.
+    ///   Closed - non-recoverable prducer error (disconnected, protocol, etc.); caller should fail the query.
```
yschroeder97:
```suggestion
    ///   Closed - non-recoverable producer error (disconnected, protocol, etc.); caller should fail the query.
```

If its an unrecoverable error, call it `Err` or `Error` instead of `Closed`.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -0,0 +1,123 @@
...
+        Full,
+        Closed,
+    };
+
+    class DeliveryReportCallback final : public RdKafka::DeliveryReportCb
+    {
+    public:
+        void dr_cb(RdKafka::Message& message) override;
+        std::atomic<bool> failed{false};
+        std::string lastError;
```
yschroeder97: This is a data race: one thread could write to `lastError` string while the other reads it to build the exception. 

You could combine the two fields into an atomic error code and format it when you throw.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -0,0 +1,123 @@
...
+    std::string brokers;
+    std::string topic;
+    int32_t maxOutstandingMessages;
+
+    DeliveryReportCallback deliveryReportCallback;
+    std::unique_ptr<RdKafka::Producer> producer;
+    BackpressureHandler backpressureHandler;
+};
+
+/// Defines the names, (optional) default values, (optional) validation & config functions for all Kafka sink config parameters.
```
yschroeder97:
```suggestion
```

Stating the obvious.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -0,0 +1,123 @@
...
+    {
+    public:
+        void dr_cb(RdKafka::Message& message) override;
+        std::atomic<bool> failed{false};
+        std::string lastError;
+    };
+
+    ProduceResult tryProduce(const TupleBuffer& buffer);
+
+    std::string brokers;
```
yschroeder97: It's a bit confusing that "brokers" (plural) is stored in a single string. Can there be multiple? Are they addressed independently?

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include <Util/Logger/Formatter.hpp>
+#include <librdkafka/rdkafkacpp.h>
+#include <BackpressureChannel.hpp>
+#include <PipelineExecutionContext.hpp>
+
+namespace NES
+{
+
+/// A sink that produces the bytes of every incoming TupleBuffer as a single Kafka message
+/// to the configured topic.
```
yschroeder97, line 42: On engine shutdown, `stop` is skipped, so we may have outstanding messages in the outbound buffer. 

You could add a destructor that flushes with a bounded timeout, or at least log that there are X outstanding messages that will be lost.

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
```diff
@@ -0,0 +1,198 @@
...
+            case ProduceResult::Full: {
+                if (const auto emit = backpressureHandler.onFull(*currentBuffer, backpressureController))
+                {
+                    pec.repeatTask(*emit, BACKPRESSURE_RETRY_INTERVAL);
+                }
+                return;
+            }
+            case ProduceResult::Closed: {
+                [[maybe_unused]] auto droppedBuffer = backpressureHandler.onFull(*currentBuffer, backpressureController);
+                throw CannotOpenSink("KafkaSink connection to brokers {} was closed", brokers);
```
yschroeder97: All `tryProduce` manifest as `CannotOpenSink` (misleading!) and "connection closed" (also misleading), because the actual message could be "message too large" or "unknown topic", which does not have anything to do with the connection being closed. 

You should rather throw with the real error string and drop the closed state. 
As far as I read the docs, synchronous produce does not report broker disconnects anyways.

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
```diff
@@ -0,0 +1,198 @@
...
+void KafkaSink::start(PipelineExecutionContext&)
+{
+    NES_INFO("Opening KafkaSink at {} for topic {}.", brokers, topic);
+    auto conf = std::unique_ptr<RdKafka::Conf>(RdKafka::Conf::create(RdKafka::Conf::CONF_GLOBAL));
+
+    std::string error;
+    if (conf->set("bootstrap.servers", brokers, error) != RdKafka::Conf::CONF_OK
+        || conf->set("dr_cb", &deliveryReportCallback, error) != RdKafka::Conf::CONF_OK
+        || conf->set("queue.buffering.max.messages", std::to_string(maxOutstandingMessages), error) != RdKafka::Conf::CONF_OK
+        || conf->set("message.timeout.ms", std::to_string(BROKER_TIMEOUT_IN_MILLISECONDS), error) != RdKafka::Conf::CONF_OK)
```
yschroeder97: I don't think the startup metadata timeout and delivery timeout should be identical. 

I'd suggest separating them and making the latter configurable. 
We could have a perfectly healthy, but busy broker that will terminate the query because the 5s are not configurable by the user.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -0,0 +1,123 @@
...
+    static inline const DescriptorConfig::ConfigParameter<std::string> TOPIC{
+        "TOPIC",
+        std::nullopt,
+        [](const std::unordered_map<std::string, std::string>& config) { return DescriptorConfig::tryGet(TOPIC, config); }};
+
+    /// Maximum number of messages librdkafka may hold in its outbound queue (queue.buffering.max.messages).
+    /// Once reached, produce() fails with ERR__QUEUE_FULL and the sink retries via the BackpressureHandler.
+    static inline const DescriptorConfig::ConfigParameter<int32_t> MAX_OUTSTANDING_MESSAGES{
+        "MAX_OUTSTANDING_MESSAGES",
+        100000,
```
yschroeder97, line 104: You should add range validation to this config param.

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
```diff
@@ -0,0 +1,198 @@
...
+    }
+    if (err != RdKafka::ERR_NO_ERROR)
+    {
+        NES_ERROR("KafkaSink produce to topic {} failed: {}", topic, RdKafka::err2str(err));
+        return ProduceResult::Closed;
+    }
+    return ProduceResult::Ok;
+}
+
+void KafkaSink::execute(const TupleBuffer& inputTupleBuffer, PipelineExecutionContext& pec)
```
yschroeder97, line 137: We now have three near-identical execute and stop loops: Kafka, MQTT, and Network. 

They are starting to drift apart. 

At least add an issue to provide a shared base with a virtual send hook or similar.

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
```diff
@@ -0,0 +1,198 @@
...
+{
+    PRECONDITION(producer, "KafkaSink producer is not initialized");
+    PRECONDITION(inputTupleBuffer, "Invalid input buffer in KafkaSink.");
+
+    if (deliveryReportCallback.failed)
+    {
+        throw CannotOpenSink("KafkaSink delivery to brokers {} failed: {}", brokers, deliveryReportCallback.lastError);
+    }
+
+    auto currentBuffer = std::optional(inputTupleBuffer);
```
yschroeder97: nit: prefer `{}` for ctors instead of `()`.

nes-plugins/Sinks/KafkaSink/tests/distributed.bats
```diff
@@ -0,0 +1,155 @@
...
+        4 AS "SINK"."MAX_OUTSTANDING_MESSAGES",
+        '2' AS "SINK"."BACKPRESSURE_LOWER_THRESHOLD",
+        '4' AS "SINK"."BACKPRESSURE_UPPER_THRESHOLD"
+    )
+EOF
+)"
+  assert_success
+  wait_until_status tests/good/example.yaml "Stopped" $output --require-healthy "worker-1"
+
+  assert_file_line_count results.csv 1000 --ignore-empty-lines
```
yschroeder97: The consumer writes results.csv in the background, so the count can lag the Stopped status. 
Wrap this in wait_until, like the checks at lines 39 and 103.

nes-plugins/Sinks/KafkaSink/tests/distributed.bats
```diff
@@ -0,0 +1,155 @@
...
+        'CSV' AS "SINK"."OUTPUT_FORMAT",
+        -- provoke backpressure
+        4 AS "SINK"."MAX_OUTSTANDING_MESSAGES",
+        '2' AS "SINK"."BACKPRESSURE_LOWER_THRESHOLD",
+        '4' AS "SINK"."BACKPRESSURE_UPPER_THRESHOLD"
+    )
+EOF
+)"
+  assert_success
+  wait_until_status tests/good/example.yaml "Stopped" $output --require-healthy "worker-1"
```
yschroeder97: This test was set up from single-worker-with-4k-buffers.yaml but polls status with example.yaml. It only works because both declare the same worker. 

Pass the topology the cluster was started with.

nes-plugins/Sinks/KafkaSink/tests/distributed.bats
```diff
@@ -0,0 +1,155 @@
...
+  setup_distributed tests/good/example.yaml
+
+  docker_kafka_consume "kafka-sink-test" results.csv
+  run docker_nes_cli -t tests/good/example.yaml start
+  assert_success
+
+  # wait until the query stops
+  wait_until_status tests/good/example.yaml "Stopped" $output
+
+  wait_until awk 'NF { count++ } END { exit(count != 9) }' results.csv
```
yschroeder97, line 40: `SEQUENCE UINT64 1 10 1` is end-exclusive, so this yields 9 tuples, not (1, 1000].

nes-plugins/Sinks/KafkaSink/tests/util/create_compose.sh
```diff
@@ -0,0 +1,192 @@
...
+# Single-node KRaft broker: no zookeeper, combined broker+controller process.
+cat <<EOF
+services:
+  kafka-broker:
+    image: apache/kafka:3.8.0
+    stop_grace_period: 0s
+    environment:
+      KAFKA_NODE_ID: 1
+      KAFKA_PROCESS_ROLES: broker,controller
+      KAFKA_LISTENERS: PLAINTEXT://0.0.0.0:9092,CONTROLLER://0.0.0.0:9093
```
yschroeder97, line 71: The MQTT compose throttles its broker with cpus: 0.01, so the backpressure test fills the queue. 

Without a throttle here, a fast broker can keep the deque below 4 and the "Backpressure acquired" grep at line 104 fails.

nes-plugins/Sinks/KafkaSink/tests/util/create_compose.sh
```diff
@@ -0,0 +1,192 @@
...
+      KAFKA_CONTROLLER_LISTENER_NAMES: CONTROLLER
+      KAFKA_LISTENER_SECURITY_PROTOCOL_MAP: CONTROLLER:PLAINTEXT,PLAINTEXT:PLAINTEXT
+      KAFKA_CONTROLLER_QUORUM_VOTERS: 1@kafka-broker:9093
+      KAFKA_INTER_BROKER_LISTENER_NAME: PLAINTEXT
+      KAFKA_OFFSETS_TOPIC_REPLICATION_FACTOR: 1
+      CLUSTER_ID: nes-kafka-sink-test
+    healthcheck:
+      test: ["CMD", "/opt/kafka/bin/kafka-topics.sh", "--bootstrap-server", "localhost:9092", "--list"]
+      interval: 10s
+      start_interval: 10s
```
yschroeder97: A 10 s start_interval with a JVM probe and JOBS 1 adds several idle seconds per test. 

MQTT uses `start_interval: 100ms, interval: 1s` and `JOBS 4`. Consider a TCP probe on 9092 with a 1 s interval.

yschroeder97 (changes requested):
I think it's nearly done. 

The main issues to be handled are a data race in the callback and confusing error messages from the `tryProduce` method.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
yschroeder97: --> E.g., 0 means infinite to rdkafka.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
ls-1801: Most of the sinks will not flush on engine stop, same as the network channel. We have to draw the line at some point otherwise we can never get a clean stop. If you need the flushing behavior you have to stop the query first, wait until it's stopped and than shutdown the server.

nes-plugins/Sinks/KafkaSink/tests/util/create_compose.sh
ls-1801: Kafka is just way slower, throttling it made the ci super flakey as Kafka wouldn't come up within minutes

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
ThanGeo: I switched to using the same enum used in the NetworkSink. Changed it also for the MQTTSink.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
ThanGeo: Redundant now since we're reusing the same enum as NetworkSink

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
ThanGeo: Done

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
ThanGeo: changed it to ```bootstrapServers``` as this is the exact same name used in rdKafka internally. Also reflects the ```conf->set("bootstrap.servers")``` usage for when the field is actually passed to rdKafka

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
ThanGeo: I can confirm that the other sinks also don't flush. Maybe we could discuss the desired behavior and open an issue for this, then change all sinks to behave similarly wrt flushing and shutdown

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
ThanGeo: Changed it so that it instead throws an error, but the caller of the method was in a switch clause. I marked that as an INVARIANT (like unreachable), since it is never supposed to reach there. I did the same for the MQTTSink. Let me know if you agree/disagree

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
ThanGeo: Done

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
ThanGeo: Done

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
ThanGeo: Yeah this should probably be a separate issue but if you take a closer look, they aren't exactly the same across sinks. So maybe it makes sense to have local implementations of execute even though they are called the same but behave slightly different?

nes-plugins/Sinks/KafkaSink/tests/util/create_compose.sh
ThanGeo: Yeah it was a pain to make the kafka broker work on the CI. Haven't changed anything based on this comment alone yet. Not sure how to proceed

nes-plugins/Sinks/KafkaSink/tests/util/create_compose.sh
ThanGeo: I changed it to 1s. If it fails i'll revert to 10s.

nes-plugins/Sinks/MQTTSink/MQTTSink.hpp
```diff
@@ -127,11 +132,20 @@ struct ConfigParametersMQTTSink
...
         "MAX_OUTSTANDING_MESSAGES",
         128,
-        [](const std::unordered_map<std::string, std::string>& config)
-        { return DescriptorConfig::tryGet(MAX_OUTSTANDING_MESSAGES, config); }};
+        [](const std::unordered_map<std::string, std::string>& config) -> std::optional<int32_t>
+        {
+            auto value = DescriptorConfig::tryGet(MAX_OUTSTANDING_MESSAGES, config);
+            if (!value || value.value() <= 0)
+            {
+                NES_ERROR("MQTTSink: MAX_OUTSTANDING_MESSAGES must be positive, got {}.", value.value_or(-1));
```
yschroeder97: Remove `MQTTSink`, the logging framework provides this context already.

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
```diff
@@ -124,45 +125,45 @@ KafkaSink::ProduceResult KafkaSink::tryProduce(const TupleBuffer& buffer)
...
     if (err == RdKafka::ERR__QUEUE_FULL)
     {
-        return ProduceResult::Full;
+        return SendResult::Full;
     }
     if (err != RdKafka::ERR_NO_ERROR)
     {
-        NES_ERROR("KafkaSink produce to topic {} failed: {}", topic, RdKafka::err2str(err));
-        return ProduceResult::Closed;
+        throw CannotOpenSink("KafkaSink produce to topic {} failed: {}", topic, RdKafka::err2str(err));
```
yschroeder97, line 132: Could you add an issue to add a proper exception type for errors during sending in the sink? 

This really has nothing to do with "opening".

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
```diff
@@ -124,45 +125,45 @@ KafkaSink::ProduceResult KafkaSink::tryProduce(const TupleBuffer& buffer)
...
                 }
                 return;
             }
-            case ProduceResult::Closed: {
-                [[maybe_unused]] auto droppedBuffer = backpressureHandler.onFull(*currentBuffer, backpressureController);
-                throw CannotOpenSink("KafkaSink connection to brokers {} was closed", brokers);
+            case SendResult::Closed: {
+                /// tryProduce() can only return SendResult::Full or SendResult::Ok. If this point is reached, it means something went wrong.
+                INVARIANT(false, "KafkaSink::tryProduce unexpectedly returned SendResult::Closed");
+                throw CannotOpenSink("KafkaSink::tryProduce unexpectedly returned SendResult::Closed");
```
yschroeder97: Use either the INVARIANT or throw, not both. 

Whenever a programming bug is guarded (as here), use INVARIANT/PRECONDITION.

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -107,14 +98,40 @@ struct ConfigParametersKafkaSink
...
         "MAX_OUTSTANDING_MESSAGES",
         100000,
-        [](const std::unordered_map<std::string, std::string>& config)
-        { return DescriptorConfig::tryGet(MAX_OUTSTANDING_MESSAGES, config); }};
+        [](const std::unordered_map<std::string, std::string>& config) -> std::optional<int32_t>
+        {
+            auto value = DescriptorConfig::tryGet(MAX_OUTSTANDING_MESSAGES, config);
+            if (!value || value.value() <= 0)
+            {
+                NES_ERROR("KafkaSink: MAX_OUTSTANDING_MESSAGES must be positive, got {}.", value.value_or(-1));
```
yschroeder97: Remove "KafkaSink".

nes-plugins/Sinks/KafkaSink/KafkaSink.hpp
```diff
@@ -107,14 +98,40 @@ struct ConfigParametersKafkaSink
...
+    /// Must be positive: librdkafka treats 0 as "infinite", which could block a query forever.
+    static inline const DescriptorConfig::ConfigParameter<int32_t> DELIVERY_TIMEOUT_MS{
+        "DELIVERY_TIMEOUT_MS",
+        5000,
+        [](const std::unordered_map<std::string, std::string>& config) -> std::optional<int32_t>
+        {
+            auto value = DescriptorConfig::tryGet(DELIVERY_TIMEOUT_MS, config);
+            if (!value || value.value() <= 0)
+            {
+                NES_ERROR("KafkaSink: DELIVERY_TIMEOUT_MS must be positive, got {}.", value.value_or(-1));
```
yschroeder97: Same here, spdlog provides this.

nes-plugins/Sinks/MQTTSink/MQTTSink.cpp
```diff
@@ -132,38 +138,43 @@ MQTTSink::PublishResult MQTTSink::tryPublish(const TupleBuffer& buffer)
...
                 }
                 return;
             }
-            case PublishResult::Closed: {
-                [[maybe_unused]] auto droppedBuffer = backpressureHandler.onFull(*currentBuffer, backpressureController);
-                throw CannotOpenSink("MQTTSink connection to broker {} was closed", serverURI);
+            case SendResult::Closed: {
+                /// tryPublish() can only return SendResult::Full or SendResult::Ok. If this point is reached, it means something went wrong.
+                INVARIANT(false, "MQTTSink::tryPublish unexpectedly returned SendResult::Closed");
+                throw CannotOpenSink("MQTTSink::tryPublish unexpectedly returned SendResult::Closed");
```
yschroeder97: Again, remove the exception throw.

yschroeder97 (changes requested):
Few minor things to cleanup.

greenBene (approved):

nes-plugins/Sinks/KafkaSink/KafkaSink.cpp
ThanGeo: Done

yschroeder97 (approved):

PR #2005, opened 2026-09-10
build: robust vcpkg clone and python3 fallback for the dependency hash

yschroeder97:

Two fixes by @keyseven123 that showed up on the coordinator stack because they change the dependency hash and therefore rebuild the dev images. They are independent of the coordinator work and useful to every PR that touches dependencies.

- `docker/dependency/Dependency.dockerfile`: the vcpkg clone now runs with `http.version=HTTP/1.1` and `GIT_TERMINAL_PROMPT=0`. The dependency image needs a full clone of the vcpkg repository. Under load, GitHub aborts large HTTP/2 clones with `curl 92 HTTP/2 stream was not closed cleanly`, and HTTP/1.1 avoids that. The disabled prompt makes git fail fast instead of hanging on an auth challenge in a Docker build with no terminal. Ported onto main's Dockerfile; the original branch had this as two commits, first a retry loop and then this replacement.
- `scripts/check_rust_dependency_policy.py`: falls back to an installed `python3.11` or newer when the default `python3` has no `tomllib`. The script runs on every CMake configure through the dependency hash and in the local image install script, so a host whose default Python is older (Ubuntu 22.04 ships 3.10) could not configure at all. An environment variable guards against a re-exec loop. CI runners and the dev container already run Python 3.12, so this only affects developer hosts.

## Verifying this change

This change is tested by
- the dev image rebuild that this PR triggers itself, since the dockerfile is part of the dependency hash. The `get-dev-images` jobs exercise the changed clone on every image variant.
- the same two changes on the coordinator branches on 2026-09-02, where every `Build Dev Images` job passed after they were added.
- the policy script on a host with Python 3.14, unchanged behaviour when `tomllib` is present.

- Dependencies: the dependency image build only. No dependency is added or upgraded.
- Build: the dependency hash script, and with it CMake configuration on hosts with an old default Python.

No issue. The failures were observed on the coordinator stack CI.

### DIFF
--- a/docker/dependency/Dependency.dockerfile
+++ b/docker/dependency/Dependency.dockerfile
@@ -83,6 +83,8 @@ RUN --mount=type=secret,id=VCPKG_CACHE_ACCESS_KEY \
     if [ "$STDLIB" = "libcxx" ]; then VCPKG_STDLIB="libcxx"; else VCPKG_STDLIB="local"; fi; \
     \
     cd /vcpkg_input; \
+    # An image build has no terminal, so git would wait on an auth challenge until the build times out.
+    export GIT_TERMINAL_PROMPT=0; \
     git clone https://github.com/microsoft/vcpkg.git vcpkg_repository; \
     git -C vcpkg_repository checkout "$(grep -o "\"builtin-baseline\": \"[0-9a-f]*\"" vcpkg.json | cut -d\" -f4)"; \
     ./vcpkg_repository/bootstrap-vcpkg.sh --disableMetrics; \
--- a/scripts/check_rust_dependency_policy.py
+++ b/scripts/check_rust_dependency_policy.py
@@ -25,8 +25,10 @@ from typing import Any, Iterable
 try:
     import tomllib
 except ModuleNotFoundError:
+    version = ".".join(str(part) for part in sys.version_info[:3])
     print(
-        "Python 3.11 or newer is required to validate Rust workspace dependencies.",
+        f"Python 3.11 or newer is required to validate Rust workspace dependencies, "
+        f"but {sys.executable} is {version}.",
         file=sys.stderr,
     )
     sys.exit(2)
--- a/cmake/EnableRust.cmake
+++ b/cmake/EnableRust.cmake
@@ -10,6 +10,20 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# The dependency image vendors the workspace from the root manifest alone and then builds offline.
+# That vendoring never sees a member that declares its own dependency, so the policy has to hold before a build starts.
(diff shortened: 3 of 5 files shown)


docker/dependency/Dependency.dockerfile
```diff
@@ -83,7 +83,8 @@ RUN --mount=type=secret,id=VCPKG_CACHE_ACCESS_KEY \
     if [ "$STDLIB" = "libcxx" ]; then VCPKG_STDLIB="libcxx"; else VCPKG_STDLIB="local"; fi; \
     \
     cd /vcpkg_input; \
-    git clone https://github.com/microsoft/vcpkg.git vcpkg_repository; \
+    export GIT_TERMINAL_PROMPT=0; \
+    git -c http.version=HTTP/1.1 clone https://github.com/microsoft/vcpkg.git vcpkg_repository; \
```
ls-1801: I don't mind the git terminal prompt = 0, but downgrading the HTTP version should not be the solution. We currently don't run into this issue anymore, where the unauthenticated git clone is rate-limited. We also reduced the number of git clones in the other workflows

scripts/check_rust_dependency_policy.py
```diff
@@ -18,17 +18,32 @@
...
 from typing import Any, Iterable
 
 try:
     import tomllib
 except ModuleNotFoundError:
-    print(
-        "Python 3.11 or newer is required to validate Rust workspace dependencies.",
-        file=sys.stderr,
-    )
+    interpreter = None
```
ls-1801: I'm not a big fan of this change; I think the user should just fix their Python version, and we shouldn't try to fix it ourselves. We could extend the error message to display the current Python version used, but I think the error that we require 3.11 is good enough

ls-1801 (changes requested):
i would

scripts/check_rust_dependency_policy.py
greenBene: I'm here with Lukas. While we automate away some potential issues, we are hiding the behaviour behind recursive Python script calls. A clear error message should imo be good enough to point the user towards requiring a different Python version

greenBene (changes requested):

scripts/check_rust_dependency_policy.py
yschroeder97: I as a user would be happy that the tool picks whatever it needs, especially when it even is available. 
You just said that you are not a fan of the change, not arguing why not.

ls-1801: My issue is that the fix that Nils used worked on his machine. In general, there are so many ways a Python installation can be misconfigured that this is a battle we will always lose.

yschroeder97:
Closing: no longer required, the changes are superseded on main.

PR #2007, opened 2026-09-11
ci: unify Docker login with retries

ls-1801:

The [x64 libstdcxx Debug CI job for #1997](https://github.com/nebulastream/nebulastream/actions/runs/34578364023/job/103199348087?pr=1997) failed in `prepare-github` when Docker Hub authentication returned `connection reset by peer`, before the build or tests could start.

CI currently has several ways to log in to Docker: a shell command in `prepare-github`, another login inside `run-in-container`, and six uses of `docker/login-action`. This change consolidates authentication into one local `docker-login` action, inspired by [docker/login-action](https://github.com/docker/login-action), with retries added.

- Make up to five login attempts with exponential backoff of 5, 10, 20, and 40 seconds, and fail if all attempts fail.
- Pass the password through stdin and preserve automatic post-job logout, including after failures, using the original Docker config location.
- Replace all six external login actions and the login in `prepare-github` with the shared action.
- Remove login and credential forwarding from `run-in-container`, which launches an already downloaded image with `--pull=never`.
- Add checkout to the executable-image merge job so it can load the local action.

## Verifying this change

Validated YAML and local action inputs, checkout ordering, JavaScript and Bash syntax, and `git diff --check`. Local mocked checks covered successful login, retries, exhausted retries, missing credentials, and post-job logout. GitHub Actions execution is still pending.

GitHub Actions CI setup, Docker image builds/publishing, and container-based build/test steps.


.github/steps/docker-login/main.cjs
```diff
@@ -0,0 +1,51 @@
+const { spawnSync } = require('node:child_process');
+const { appendFileSync } = require('node:fs');
+const { homedir } = require('node:os');
+const { join, resolve } = require('node:path');
+const { setTimeout: sleep } = require('node:timers/promises');
+
+async function login() {
+  const registry = process.env.INPUT_REGISTRY || 'docker.io';
```
greenBene, line 8: Isn't the default already defined in the action.yml?

.github/steps/docker-login/main.cjs
```diff
@@ -0,0 +1,51 @@
...
+  if (!username || !password) {
+    throw new Error('Docker registry username and password are required');
+  }
+
+  const configDir = resolve(process.env.DOCKER_CONFIG || join(homedir(), '.docker'));
+  // Save the original config location before attempting login so post-job cleanup
+  // also runs after a failed login and is unaffected by later DOCKER_CONFIG changes.
+  appendFileSync(process.env.GITHUB_STATE, `docker_login=${JSON.stringify({ registry, configDir })}\n`);
+
+  const maxAttempts = 5;
```
greenBene, line 20: Should we move the maxAttempts variable with a default value to the actions.yml? I think it would be a more adaptable interface, it's no major concern :) Do as you think is best!

greenBene (changes requested):
Just two minor nits :) Otherwise the Pr looks good and should make our CI more stable 🥳

.github/steps/docker-login/main.cjs
ls-1801: yes it is

.github/steps/docker-login/main.cjs
ls-1801: i moved it into the yaml, but the js still has to verify that it is a valid integer

greenBene (approved):

Issue #1833, opened 2026-07-31
Make SQL CHAR semantics architecture-independent on ARM64 and x86-64


ls-1801:
## Problem

`CHAR` behavior currently depends on the target architecture's C++ plain-`char` signedness. The same SQL query produces different results on x86-64 and ARM64 when a byte with its high bit set is cast to `CHAR` and then widened.

This is a runtime/SQL correctness issue independent of any particular operator: a logical NES data type should have identical semantics on every supported architecture.

## Reproducer

The following system test makes the representation observable without writing non-UTF-8 bytes to the result file:

```text
CREATE LOGICAL SOURCE stream(ascii_max UINT8 NOT NULL, high_bit UINT8 NOT NULL, max_byte UINT8 NOT NULL);
CREATE PHYSICAL SOURCE FOR stream TYPE File;
ATTACH INLINE
127,128,255

CREATE SINK sink(ascii_char INT16 NOT NULL, high_bit_char INT16 NOT NULL, max_byte_char INT16 NOT NULL) TYPE File;

SELECT
    CASTTOTYPE(CASTTOTYPE(ascii_max AS CHAR) AS INT16) AS ascii_char,
    CASTTOTYPE(CASTTOTYPE(high_bit AS CHAR) AS INT16) AS high_bit_char,
    CASTTOTYPE(CASTTOTYPE(max_byte AS CHAR) AS INT16) AS max_byte_char
FROM stream INTO sink;
----
127,-128,-1
```

Observed results:

| Architecture | Result |
| --- | --- |
| x86-64 | `127,-128,-1` |
| ARM64 | `127,128,255` |

ARM64 therefore treats the high-bit values as unsigned while x86-64 treats them as signed.

## Expected behavior

NES should define one architecture-independent representation and conversion contract for `CHAR`. Nautilus currently models `CHAR` as an 8-bit value, so if the intended contract is signed `i8`, both architectures should produce `127,-128,-1`.

## Acceptance criteria

- Define whether NES `CHAR` is signed or unsigned.
- Use a fixed-width type for its runtime representation and conversions instead of relying on host plain-`char` signedness.
- Make the reproducer pass with the same result on x86-64 and ARM64 in interpreter and compiler execution modes.
- Audit memory reads/writes and casts involving `CHAR` for other uses of host-dependent plain `char`.

PR #2012, opened 2026-09-14
fix: link UBSan runtime in undefined-sanitizer Rust cxx bridge builds

oczerwia-2010:
The undefined sanitizer was not defined within rust.

Fixes the failed compilation at NES Nightly #1833


ls-1801 (changes requested):
Did you test that, I would be very surprised if that works. AFAIK there is no ubsan for rust. We would need to supply the correct linker flags to rust.

ls-1801 (approved):

ls-1801:
I tested it here: 
https://github.com/nebulastream/nebulastream/actions/runs/35141552360

greenBene (approved):

PR #2015, opened 2026-09-15
perf(formatters): select CSV delimiters during tracing

adrianmichalke:
Moves csv delimiter decision into tracing.

IR operations before: 4,785,739
After: 4,296,915


LeoRaasch (approved):
Good catch!

yschroeder97 (approved):

PR #2017, opened 2026-09-15
Perf/handler lookup no copy

adrianmichalke:
Do not copy the handlers during `getGlobalOperatorHandlerProxy(...)` lookup. Its called quite often during query execution. For a full systest run ~10'000 times.

Also the old impl with handlers[index] inserted missing keys into a temporary copy. Added a regression test.


ls-1801 (approved):
good catch

ThanGeo (approved):

Issue #30, opened 2024-06-20
[SUB-EPIC] Moving OperatorHandlers into the NautilusExecutablePipelineStage


ls-1801:
### Description

Currently, the OperatorHandlers are created during lowering to physical operators and passed to the lowering to nautilus.

After discussing this with @PhilippGrulich and @keyseven123, we concluded that operator handlers should be an implementation detail for the Nautilus operator and should not be exposed to other system components.

We keep operator handlers at the physical level to give all relevant operators access to the global operator state (accessed via an OperatorHandler) 

OperatorHandlers were previously accessed through the PipelineExecutionContext (PEC). This issue tracks the integration of the access to operator handlers via static memory access within the generated operator.

During Lowering, we emit the address of the heap-allocated operator handler to the tracing context. We guarantee that the pointer is always valid by:
1. Allocating on the Heap
2. Managing the lifetime of the OperatorHandler through a shared pointer in the the owinging pipeline.

### Use case/motivation

We want to reduce the interface between runtime and query lowering.
OperatorHandlers have previously been exposed to the runtime, especially due to the runtime reconfigurability interface. However, we will pursue actions required for operator state serialization using other mechanisms, possibly controlled by the owning pipeline.

When creating the PEC the runtime does not need information about operator handlers, thus the context can be easily constructed with local information of the worker thread.

### Related issues

#27

### Are you willing to submit a PR?

- [X] Yes I am willing to submit a PR!

### Code of Conduct

- [ ] I agree to follow this project's [Code of Conduct](https://github.com/nebulastream/nebulastream/blob/master/CODE_OF_CONDUCT.md)

```[tasklist]
### Tasks
- [ ] https://github.com/nebulastream/nebulastream-public/issues/78
```

alepping:
@keyseven123 and @ls-1801 is this still relevant?

PR #2023, opened 2026-09-16
Query Compilation Cache & Plan Serialization

adrianmichalke:
Adds query-compilation caching through Nautilus. Cache hits reuse compiled machine code and skip tracing entirely.

To make physical operators cacheable, we introduce runtime bindings ro replace replace process-local addresses.

 - 100% warm-cache hit rate
 - 73% lower full-systest wall-clock time with a warm cache
 
Also addresses #30


PhilippGrulich:
looks quite interesting, would be cool if we can have the nautilus part on main. Let's maybe have a chat in the next days.

vcpkg/vcpkg-registry/ports/nautilus/0005-object-code-cache.patch
```diff
@@ -1927,10 +1955,16 @@ index c5cda3fe..120e97d4 100644
...
 +	builder.setJITTargetMachineBuilder(std::move(*tmBuilderOrError));
- 	if (!targetTriple.isAArch64()) {
--		jitBuilder.setObjectLinkingLayerCreator(objectLinkingLayerCreator);
++	if (targetTriple.isAArch64()) {
++		builder.setObjectLinkingLayerCreator(
++		    [](llvm::orc::ExecutionSession& session) -> std::unique_ptr<llvm::orc::ObjectLayer> {
++			    return std::make_unique<llvm::orc::ObjectLinkingLayer>(session);
++		    });
++	} else {
 +		builder.setObjectLinkingLayerCreator(objectLinkingLayerCreator);
```
ls-1801: @PhilippGrulich I think we wanted to do that anyways, use the new LLJit instead of the legacy linker

nes-physical-operators/include/PhysicalPlan.hpp
```diff
@@ -26,6 +27,17 @@
 
 namespace NES
 {
+class PhysicalPlanSignature final
```
ls-1801, line 30: Is there something wrong with using NESStrongStringType here?

nes-physical-operators/src/Functions/ConstantValueVariableSizePhysicalFunction.cpp
```diff
@@ -33,11 +34,13 @@ ConstantValueVariableSizePhysicalFunction::ConstantValueVariableSizePhysicalFunc
...
 {
-    /// NOLINTBEGIN(cppcoreguidelines-pro-type-const-cast) - VariableSizedData requires non-const pointer but data is not modified
-    VariableSizedData result(const_cast<int8_t*>(data.data()), data.size());
-    /// NOLINTEND(cppcoreguidelines-pro-type-const-cast) - VariableSizedData requires non-const pointer but data is not modified
+    const auto result = arena.allocateVariableSizedData(nautilus::val<uint64_t>(data.size()));
+    for (nautilus::static_val<size_t> i = 0; i < data.size(); ++i)
+    {
+        VarVal{nautilus::val<int8_t>(data[i])}.writeToMemory(result.getContent() + nautilus::val<uint64_t>(i));
+    }
     return result;
```
ls-1801, line 44: This should definitely also use the runtime bindings. Currently, this allocates everytime we use a constant string in the queries and writes that string into the memory.

Assuming the constant string is large and we have a lot of tuples, this will add up quite a bit until we can release the arena again.

ls-1801 (changes requested):
I worry this adds complexity for little gain. The potential to screw up the cache key is quite large; every change to every logical operator needs to ensure that the cache key is correctly updated.

The risk of using addresses in the Nautilus code that would fail to recompile is also high, and nothing would catch it, which could lead to hours of debugging a random crash. I think the runtime bindings are the correct approach.

I don't see any blocking issues that would prevent this from getting merged, but I don't see the value of the compilation cache, at least not at this point. We should focus on improving the compile time instead of masking it. I would add caching later, maybe with a dedicated compilation service. Regarding the Fault-Tolerance use case, I think recovering the data would be the more time-consuming part anyway (at least for large, long-running queries), so compiling for 1 second wouldn't be the end of the world.

So if you think we need this, I will support it. 

I would strip out all systest changes tho, and I am not too happy with the Python code used to warm up the cache. Can we integrate that into the endless test-mode CI run?

PR #2027, opened 2026-09-17
refactor(descriptors): construct source and sink descriptors without a catalog

### LINKED ISSUE #2031: Construct source and sink descriptors without a catalog
Preparation for #1324 (#1577).

SourceCatalog and SinkCatalog are the only way to obtain a descriptor and run the type and format validation inline. Move that validation into factories on SourceDescriptor and SinkDescriptor so a descriptor can be built without a catalog, give named sinks a SinkId, and let the catalogs delegate. The coordinator later constructs descriptors from the Rust catalog through the same factories. Implemented in #2027.

yschroeder97:
Preparation for the coordinator bridge (#1577).

Until now, the catalogs were the only way to get a descriptor, so we ran validation inline.
The validation moves into descriptor factories, so a descriptor can be built without a catalog.
- `SourceDescriptor::create` validates the source type and the input-formatter config.
- `SinkDescriptor::createNamed` and `createAnonymous` validate the sink config; a named sink gets a `SinkId` like an anonymous one.
- `LogicalSource` has a public constructor.
- Tests that only need a descriptor build it directly.

**This is required because the coordinator (in Rust) owns the catalog, whereas descriptors and their configuration/validation should stay in their respective C++ libs.**
We still want to test the validation/registration logic; therefore, we need C++ unit tests to construct descriptors without the Rust catalog.

## Verifying this change
- The existing tests of the catalogs, the binder, the planning test, and the four components whose tests changed.

- Sources, Sinks, Identifiers.

## Documentation
- None required.

This PR closes #2031.


Artraxon (approved):

greenBene (approved):
Huge fan of these changes! :) The code becomes, imo, way easier to reason about.

PR #2028, opened 2026-09-17
refactor(inference): validate a model in RegisteredModel::create

### LINKED ISSUE #2032: Validate a model in RegisteredModel::create
Preparation for #1324 (#1577).

ModelCatalog::registerModel imports a model and checks the declared schema against the tensor shapes before storing it. Move that validation into a factory on RegisteredModel and the type out of the catalog header, so a validated model can be built without a ModelCatalog. The coordinator later serves RegisteredModel from the Rust catalog. Implemented in #2028.

yschroeder97:
Preparation for the coordinator bridge (#1577).
The import and schema checks that `ModelCatalog::registerModel` ran move into `RegisteredModel::create`, and the type moves out of the catalog header, so a validated model can be built without a catalog.
The catalog only stores what the factory returns.

**This is required because the catalog is owned entirely by the coordinator (in Rust), whereas a registered model and its configuration/validation should (for now) stay within the logical operators lib.**
We still want to test the validation/registration logic, therefore we need C++ unit tests to be able to construct a registered model without the Rust catalog. 

## Verifying this change
- `ModelCatalogTest` becomes `ModelValidationTest` with the same cases through the factory.
- `InferModelLogicalOperatorTest` builds its models and sources directly.

- Inference, LogicalOperators.

## Documentation
- None required.

This PR closes #2032.


nes-inference/include/ModelCatalog.hpp
```diff
@@ -29,71 +29,10 @@
 namespace NES
 {
 
-class ModelCatalog;
-
 /// Catalog-side field schema for a model — ordered list of name + DataType pairs.
 /// Ordered because the runtime indexes inputs/outputs positionally by tensor slot.
 using ModelFieldList = Schema<UnqualifiedUnboundField, Ordered>;
```
gaturchenko, line 34: Does this alias have any usages left anywhere? If not, then I'd suggest removing it altogether

nes-inference/src/Model.cpp
```diff
@@ -156,21 +160,77 @@ Reflected Reflector<RegisteredModel>::operator()(const RegisteredModel& model, c
...
+            {
+                throw CannotLoadModel(
+                    "Model '{}' {}: declared {} field(s) but tensor has {} element(s)", name, role, fields.size(), elementCount);
+            }
+        }
+    };
+    validateSide(schema.inputs, imported->getInputShape(), "input");
+    validateSide(schema.outputs, imported->getOutputShape(), "output");
+
+    return {name, std::move(path), std::move(*imported), std::move(schema)};
```
gaturchenko, line 233: Maybe move `name` as well, since it's taken by value?

nes-inference/tests/ModelValidationTest.cpp
gaturchenko, line 1: Could you explain why `catalog.hasModel` checks are removed?

gaturchenko (changes requested):
Hey @yschroeder97 submitted my review, minor things pretty much

greenBene (approved):

nes-inference/include/ModelCatalog.hpp
yschroeder97: Removed it.

nes-inference/src/Model.cpp
yschroeder97: Applied.

nes-inference/tests/ModelValidationTest.cpp
yschroeder97: Reintroduced them. 

In the coordinator, the catalog has its own tests to validate such things (there is no "model catalog" anymore, just one catalog with a table for ML models). 

But until then, the `ModelCatalog` would be uncovered/untested, so I should remove the test then and not now.

gaturchenko (approved):
LGTM

ls-1801 (approved):

